PR #3, opened 2026-05-13
Design Change, please check the html chart

t2o2:


dbrowneup:
This looks great! I'm not at home so I can't pop into Discord, but I can see this here 😁

dbrowneup:
Your diagram is making me think.... blockchain as memory? Essentially, that's what it is.

Something I shared with Marten earlier today: https://thinks.lol/2025/01/memory-makes-computation-universal/

dbrowneup:
This gives me a new perspective on Eric's conclusion in that blog post:

A Path Forward
As a sandwich-eating universal intelligence, I’m not convinced. Humans got smarter not by growing our brains, but by distributing memory across society—spoken word, libraries, the internet—finding ways to preserve and share knowledge without a ruinous overhead. 


**"distributing memory across society"**


Paired with our on-chain reasoning hashes, blockchain distributes memory across computers.

dbrowneup:
Granted, humans **did** in fact get smarter by growing our brains over a few million years of evolution. But his point stands when time-bound by the last 10,000 years or so.

t2o2:
> This gives me a new perspective on Eric's conclusion in that blog post:
> 
> A Path Forward As a sandwich-eating universal intelligence, I’m not convinced. Humans got smarter not by growing our brains, but by distributing memory across society—spoken word, libraries, the internet—finding ways to preserve and share knowledge without a ruinous overhead.
> 
> **"distributing memory across society"**
> 
> Paired with our on-chain reasoning hashes, blockchain distributes memory across computers.

Language enables human to pass down knowledge through generations with much higher fidelity than words or drawings. And for its simplicity, and human has made LLM to mimic the way we generate words that we prefer and recognise. Under the hood, it's still only probability prediction. It doesn't really understand what it spits out. There is a really good paper called The Bitter Lesson, really fundamental to the approach to AI.

dbrowneup:
There was also a nice paper from Apple called The Illusion of Thinking that touches on this I believe. Good food for thought. Sometimes I wonder though. Do humans understand any better than machines? In some ways the answer is obviously yes. In other ways... well, let's just say there are some humans who don't seem to understand much at all. 🤣

t2o2:
Interesting paper. This is probably why you can only efficiently use up to
~60% of the context window. The return after that is a slippery slope.

PR #87, opened 2026-05-19
[fix] backtest_repository NameError + import/dependency guards

dbrowneup:
## What

Three small, bounded fixes from this session's dependency/import audit. Code + CI + editor config; no behavior change beyond fixing a latent bug.

## Changes

1. **Fix a latent `NameError`** — `backend/archimedes/services/backtest_repository.py:77`: `except (json.JSONDecodeError, KeyError)` referenced bare `json`, but it's imported as `import json as _json` (line 69). On any malformed `artifact_json` this raised `NameError` instead of falling through to the equity-curve fallback — a silent break in the backtest read path. → `_json.JSONDecodeError`.
2. **`.vscode/settings.json`** — points Pylance/ruff/pytest at the `archimedes` conda env. Fixes the "cannot resolve `import anthropic`" false positive (the dep *is* declared + installed; VS Code was just on the wrong interpreter). Function-local imports like `strategy_fusion.py:128` now resolve.
3. **`.github/workflows/import-guard.yml`** — a deliberately narrow PR-only guard against the silent-failure class: `ruff F821` (catches aliased-import `NameError`s like #1), `compileall`, and `deptry` (AST-based — sees function-local/lazy imports; flags missing/undeclared deps). Not a full lint regime (that stays Chuan's call); does not touch the deploy workflow.

## Audit result (context)

Full dependency posture is **clean** — every runtime third-party import (incl. `anthropic`) is declared in both `requirements.txt` and `environment.yml` and installed. Only genuine finding was bug #1 plus one low-severity deploy edge (`archimedes_analytics_engine.cli` sys.path injection, separate).

## Not touched (flagged for a separate, owner-approved change)

`.env.example` is **inconsistent with how the intelligence actually runs** (it says "get an Anthropic key", but the working path is GLM via `ANTHROPIC_AUTH_TOKEN` + `ANTHROPIC_BASE_URL`, which it doesn't list; chain-ID comment also stale). Editing `.env.example` is ask-first per CLAUDE.md — recommended changes are written up separately, not in this PR.

## Verify

```
ruff check backend/archimedes --select F821   # clean
pytest                                          # 0 failed
```

### DIFF
--- a/.github/workflows/import-guard.yml
+++ b/.github/workflows/import-guard.yml
@@ -1,0 +1,40 @@
+# Import / dependency guard — catches the silent-failure class:
+# function-local ("lazy") imports of packages that are not declared in
+# requirements.txt / environment.yml, and undefined-name bugs (F821) like
+# `except json.X` when json was imported `as _json`.
+#
+# Scoped deliberately narrow: this is NOT a full lint regime (that stays
+# Chuan's call) — it only guards imports/deps so a missing dependency or
+# aliased-name NameError can't reach the live deploy silently. Runs on PRs
+# only; does not touch the deploy workflow.
+
+name: import-guard
+
+on:
+  pull_request:
+    paths:
+      - "backend/**.py"
+      - "backend/requirements.txt"
+      - "environment.yml"
+      - ".github/workflows/import-guard.yml"
+
+jobs:
+  import-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install guards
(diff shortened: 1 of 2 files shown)


dbrowneup:
Correction: `.vscode/settings.json` is **not** in this PR — `.vscode/` is intentionally gitignored (`.gitignore:33`). I won't override that policy unilaterally. PR #87 = the `backtest_repository` NameError fix + the `import-guard` workflow only. The Pylance fix should be either (a) a documented "Python: Select Interpreter → archimedes env" step in the README/CLAUDE setup, or (b) an explicit `!/.vscode/settings.json` gitignore negation if the team wants it tracked — owner's call (gitignore policy).

PR #123, opened 2026-05-22
[quant] Self-review fixes: 14 issues from internal critique sub-agents

onder-akkaya:
## Summary

After PR #121 merged, three independent reviewers (quant-correctness, paranoid-backend, UI/API-contract) tore the work apart. This PR addresses every substantive finding.

### Quant correctness (the embarrassing ones)
- **Kelly was solved on TOTAL returns; should be EXCESS.** Now subtracts `rf_annual = 5%/yr` before the QP. User-facing μ stays gross.
- **`exp_max_dd = 2.0 × portfolio_vol` was nonsense as a max-drawdown estimator** (at best the median 1y down-move for a zero-Sharpe asset). Replaced with the Magdon-Ismail & Atiya (2004) closed-form expected max-DD over 1y. Also surfaces 1y parametric 95% VaR alongside. New API field `max_drawdown_method` says which estimator was used (`magdon_ismail_2004` or `weighted_strategy_avg`).
- **μ-override double-promised paper returns** — strategy CAGR applied to every asset that strategy voted for. An equity-momentum strategy could put 25% CAGR on BIL and the optimizer would load up on T-bills. Added `mu_shrinkage` (default 0.5) that blends override with sample mean. Bayesian prior framing instead of hard override.
- **"Ledoit-Wolf-style identity shrinkage" was a docstring lie** — actual impl is fixed-α=0.10 diagonal shrinkage. Renamed honestly; sklearn LW upgrade flagged as TODO.

### Backend correctness (the silent failures)
- **`trace_hash` was non-deterministic across processes**: `list({set})` order varies by `PYTHONHASHSEED`. Two identical requests produced different hashes — silently broke the on-chain "anyone can re-derive" verifiability promise. Now `sorted({...})`. Verified deterministic in isolation: identical content → identical hash.
- **TSMOM paper-anchor `moskowitz_2012_tsmom`** advertised in the system prompt is NOT a substring of the filename `moskowitz_ooi_pedersen_2012_tsmom.py`, so every TSMOM-anchored pick was silently falling through to `strategies[0]` and getting the wrong rigor metrics. Added `_ANCHOR_ALIASES` table + updated the prompt to use canonical strings.
- **`stress_engine` returned 0.0 silently** for any `asset_class` missing from the scenario dict — including the no-yfinance fallback path's `asset_class="unknown"`. That made every scenario report 0% loss (the worst possible demo-day failure mode). `_shock_for` now returns `(shock, covered)` tuples; `StressResult` tracks `coverage_pct` + `uncovered_symbols`; UI flags <80% coverage with a yellow warning.
- **`paper_delta_max_dd` aggregation used `min()`** with a comment claiming "smaller is better" — but `delta = real - claimed` (both negative); HIGHER delta = strategy under-delivered LESS. Fixed to `max()`.
- **Un-converged SLSQP iterate was returned as a valid result** and could even violate bounds before the clip. Now returns `None` to trigger fallback.

### Response contract (UI broken without crashing)
- `agent` dict in the rule-based path was missing `iterations` and `tool_calls` keys → fixed.
- `_build_rigor_summary` empty branch was missing threshold/avg keys → fixed.
- `risk_decomposition` + `correlation_pairs` returned synth codes (`sSPY`, `sQQQ`) instead of display symbols — UI table didn't line up with the per-pick table on the same screen. Now use `_display_for()`. `risk_decomposition` still carries `synth` for downstream callers.

### Code hygiene
- Dead import (`_build_mu_sigma_from_prices` in portfolio_agent.py).
- Dead param (`agent_obj` in `_build_and_anchor_trace`).

### UI
- `fmt()` / `fmtPct()` guard against Infinity/NaN (would otherwise render literal "Infinity" on the page).
- Advisor surfaces backend 200-with-error bodies as user-facing errors instead of rendering a half-empty card.
- Stress panel handles non-finite pnl + flags low-coverage scenarios.

## What I did NOT fix (deliberate)

- **Rule-based aggregation rigor cherry-picking** (reviewer #2 finding): the agg row keeps the best contributing strategy's DSR/PBO. This is honest because the pick IS supported by that strategy — but the alternative (averaging) is also defensible. Left as-is pending product call.
- **Real Ledoit-Wolf intensity from data**: the fixed-α=0.10 still under/over-shrinks per window length. TODO commented; one-line swap with sklearn.
- **Per-pick table dropping Sharpe + CAGR** (UX regression flagged by reviewer #3): the aggregate Expected Portfolio Metrics still shows Sharpe at the top; per-pick now shows DSR/PBO/OOS/Δ-vs-paper which I think is more informative. Open to feedback.

## Test plan

- [x] `ruff check backend/archimedes --select F821` → all checks passed
- [x] `python -m pytest backend/tests/test_api_routes.py backend/tests/services/ -x -q` → 133 passed, 2 skipped
- [x] Verified trace_hash determinism: same content → same hash across runs; previously different due to set-ordering
- [x] End-to-end smoke: Sharpe 0.89, CAGR 18.9%, vol 16.7%, MaxDD 4.9% (Magdon-Ismail), VaR_95 8.6%, stress coverage 100% on rule-based path
- [ ] Post-merge: confirm `/api/strategies/advisor` on live EC2 shows the new max-DD method + coverage_pct fields

## Files

- `backend/archimedes/api/routes.py` — Kelly excess, μ-shrinkage wiring, MaxDD/VaR call sites, response symmetry, error path, dead param
- `backend/archimedes/services/portfolio_optimizer.py` — excess returns, Magdon-Ismail, VaR_95, μ-shrinkage, un-converged → None, display symbols, honest shrinkage docstring
- `backend/archimedes/services/portfolio_agent.py` — anchor prompt fix, dead import
- `backend/archimedes/services/stress_engine.py` — `(shock, covered)` tuple, `coverage_pct` + `uncovered_symbols`
- `ui/src/components/PortfolioAdvisor.jsx` — fmt guards, data.error path, low-coverage warning

### DIFF
--- a/backend/archimedes/api/routes.py
+++ b/backend/archimedes/api/routes.py
@@ -761,7 +761,6 @@ async def get_portfolio_advisor(
     async def _build_and_anchor_trace(
         allocations_for_trace: list[dict],
         thesis_for_trace: str,
-        agent_obj,
     ) -> dict:
         """Build a ReasoningTrace from the recommendation, compute the
         keccak256 hash, and best-effort anchor it on-chain via
@@ -817,7 +816,11 @@ async def get_portfolio_advisor(
                 confidence=float(regime_confidence or 0.0),
                 expected_outcome="Portfolio constructed; pending user vault deployment",
                 trades_executed=[],
-                strategies_referenced=list({
+                # sorted() — set iteration order is non-deterministic across
+                # Python processes, which would produce a different trace_hash
+                # for identical inputs and silently break the on-chain
+                # verifiability promise.
+                strategies_referenced=sorted({
                     a.get("paper_anchor") for a in allocations_for_trace if a.get("paper_anchor")
                 }),
             )
@@ -893,8 +896,11 @@ async def get_portfolio_advisor(
         n = len(active_rows)
         if n == 0:
             return {
-                "total_picks": 0, "passes_rigor_gate": 0, "dsr_significant": 0,
-                "pbo_acceptable": 0, "oos_positive": 0,
+                "total_picks": 0, "passes_rigor_gate": 0,
+                "dsr_significant": 0, "dsr_significant_threshold": 0.05,
+                "pbo_acceptable": 0, "pbo_acceptable_threshold": 0.50,
+                "oos_positive": 0,
+                "avg_dsr_p_value": None, "avg_pbo_score": None,
             }
         passes = sum(1 for r in active_rows if r.get("passes_rigor_gate"))
         dsr_sig = sum(
@@ -927,15 +933,52 @@ async def get_portfolio_advisor(
 
     # ── Agent path: convert AgentPicks into `scored` rows ──────────
     if agent_portfolio and agent_portfolio.picks:
+        # Loose-match map: common abbreviations the LLM may emit → canonical
+        # substrings of strategy_code_path.  Without this, "tsmom" alone
+        # won't match "moskowitz_ooi_pedersen_2012_tsmom.py" via in-check
+        # because the LLM tends to drop author names.
+        _ANCHOR_ALIASES = {
+            "tsmom":        "tsmom",
+            "moskowitz":    "moskowitz",
+            "moreira":      "moreira_muir",
+            "vol_managed":  "moreira_muir",
+            "vol-managed":  "moreira_muir",
+            "faber":        "faber_2007_sma200",
+            "sma200":       "sma200",
+            "george":       "george_hwang",
+            "hwang":        "george_hwang",
+            "52w":          "52w",
+            "52_week":      "52w",
+            "52-week":      "52w",
+            "capital_preservation": "capital_preservation",
+            "tbill":        "capital_preservation",
+            "t_bill":       "capital_preservation",
+            "buy_hold":     "buy_hold",
+            "buy-hold":     "buy_hold",
+            "buy_and_hold": "buy_hold",
+        }
+
         def _find_strategy_for_anchor(anchor: str):
-            anchor_l = (anchor or "").lower()
+            anchor_l = (anchor or "").lower().strip()
             if not anchor_l:
                 return strategies[0] if strategies else None
+            # Try direct substring against path / title / id first.
             for st in strategies:
-                if (anchor_l in (st.strategy_code_path or "").lower()
-                        or anchor_l in (st.paper_title or "").lower()
-                        or anchor_l in st.id.lower()):
+                hay = (
+                    (st.strategy_code_path or "").lower()
+                    + " " + (st.paper_title or "").lower()
+                    + " " + st.id.lower()
+                )
+                if anchor_l in hay:
                     return st
+            # Fall back to alias resolution — let "tsmom"/"moskowitz_2012_tsmom"
+            # still resolve to moskowitz_ooi_pedersen even though it's not
+            # a substring of the filename.
+            for alias, canonical in _ANCHOR_ALIASES.items():
+                if alias in anchor_l:
+                    for st in strategies:
+                        if canonical in (st.strategy_code_path or "").lower():
+                            return st
             return strategies[0] if strategies else None
 
         for pick in agent_portfolio.picks:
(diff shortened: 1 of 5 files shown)


dbrowneup:
## ✅ Reviewed + rebased + one cosmetic-alignment commit added — ready to merge

### Rebase
Branch was 2 days behind main (base `7153177` from PR #122 merge). Rebased against latest main `e9b1cc5`. **Mechanically clean — no manual conflict resolution needed.** Surprising given all 5 files were touched on main since:
- `routes.py` — Chuan's Day-10 advisor/stress/agent endpoints + empty_vault filter
- `portfolio_agent.py` — Chuan's Day-10 new 850-line file
- `portfolio_optimizer.py` — Chuan's Day-10 refactor
- `stress_engine.py` — Chuan's Day-10 new 380-line file
- `PortfolioAdvisor.jsx` — Daniel R's PR #124 surgical UnoCSS pass

Your edits landed in non-overlapping line regions. `git merge-tree` dry-run was clean before I pushed.

### One cosmetic-alignment commit added: `421d771`
PR #124 (Daniel R, merged 8h ago) added the convention "every emoji → Lucide icon span" for the frontend. Your low-coverage warning in `PortfolioAdvisor.jsx` had a literal `⚠` glyph and an inline-style block — the glyph specifically violates the new convention. Narrow fix:

```diff
-  <div style={{
-    fontSize: '0.65rem', fontWeight: 600, color: '#f59e0b',
-    marginTop: 2,
-  }}>
-    ⚠ only {(coverage * 100).toFixed(0)}% of book modeled
-  </div>
+  <div className="inline-flex items-center gap-1 mt-0.5 text-[0.65rem] font-semibold text-[#f59e0b]">
+    <span className="i-lucide-alert-triangle w-3 h-3" />
+    only {(coverage * 100).toFixed(0)}% of book modeled
+  </div>
```

No behavior change — same threshold (< 0.8), same warning text, same color, same conditional. `i-lucide-alert-triangle` was already in `uno.config.js`'s safelist so no churn there. Verified with `docker compose build nginx --no-cache` (vite: ✓ built in 2.54s).

Deliberately narrow: 71 other inline-style blocks remain in `PortfolioAdvisor.jsx` from before #124 — Daniel R's pass was surgical (3+/4-) on this specific file, not wholesale. Converting just your edits without touching those would create inconsistency in the opposite direction; only the emoji rule was an unambiguous violation.

### On your three "open to feedback" items

1. **Rule-based aggregation rigor cherry-picking** — your framing is right. The pick *is* supported by that strategy; using its DSR/PBO is honest attribution. Averaging would silently dilute attribution from a passing strategy with a barely-failing one. **Keep as-is.**

2. **Fixed-α=0.10 vs real Ledoit-Wolf intensity** — fine for the demo window. The honest docstring + the TODO is the right posture; sklearn LW swap is a one-liner post-hackathon. **Keep as-is.**

3. **Per-pick table dropping Sharpe + CAGR for DSR/PBO/OOS/Δ-vs-paper** — your call is correct. Sharpe alone is what gets gamed (multiple-testing inflation, in-sample fit, etc.); DSR/PBO/OOS/Δ is the rigor wedge surfaces and the exact set the demo needs to land. The aggregate Expected Portfolio Metrics still shows Sharpe at the top so it's not hidden. **Keep as-is.**

### Substantive fixes I want to call out

The quant-correctness fixes are the kind of things that get a deck destroyed in Q&A if a sharp judge spots them:

- **Kelly on excess returns, not total** — fundamental. Solving Kelly on gross returns leads to over-allocation in any positive-Sharpe-floor regime.
- **`exp_max_dd = 2.0 × portfolio_vol`** replaced by **Magdon-Ismail & Atiya (2004)** closed-form — well-cited, defensible under a real quant question. Surfacing 1y parametric 95% VaR alongside is the right move.
- **μ-override double-promised paper returns** — your `mu_shrinkage = 0.5` Bayesian-blend fix is exactly the right framing. The hard-override version was a CAGR-laundering bug.
- **`trace_hash` non-determinism via `list({set})`** — silent on-chain verifiability break. Fix is `sorted({...})` which is the correct one-liner; the fact that you found it through review (not a test failure) is concerning but well-handled.
- **TSMOM anchor `moskowitz_2012_tsmom` not a substring of `moskowitz_ooi_pedersen_2012_tsmom.py`** — silent paper-attribution failure mode. Verified all 6 new anchor IDs in `_agent_tools()` are substrings of the actual `analytics-engine/strategies/*.py` filenames. Your `MUST be substring` comment is load-bearing — should probably move to a docstring on `_find_strategy_for_anchor` in the next pass.
- **`stress_engine` returned `0.0` silently** for uncovered asset classes — worst possible demo-day failure mode. `(shock, covered)` tuple + `coverage_pct` + `uncovered_symbols` is the correct fix; the UI low-coverage warning closes the loop.

### Tests
- Your test plan: 133 passed / 2 skipped in `test_api_routes.py` + `services/`. Matches my full-suite run on main: 300 passed / 2 failed (pre-existing Redis-state-bleed flakes in `TestAgentRoutes::test_agent_status_redis_down_defaults` + `TestAdvisorRoutes::test_advisor_redis_unavailable`) / 2 skipped. Neither flake introduced by this PR — same exact failures on main pre-rebase.
- UI build clean on the cosmetic-alignment commit.
- Post-merge confirmation step (`/api/strategies/advisor` on live EC2 shows new `max_drawdown_method` + `coverage_pct` fields) — that one can be checked once CI deploy completes.

### Merge plan
Merging #126 first (docs-only, can't interact), then re-verifying this PR remains mergeable, then merging. No conflicts predicted between the two.

Strong substantive PR, Önder — these are the kind of fixes that make the difference between a deck that survives Q&A and one that doesn't.

PR #324, opened 2026-05-25
[backend] Per-vault strategy scoping in agent_runner (Issue #307) !minor

### LINKED ISSUE #307: APIN - Backend - agent_runner applies global strategy consensus to ALL vaults; needs to read each vault's VaultMetadata.strategy_ids and derive per-vault allocations

t2o2:
Closes #307. Wire _get_vault_strategy_ids() into the rebalance loop. Each vault now executes only its selected strategies (skip if no metadata). 859 tests pass.

### DIFF
--- a/backend/archimedes/chain/agent_runner.py
+++ b/backend/archimedes/chain/agent_runner.py
@@ -185,13 +185,55 @@ class StrategyRunner:
 
             logger.info("[tick %s] Managing %d vaults", tick_id, len(vaults))
 
-            # 6. Process each vault
+            # 6. Process each vault with per-vault strategy scoping (Issue #307)
             for vault_addr in vaults:
                 try:
+                    # Per-vault scoping: each vault executes only its selected strategies
+                    vault_strategy_ids = self._get_vault_strategy_ids(vault_addr)
+
+                    if vault_strategy_ids is None:
+                        # No metadata — skip. Never apply global consensus to
+                        # a vault the user hasn't configured.
+                        logger.warning(
+                            "[tick %s] Vault %s has no metadata (strategy_ids=None) — "
+                            "skipping rebalance. Deploy via UI to set strategies.",
+                            tick_id,
+                            vault_addr[:10],
+                        )
+                        continue
+
+                    # Filter signals to this vault's strategies
+                    scoped_signals = [ss for ss in all_signals if ss.strategy_id in vault_strategy_ids]
+
+                    if not scoped_signals:
+                        logger.warning(
+                            "[tick %s] Vault %s strategy_ids=%s — none produced signals, skipping",
+                            tick_id,
+                            vault_addr[:10],
+                            vault_strategy_ids,
+                        )
+                        continue
+
+                    # Aggregate scoped signals into per-vault target weights
+                    vault_weights = strategy_evaluator.aggregate_signals(
+                        scoped_signals,
+                        usdc_floor=USDC_FLOOR,
+                    )
+                    vault_targets = self._weights_to_targets(vault_weights, scoped_signals)
+
+                    logger.info(
+                        "[tick %s] Vault %s: scoped to %d/%d strategies → %s",
+                        tick_id,
+                        vault_addr[:10],
+                        len(scoped_signals),
+                        len(all_signals),
+                        " | ".join(f"{k}={v:.0%}" for k, v in vault_weights.items()),
+                    )
+
                     await self._process_vault(
                         vault_addr,
-                        targets,
-                        all_signals,
+                        vault_targets,
+                        scoped_signals,
                         regime,
                         tick_id,
                     )
@@ -207,6 +249,29 @@ class StrategyRunner:
 
         except Exception:
             logger.exception("[tick %s] Strategy tick failed — will retry", tick_id)
+
+    # ─── Per-vault strategy scoping (Issue #307) ─────────────────────
+
+    def _get_vault_strategy_ids(self, vault_address: str) -> list[str] | None:
+        """Load strategy_ids from VaultMetadata for a vault.
+
+        Returns:
+            list[str] — the user's selected strategies for this vault.
+            None — if no metadata exists (vault deployed outside UI).
+        """
+        try:
+            from archimedes.db import get_session
+            from archimedes.models.chat import VaultMetadata
+
+            with get_session() as session:
+                meta = session.query(VaultMetadata).filter(VaultMetadata.vault_address == vault_address).first()
+                if meta is None:
+                    return None
+                ids = meta.get_strategy_ids()
+                return ids if ids else None
+        except Exception as exc:
+            logger.debug("_get_vault_strategy_ids(%s) failed: %s", vault_address[:10], exc)
+            return None
 
     async def _process_vault(
         self,
--- a/backend/tests/test_health_amm.py
+++ b/backend/tests/test_health_amm.py
@@ -1,0 +1,107 @@
+"""Tests for /health/amm endpoint (Issue #309)."""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from httpx import ASGITransport, AsyncClient
+
+
+@pytest.mark.asyncio
+async def test_health_amm_returns_200_or_503_never_404():
+    """The endpoint exists and never returns 404."""
+    from archimedes.main import app
+
+    async with AsyncClient(transport=ASGITransport(app=app), base_url="http://test") as client:
+        resp = await client.get("/health/amm")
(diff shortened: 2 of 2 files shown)


dbrowneup:
## 🚨 Demo-breaking change if merged as-is — please read before any further iteration

Thanks for the rebase + the dropped duplicates. The narrowed diff is clean. **However:** the new behavior in this PR vs the older #321 you closed has a critical regression I need to flag for the submission-day demo window.

### The behavior change

Old #321 (closed): "no metadata → fallback to global consensus"
This #324: **"no metadata → explicit skip with warning log + `continue`"**

```python
if vault_strategy_ids is None:
    logger.warning(...)
    continue   # ← THIS LINE stops rebalancing for legacy vaults
```

### Live state right now

```sh
$ curl -s 'https://archimedes-arc.app/api/vaults/' | jq '.vaults[].strategy_ids'
[]
[]
[]
[]
[]
[]
```

**All 6 live vaults return `strategy_ids: []`**, meaning **every existing vault hits the `_get_vault_strategy_ids() is None` path** under this PR's logic. After deploy, the agent's tick loop would skip all 6 vaults — no more rebalances, no more new on-chain traces, no more `arc_tx_hash` values landing in `/api/traces`.

The 4 real rebalance traces we've shown to judges (visible at `/api/traces/?limit=10`, including the just-landed one at `2026-05-25T17:46:44+00:00` with commit/reveal/trade tx hashes 43987404 / 43987418 / 43987412) all came from the global-consensus fallback path. After this PR, that path is gone.

This is **strictly worse for the submission-day demo** than the current global-consensus behavior, even though the per-vault scoping is architecturally correct for non-legacy vaults.

### Path forward (suggested)

Two options, pick whichever you prefer:

**Option A — Restore the soft fallback for legacy vaults** (the #321 behavior):

```python
if vault_strategy_ids is None:
    # Legacy vault (deployed before strategy-selection flow shipped) — 
    # fall back to global consensus for the demo window. New vaults
    # deployed via the UI will set strategy_ids and follow the scoped path.
    logger.info("[tick %s] Vault %s: no metadata, using global consensus", tick_id, vault_addr[:10])
    await self._process_vault(vault_addr, targets, all_signals, regime, tick_id)
    continue
```

**Option B — Backfill `VaultMetadata` rows for the 6 existing vaults in the same PR** so they get the scoped path with a sensible default (e.g. all curated strategies).

Either is fine; Option A is the smaller surgical change and matches the spirit of "don't change live behavior on demo day unless you've verified the new path produces equivalent or better output."

### Other notes (lower priority)

- `test_health_amm.py:80` mocks `mock_loader.amm_router.return_value = mock_router` (treats `amm_router` as a callable). That mirrors the AsyncContract-not-callable bug in #318's live implementation rather than the correct property-access pattern. Tests pass for the same reason live fails. See follow-up issue #325 for the underlying bug. Not a #324 blocker — but worth fixing in whatever PR closes #325.
- `getAllPools()` — be sure this method exists on the deployed AMM router. The live error message from #318 is `'AsyncContract' object is not callable` which suggests the issue is the call pattern, not the method name, but worth a sanity check against the cached ABI in `contracts/abis/`.

### Acceptance for re-review

- [ ] After Option A or B applied, `pytest backend/tests/chain/test_agent_runner.py` adds explicit coverage that **legacy vaults (`VaultMetadata is None`) still rebalance** post-merge.
- [ ] On a deploy preview (or post-merge with a fast revert plan), `curl 'https://archimedes-arc.app/api/traces/?limit=5'` produces at least one new `arc_tx_hash` within ~10 minutes of the deploy completing, proving rebalancing continues.

Sorry to flag this so late in the process — I'd been validating live state alongside the diff and only caught the gap when I cross-checked `strategy_ids` on the live vaults.

PR #341, opened 2026-05-25
[frontend] Generate: 'View in Library' button now actually navigates

dbrowneup:
## Summary

The post-generation **View in Library** button was setting `window.location.hash = #/library?highlight=...` — that only changes the URL fragment without telling the SPA router, so the user stayed on `/generate` and watched a phantom hash get appended to the address bar. A happy-path clicker hits this and the canonical flow stalls at the most important moment (the just-generated strategy → review surface).

Fix: thread the parent `onNavigate` through `Generate` → `GenerationStream` and call it correctly with the strategy_id as the `highlight` param. Falls back to a full-page nav (`window.location.href = ...`) if the prop isn't wired so the button never silently does nothing.

## Test plan

- [x] `npm run build` passes.
- [x] `npm run lint` clean.
- [ ] Manual after merge: Generate → submit a brief → wait for SSE done → click "View in Library →" on a candidate card → land on `/library?highlight=<strategy_id>` with the row scrolled into view.

### DIFF
--- a/ui/src/components/Generate.jsx
+++ b/ui/src/components/Generate.jsx
@@ -378,6 +378,7 @@ export default function Generate({ onNavigate }) {
           onDone={onJobDone}
           onReset={resetJob}
           onPipelineSelected={handlePipelineEvent}
+          onNavigate={onNavigate}
         />
       )}
 
--- a/ui/src/components/GenerationStream.jsx
+++ b/ui/src/components/GenerationStream.jsx
@@ -72,7 +72,7 @@ function summarizeEvent(name, data) {
   }
 }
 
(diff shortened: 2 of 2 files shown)


dbrowneup:
Closing as duplicate — my PR #343 (https://github.com/a-apin/archimedes/pull/343) ships the same View-in-Library navigation fix using identical onNavigate threading, AND also includes two additional bug fixes Dan surfaced in his SPEC-1 walkthrough (rigor-failed strategies polluting the Library Generated tab + RejectedCandidates modal contrast). Letting the broader bundle land instead. Thanks for the parallel work — same diagnosis confirms it's the right fix.

PR #403, opened 2026-05-26
[backend] executor: confirm rebalance receipt + raise on revert (#6)

onder-akkaya:
## Summary

`execute_trades` sent the rebalance tx and returned the hash **without waiting for a receipt** (both the Circle and raw-key paths). A reverted trade was logged as "sent" and returned as a success hash — so the agent then published a **REBALANCE** reasoning trace for a trade that never actually settled. Meanwhile `create_vault` already waits for its receipt; only the rebalance path skipped it.

## Fix

- New `_confirm_receipt(tx_hash)` — waits for the receipt, raises `TradeRevertedError` on `status == 0`, returns the normalized 0x hash on success. Mirrors the existing `create_vault` pattern.
- Wired into both `execute_trades` submission paths (Circle + raw key).
- The `agent_runner` generic handler already catches this and records an honest `execution_failed` SKIP trace instead of a false REBALANCE.
- Raw-key path now uses the **pending-block nonce** to reduce nonce-collision drops on a quick second rebalance.

## Tests

`backend/tests/chain/test_executor_receipt.py` (mocked w3):
- reverts (status 0) → `TradeRevertedError`
- success (status 1) → returns the hash
- unprefixed hash → normalized to `0x…`

All green; blocking ruff checks pass.

## Scope / review

⚠️ **On-chain rebalance path — requesting @Chuan's review.** No contract change; backend-only. The behavior change is: a reverted rebalance now surfaces as a failed trace rather than a false success. This is strictly more honest.

## Note on nonce handling

This addresses the receipt-confirmation half of the audit's "fire-and-forget + nonce race" finding, plus a pending-nonce improvement. A full nonce-manager (serializing concurrent submissions per signer) is a larger change left for a follow-up if concurrency on a single signer becomes real.

### DIFF
--- a/backend/archimedes/chain/executor.py
+++ b/backend/archimedes/chain/executor.py
@@ -33,6 +33,15 @@ class InsufficientLiquidityError(RuntimeError):
     """Raised when an AMM pool's USDC reserve is below MIN_HEALTHY_LIQUIDITY_USDC."""
 
 
+class TradeRevertedError(RuntimeError):
+    """Raised when a submitted rebalance transaction reverts on-chain (status 0).
+
+    Without this, a reverted rebalance is logged as "sent" and returned as a
+    success hash — the agent then records a reasoning trace as if the trade
+    executed when it actually failed.
+    """
+
+
 class ChainExecutor:
     """Executes on-chain vault operations: read portfolio, rebalance, create vault."""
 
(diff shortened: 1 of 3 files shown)


t2o2 (commented):
Reviewed per Dan's notes. Two items before merge:

1. **HexBytes test gap**: Both new tests for `_confirm_receipt` pass `str` tx hashes, but the raw-key signer path emits `HexBytes`. Add one test with `bytes.fromhex(...)` input to cover that branch.

2. **Call-site audit**: `execute_trades` now raises `TradeRevertedError` where it previously returned a hash. `agent_runner.py:607` catches via generic `except` (fine). Verified no other call sites assume success-on-return without exception handling.

Dan's read: fix (1), then merge. Holding merge until the HexBytes test is added.

PR #419, opened 2026-05-27
[infra] ALB + WAF: load balancer with managed security rules !minor

t2o2:
## Summary

Application Load Balancer with AWS WAF, completing the public-facing
security layer. After DNS cutover, the ALB becomes the ONLY entry point.

## ALB Config

| Setting | Value | Why |
|---------|-------|-----|
| idle_timeout | 300s | SSE `/api/generate/stream/` needs long connections |
| TLS policy | ELBSecurityPolicy-TLS13-1-2-2021-06 | TLS 1.3 preferred |
| Health check | `/health` on port 8080, 30s interval | Not the SSE endpoint |
| Access logs | S3 bucket, 30-day lifecycle | Audit trail |
| HTTP→HTTPS | Redirect listener on port 80 | Force encryption |

## WAF Rules (per locked spec)

| Rule | Priority | Action |
|------|----------|--------|
| Rate limit (1000 req/5min/IP) | 1 | Block |
| Core Rule Set | 10 | Managed |
| Known Bad Inputs | 20 | Managed |
| IP Reputation | 30 | Managed |
| SQL Injection | 40 | Managed |

No Bot Control, no geo-blocking (per spec).

## Terraform plan

```
Plan: 40 to add, 1 to change, 0 to destroy.
(includes VPC+Aurora from #416 not yet applied)
```

## DNS cutover plan

After `terraform apply`:
1. Set Route 53 TTL to 60s
2. Change A record to ALIAS → ALB DNS name
3. Verify HTTPS works through ALB
4. Restore normal TTL

## Sequence

PR 5 of the security architecture.
Previous: #396, #409, #416, #417.
Next: EC2 → private subnet + port 22 off.

### DIFF
--- a/infra/alb.tf
+++ b/infra/alb.tf
@@ -1,0 +1,197 @@
+# ── Application Load Balancer ─────────────────────────────────────────
+#
+# The ALB is the ONLY public-facing resource once EC2 moves to the
+# private subnet. Terminates TLS via ACM certificate, forwards to
+# the EC2 target group.
+#
+# Key config per the locked spec:
+#   - idle_timeout = 300s (SSE /api/generate/stream/ survival)
+#   - Access logs → S3
+#   - HTTP → HTTPS redirect (listener rule)
+
+# ── S3 bucket for ALB access logs ────────────────────────────
+
+resource "aws_s3_bucket" "alb_logs" {
+  bucket = "${var.project_name}-alb-logs-159903201072"
+
+  tags = {
+    Project = var.project_name
+  }
+}
+
+resource "aws_s3_bucket_lifecycle_configuration" "alb_logs" {
+  bucket = aws_s3_bucket.alb_logs.id
+
+  rule {
+    id     = "expire-old-logs"
+    status = "Enabled"
+
+    filter {}
+
+    expiration {
+      days = 30
+    }
+  }
+}
+
+# ALB requires a specific bucket policy for log delivery
+data "aws_elb_service_account" "main" {}
+
+resource "aws_s3_bucket_policy" "alb_logs" {
+  bucket = aws_s3_bucket.alb_logs.id
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Sid       = "ALBLogDelivery"
+        Effect    = "Allow"
+        Principal = { AWS = data.aws_elb_service_account.main.arn }
+        Action    = "s3:PutObject"
+        Resource  = "${aws_s3_bucket.alb_logs.arn}/*"
+      }
+    ]
+  })
+}
+
+# ── ACM Certificate ──────────────────────────────────────────
+# Use the existing Let's Encrypt cert on the EC2 for now.
+# For ALB, we need an ACM certificate. Import or create one.
+# Using DNS validation via Route 53.
+
+resource "aws_acm_certificate" "main" {
+  domain_name       = "archimedes-arc.app"
+  validation_method = "DNS"
+
+  tags = {
+    Project = var.project_name
+  }
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+# ── Security Group ───────────────────────────────────────────
+
+resource "aws_security_group" "alb" {
+  name        = "${var.project_name}-alb-sg"
+  description = "ALB — public HTTPS only"
+  vpc_id      = aws_vpc.main.id
+
+  # HTTPS from internet
+  ingress {
+    description = "HTTPS"
+    from_port   = 443
+    to_port     = 443
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  # HTTP (for redirect to HTTPS)
+  ingress {
+    description = "HTTP redirect"
+    from_port   = 80
+    to_port     = 80
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  # Outbound to targets
+  egress {
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
+  }
+
+  tags = {
+    Name    = "${var.project_name}-alb-sg"
+    Project = var.project_name
+  }
+}
+
+# ── ALB ──────────────────────────────────────────────────────
+
+resource "aws_lb" "main" {
+  name               = "${var.project_name}-alb"
+  internal           = false
+  load_balancer_type = "application"
+  security_groups    = [aws_security_group.alb.id]
+  subnets            = aws_subnet.public[*].id
+
+  idle_timeout = 300 # SSE /api/generate/stream/ needs long-lived connections
+
+  access_logs {
+    bucket  = aws_s3_bucket.alb_logs.id
+    enabled = true
+  }
+
+  tags = {
+    Project = var.project_name
+  }
+}
+
+# ── Target Group ─────────────────────────────────────────────
+# Points at the EC2 instance. Health check on /health (not /api/generate/stream/).
+
+resource "aws_lb_target_group" "backend" {
+  name     = "${var.project_name}-backend-tg"
+  port     = 80
+  protocol = "HTTP"
+  vpc_id   = data.aws_vpc.default.id # EC2 is still in default VPC
+
+  health_check {
+    enabled             = true
+    path                = "/health"
+    port                = "8080" # nginx listens on 8080 inside the container
+    protocol            = "HTTP"
+    healthy_threshold   = 2
+    unhealthy_threshold = 3
+    interval            = 30
+    timeout             = 5
+    matcher             = "200,301"
+  }
+
+  tags = {
+    Project = var.project_name
+  }
+}
+
+resource "aws_lb_target_group_attachment" "backend" {
+  target_group_arn = aws_lb_target_group.backend.arn
+  target_id        = aws_instance.archimedes.id
+  port             = 80
+}
+
+# ── Listeners ────────────────────────────────────────────────
+
+# HTTPS listener (main)
+resource "aws_lb_listener" "https" {
+  load_balancer_arn = aws_lb.main.arn
+  port              = 443
+  protocol          = "HTTPS"
+  ssl_policy        = "ELBSecurityPolicy-TLS13-1-2-2021-06"
+  certificate_arn   = aws_acm_certificate.main.arn
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.backend.arn
+  }
+}
+
+# HTTP → HTTPS redirect
+resource "aws_lb_listener" "http_redirect" {
+  load_balancer_arn = aws_lb.main.arn
+  port              = 80
+  protocol          = "HTTP"
+
+  default_action {
+    type = "redirect"
+    redirect {
+      port        = "443"
+      protocol    = "HTTPS"
+      status_code = "HTTP_301"
+    }
+  }
+}
--- a/infra/alb.tf
+++ b/infra/alb.tf
@@ -19,6 +19,25 @@ resource "aws_s3_bucket" "alb_logs" {
   }
 }
 
+resource "aws_s3_bucket_server_side_encryption_configuration" "alb_logs" {
+  bucket = aws_s3_bucket.alb_logs.id
+
+  rule {
+    apply_server_side_encryption_by_default {
+      sse_algorithm = "AES256"
+    }
+  }
+}
+
+resource "aws_s3_bucket_public_access_block" "alb_logs" {
+  bucket = aws_s3_bucket.alb_logs.id
+
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
 resource "aws_s3_bucket_lifecycle_configuration" "alb_logs" {
   bucket = aws_s3_bucket.alb_logs.id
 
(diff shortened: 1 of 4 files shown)


dbrowneup:
## Review — meticulous pass

Architecture is sound and structure matches the locked spec. But I found **two blockers** that will break the cutover and several hardening gaps worth tightening before apply. Walking through file by file.

---

### 🔴 Critical 1: ALB health check on port 8080 is unreachable

`alb.tf:147` sets the target group health check to port 8080. The target attachment uses `aws_instance.archimedes.id` so the ALB targets the EC2 host port, not the docker container port. Two things make port 8080 unreachable on the host:

**(a)** `docker-compose.yml` only binds host port 80 → container 8080 and host 443 → container 8443. **Host port 8080 is not bound.** Nothing listens there.

```yaml
# docker-compose.yml — nginx ports block
ports:
  - "80:8080"
  - "443:8443"
```

**(b)** Even if something were listening on host 8080, `infra/main.tf` security group `aws_security_group.archimedes` only allows ingress on 22, 80, 443 — port 8080 is closed at the SG layer.

**Result:** Every health check probe gets connection-refused. The target group will mark the instance unhealthy and the ALB returns 503 to every request after DNS cutover.

**Recommended fix (cheapest):** change the target group health check to port 80 path `/health`, and update `nginx.conf` so the port-8080 listener (inside the container — host port 80) either:
- Forwards `/health` to backend instead of returning a 301, OR  
- Checks `$http_x_forwarded_proto` and only 301's when it's not `https` (see Critical 2 below — same root issue).

Matcher `200,301` already accepts the redirect, but a 301 from a health probe is a bad signal: it means the ALB is being told to follow a redirect rather than reaching the actual service. Better to have nginx serve `/health` directly without redirect on the port-8080 block.

---

### 🔴 Critical 2: nginx will create an HTTP→HTTPS redirect loop behind the ALB

After cutover the path is:

```
Browser → ALB:443 (HTTPS, TLS terminated) → EC2:80 (HTTP) → nginx container:8080
```

`nginx/nginx.conf:7-20` server block on container port 8080 has only one location: `return 301 https://$host$request_uri`. Every request from the ALB will get 301 → browser follows the redirect → ALB:443 → EC2:80 → 301 → loop.

**Fix:** nginx must check `X-Forwarded-Proto` (set by ALB) and only redirect when it's not `https`:

```nginx
server {
    listen 8080;
    server_name archimedes-arc.app;

    if ($http_x_forwarded_proto = "https") {
        # Behind ALB — serve the real app
        # (Move the proxy + static-serve blocks from the 8443 server here, 
        #  or proxy to the 8443 server, etc.)
    }
    # Direct hit on HTTP (no X-Forwarded-Proto) — redirect
    location /.well-known/acme-challenge/ {
        root /usr/share/nginx/html;
    }
    location / {
        return 301 https://$host$request_uri;
    }
}
```

This and Critical 1 should be fixed in the same nginx-config change.

---

### 🟠 Critical-adjacent: ACM cert will hang in PENDING_VALIDATION

`alb.tf:62-73` creates `aws_acm_certificate.main` with DNS validation, but there are no `aws_route53_record` for the validation records and no `aws_acm_certificate_validation` resource in `infra/*.tf`.

The Route 53 zone exists (`Z03812612E5OLGK1YGZSR`, documented in `infra/terraform/route53.tf` — but that file is in a separate root module without a backend, so it's just documentation, not active TF state).

Two paths forward:

**Option A (cleanest):** Add the validation records to `infra/alb.tf` using a data source for the existing zone:

```hcl
data "aws_route53_zone" "main" {
  name = "archimedes-arc.app."
}

resource "aws_route53_record" "acm_validation" {
  for_each = {
    for dvo in aws_acm_certificate.main.domain_validation_options : dvo.domain_name => {
      name   = dvo.resource_record_name
      record = dvo.resource_record_value
      type   = dvo.resource_record_type
    }
  }
  zone_id = data.aws_route53_zone.main.zone_id
  name    = each.value.name
  type    = each.value.type
  ttl     = 60
  records = [each.value.record]
}

resource "aws_acm_certificate_validation" "main" {
  certificate_arn         = aws_acm_certificate.main.arn
  validation_record_fqdns = [for r in aws_route53_record.acm_validation : r.fqdn]
}
```

**Option B:** Document explicitly that the apply will exit with the cert in PENDING_VALIDATION state and a manual Route 53 record-add step is required between `terraform apply` and DNS cutover. The risk with this path: the ALB HTTPS listener references the cert ARN — AWS may refuse to create the listener while the cert is unvalidated. Worth verifying behavior on a clean apply before relying on it.

Option A is strongly preferred; it makes the apply fully self-validating.

---

### 🟡 High: WAF managed rules in BLOCK mode without a count-mode trial

`waf.tf:46-132` runs all four AWS managed rule groups (Core, KnownBadInputs, IPReputation, SQLi) with `override_action { none {} }` — meaning the rules' default actions (block) fire as-is. Two concrete false-positive risks for us:

- **CoreRuleSet** flags non-ASCII characters in request bodies and a long list of other patterns. Our LLM endpoints accept free-form prompts; international users or special-character prompts could trip it.
- **SQLiRuleSet** flags SQL-like patterns. A user asking the Strategy Agent "show me top strategies by sharpe ratio" or pasting a paper title that contains the word "select" or "where" could match.

**Recommendation:** ship in COUNT mode for the first 24–48h, watch CloudWatch metrics for actual hits on our traffic shape, then flip to BLOCK once we know the false-positive rate is acceptable. Change is one line per rule:

```hcl
override_action {
  count {}  # instead of none {} — count but don't block
}
```

Not strictly a blocker (the rules can be flipped post-apply), but the failure mode if we get this wrong is "every WAF block looks like a backend bug on the demo".

---

### 🟡 High: ALB access-logs S3 bucket missing standard hardening

`alb.tf:14-55` creates the access-logs bucket without the three hygiene resources we already have on the tfstate bucket per `infra/README.md`:

- No `aws_s3_bucket_server_side_encryption_configuration` (the tfstate bucket has SSE-AES256)
- No `aws_s3_bucket_public_access_block` (the tfstate bucket has all 4 booleans set)
- No deny-non-TLS statement in the bucket policy (the tfstate bucket has this)

These are 1-resource-each fixes. Worth landing them in this PR rather than a follow-up so the access-logs bucket starts compliant with the rest of our state.

---

### 🟡 High: ALB hardening flags

Two recommended ALB-level settings missing on `aws_lb.main`:

```hcl
drop_invalid_header_fields = true   # drops malformed headers (security)
enable_deletion_protection = true   # prevents accidental deletion
```

`drop_invalid_header_fields` is specifically called out in AWS Well-Architected security guidance. `enable_deletion_protection` is operational hygiene — once this is the only public ingress, accidental destroy would take the whole site down.

---

### 🟢 Medium: ALB SG egress wide open

`alb.tf:101-106` allows egress to `0.0.0.0/0`. The ALB only needs to talk to the EC2 target SG. Tighten to:

```hcl
egress {
  description     = "to backend"
  from_port       = 80
  to_port         = 80
  protocol        = "tcp"
  security_groups = [aws_security_group.archimedes.id]
}
```

Same critique I raised on the Aurora/Redis SGs in #416. Defense in depth.

---

### 🟢 Medium: WAF rate limit might be tight under demo conditions

1000 req / 5 min / IP = ~3.3 req/s per IP sustained. For a single user page-loading the app (frontend pulls bundle + several `/api/*` calls + WS), they'd burn ~30 reqs in a few seconds, which is fine. But during a demo with multiple judges behind one corporate NAT/CGNAT, this could trip. Not a code bug, but worth dry-running before going live.

---

### 🔵 Low / Nice-to-have

- `aws_acm_certificate.main` has no `subject_alternative_names`. If we ever want `www.archimedes-arc.app` it'll need a new cert. Adding `subject_alternative_names = ["www.archimedes-arc.app"]` now is cheap.
- No output for `aws_s3_bucket.alb_logs.id` — handy when wiring observability.
- `aws_lb_listener.https` could explicitly set `alpn_policy = "HTTP2Preferred"` — currently default behavior is fine but explicit is better for posterity.

---

### What I verified is right

- ✅ ALB `idle_timeout = 300` matches the SSE survival spec
- ✅ TLS policy `ELBSecurityPolicy-TLS13-1-2-2021-06` is current AWS-recommended
- ✅ HTTP→HTTPS redirect listener on port 80 — correct pattern
- ✅ Multi-AZ deployment via `aws_subnet.public[*].id`
- ✅ Target attached to existing EC2 in default VPC via VPC peering set up in #416
- ✅ WAF rule priorities + the explicit no-BotControl, no-geo-block choices match the locked spec
- ✅ S3 lifecycle 30-day expiration matches the spec
- ✅ All 4 CI checks green
- ✅ `default_action { allow {} }` on the WAF web ACL (deny-by-rule, not deny-by-default — correct for our use case)

---

### Recommendation

**Kick back to Pi for the two Critical fixes** (health-check port + nginx redirect-loop awareness) and the ACM validation gap. The S3 hardening + ALB hardening flags + WAF count-mode are also worth landing in the same PR rather than as follow-ups, since they're one-line changes each and we'd just be filing immediate-issue debt otherwise.

I'm explicitly not requesting changes via the GitHub review system (so this doesn't block the merge button) — these need a judgment call from you and Pi on whether to fix in-PR or split.

cc @dbrowneup

dbrowneup (approved):
## Re-review — rework `c4105b5`

All seven findings addressed correctly. Walking each, then one minor observation and a merge verdict.

### 🔴 Critical 1: Health check port — ✅ fixed

`alb.tf:212`: `port = "traffic-port"` + `matcher = "200"`. The ALB now health-checks on the same port traffic uses (80), so health probes hit nginx's port-8080 listener → `location /health` → proxy to backend → 200. Verified path: ALB:443 → EC2:80 (host) → docker maps to nginx container 8080 → nginx routes `/health` → backend:8000 → 200.

### 🔴 Critical 2: nginx redirect loop — ✅ fixed

`nginx/nginx.conf:23-31`: server-level `set $do_redirect 1` defaults to redirect; `if ($http_x_forwarded_proto = "https")` and `if ($scheme = "https")` clear the flag. `location /:34-36` only 301s when `$do_redirect = 1`. Verified the four legit-via-ALB paths all work without loop:

- `/` with X-F-P=https → `$do_redirect=0` → serves SPA from `/usr/share/nginx/html` (line 39-41)
- `/api/`, `/api/generate/stream/`, `/health` → match more-specific locations (lines 45/58/71) which proxy unconditionally, bypassing the redirect logic entirely

The `if` usage stays within the nginx-safe pattern (`return` is one of the two directives explicitly listed as safe inside `location` per the official "if is evil" guidance). The server-level `set` inside `if` pattern is also fine in modern nginx; it's the standard idiom for multi-condition variable setting.

### 🟠 ACM cert validation — ✅ fixed

`alb.tf:108-135`: `data "aws_route53_zone" "main"` with `private_zone = false` ensures it finds the public zone; `for_each` over `domain_validation_options` generates the CNAME records; `aws_acm_certificate_validation` blocks until ACM confirms. Apply is now fully self-validating.

Bonus: `allow_overwrite = true` (line 129) is a nice defensive touch in case validation records pre-exist from an earlier manual attempt.

### 🟡 ALB logs bucket parity — ✅ fixed

- SSE-AES256: `aws_s3_bucket_server_side_encryption_configuration.alb_logs` (lines 22-30) ✅
- Public access block: all 4 booleans set on `aws_s3_bucket_public_access_block.alb_logs` (lines 32-39) ✅
- DenyNonTLS: added as second statement in bucket policy (lines 72-84), correctly applies to both bucket and objects, condition `aws:SecureTransport: "false"` ✅

### 🟡 ALB hardening — ✅ fixed

`alb.tf:187-188`: `drop_invalid_header_fields = true` + `enable_deletion_protection = true`. Both correct.

### 🟡 ALB SG egress — ✅ fixed

`alb.tf:162-169`: tightened to `security_groups = [aws_security_group.archimedes.id]` on port 80 only. Clean.

### 🟡 WAF COUNT mode — ✅ fixed (with reasonable deviation)

- Core (10): `count {}` ✅
- Known Bad Inputs (20): `count {}` ✅ (I'd flagged only Core+SQLi; Pi's added Known Bad Inputs to count too — defensible, that rule group also has some pattern-match false positives)
- IP Reputation (30): `none {}` (BLOCK) — Pi's judgment call, **I agree.** IP reputation is AWS threat-intel-driven, not pattern-matching; false-positives are rare and the action is targeted at well-known bad IPs.
- SQLi (40): `count {}` ✅
- Rate limit (1): `action { block {} }` ✅ (this is its own rule action, not an override — correct)

The comment on the Core rule (lines 54-57) explicitly notes the flip-to-`none {}` plan after the 24-48h observation window. Good operational hygiene.

---

### 🟢 Minor observation — not blocking, worth noting

The HTTP server block (port 8080) now proxies `/api/`, `/api/generate/stream/`, `/health` unconditionally, regardless of `$do_redirect`. This is correct for ALB-forwarded traffic (X-F-P=https → no redirect needed), and it's correct for the health-check path (no X-F-P header → proxies normally). But during the migration window — between ALB cutover (Route 53 → ALB) and PR 6 (EC2 → private subnet) — the EC2 instance still has a public IP, so anyone hitting `http://<ec2-public-ip>/api/...` directly would get a proxied response without TLS.

Concrete severity: low. The risk is bounded by (a) needing the EC2 IP, which is harder once DNS points at the ALB but still discoverable from DNS history; (b) the migration window being short (single sprint to PR 6). The old config 301'd `/api/` on HTTP, so this is technically a regression for that narrow window.

If Pi wants to close it pre-merge, add the same guard inside each proxy location:

```
location /api/ {
    if ($do_redirect = 1) {
        return 301 https://$host$request_uri;
    }
    # ... existing proxy block ...
}
```

Otherwise, accept it as a transient and prioritize PR 6. **I'd take either path.**

---

### Verdict: ✅ approve for merge

All four CI gates green, every flagged finding fixed correctly, no new issues introduced beyond the minor migration-window note above. After merge: terraform apply, ACM validation (now automatic), DNS cutover, then PR 6 to close the loop.

PR #424, opened 2026-05-27
[infra] Fix SSM deploy: git safe.directory for root execution

t2o2:
SSM RunShellScript runs as root. `/opt/archimedes` is owned by ubuntu.
Git 2.35.2+ rejects cross-user repo operations with 'dubious ownership'.

One-line fix: `git config --global --add safe.directory /opt/archimedes`

Evidence from deploy log:
```
fatal: detected dubious ownership in repository at '/opt/archimedes'
```

This is the third SSM deploy fix (after IAM document ARN + bash shebang).
The root cause of all three is that SSM's execution environment differs
from SSH's in ways that aren't visible until runtime. Debt item #417.2
(codify IAM + test deploy in staging) would prevent this class of bug.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -85,6 +85,7 @@ jobs:
               "#!/bin/bash",
               "set -eux",
               "cd /opt/archimedes",
+              "git config --global --add safe.directory /opt/archimedes",
               "mkdir -p analytics-engine/artifacts",
               "sudo chown -R ubuntu:ubuntu analytics-engine/artifacts/ 2>/dev/null || true",
               "git fetch origin main",


dbrowneup (approved):
## Review — approve, ship it

Verified. The error message from the failed deploy is *literally* the AWS-suggested remediation:

```
fatal: detected dubious ownership in repository at '/opt/archimedes'
To add an exception for this directory, call:
    git config --global --add safe.directory /opt/archimedes
```

Pi's fix adds exactly that line before any git operation in the script. Diff is one line. CI green. Approving.

### Validating progress

The previous failure ran 4 commands successfully before the git error:
```
+ cd /opt/archimedes
+ mkdir -p analytics-engine/artifacts
+ sudo chown -R ubuntu:ubuntu analytics-engine/artifacts/
+ git fetch origin main      ← failed here
```

The `+` trace confirms `set -eux` is executing — which means the bash/shell-compat fix from #422 IS working. The shebang either worked (and we're in bash) or didn't matter (we're in dash with no `pipefail`). Either way, we're past the shell issue. This is the third (and hopefully last) SSM-vs-SSH environment difference to surface.

### One observation for future hardening (not blocking)

The deploy script now runs as **root** under SSM (vs `ubuntu` under SSH). Files touched by the script — `.env`, `analytics-engine/artifacts/*`, anything created during `docker compose build` — end up root-owned where they were ubuntu-owned before. Subtle implications:

- `.env` becomes root-owned after `sed -i` / `echo >> .env`. If anyone SSHes in as ubuntu later and tries to edit it manually, they'd hit permission denied.
- The script already has `sudo chown -R ubuntu:ubuntu analytics-engine/artifacts/` for the artifacts dir — that's good. Worth considering whether `.env` deserves the same treatment.

Long-term fix would be to make SSM run the deploy as ubuntu (via the `RunAs` parameter on RunShellScript, requires SSM agent config), or move secrets out of `.env` entirely (per #417 debt item 1). Not blocking.

Pi's framing in the PR body is on-point: this is the **third** SSM deploy bug after IAM document ARN + bash shebang, all from SSM-vs-SSH environment differences that weren't visible until runtime. The pattern argues strongly for a staging deploy target before any future deploy-pipeline changes — would have caught all three issues in one test run.

Approving — Dan, merge when you see this.

Issue #425, opened 2026-05-27
APIN - Infra - Staging deploy target to prevent SSM-vs-SSH runtime regressions


dbrowneup:
## Summary

Establish a staging deploy target that exercises the full `deploy.yml` pipeline before any change to it can merge to `main`. Catalyzed by **four consecutive production deploy failures** today (2026-05-27), each surfacing a different SSM-vs-SSH environment difference that wasn't visible until production runtime:

1. **#417**: missing `ssm:SendCommand` document ARN in the IAM policy
2. **#422**: `set -o pipefail` failed because SSM's default shell on this instance is dash, not bash
3. **#424**: `git fetch` failed with "dubious ownership" because SSM runs the script as `root`, not `ubuntu`
4. **(current)**: `git config --global --add safe.directory` failed with `$HOME not set` because SSM's root-context doesn't inherit a login shell's env

Aggregate cost: ~7 hours of frozen production deploys, with each fix-cycle averaging 30-60 min wall-clock (CI run + PR review + merge + redeploy retry). The cost would have been near-zero with a staging target where Pi could iterate cheaply.

## Why this matters now

- We have several more deploy-touching items in the pipeline: ALB cutover (per PR #419), EC2 → private subnet (PR 6), Aurora/ElastiCache env-var cutover, future `RunAs ubuntu` change to fix the file-ownership regression I flagged in PR #424. Each is another opportunity for the same class of failure.
- The pattern is the dual of the [#421](https://github.com/a-apin/archimedes/issues/421) / [#417 debt item 2](https://github.com/a-apin/archimedes/blob/main/.github/workflows/deploy.yml#L30) ("codify IAM in Terraform") argument: *both* the policy AND the runtime environment need to be testable before merge.

## Options (cost vs realism tradeoff)

| Option | Cost | Catches what | Catches what NOT |
|---|---|---|---|
| **A. Second EC2 mirroring prod** (same AMI, same VPC, separate Aurora/Redis, staging.archimedes-arc.app subdomain) | ~$25-50/mo | Everything we hit today (IAM, shell, ownership, env) | Hardware-specific edge cases (rare) |
| **B. Dev EC2 (existing or new) via SSM RunCommand** | $0–$10/mo (reuse a hackathon-dev box; or a t3.micro spot) | Most SSM env differences | DNS/cert specifics; ALB integration |
[...]

PR #426, opened 2026-05-27
[infra] Fix SSM deploy #4: export HOME=/root

t2o2:
Bug #4 in the SSH → SSM series: `git config --global` fails with `fatal: $HOME not set`.

SSM runs scripts in a non-login context without HOME. One-line fix sets it at script top.

Long-term: `sudo -u ubuntu -i` wrapping or SSM RunAs (per #425 staging deploy target).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -84,6 +84,7 @@ jobs:
             --parameters commands='[
               "#!/bin/bash",
               "set -eux",
+              "export HOME=/root",
               "cd /opt/archimedes",
               "git config --global --add safe.directory /opt/archimedes",
               "mkdir -p analytics-engine/artifacts",


dbrowneup (approved):
## Review — approve, ship it

Verified — exactly the proposed one-line fix in the exact placement we discussed. `export HOME=/root` lands after `set -eux` so any subsequent HOME-sensitive tool (docker creds, pip caches, etc.) is also covered.

The previous failure showed the trace getting to `git config --global --add safe.directory /opt/archimedes` before failing on `$HOME not set`. With this fix, that command now writes to `/root/.gitconfig` and the subsequent git operations should honor it.

CI green on all 4 gates. Approving. Dan can merge whenever.

### Reminder of the larger refactor (not in this PR)

This is the fourth tactical fix to the SSM-vs-SSH environment-difference problem. The durable answer is one of the two alternatives I flagged on the #424 PR — either `sudo -i -u ubuntu bash -c "..."` wrapping or SSM `RunAs` parameter — both of which would match the SSH model exactly (HOME, USER, LOGNAME, file ownership, etc. all set correctly). I'll be filing a follow-up issue for that refactor since it's larger than a one-shot PR (it interacts with secrets passing and would benefit from review of the approach before implementation).

Also see #425 (just filed) — staging deploy target as the structural fix to "no more bug #5/#6/#N at production runtime."

PR #429, opened 2026-05-27
[security] Drop X-Wallet-Address fallback for profile writes (#402)

### LINKED ISSUE #402: [security] Require SIWE session for all write operations — drop X-Wallet-Address header fallback
## Summary

Non-PII profile/chat **write** operations still accept a forgeable
\`X-Wallet-Address\` header as a fallback when no SIWE session is present
(\`_extract_caller_wallet\` in \`backend/archimedes/api/user_routes.py\`). An
attacker can set both the header and the payload \`wallet_address\` to a
victim's address and write to it. Documented in
[\`docs/security/auth-model.md\`](docs/security/auth-model.md).

**Blast radius is limited** — non-PII fields (display_name, interests,
attribution, marketing_opt_in) and user chat messages only. No funds, no PII
reads, no reasoning traces (those are already SIWE-only / agent-key-gated).

## Why it's not already fixed

The live frontend lets a MetaMask user write (e.g. save a display name) before
completing a SIWE signature. Removing the header fallback outright breaks that
path until the frontend requires SIWE login first. So this is a **coordinated
frontend + backend change**, not a one-line backend edit.

## Scope

**Backend:**
- [ ] \`_extract_caller_wallet\` returns only \`get_verified_wallet(request)\` (drop the header fallback).
- [ ] All write endpoints (profile upsert, chat post) return 401 when there is no SIWE session.

**Frontend:**
- [ ] Prompt SIWE sign-in (challenge → sign → \`/api/auth/siwe/verify\`) before the first write; reuse the session cookie thereafter.

## Acceptance criteria

- [ ] A profile upsert with a forged \`X-Wallet-Address\` header and no SIWE cookie returns **401** (add a test mirroring \`tests/test_auth_guard.py\`).
- [ ] A logged-in (SIWE) user can still upsert their own profile.
- [ ] \`grep -n "X-Wallet-Address" backend/archimedes/api/user_routes.py\` shows no write-path fallback remaining.

## Anti-goals

- Don't touch the PII-read path (\`_extract_caller_wallet_siwe\`) — already correct.
- Don't change the internal-agent-key gating on traces / chat events / bootstrap — already correct.

## Source

Audit follow-up, 2026-05-27. Companion to PR #401 (the documentation).

dbrowneup:
## Summary

Closes #402. Profile writes (POST /api/user/profile) no longer accept the
forgeable X-Wallet-Address header as an auth fallback. SIWE session is now
required for the write — the same cryptographic-proof bar that already
gates PII reads. The boundary was the same shape on both sides of the
read/write split before; now it's actually enforced symmetrically.

## What changes

**Backend** (`backend/archimedes/api/user_routes.py`):
- Removed `_extract_caller_wallet` (the function that fell through to the header).
- `upsert_profile` now uses `_extract_caller_wallet_siwe` and returns two distinct 403s — "SIWE session required for profile writes" (no session present) and "SIWE session wallet does not match payload wallet" (session-vs-payload mismatch). Clearer signals for the frontend, simpler test assertions.

**Frontend** (`ui/src/components/WelcomeProfileModal.jsx`):
- Added `credentials: 'include'` to the POST so the SIWE session cookie is actually sent. Previously the welcome flow relied on the X-Wallet-Address header which is no longer trusted.
- Removed the `'X-Wallet-Address'` header from that fetch since the backend ignores it.

**Tests** (`backend/tests/test_user_routes.py`, `backend/tests/test_user_profile_privacy.py`):
- Every profile-write test was switched from `headers={"X-Wallet-Address": …}` to `cookies=_siwe_cookies(…)` via the existing helper.
- Two new tests lock in the #402 behavior:
  - `test_write_with_header_only_fails` — POST with header but no SIWE → 403 with "SIWE session required" in the detail.
  - `test_write_with_wrong_siwe_session_fails` — POST with mismatched-wallet SIWE session → 403 with "does not match" in the detail.
- The `test_upsert_encrypts_email_before_storage` integration test was updated to patch `archimedes.api.auth_siwe.get_verified_wallet` instead of setting a mock header (since the header is no longer the auth source).
- Module docstring updated to describe the new auth model.

## Out of scope (deferred to follow-up cleanup)

- `X-Wallet-Address` is still in the CORS `allow_headers` list (`main.py:113-119`) and still sent by `Layout.jsx:71-74` on the profile GET. Both are dead config now (backend ignores the header on both reads and writes), but removing them is unrelated to the security boundary closure. Worth a small cleanup PR — flagging here so it's not forgotten.

## Test results

```
backend/tests/test_user_routes.py            18 passed
backend/tests/test_user_profile_privacy.py   12 passed
backend/tests/test_security_hardening.py     20 passed
```

Hermetic — runs with `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest` (no .env, no docker, no network).

## Verify

```bash
# Header-only POST is rejected
curl -X POST https://archimedes-arc.app/api/user/profile \
  -H 'Content-Type: application/json' \
  -H 'X-Wallet-Address: 0x...' \
  -d '{"wallet_address": "0x...", "display_name": "spoofed"}'
# → 403 {"detail":"Forbidden: SIWE session required for profile writes"}

# SIWE-session POST succeeds (with valid cookie)
curl -X POST https://archimedes-arc.app/api/user/profile \
  -H 'Content-Type: application/json' \
  --cookie 'archimedes_session=<valid-signed-cookie>' \
  -d '{"wallet_address": "0x<matching-wallet>", "display_name": "ok"}'
# → 200 {"wallet_address": ..., "display_name": "ok", ...}
```

### DIFF
--- a/backend/archimedes/api/user_routes.py
+++ b/backend/archimedes/api/user_routes.py
@@ -76,23 +76,6 @@ def _extract_caller_wallet_siwe(request: Request) -> str | None:
     return get_verified_wallet(request)
 
 
-def _extract_caller_wallet(request: Request) -> str | None:
-    """Extract wallet from SIWE session or X-Wallet-Address header.
-
-    Used for write operations (profile upsert, chat posting) where the
-    caller is modifying their OWN data. The header is acceptable here
-    because the write check enforces caller == payload.wallet_address.
-
-    PII reads use _extract_caller_wallet_siwe() instead.
-    """
-    from archimedes.api.auth_siwe import get_verified_wallet
-
-    verified = get_verified_wallet(request)
-    if verified:
-        return verified
-    return request.headers.get("X-Wallet-Address", "").lower().strip() or None
-
-
 @user_router.get("/profile/{wallet}", response_model=UserProfileResponse)
 async def get_profile(wallet: str, request: Request):
     """Retrieve a wallet's profile. Returns 404 if not set.
@@ -128,14 +111,17 @@ async def get_profile(wallet: str, request: Request):
 async def upsert_profile(payload: UserProfileCreate, request: Request, response: Response):  # noqa: ARG001 — response param threaded for downstream cookie/header setting; not used in current body
     """Create or update a wallet's profile. All fields optional except wallet.
 
-    Email is encrypted at rest before storage. Caller must supply an
-    `X-Wallet-Address` header matching `payload.wallet_address` — prevents one
-    wallet from writing another wallet's profile.
+    Email is encrypted at rest before storage. Caller must hold a valid SIWE
+    session matching `payload.wallet_address` — prevents one wallet from
+    writing another wallet's profile. The X-Wallet-Address header fallback
+    was dropped per Issue #402 because it is forgeable.
     """
-    caller = _extract_caller_wallet(request)
+    caller = _extract_caller_wallet_siwe(request)
+    if caller is None:
+        raise HTTPException(status_code=403, detail="Forbidden: SIWE session required for profile writes")
     wallet = payload.wallet_address.lower()
     if caller != wallet:
-        raise HTTPException(status_code=403, detail="Forbidden: X-Wallet-Address header must match payload wallet")
+        raise HTTPException(status_code=403, detail="Forbidden: SIWE session wallet does not match payload wallet")
 
     session: Session = get_session()
     try:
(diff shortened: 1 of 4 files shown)


t2o2 (approved):
Reviewed. Clean removal of the header fallback — profile writes now strictly require SIWE session. Tests cover all rejection paths (no auth, header-only, wrong wallet session). Frontend credentials: 'include' is correct. Error messages are clear and actionable. CI green on 5 checks. Merging.

Issue #408, opened 2026-05-27
APIN - Tests - Comprehensive hermetic coverage for on-chain execute_* methods (HexBytes, reverts, timeouts, concurrent)


dbrowneup:
## Summary

Critical on-chain methods in `backend/archimedes/chain/executor.py` (`execute_trades`, `_confirm_receipt`, `create_vault`, `set_token_oracles`, `set_oracle_price`) each have at-most-a-few hermetic unit tests today. More importantly, **the production raw-key signing path uses `HexBytes` inputs while every existing test uses `str` inputs** — meaning the actual code path that runs against live funds is currently uncovered.

This is a **focused depth-coverage** issue, distinct from the broader breadth-coverage in [#405](https://github.com/a-apin/archimedes/issues/405). Where #405 targets ≥85% line coverage across 5 modules, this issue targets *scenario coverage* on the on-chain `execute_*` methods specifically — the surface area where a missed edge case can cost real funds.

## Why now (after-hackathon)

Surfaced by review of PR #403 (Önder, [`chain/executor.py`](backend/archimedes/chain/executor.py)). The `_confirm_receipt` helper added there handles both `str` and `bytes` (HexBytes) inputs — but the two new tests only cover the `str` branch. The `bytes` branch is what the raw-key signer actually emits in production. A reviewer can't tell from CI alone that the production path is exercised.

## Scope

Add hermetic unit tests covering the following scenarios across the five `execute_*` methods. **One PR per method** — no bundling.

### Per-method scenario matrix (all must be tested hermetically)

| Method | str input | HexBytes input | revert (status=0) | timeout | rate-limit error | network failure |
|--------|-----------|----------------|-------------------|---------|------------------|-----------------|
| `execute_trades` | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| `_confirm_receipt` | ✓ | ✓ | ✓ | ✓ | — | ✓ |
| `create_vault` | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| `set_token_oracles` | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |
| `set_oracle_price` | ✓ | ✓ | ✓ | ✓ | ✓ | ✓ |

Also include for `execute_trades` specifically:
- Empty `trades` list → returns early without RPC calls
- All-USDC legs → filtered out (per #400 fix), no `getPool` calls fire
- Mixed USDC + synth legs → synth leg evaluated, USDC leg filtered
[...]

PR #430, opened 2026-05-27
[tests] execute_trades depth coverage: HexBytes + revert + timeout + network (#408 sub-task 1)

dbrowneup:
## Summary

First of five #408 sub-task PRs — depth-coverage on `ChainExecutor.execute_trades`. Adds 11 new tests in a new `TestExecuteTradesDepth` class plus the two test-infrastructure helpers needed to make the raw-key signer path testable. The remaining four sub-tasks (`_confirm_receipt`, `create_vault`, `set_token_oracles`, `set_target_allocations`) will follow as separate PRs.

## What's covered

| Scenario | str input | HexBytes input | revert (status=0) | timeout | rate-limit | network failure |
|---|---|---|---|---|---|---|
| execute_trades Circle path | ✅ (existing) | n/a (Circle returns str) | ✅ | n/a in Circle layer | ✅ | ✅ |
| execute_trades raw-key path | n/a (raw-key returns HexBytes) | ✅ **NEW** | ✅ | ✅ | n/a (RPC not rate-limited) | ✅ |

Plus the four execute_trades-specific scenarios from the #408 issue:
- ✅ Empty `trades` list → no `getPool` calls (`test_empty_trades_no_amm_calls`)
- ✅ All-USDC legs → filtered out, no AMM calls (`test_all_usdc_legs_filtered_no_pool_calls`)
- ✅ Mixed USDC + synth → only synth pool-checked (`test_mixed_usdc_and_synth_only_synth_validated`)
- ✅ Both Circle and raw-key signer paths exercised

Bonus: also closes the existing SELL-direction coverage gap (lines 216-223 of executor.py), which was uncovered because all prior tests used BUY direction (`test_sell_direction_calls_usdc_value_to_token_raw`).

## The critical new test

`test_raw_key_path_succeeds_with_hexbytes_tx_hash` is the test #408 was filed for. The production raw-key path emits `HexBytes` from `chain_client.w3.eth.send_raw_transaction`, and that's what reaches `_confirm_receipt`. Before this PR, every existing test for `execute_trades` used the Circle path (returns `str`). This test asserts:

```python
mock_confirm.assert_called_once()
(call_arg,) = mock_confirm.call_args.args
assert isinstance(call_arg, (bytes, bytearray))  # not str
```

The production path that runs against live funds is now under regression guard.

## Test infrastructure: why the new helpers exist

The existing `executor` fixture sets `chain_client.w3.eth.gas_price = AsyncMock(return_value=...)`. The production code does `await chain_client.w3.eth.gas_price` (property-access pattern, not method call). `AsyncMock` instances are only awaitable when *called*, not on attribute access — so the existing fixture happens to work for tests that never exercise the raw-key path, but the moment a test does reach line 255 in `executor.py`, the mock fails with `TypeError: object AsyncMock can't be used in 'await' expression`.

Rather than modify the shared fixture (which would touch every other test in the module), this PR adds two per-test helpers:

- `_Awaitable(value)` — a tiny wrapper class with `__await__` so `await _Awaitable(x)` returns `x`. Re-awaitable (creates a fresh coroutine on each `__await__`).
- `_install_raw_key_eth(executor, tx_hash)` — replaces `executor._mock_cc.w3.eth` with a real `types.SimpleNamespace` whose methods are actual `async def` functions and whose `gas_price` is `_Awaitable(...)`. Returns the namespace so individual tests can override (e.g. for the network-failure test where `send_raw_transaction` should raise).
- `_install_raw_key_vault(mock_loader, mock_account)` — same pattern for the `vault.functions.rebalance(...).build_transaction(...)` chain, which has the same `await`-attribute issue.

These bypass the fixture's broken mocks for the raw-key tests while leaving the existing Circle-path tests untouched.

## Test results

```
backend/tests/chain/test_chain_executor.py      27 passed
backend/tests/chain/test_executor_receipt.py     5 passed (existing, unchanged)
```

Hermetic — `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain/test_chain_executor.py -q` ends with `27 passed, 0 failed`. No `asyncio.get_event_loop()` (verified clean via grep). No production code changes.

## Verify

```bash
gh pr checkout 430  # or whatever this PR's number ends up
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain/test_chain_executor.py -v
grep -E "HexBytes|bytes\.fromhex" backend/tests/chain/test_chain_executor.py  # must show ≥1
grep -E "asyncio\.get_event_loop" backend/tests/chain/test_chain_executor.py  # must be empty
```

## Out of scope (follows in separate PRs)

- `_confirm_receipt` (mostly covered by `test_executor_receipt.py` already; small gap-fill PR may follow)
- `create_vault` — sub-task 2 PR
- `set_token_oracles` — sub-task 3 PR
- `set_target_allocations` — sub-task 4 PR (note: #408 issue says `set_oracle_price` but the production method is actually `set_target_allocations`; will flag in that sub-task's PR)

The ≥85% executor.py coverage gate from #408 lands once all five sub-tasks are in.

### DIFF
--- a/backend/tests/chain/test_chain_executor.py
+++ b/backend/tests/chain/test_chain_executor.py
@@ -12,14 +12,14 @@ import asyncio
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
 import pytest
-
 from archimedes.chain.executor import (
     MIN_HEALTHY_LIQUIDITY_USDC,
     ChainExecutor,
     InsufficientLiquidityError,
+    TradeRevertedError,
 )
 from archimedes.models.portfolio import TradeDirection, TradeOrder
-
+from hexbytes import HexBytes
 
 # ── Fixtures ──────────────────────────────────────────────────
 
@@ -178,11 +178,15 @@ class TestExecuteTrades:
         """execute_trades propagates InsufficientLiquidityError."""
         trade = _make_trade()
 
-        with patch.object(
-            executor, "_validate_trade_liquidity", new=AsyncMock(side_effect=InsufficientLiquidityError("thin pool"))
+        with (
+            patch.object(
+                executor,
+                "_validate_trade_liquidity",
+                new=AsyncMock(side_effect=InsufficientLiquidityError("thin pool")),
+            ),
+            pytest.raises(InsufficientLiquidityError),
         ):
-            with pytest.raises(InsufficientLiquidityError):
-                asyncio.run(executor.execute_trades("0xvault", [trade]))
+            asyncio.run(executor.execute_trades("0xvault", [trade]))
 
     def test_circle_signer_path(self, executor, mock_loader):
         """When Circle signer is configured, uses circle_signer.execute_contract.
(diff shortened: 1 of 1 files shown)


t2o2 (approved):
Reviewed. The _Awaitable helper for gas_price and _install_raw_key_eth/vault setup functions are well-designed workarounds for the AsyncMock-on-property issue. HexBytes test exercises the actual production code path. SELL direction coverage gap closed. CI green on 5 checks. Merging.

Note: the shared executor fixture's gas_price mock issue is a latent bug worth a follow-up (per PR description). Not blocking.

PR #435, opened 2026-05-28
[infra] Fix terraform apply: ASCII descriptions + cross-VPC routing + IP targets

t2o2:
Fixes discovered during `terraform apply`. All infrastructure is now live.

## Fixes
1. Em-dashes → ASCII dashes in SG descriptions (AWS rejects non-ASCII)
2. Target group: instance-based → IP-based for cross-VPC ALB→EC2
3. Public route table: added peering route to 172.31.0.0/16

## Verified live
- Site: `curl https://archimedes-arc.app/api/health` → `ok`
- ALB target: healthy
- ACM cert: ISSUED
- WAF: attached

### DIFF
--- a/infra/alb.tf
+++ b/infra/alb.tf
@@ -138,7 +138,7 @@ resource "aws_acm_certificate_validation" "main" {
 
 resource "aws_security_group" "alb" {
   name        = "${var.project_name}-alb-sg"
-  description = "ALB — public HTTPS only"
+  description = "ALB - public HTTPS only"
   vpc_id      = aws_vpc.main.id
 
   # HTTPS from internet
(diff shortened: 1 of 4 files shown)


dbrowneup (approved):
## Review — approve, merge

All three fixes are real apply-time issues with correct minimal fixes:

1. **Em-dashes → ASCII in SG descriptions** — AWS API rejects non-ASCII in SG `description` fields; my prior PRs (the ones that introduced em-dashes) didn't surface this because they weren't applied at the time. Trivially correct.
2. **Target group IP-based + cross-VPC AZ** — standard pattern for cross-VPC ALB→target reachability via peering. `instance` target type requires same-VPC; `ip` works across the peering boundary. The `availability_zone = "all"` is the AWS-required incantation for IP targets that don't fall under the ALB's own AZs. Correct.
3. **Public route table peering entry** — required return path for ALB→EC2 traffic. Without it, packets reach EC2 but the EC2's response packets to ALB's private IP have no route back.

No file overlap with #434 (which touched test files only); the just-merged #434 doesn't create a conflict. CI green on all 4 gates. The "all infrastructure is now live" verification (`/api/health` returns `ok` through the ALB, ACM ISSUED, WAF attached) confirms the fixes work end-to-end.

Approving + merging on Dan's behalf since Pi is offline (Chuan commuting with the laptop).

PR #475, opened 2026-06-06
[backtest] Seed only latest artifact per strategy with fallback (#462)

### LINKED ISSUE #462: APIN - Artifacts - ./analytics-engine/artifacts
https://github.com/a-apin/archimedes/tree/main/analytics-engine/artifacts produces static assets that are loaded hardcoded. Should be result of a dynamic pattern to a single resulting json of the current state and a backup solution if broken.

onder-akkaya:
Closes #462 ("APIN - Artifacts - ./analytics-engine/artifacts").

## Problem
`analytics-engine/artifacts/` accumulates one timestamped JSON per backtest re-run, so a single strategy can have several files (today: 3 for "Buy-and-Hold Baseline", 2 for "Tactical Asset Allocation"). `seed_from_artifacts` globbed **all** of them and seeded each, persisting stale historical runs alongside the current one — there was no notion of "current state," and a single corrupt file just raised and got logged per-file.

## Fix
Make seeding resolve to a **single current-state run per strategy**, with a backup:
- Group artifact files by `paper_title`.
- `run_id` is a sortable ISO-8601 UTC timestamp (`20260518T223800Z`), so a reverse sort gives newest-first.
- Seed the **newest** artifact that parses + maps cleanly; if the newest is broken, **fall back** to the next-newest. A single corrupt artifact never silently drops a strategy's backtest.
- Unreadable JSON is skipped (not fatal); a strategy whose runs all fail is counted once as `failed`.

This matches the issue's ask: "a dynamic pattern to a single resulting json of the current state and a backup solution if broken."

## Tests
New hermetic suite `backend/tests/scripts/test_seed_backtests_from_artifacts.py` (6 tests, all DB/provider/mapper boundaries mocked — no Postgres/Redis/analytics-engine):
- seeds only the latest run per strategy
- falls back to an older run when the newest is broken
- counts `failed` once when all runs are broken
- unmatched strategy counted once as failed
- unreadable JSON skipped, not fatal
- `was_inserted=False` → `skipped`, not `failed`

```
python -m pytest backend/tests/scripts/test_seed_backtests_from_artifacts.py -q   # → 6 passed
```

## Anti-goals
- Doesn't change the artifact file format or the analytics-engine writer; doesn't touch contracts/infra. Pure seeding-selection logic + tests.

### DIFF
--- a/backend/archimedes/scripts/seed_backtests_from_artifacts.py
+++ b/backend/archimedes/scripts/seed_backtests_from_artifacts.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 
 import json
 import logging
+from collections import defaultdict
 from pathlib import Path
 
 from archimedes.db import get_session, init_db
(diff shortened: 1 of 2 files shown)


onder-akkaya:
### Re: the Quality Gate / Complexity bot comments — not introduced by this PR ✅

I checked the two informational (`continue-on-error`) findings; both are **pre-existing repo-wide debt**, not caused by this change. The hard-block gates all passed.

**ruff — "45 errors":** That number comes from the informational `ruff check .` job, which runs the broader ruleset across the **whole repo**. This PR's two files pass the full ruleset cleanly:
```
$ ruff check backend/archimedes/scripts/seed_backtests_from_artifacts.py \
             backend/tests/scripts/test_seed_backtests_from_artifacts.py
All checks passed!   # exit 0
```
This PR's diff is only those two backend files (`git diff --name-only main...HEAD`), so none of the 45 live in changed code — they're in files this PR never touches. The **hard-block** `Ruff — format + critical lint rules` gate (`ruff format --check` + `E9,F63,F7,F40`) is ✅.

**eslint — "1 error":** It's in the frontend (`ui/`). This PR touches **zero** frontend files, so it's pre-existing.

**Complexity:** Actually *improved* — Δ CC **−2.9** ✅. The "11 orphan functions ⚠️" are the pytest test functions + helpers in the new test file; static analysis flags test functions as orphans because nothing calls them in-repo (pytest invokes them at runtime). Expected for a test module, informational-only.

**Required checks:** Backend unit tests ✅ · Ruff format+critical ✅ · import-guard ✅ · complexity ✅ — `mergeStateStatus: CLEAN`.

The 45 repo-wide ruff findings (19 auto-fixable) + the 1 eslint error are worth a separate cleanup PR so we don't bundle unrelated changes here per our one-logical-change rule — happy to open one if there's appetite.

PR #479, opened 2026-06-07
[portfolio] Realistic backtest execution: look-ahead guard + Almgren market impact

onder-akkaya:
## Summary

Re-lands the **execution-realism** work through a reviewable PR (it had been pushed direct to `main`; reverted in #478). Two coherent pieces, both in `portfolio_backtester.py`:

### 1. Look-ahead execution guard
- Rolling σ and ADV use `.shift(1)` so the simulator never sees same-bar information.
- Missing data is **strictly inner-joined** — forward-filling halted / not-yet-public symbols would leak the survival fact to the simulator.
- Bankruptcy handling: equity ≤ 0 → `r_t = -1`, holdings zeroed.

### 2. Almgren square-root market impact + capacity decay
- `Impact = γ·σ·Q·√(Q/ADV)`, additive to the linear bps cost.
- Capacity-decay loop re-runs the backtest across AUM tiers ($1M → $10B) to surface Sharpe degradation at scale.
- `γ` now carries a citation (Almgren-Chriss 2000 / Almgren et al. 2005 square-root law) and a conservative `0.1` default with a calibration note — no longer an unexplained magic constant.

## Verify
`PYTHONPATH=backend python -m pytest backend/tests/services/test_portfolio_backtester.py -q` → 19 passed
Full suite → 1178 passed, 2 skipped.

## Anti-goals
- No change to the rigor gate (DSR/CPCV) — that's the separate `rigor-gate-hardening` PR.
- No new dependencies.

### DIFF
--- a/backend/archimedes/services/portfolio_backtester.py
+++ b/backend/archimedes/services/portfolio_backtester.py
@@ -42,6 +42,7 @@ import numpy as np
 import pandas as pd
 
 from archimedes.models.backtest import BacktestResult
+from archimedes.services.rigor_evaluator import compute_dsr, compute_oos_sharpe
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


onder-akkaya:
Addressed the one actionable lint finding in this PR's files: ruff `I001` (unsorted import block in `test_portfolio_backtester.py`) — fixed in 28043ee. The blocking `ruff-gate` (format + E9/F63/F7/F40) was already green.

The remaining quality-gate counts are pre-existing and out of scope here:
- The repo-wide broad-`ruff check` total and the 1 eslint error are not introduced by this PR (no frontend files touched). They're the existing informational debt the gate reports on every PR.
- Complexity gate is informational only; the 22 "orphan functions" are pytest test methods, which are invoked by the runner rather than called in-tree — expected, not a regression.

PR #481, opened 2026-06-07
[rigor] Align fusion-path DSR gate to p-value >= 0.95 threshold

onder-akkaya:
## Summary

The fusion path in `apply_rigor_gate()` (`fusion_evaluator.py:423`) was using `dsr > 0.0` as its DSR passing criterion — meaning any strategy with a positive Deflated Sharpe Ratio (even one with p ≈ 0.5) would pass.

The curated-path gate in `run_rigor_gate` (`rigor_evaluator.py`) uses `dsr_p_value >= 0.95`. These two paths must agree: a strategy generated via fusion and admitted Tier-1 should face the same statistical bar as one admitted via the curated library.

**Concrete failure mode:** a strategy with z = 0.5 (p ≈ 0.69) would be promoted Tier-1 by fusion, then fail if re-evaluated through the API gate. Users could end up with a Tier-1 badge on a strategy that would not pass the advertised admission criteria.

## Changes

- `backend/archimedes/services/fusion_evaluator.py`
  - `dsr_pass`: `dsr is not None and dsr > 0.0` → `dsr_p is not None and dsr_p >= 0.95`
  - `passing`: removed redundant `metrics.sharpe_ratio > 0.0` guard — `dsr_p >= 0.95` is strictly stronger and already implies a positive Sharpe with statistical confidence

## Test plan

- [ ] Confirm existing unit tests for `apply_rigor_gate` still pass (`pytest -k fusion`)
- [ ] Manually verify: a strategy with p = 0.94 is now rejected by fusion (was previously admitted)
- [ ] Manually verify: a strategy with p = 0.95 still passes

### DIFF
--- a/backend/archimedes/services/fusion_evaluator.py
+++ b/backend/archimedes/services/fusion_evaluator.py
@@ -415,14 +415,14 @@ def apply_rigor_gate(
     # rejected at validation time, long before reaching this evaluator).
     look_ahead_clean = True
 
-    # DSR-passing convention follows the seed strategies' implementation:
-    # the canonical Bailey-LdP DSR is "credibly > 0" via the test
-    # statistic z-score, where ``dsr_p_value`` is the p-value under the
-    # null Sharpe=0. We pass when ``dsr > 0`` (the DSR itself is positive
-    # after correcting for the trials-correction).
-    dsr_pass = dsr is not None and dsr > 0.0
+    # DSR gate: require p-value >= 0.95 (same threshold enforced by the curated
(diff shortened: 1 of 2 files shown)


onder-akkaya:
The two failing unit tests ( and ) have been fixed — they were relying on Faber/SPY producing dsr_p_value >= 0.95 with num_trials=10 (it's 0.82, which passed the old dsr > 0.0 threshold). Both tests now pass num_trials=1 since they test the provenance gate, not library-level multiple-testing correction. The ruff/eslint errors in the quality gate table are pre-existing informational issues (the hard ruff gate passed).

PR #15, opened 2026-05-16
[strategy] Interactive Claude strategy architect (provider→architect→guardrail→trace→/construct + arxiv)

dbrowneup:
## What

The interactive **"design me a portfolio"** path — Dan's strategy-engine lane. Free-text intent + risk profile → Claude (Maestro) selects/weights paper-grounded strategies → deterministic guardrail → hashed reasoning trace, exposed at `POST /api/strategies/construct`. Plus the previously-stubbed strategy endpoints now serve real data, and `extract_from_paper` is implemented.

Distinct from Chuan's autonomous `IAgentOrchestrator` (the *maintain-it* loop); they share `LocalStrategyProvider` + the guardrail. Önder's `IPortfolioConstructor` is the documented drop-in upgrade for the guardrail.

## Commits (8, atomic, each smoke-tested)

| | |
|---|---|
| `5b54219` | Serve strategy library from `LocalStrategyProvider` (was `[]`/404 stubs) |
| `003a678` | Claude strategy architect — `LLMBackend` seam (Claude + offline canned fallback), anti-hallucination |
| `8fcab55`→`29237f6` | Deterministic guardrail — normalize / 30% cap / USYC **floor & ceiling** honest-tension flag |
| `218c929` | Hashed `ReasoningTrace`; stops at the on-chain seam (`arc_tx_hash=None` for `ITracePublisher`) |
| `850b52a` | `POST /api/strategies/construct` — **additive** `architect_schemas.py`, shared `schemas.py` untouched |
| `2fa4351` | arxiv extraction pipeline — KnowledgeBase pattern w/ **pypdf (BSD), not PyMuPDF (AGPL)**; produces CANDIDATE passports |
| `73c3eb3` | Container-robust strategies-dir resolution (found via live docker test) |

## Verification

- **Offline:** every step's logic exercised by real execution (parse robustness, anti-hallucination, guardrail math across all 4 profiles incl. USYC-ceiling flag, trace hash determinism + tamper-sensitivity, arxiv render→provider AST roundtrip).
- **Live (docker, in-container):** `GET /api/strategies/` (4), `GET /{id}` (200), `POST /construct` (200, well-formed) all pass.

### Honest caveats
- `POST /construct` was verified using the **canned fallback**, not real Claude — the repo `.env` has an **empty `ANTHROPIC_API_KEY`**. Pipeline/HTTP/schema/guardrail/trace are all real; only model reasoning is unverified here (config, not code). Set a key to exercise live Claude.
[...]

PR #500, opened 2026-06-09
[rigor] Enforce OOS Sharpe + use real trial count in fusion rigor gate

onder-akkaya:
## Summary

Fixes **High finding #5** and **Medium finding #15** from `AUDIT_2026-06-10.md`. Both are corrections to the same function, `fusion_evaluator.apply_rigor_gate`, on adjacent lines — they form one logical unit (making the fusion gate enforce the four-primitive admission spec), so they ship together rather than as two PRs that would conflict on the same function.

### #5 — OOS Sharpe was computed but never enforced
`oos_sharpe` was computed and stored on the verdict, but the `passing` condition was:
```python
passing = dsr_pass and look_ahead_clean and (pbo_score is None or pbo_score < 0.5)
```
Only three of the four advertised primitives gated admission. A strategy with a non-positive out-of-sample Sharpe could be certified Tier-1 through the fusion path. Added an absolute OOS floor (`oos_sharpe > 0`), mirroring the curated path's [`RigorGateResult.passes_all`](backend/archimedes/services/rigor_evaluator.py).

### #15 — num_trials hardcoded to 10
The DSR was deflated with `num_trials=10` regardless of the real variant-grid size — a 50-variant sweep got only a 10-trial penalty (5× undercount); a tiny grid was over-penalized. Now `effective_trials = len(variants_metrics)` when the variant matrix is known (≥2), falling back to the passed count otherwise, and the verdict reports it.

## Verify

```bash
PYTHONPATH=backend pytest backend/tests/services/test_fusion_evaluator.py -q   # 21 passed
```

New tests: `TestFusionGateEnforcesOosSharpe` (negative OOS fails even when DSR passes; positive OOS passes), `TestFusionGateUsesRealTrialCount` (num_trials tracks variant count and falls back correctly).

## Anti-goals

- No change to the DSR / PBO / correlation math.
- No change to the curated `run_rigor_gate` path.

Lane: quant rigor (Önder).

### DIFF
--- a/backend/archimedes/services/fusion_evaluator.py
+++ b/backend/archimedes/services/fusion_evaluator.py
@@ -395,10 +395,18 @@ def apply_rigor_gate(
                 (curve[i] - curve[i - 1]) / curve[i - 1] for i in range(1, len(curve)) if curve[i - 1] > 0
             ]
 
+    # num_trials = actual size of the multiple-testing selection set. A fixed
+    # default (10) under-deflates a large variant grid (e.g. a 50-variant sweep
+    # gets only a 10-trial penalty) and over-deflates a small one. When the
+    # variant matrix is known, use its real cardinality as the trial count.
+    effective_trials = num_trials
+    if variants_metrics is not None and len(variants_metrics) >= 2:
+        effective_trials = len(variants_metrics)
+
     # Correlated variants carry fewer independent trials than their nominal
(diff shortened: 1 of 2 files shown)


onder-akkaya:
**Batch 1 (quant).** Independent. Verify: `PYTHONPATH=backend pytest backend/tests/services/test_fusion_evaluator.py -q`.

PR #511, opened 2026-06-09
[backend] SIWE-gate unauthenticated vault create + metadata endpoints

onder-akkaya:
## Summary

Fixes **High finding #8** from `AUDIT_2026-06-10.md`.

Two endpoints triggered backend-signed on-chain transactions **without SIWE authentication** (rate-limited per IP only):

- `POST /api/vaults/create` — deploys a vault, charging the backend signer's gas → unauthenticated gas-drain.
- `POST /api/vaults/metadata` — lets anyone overwrite any vault's metadata and fire `_anchor_strategies_async` (backend-signed on-chain tx) → metadata tampering + unauthenticated on-chain spend.

The same signer (`ARC_AGENT_PRIVATE_KEY`) also drives `execute_trades`, so weak auth here is funds-adjacent.

## Changes

- Add `Depends(require_verified_wallet)` (existing SIWE session dependency) to both endpoints — **401 without a valid session**.
- **Owner-scope** metadata writes: the authenticated wallet becomes the metadata owner on first write; `creator_address` is bound to the session wallet (never a caller-supplied value, which was the spoofing vector); later writes require the caller to be that owner (**403** otherwise).
- Let `HTTPException` (401/403) pass through the metadata handler instead of being masked as 500.

## Verify

```bash
PYTHONPATH=backend pytest backend/tests/test_vault_metadata_anchor.py -q   # 7 passed
# unauthenticated create now 401:
curl -X POST .../api/vaults/create -d '{"name":"x",...}'   # 401
```

New tests: 401 unauthenticated (create + metadata); 403 non-owner overwrite (owner can still edit). Existing anchor tests now send a SIWE cookie.

## Note

The audit's `:131` blocking `session.query` (finding #24) is addressed separately. Lane: backend (Daniel R.); funds-adjacent, so flagged for prompt review.

PR #512, opened 2026-06-10
[auth] Wire SIWE in the UI + bind domain/chain-id/expiry in verify

onder-akkaya:
## Summary

Fixes **High finding #9** from `AUDIT_2026-06-10.md` — SIWE was dead code in the UI and under-validated on the server.

### UI: SIWE never ran
`WalletConnect.jsx:102` called `getWalletClient()` **without importing it**. That threw `ReferenceError`, was swallowed by the surrounding `try/catch` (`console.warn('SIWE auth failed (non-fatal)')`), and returned early — so no session cookie was ever established and the entire PII/auth-gated layer was unreachable from the UI. `getWalletClient` is already exported from `../config`; the fix is to import it.

(This is why the SIWE gates in #511 matter — without this, no UI user could ever authenticate.)

### Server: verify only checked the nonce
`verify_signature` validated the nonce but not the message itself, so a signature from any dApp on the same address — if it happened to carry a current Archimedes nonce — would verify. Added EIP-4361 binding:

- **domain** — normalized (scheme/trailing-slash-insensitive) compare against `PUBLIC_DOMAIN`.
- **chain-id** — must equal `ARC_CHAIN_ID` (default 5042002).
- **issued-at freshness** — within the nonce TTL, with a small clock-skew tolerance.

## Verify

```bash
PYTHONPATH=backend pytest backend/tests/test_auth_siwe.py -q   # 28 passed
grep -n "getWalletClient" ui/src/components/WalletConnect.jsx    # now imported
```

New tests: wrong-domain / wrong-chain / stale-issued-at each 401. The existing full-flow tests hardcoded `Issued At: 2026-05-26` (now stale under the freshness check) and were updated to use a current timestamp.

## Lanes

UI import → Marten; backend validation → Daniel R. Filed together because they are one finding (the UI fix is meaningless without the round-trip working, and vice-versa).

### DIFF
--- a/backend/archimedes/api/auth_siwe.py
+++ b/backend/archimedes/api/auth_siwe.py
@@ -27,6 +27,7 @@ import logging
 import os
 import secrets
 import time
+from datetime import datetime, timedelta
 
 from fastapi import APIRouter, HTTPException, Request, Response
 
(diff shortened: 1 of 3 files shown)


onder-akkaya:
Self-audit (backend/security reviewer subagent) flagged that the domain/chain-id/issued-at checks were *skip-if-absent* — a crafted message omitting those lines bypassed binding (low risk, since the nonce + signer-recovery still block cross-dapp replay, but it left the binding non-self-sufficient). Hardened: all three fields are now **required** (missing → 400), so the binding stands on its own. Added a regression test for the missing-field case; updated the minimal-message tests to carry Chain ID + a fresh Issued At. 29 auth tests pass.

onder-akkaya:
**Batch 2 (auth).** Land **before or with #511** so the SIWE gates there aren't stranded (this PR is what makes SIWE actually run in the UI). Reviewers: Marten (UI) + Daniel R. (backend). ℹ️ Force-pushed during self-audit (mandatory domain/chain/expiry binding) — current HEAD is correct, 29 auth tests pass.

PR #515, opened 2026-06-10
[infra] Stop hardcoding the Postgres password; generate at boot (audit #11)

onder-akkaya:
## Summary

Fixes **High finding #11** from `AUDIT_2026-06-10.md`.

`POSTGRES_PASSWORD=archimedes-hackathon-2026` was committed verbatim in **both** `.env.example` and `infra/user-data.sh` — the production Postgres credential was world-readable in the repo.

## Changes

- `infra/user-data.sh`: generate a random password at boot (`openssl rand -hex 24`, URL-safe) and reuse it for `POSTGRES_PASSWORD` and `DATABASE_URL`; `chmod 600` the `.env`.
  - **Templatefile escaping:** `user-data.sh` is rendered by Terraform `templatefile()` (main.tf), which treats `${...}` as template syntax. The shell var is written `$${DB_PASS}` so Terraform emits a literal `${DB_PASS}` for the boot-time shell. (Only `${repo_url}` is a real template var.)
- `.env.example`: obviously-local throwaway value (`local-dev-only-change-me`) so local `docker compose up` still works one-command, with a comment that it is not a production secret.

## ⚠️ Rotation does not undo the leak (CLAUDE.md)

The old literal remains in `git log -p` and every clone. **The live DB password must be rotated out-of-band regardless.** An existing deployment whose Postgres volume was initialized with the old literal needs an `ALTER USER ... PASSWORD` (a fresh-boot random password won't match an already-initialized data dir). Chuan should coordinate the rotation + `terraform apply`.

Longer term: pull the password from AWS SSM Parameter Store at container start.

## Verify

```bash
grep -rn "archimedes-hackathon-2026" .env.example infra/user-data.sh   # only in an explanatory comment
```

Lane: infra (Chuan).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,10 +12,13 @@
 # wallet / RPC / Circle variables are optional locally (skip on-chain features).
 
 # ── PostgreSQL (the local docker compose stack starts a server with these) ──
+# LOCAL-DEV ONLY. This is a throwaway local password, NOT a production secret —
+# the EC2 bootstrap (infra/user-data.sh) generates a random password at boot.
+# Never reuse this value anywhere real. Keep the two lines in sync if you change it.
 POSTGRES_USER=archimedes
-POSTGRES_PASSWORD=archimedes-hackathon-2026
+POSTGRES_PASSWORD=local-dev-only-change-me
 POSTGRES_DB=archimedes
-DATABASE_URL=postgresql://archimedes:archimedes-hackathon-2026@postgres:5432/archimedes
+DATABASE_URL=postgresql://archimedes:local-dev-only-change-me@postgres:5432/archimedes
 
 # ── Redis ──────────────────────────────────────────────────────────────────
 REDIS_URL=redis://redis:6379/0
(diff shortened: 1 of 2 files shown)


onder-akkaya:
Self-audit (contracts/infra reviewer subagent) caught a ship-blocker I'd missed: my explanatory comment in `user-data.sh` contained a bare `${...}`, which Terraform `templatefile()` parses (it does not skip comments) → `terraform apply` would error with an invalid-expression. Fixed: the comment now uses escaped `$${...}` and avoids any bare `${...}`; the only bare interpolation left is the legitimate `${repo_url}` template var. Force-pushed the amend.

onder-akkaya:
**Batch 4 (infra, Chuan).** ⚠️ **Merging does NOT apply this** (user-data.sh runs only on new instance boot). Chuan must `terraform apply` **and rotate the live DB password out-of-band** (`ALTER USER` — the running Postgres volume still holds the old password). ℹ️ Force-pushed during self-audit to fix a Terraform `templatefile()` escaping bug — current HEAD is correct.

Issue #27, opened 2026-05-17
APIN - Portfolio - Liquidity


martenwindler:
APIN - Portfolio - Liquidity

Reference
http://18.171.230.205/ 

### **User Story: Manage Portfolio Liquidity on APIN**

**As a** DeFi portfolio manager or liquidity provider,
**I want to** view, deposit, withdraw, and monitor liquidity positions in my APIN portfolio,
**So that** I can optimize my asset allocation and earn yields efficiently.

Iteration 16.05.2026

Iteration 17.05.2026
<img width="1244" height="676" alt="Image" src="https://github.com/user-attachments/assets/4972aca2-3223-473c-8d2a-c7f0c7f05872" />

Iteration 18.05.2026

Iteration 19.05.2026

Iteration 20.05.2026

Iteration 21.05.2026

Iteration 22.05.2026

Iteration 23.05.2026

---
---

### **Acceptance Criteria:**
- [ ] User can **view their current liquidity positions** (e.g., pools, tokens, APY, TVL).
- [ ] User can **filter liquidity pools** by:
  - Token pair (e.g., ETH/USDC, BTC/USDT).
  - APY range (e.g., 5%-20%).
  - TVL (Total Value Locked) size.
- [ ] User can **deposit assets** into a liquidity pool:
  - Select a pool from the list.
  - Input the amount of each token to deposit (e.g., 1 ETH + 2000 USDC).
  - Preview the expected LP tokens and APY before confirming.
- [ ] User can **withdraw assets** from a liquidity pool:
  - Select a pool and input the amount of LP tokens to burn.
  - Preview the expected tokens to receive (e.g., 0.5 ETH + 1000 USDC).
  - Confirm withdrawal with a transaction signature.
- [ ] User can see **real-time data** for each pool:
  - Current APY.
  - TVL.
  - 24h volume.
  - Impermanent loss risk (if applicable).
- [ ] User can **connect their wallet** (e.g., MetaMask, WalletConnect) to interact with liquidity pools.
- [ ] System **validates inputs** (e.g., insufficient balance, invalid amounts).
- [ ] User receives **success/error notifications** for all actions (e.g., "Deposit successful! TX: [hash]").
- [ ] All liquidity actions are **recorded on-chain** (e.g., Arc Testnet).

---
---

### **Demo Screen Reference:**
> **Screen**: APIN Portfolio - Liquidity Dashboard (`http://18.171.230.205/`).
> **Elements**:
[...]

Issue #20, opened 2026-05-17
APIN - Markets - Strategies


martenwindler:
APIN - Markets - Strategies

Reference
http://18.171.230.205/ 

Iteration 16.05.2026

Iteration 17.05.2026
<img width="1244" height="674" alt="Image" src="https://github.com/user-attachments/assets/6f0c5ca5-4df1-4c2a-9f72-6cc4bb91b087" />

Iteration 18.05.2026

Iteration 19.05.2026

Iteration 20.05.2026

Iteration 21.05.2026

Iteration 22.05.2026

Iteration 23.05.2026

PR #517, opened 2026-06-10
[infra] Serve security headers on the live HTTP (:8080) server block (audit #20)

onder-akkaya:
## Summary

Fixes **Medium finding #20** from `AUDIT_2026-06-10.md`.

Production traffic is served from the `listen 8080` block (the ALB terminates TLS and forwards to `:8080` as HTTP with `X-Forwarded-Proto: https`). The security headers (`CSP`, `X-Frame-Options: DENY`, `HSTS`, `X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`) existed **only in the unused `:8443` block** — so the live app served none and was clickjackable with no CSP.

## Change

Add the same `add_header ... always` block to the `:8080` server (server-level → inherited by all its locations, none of which define their own `add_header`). Kept identical to the `:8443` block.

CSP content is **unchanged** — it still allows `'unsafe-inline'`/`'unsafe-eval'` (needed today by the Vite bundle + Circle SDK). Tightening that is **finding #27**, filed separately because it needs inline-script mapping + browser testing to avoid breaking passkey signup.

## ⚠️ Needs Chuan to validate + redeploy

Not `nginx -t`'d locally. Verify after deploy:
```bash
curl -sI https://archimedes-arc.app/ | grep -iE "content-security-policy|x-frame-options|strict-transport"
```

Lane: infra (Chuan).

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -13,6 +13,20 @@ server {
     listen 8080;
     server_name archimedes-arc.app;
 
+    # ── Security headers ──────────────────────────────────────────────────
+    # Production traffic is served from THIS block (ALB terminates TLS and
+    # forwards to :8080), so the headers must live here — they previously
+    # existed only in the unused :8443 block, leaving the live app with none
+    # (clickjacking + no CSP). See AUDIT_2026-06-10.md finding #20. Kept in sync
+    # with the :8443 block below; CSP still allows 'unsafe-inline'/'unsafe-eval'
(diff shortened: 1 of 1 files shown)


onder-akkaya:
**Batch 4 (infra, Chuan).** Self-applies on next deploy. ⚠️ **Not `nginx -t`'d locally** — Chuan should validate the config before/at merge.

PR #531, opened 2026-06-11
[strategy] Add 3 second-wave economic pairs (KO/PEP, EWA/EWC, GLD/SLV) !minor

onder-akkaya:
## What this adds (in plain terms)

Three new **pairs-trading** strategies to the library, taking it from **13 → 16**:

| Strategy | The two assets | Why they move together |
|---|---|---|
| `gatev_2006_pairs_ko_pep` | **KO / PEP** (Coca-Cola / PepsiCo) | Same industry — both global beverage / consumer-staples companies |
| `gatev_2006_pairs_ewa_ewc` | **EWA / EWC** (Australia / Canada ETFs) | Both commodity-exporter economies (the textbook cointegration pair) |
| `gatev_2006_pairs_gld_slv` | **GLD / SLV** (gold / silver) | Both precious metals driven by real rates + risk sentiment |

A "pairs trade" is **market-neutral**: instead of betting an asset goes up, it bets that two normally-linked assets that have drifted apart will converge again — buy the cheap one, short the rich one. This is the same logic as the existing GLD/GDX flagship pair; we're just applying it to three new, economically-sensible pairs. **No engine changes** — this is the cheapest tier of the second-wave roadmap (Phase 1.3), proving the loop end-to-end.

This is the first PR implementing [`docs/specs/second-wave-multi-asset-strategies.md`](https://github.com/hackagora/archimedes-arcadia/blob/main/docs/specs/second-wave-multi-asset-strategies.md) (companion: [`second-wave-handover.md`](https://github.com/hackagora/archimedes-arcadia/blob/main/docs/specs/second-wave-handover.md)).

## Honest result — all three FAIL the rigor gate (this is correct, not a bug)

On real 2004–2026 data, every one of these pairs has a **negative Sharpe ratio** and does **not** pass our DSR/PBO/out-of-sample rigor gate. So all three are admitted as **`CANDIDATE`**, not `VALIDATED` — exactly the behaviour the roadmap calls for:

> *"Expect many strategies to fail the gate — that's correct. Admit as CANDIDATE; never weaken thresholds to force a pass."*

A naive fixed-ratio pairs trade on these ETFs doesn't beat transaction costs over this window. The product surfaces that honestly rather than hiding it. (A smarter cointegration-tested and Kalman dynamic-hedge variant is Phase 1.1/1.2 — the next PR.) **Nothing was tuned to make these pass.**

The metrics are **real backtest output**, computed by `scripts/regen_fixtures.py` against live market data — never hand-typed. `paper_claimed_*` are left **null** because the source paper reports a diversified-portfolio return, not a clean single-pair number, so there's nothing honest to compare against.

## How it works under the hood (for reviewers)

- The backend is **directory-driven**: each strategy is one `.py` file in `analytics-engine/strategies/` with a metadata block the backend reads via AST. These three files reuse the shared `PairsDistanceTrading` class (via import) but each carries its own passport metadata, so they get **distinct deterministic IDs** despite citing the same paper. (Verified by a test.)
- `regen_fixtures.py` is now **add-only AND idempotent**: it skips any strategy already in `backtest_fixtures.json` and re-running it is a no-op. `num_trials_in_selection` (the DSR multiple-testing penalty) is computed from the post-add library size (16). The legacy fixture entries are byte-for-byte untouched — verified: the fixture diff is **87 insertions, 0 deletions**.

## Files
- **New strategies:** `gatev_2006_pairs_{ko_pep,ewa_ewc,gld_slv}.py`
- **Instruments:** `KO, PEP, EWA, EWC, SLV` added to `OPERATION_TO_SYMBOL`
- **Fixtures:** 3 new real-metric entries in `backtest_fixtures.json`
- **Script:** `regen_fixtures.py` made add-only idempotent
- **Tests:** `test_pairs_engine.py` (load+run each new pair, hermetic), `test_instruments.py`, `test_strategy_endpoints.py` (presence / honest-candidate / distinct-id)

## Verification (all green locally)
```
cd analytics-engine && uv run pytest          # 33 passed
pytest backend/tests/test_strategy_endpoints.py backend/tests/services/test_strategy_provider.py   # 62 passed
ruff format --check . && ruff check --select E9,F63,F7,F40 .   # clean
```

## Out of scope / cross-lane note
Quant-lane work (Önder). No engine, contract, or infra changes. Reviewer eyes most useful from **Dan** (strategy-engine lead). The negative-Sharpe CANDIDATE outcome is intentional — please don't read it as "broken."


onder-akkaya:
Merging (first of the stacked second-wave chain, → `main`).

**Why merge:** full quality-gate is green here — Backend unit tests, Ruff format+critical-lint, Complexity, Lint-report, and import-guard all pass; locally the analytics-engine suite (33) and backend strategy suite are green. The change is purely additive: three new distance pairs reusing the vetted `PairsDistanceTrading` logic, real backtest fixtures (add-only — 87 insertions, 0 deletions, legacy entries untouched), and distinct deterministic strategy IDs verified by test. All three are admitted as honest **CANDIDATE** (negative Sharpe on real data) with null `paper_claimed_*` — nothing tuned to pass the rigor gate, which is the intended discipline. Library 13 → 16. Merge commit per repo convention.

Issue #547, opened 2026-06-11
DSR convention divergence: backend rigor_evaluator subtracts risk-free, fixture-side compute_dsr does not (team decision)


onder-akkaya:
## Summary

While assembling the #537 decision package I found that the two `compute_dsr` implementations disagree on the Sharpe convention, and the disagreement is large enough to flip gate outcomes:

- **Backend** (`backend/archimedes/services/rigor_evaluator.py::compute_dsr`, the live gate for generated strategies): `SR_hat = (mean − rf_daily) / σ` with rf = 5%/252 — **excess-return Sharpe**.
- **Fixture-side** (`analytics-engine/scripts/regen_buy_hold_fixture.py::compute_dsr`, the single source for every `dsr_p_value`/`deflated_sharpe_ratio` stored in `backtest_fixtures.json`, i.e. for every served library verdict): `sr_hat = mean / σ` — **raw Sharpe, no rf subtraction**.

At today's 5% rf the raw convention is systematically more lenient. Both paths are live: library strategies get the fixture-side numbers; pipeline-generated strategies get the backend numbers. The same strategy can pass one path and fail the other.

## Reproduce (fresh 2026-06-11 daily-returns store from PR #545)

`maillard_2010_risk_parity` full fresh series (n=5268, Sharpe +0.3492 — matches its fixture value, so this is convention, not data drift):

| num_trials | fixture-side p (raw) | backend p (excess, ρ=0) |
|---|---|---|
| 1 | 0.9996 | 0.8882 |
| 13 | 0.9630 | 0.3214 |
| 22 | 0.9414 | 0.2421 |

```python
# fixture side:
from regen_buy_hold_fixture import compute_dsr            # analytics-engine/scripts
# backend side:
from archimedes.services.rigor_evaluator import compute_dsr
# both fed strategies/daily_returns/maillard_2010_risk_parity.json -> daily_returns
```

The fixture-side sweep on fresh data reproduces #537's N-sweep table almost exactly (0.963 @ N=13, 0.941 @ N=22) — confirming the stored table was computed under the raw convention.

## Why it matters

1. **#537 is entangled with this.** Under the raw convention, a provenance-based N recovers risk parity (N=3 → p=0.995). Under the excess convention, no N recovers it (N=1 → p=0.888). The 'what N?' decision needs the 'which Sharpe?' decision first.
[...]

Issue #537, opened 2026-06-11
Rigor gate: should DSR num_trials be the full library size for individually-specified strategies? (team decision)


onder-akkaya:
## Summary

The Deflated Sharpe Ratio (DSR, Bailey & López de Prado 2014) penalizes a strategy's Sharpe by `num_trials_in_selection` (N), the number of trials the strategy was "selected" from. We currently set **N = the full library size** (now 22) for *every* strategy, in both fixture generators and the backend `rigor_evaluator`. This treats each strategy as if it were cherry-picked as the best of 22 independent backtests.

For strategies that were **data-mined/selected** from the library, that's correct and conservative. But for an **individually-specified, paper-grounded** strategy — one we implemented from a named paper, not picked because it scored well — penalizing it by the whole library's trial count is arguably too harsh. This is a calibration question about the rigor gate, and because the gate decides what can be promoted toward live/deployable with real USDC, it warrants a **team decision** rather than a unilateral change.

## Evidence — it blocks our single best new strategy (and only that one)

`maillard_2010_risk_parity` on the original 5-asset universe: **Sharpe +0.35, max-DD 27%, CAGR +8.4%** — the strongest of the entire second wave. Its DSR p-value vs `num_trials`, on its real backtest returns:

| `num_trials` | DSR p-value | passes p ≥ 0.95? |
|---|---|---|
| 1–13 | 0.999 → 0.963 | **yes** |
| **22 (current: full library)** | **0.941** | **no** |
| 50 | 0.896 | no |

So with N ≤ 13 it would clear the DSR criterion; at N = 22 it does not. It is the *only* near-miss — see [`docs/specs/second-wave-universe-experiment.md`](docs/specs/second-wave-universe-experiment.md): the other 8 second-wave strategies fail on **performance** (7 have negative Sharpe), not on this penalty. So fixing the penalty would not "flood" the library — it would, at most, let one genuinely-decent strategy through.

## The design question

What should `num_trials_in_selection` be for an individually-specified, paper-grounded strategy?

**Options (for the team to weigh — not implementing unilaterally):**
1. **Status quo:** N = full library size. Maximally conservative; simple; defensible as "every strategy competes against the whole shelf." Cost: blocks individually-justified strategies that weren't data-mined.
[...]

PR #548, opened 2026-06-11
[DO NOT MERGE] #537 prototype: provenance-based num_trials resolver

onder-akkaya:
**DO NOT MERGE.** Draft on purpose; gate change requiring @dbrowneup's sign-off per the #537 anti-goals.

Executable spec for [#537](https://github.com/a-apin/archimedes/issues/537) option 3 (provenance-based split), showing the diff is small: one pure function in `rigor_evaluator.py` + 5 hermetic tests documenting the semantics. No call site is wired; no threshold moves; the p ≥ 0.95 bar is untouched.

Also blocked on the DSR convention decision in [#547](https://github.com/a-apin/archimedes/issues/547) — under the backend (excess-Sharpe) convention the N input alone does not recover risk parity, so #547 should be decided first. Full decision package: see the comment on #537.

Verify: `PYTHONPATH=backend /tmp/abe/bin/python -m pytest backend/tests/test_resolve_num_trials.py backend/tests/test_rigor_evaluator.py -q` → 118 passed.

### DIFF
--- a/backend/archimedes/services/rigor_evaluator.py
+++ b/backend/archimedes/services/rigor_evaluator.py
@@ -331,6 +331,35 @@ def _annualized_sharpe_arr(arr: np.ndarray) -> float | None:
     return float(((arr.mean() - _RF_DAILY) / sigma) * math.sqrt(_ANNUALIZATION))
 
 
+def resolve_num_trials(
+    provenance: str,
+    n_variants_tried: int | None,
+    library_size: int,
+) -> int:
+    """#537 PROTOTYPE — DO NOT MERGE without team sign-off (Dan).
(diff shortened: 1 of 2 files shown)


onder-akkaya:
Closing: this prototype implemented #537's option 3 (provenance-based num_trials split). #537 was resolved the other way via #592/#547 — under the canonical excess-Sharpe convention no strategy is blocked by the N input alone, so the provenance split isn't needed and N stays at full library size (status quo). If a concrete case later forces the N question, reopen with that case. Deleting the prototype to avoid a stale DO-NOT-MERGE PR lingering.

Issue #28, opened 2026-05-17
APIN - Portfolio - Vaults


martenwindler:
APIN - Portfolio - Vaults

Reference
http://18.171.230.205/ 

### **User Story: Create a Managed Vault on Arc**

**As a** portfolio manager or DeFi user,
**I want to** create a new managed vault on Arc with customizable strategies, fees, and agent-assisted rebalancing,
**So that** I can deploy a tailored investment strategy and automate its execution on-chain.

Iteration 16.05.2026
<img width="1244" height="673" alt="Image" src="https://github.com/user-attachments/assets/516f13a3-9970-4610-b70e-93433bd7ed38" />

Iteration 17.05.2026
<img width="1243" height="676" alt="Image" src="https://github.com/user-attachments/assets/b9cfa0d6-5fd6-4a7b-b823-de9b8a516c57" />
<img width="1242" height="676" alt="Image" src="https://github.com/user-attachments/assets/393d23c1-c016-4bd6-8e9f-f7d60bdc227e" />

Iteration 18.05.2026

Iteration 19.05.2026

Iteration 20.05.2026

Iteration 21.05.2026

Iteration 22.05.2026

Iteration 23.05.2026

---

### **Acceptance Criteria:**
- [ ] User can input a **vault name** (e.g., "Archimedes Moderate Growth").
- [ ] User can input a **symbol** (e.g., "AMG") for the vault.
- [ ] User can set a **management fee (BPS)** (e.g., 50 BPS = 0.50%/year).
- [ ] User can set a **performance fee (BPS)** (e.g., 1000 BPS = 10.00% on profits).
- [ ] User can toggle **agent-assisted rebalancing** (On/Off).
- [ ] User can select **one or more strategies** from the available list (e.g., "Quantitative Approach to Tactical Asset Allocation," "Volatility-Managed Portfolios").
- [ ] Selected strategies are visually highlighted (e.g., with a "live" or "candidate" tag).
- [ ] User can see a **preview of strategy details** (e.g., description, authors, tags like `SPY`, `NIKKEI`, `GOLD`).
- [ ] User can click **"Deploy vault on Arc"** to submit the vault creation.
- [ ] System validates all inputs (e.g., fees are within allowed ranges, vault name is unique).
- [ ] User receives a **success confirmation** (e.g., transaction hash or vault address) after deployment.
- [ ] User can **connect their wallet** (e.g., via "Connect Wallet" button) to interact with the vault.

---

[...]

PR #591, opened 2026-06-13
[backend] Bind chat identity to SIWE session when present

### LINKED ISSUE #524: [backend] Chat identity is request-body-supplied, not SIWE-bound (audit #28)
## Summary
AUDIT_2026-06-10.md finding #28. `chat_routes.py` takes `wallet_address` from the request body (regex-validated only), not from the SIWE session, so any caller can post chat as any wallet. Documented as intentional ("fully open"), but chat attribution is therefore untrustworthy.

## Decision needed
If chat attribution matters for future audit trails / reputation, bind it to the SIWE session instead of trusting the body:
- Use `Depends(get_verified_wallet)`; if a session exists, override/verify `wallet_address` against it; optionally allow anonymous posts with a clearly-unverified marker.

This is a product/security trade-off (open chat vs verified attribution), not a clear bug — filing for the team to decide. Now that SIWE actually runs in the UI (#512) and vault endpoints are gated (#511), binding chat identity is low-friction if desired.

Lane: backend (Daniel R.).

onder-akkaya:
## Summary
Implements the hybrid decided for #524 (audit finding #28): vault chat stays fully open, but wallet attribution becomes trustworthy. With a valid SIWE session the message identity comes from the session (a contradicting body `wallet_address` is rejected with 403 — rejection over silent override); without a session the post is still accepted but explicitly marked `verified: false`, so attribution is never silently trusted.

## What changed
- `backend/archimedes/api/chat_routes.py` — `POST /api/vaults/{address}/chat` now takes the optional SIWE dependency (`Depends(get_verified_wallet)`, reused from `auth_siwe.py` — no new auth mechanism). Session present → attribute to session wallet, `verified=True`, 403 on body/session mismatch. No session → body wallet required + regex-validated as before, `verified=False`. Read endpoints surface the flag.
- `backend/archimedes/api/schemas.py` — `ChatPostRequest.wallet_address` optional (required only without a session); `ChatMessageResponse.verified: bool = False`.
- `backend/archimedes/models/chat.py` — `ChatMessage.verified` column, default `False` so pre-existing rows stay honest (they were body-supplied, never verified); included in `to_dict`.
- `backend/archimedes/db.py` — idempotent `ALTER TABLE chat_messages ADD COLUMN IF NOT EXISTS verified` in the existing Postgres patch list (same pattern as the `strategy_store` patches).
- `backend/archimedes/services/chat_service.py` — `post_message(..., verified: bool = False)`; AI/system messages stored `verified=True` (backend-authored identity).
- `backend/tests/test_chat_routes.py` — new route-level tests, hermetic, using the real signed-cookie `_siwe_cookies` precedent from `test_user_routes.py`: (a) session → session-wallet attribution + `verified=True` (incl. case-insensitive body match), (b) mismatched body wallet → 403 and not persisted, (c) no session → accepted with `verified=False`, plus 422 when no session and no/invalid wallet, (d) regression on list/count read endpoints incl. the `verified` flag.

## Verification
```
env -i HOME=$HOME PATH=/usr/bin:/bin PYTHONPATH=backend python -m pytest backend/tests/test_chat_routes.py -q
→ 6 passed

env -i HOME=$HOME PATH=/usr/bin:/bin PYTHONPATH=backend python -m pytest backend/tests -m "not integration" -q
→ 1492 passed, 2 skipped (pre-existing skips)

ruff format --check .   → 307 files already formatted
ruff check --select E9,F63,F7,F40 .   → All checks passed!
```

Cross-lane note: chat routes sit in Daniel R.'s backend lane — flagging for his review eyes.

Closes #524

### DIFF
--- a/backend/archimedes/api/chat_routes.py
+++ b/backend/archimedes/api/chat_routes.py
@@ -5,9 +5,12 @@ Endpoints:
   POST   /api/vaults/{address}/chat       — post a message
   GET    /api/vaults/{address}/chat/count  — message count
 
-Design (per ecosystem-design-spec.md § 16–17):
+Design (per ecosystem-design-spec.md § 16–17, identity hardened per issue #524):
   - Fully open: any connected wallet can read/write
-  - Wallet address = identity
+  - Wallet address = identity. With a SIWE session the identity comes from the
+    session and the message is stored `verified=True`; a mismatched body wallet
+    is rejected with 403. Without a session, posts are accepted but explicitly
+    marked `verified=False` — attribution is never silently trusted.
   - AI auto-responds to @archimedes mentions
 """
 
@@ -19,6 +22,7 @@ import re
 from fastapi import APIRouter, Depends, HTTPException, Query, Request, Response
 
 from archimedes.api.auth_guard import require_internal_agent_key
+from archimedes.api.auth_siwe import get_verified_wallet
 from archimedes.api.limiter import limiter
 from archimedes.api.schemas import (
     ChatMessageListResponse,
@@ -56,6 +60,7 @@ async def get_chat_messages(
                 wallet_address=m["wallet_address"],
                 message=m["message"],
                 is_ai=m["is_ai"],
+                verified=m.get("verified", False),
                 created_at=m["created_at"],
             )
             for m in messages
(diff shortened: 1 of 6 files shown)


onder-akkaya:
Self-audit before merge (re-ran the suite hermetically rather than trusting the green check): caught `test_list_and_count_surface_verified` failing on the **second** consecutive run (`assert 4 == 2`) — it asserted absolute counts against the persistent `sqlite:///./archimedes_chat.db`, so rows accumulate across runs. Fixed in the follow-up commit by switching to the before/after delta + message-specific assertions already used by `TestMismatchRejected` in the same file. Now passes twice in a row. Final: chat suite 6/6 hermetic (×2), full unit suite green in CI.

Design note for the record: AI-authored messages are stored `verified=True` (the backend itself posts them, so the attribution is trustworthy) — flag if you'd prefer a separate `system`/`ai` marker instead. Merging under solo-operator authorization (no team — Önder + agent).

PR #436, opened 2026-05-28
[infra] Cut D: teardown managed services, keep EC2-only ($133→$42/mo)

t2o2:
## Summary

Tear down all managed infrastructure. Cost: $133/mo → $42/mo.
Terraform definitions preserved in `infra/modules/security-architecture/`
for re-apply (~15 min) when real users arrive.

## Verified on live after teardown

| Service | Status |
|---------|--------|
| Health API | ✅ ok, chain connected |
| Frontend | ✅ HTTP 200 |
| 6 strategies | ✅ |
| 13 vaults | ✅ |
| 694 traces | ✅ |
| SIWE auth | ✅ |
| 84 explore assets | ✅ |
| Regime (dynamic conf) | ✅ 0.586 |
| Port 8000 blocked | ✅ |

## Still active (zero-cost security)
SIWE session auth, nginx rate limiting, closed ports (5432/6379/8000),
rotated SSH keys, OIDC deploy pipeline, detect-secrets pre-commit.

PR #676, opened 2026-06-16
[infra] WAF: exclude LLM endpoints from SQLi rules, flip SQLi group to BLOCK (#521)

### LINKED ISSUE #521: [infra] Flip WAF CRS/SQLi rules from COUNT to BLOCK once observation window passed (audit #23)
## Summary
AUDIT_2026-06-10.md finding #23. `infra/waf.tf:51,121` set the AWS Managed CRS + SQLi rule groups to `override_action { count {} }` (observe-only), not enforcing. A comment notes this is temporary ("flip to BLOCK after 24-48h observation").

## Why an issue, not a blind PR
Flipping to enforce (`override_action { none {} }`) without checking the CloudWatch WAF metrics for false positives over the observation window risks blocking legitimate traffic (e.g. the Circle SDK / RPC calls). Someone with AWS access must confirm the observation window has passed and the COUNT metrics show no legit-traffic matches first.

## Steps
1. Check WAF sampled requests / `CountedRequests` metrics for the CRS + SQLi rule groups over the last window.
2. If clean, change `override_action { count {} }` → `override_action { none {} }` for both rule groups in `infra/waf.tf`.
3. `terraform plan` + apply; re-check `BlockedRequests` for false positives.

## Verify
```bash
grep -n "override_action" infra/waf.tf   # both should be none {} after enforcing
```

Lane: infra (Chuan). Needs AWS metrics review. Do NOT auto-assign to t2o2.

onder-akkaya:
## Summary

Closes #521.

The AWS Managed SQLi rule group (`AWSManagedRulesSQLiRuleSet`, priority 40) in `infra/waf.tf` was the last managed group still in observe-only `override_action { count {} }` mode. The inline comment noted it false-positives on LLM prompt bodies (prompts routinely contain `SELECT`, quotes, `UNION`, `--` and other SQL-like tokens) and needed URI-path exclusions for `/api/strategies/generate` and `/api/chat` before it could be flipped to enforce.

This PR does exactly that, and nothing else:

1. **Adds a path-scoped exclusion** via a `scope_down_statement` on the SQLi `managed_rule_group_statement`. The scope-down is a `not_statement` wrapping an `or_statement` of two exact `byte_match_statement` matches on `uri_path` — `/api/strategies/generate` and `/api/chat`. Net effect: the SQLi managed group evaluates (and blocks) every request **except** those two LLM endpoints, so legitimate prompt traffic on those paths is never matched by the SQLi signatures.
2. **Flips `override_action` from `count {}` to `none {}`** (COUNT → BLOCK) for that group, and rewrites the comment to document that it is now enforcing with the two documented exclusions.

The other three managed rule groups (`aws-core-rules`, `aws-known-bad-inputs`, `aws-ip-reputation`) and the rate-limit rule are untouched — they were already in BLOCK mode.

> Note on the issue body: issue #521 was written when both the CRS and SQLi groups were in COUNT mode and asks to flip both. The CRS group was already flipped to BLOCK in a prior change (`none {}`, "AUDIT I4"); only the SQLi group remained. This PR resolves that one remaining group, per the corrected assignment.

## Verify

Validated offline with OpenTofu (`tofu`, drop-in terraform-compatible; the `terraform` binary is not on PATH in this environment — OpenTofu validates the same HCL against the same `hashicorp/aws ~> 5.0` provider). No AWS creds used.

```
$ tofu fmt -check waf.tf
WAF_FMT_CLEAN

$ tofu init -backend=false -input=false
- Installed hashicorp/aws v5.100.0 (signed)
OpenTofu has been successfully initialized!

$ tofu validate
Success! The configuration is valid.
```

Anti-goal grep — the three other managed groups remain `none {}` and unchanged; the only group whose `override_action` this PR changes is `aws-sqli`:

```
$ grep -n "name *=\|override_action\|none {}\|count {}" infra/waf.tf
47:    name     = "aws-core-rules"
51:      none {} # BLOCK mode active (AUDIT I4)
69:    name     = "aws-known-bad-inputs"
73:      none {} # BLOCK mode active (AUDIT I4)
91:    name     = "aws-ip-reputation"
95:      none {} # IP reputation can block immediately — known-bad IPs
113:    name     = "aws-sqli"
122:      none {}     # <-- flipped from count {} in this PR
# no `count {}` remains anywhere
```

## Anti-goals

- Did **not** touch the other 3 managed rule groups or the rate-limit rule.
- Did **not** block the two LLM endpoints — they are explicitly excluded from the SQLi group via the scope-down.
- Did **not** modify any file outside `infra/`. `tofu fmt -recursive` reformatted four unrelated `infra/` files (pre-existing style debt); those were reverted so this PR contains only the `waf.tf` change.

## Operational note: requires `terraform apply` with AWS creds to take effect — not auto-applied by CI

This is a Terraform-only change. Merging it does **not** alter the live WAF on its own. A human-credentialed `terraform plan` / `terraform apply` (eu-west-2, S3 backend) is required to push the scope-down + BLOCK flip to the live ALB-attached Web ACL. After apply, re-check WAF `BlockedRequests` / sampled requests for the SQLi metric to confirm no false-positive blocking on the prompt endpoints.

### DIFF
--- a/infra/waf.tf
+++ b/infra/waf.tf
@@ -114,15 +114,64 @@ resource "aws_wafv2_web_acl" "main" {
     priority = 40
 
     override_action {
-      # COUNT mode — SQLi rules false-positive on LLM prompts; flip to BLOCK after
-      # adding URI path exclusions for /api/strategies/generate and /api/chat.
-      count {}
+      # BLOCK mode active (AUDIT #23). The SQLi managed rule group used to
+      # false-positive on LLM prompt bodies; it now enforces, but a scope-down
+      # statement below excludes the two LLM endpoints (/api/strategies/generate
+      # and /api/chat) so legitimate prompt traffic on those paths is never
+      # evaluated by — and therefore never blocked by — the SQLi rules.
+      none {}
     }
 
     statement {
       managed_rule_group_statement {
         vendor_name = "AWS"
         name        = "AWSManagedRulesSQLiRuleSet"
+
+        # Scope-down: only evaluate requests whose URI path is NOT one of the
+        # LLM endpoints. LLM prompts routinely contain SQL-like tokens (SELECT,
+        # quotes, UNION, --) that trip the SQLi signatures; excluding these two
+        # paths lets the group block SQLi everywhere else while leaving the
+        # prompt endpoints untouched. Exact-match on the path; the backend
+        # routes are mounted at these literal paths.
+        scope_down_statement {
+          not_statement {
+            statement {
+              or_statement {
+                statement {
+                  byte_match_statement {
+                    search_string         = "/api/strategies/generate"
+                    positional_constraint = "EXACTLY"
+
+                    field_to_match {
+                      uri_path {}
+                    }
+
+                    text_transformation {
+                      priority = 0
+                      type     = "NONE"
+                    }
+                  }
+                }
+
+                statement {
+                  byte_match_statement {
+                    search_string         = "/api/chat"
+                    positional_constraint = "EXACTLY"
+
+                    field_to_match {
+                      uri_path {}
+                    }
+
+                    text_transformation {
+                      priority = 0
+                      type     = "NONE"
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
       }
     }
 


onder-akkaya:
Superseded by the EC2-only teardown (#436, owner decision post-hackathon). The live managed-AWS tier (ALB/Aurora/ElastiCache/WAF) has been torn down for cost ($133→$42/mo) and the public DNS released; the server is offline. This PR adds/extends managed-AWS resources that no longer exist, so it is contrary to the current EC2-only direction. Closing as superseded; the terraform here can be revived from history if a managed tier is ever re-applied (per #436's 're-apply when real users arrive' note).

Issue #506, opened 2026-06-09
[contracts] Rebalance/liquidation swaps pass minAmountOut=0 (audit #3)


onder-akkaya:
## Summary
From AUDIT_2026-06-10.md, Critical finding #3. Every AMM swap in `Vault.rebalance` and `Vault._liquidateToUsdc` passes `minAmountOut = 0`:
- `contracts/src/Vault.sol:258` `ammRouter.swap(tokenOut, asset, amount, 0)`
- `contracts/src/Vault.sol:278` `ammRouter.swap(asset, tokenIn, amount, 0)`
- `contracts/src/Vault.sol:414` `ammRouter.swap(heldTokens[i], asset, tokensToSell, 0)`

With a zero floor, a sandwich bot (or the agent routing through an attacker-seeded pool) can extract nearly the full trade value. This turns rebalance authority into an economic withdraw authority — the practical custody hole behind the "non-custodial, rebalance-only" claim.

## Scope
- Derive `minAmountOut` from the on-chain `PriceOracle` price for the pair, minus a bounded slippage tolerance (e.g. `maxSlippageBps`, configurable, default ~50–100 bps).
- Bound per-rebalance trade size relative to oracle price + a deviation guard.
- Handle USDC(6)/synth(18) decimal scaling in the min-out computation.

## Acceptance criteria
- [ ] All three swap call sites pass a non-zero, oracle-derived `minAmountOut`.
- [ ] `cd contracts && forge test --match-contract VaultTest` passes, incl. a new test where a manipulated pool price makes the swap revert rather than execute at a bad price.
- [ ] A `maxSlippageBps` parameter exists and is enforced.

## Anti-goals
- Don't change deposit/withdraw share math.
- Don't widen agent authority.

## Notes
Contract change — **requires Chuan's review + redeploy**. Do NOT auto-assign to t2o2 (Solidity needs human consent per CLAUDE.md). Lane: contracts (Chuan).

PR #743, opened 2026-06-25
[chain] T2.3: externalize contract addresses + #506 slippage regression test

dbrowneup:
## Summary

**T2.3 — externalize contract addresses out of `client.py`** (roadmap mitigation for the single-human contract-ownership bus factor) **+ verify #506 slippage and add a regression test.**

> ⚠️ **Funds-adjacent — needs Dan's review + Bogdan two-eyes.** This touches on-chain config (`backend/archimedes/chain/`) and the `.env.example` env contract. No funds logic was changed.

### Part 1 — Externalize contract addresses

`ChainSettings` (pydantic-settings, `env_prefix="ARC_"`) **already** reads every address field from its `ARC_<FIELD>` env var, with the deployed Arc-testnet address as the fallback default — so the "hardcoded" addresses were really pydantic defaults that env vars override. This PR makes that contract **explicit and documented** rather than implicit, and surfaces the override vars in `.env.example`:

- **`backend/archimedes/chain/client.py`** — documents the env-override-with-hardcoded-fallback model on the `ChainSettings` docstring and per-field `# ARC_*` annotations. **Values and the public API are unchanged**: field names, the `synth_addresses` / `oracle_addresses` / `agent_account` / `owner_account` properties, and the `chain_client` / `get_chain_client()` singleton all behave identically (smoke-tested). Only the source-of-truth framing moved from "hardcoded constant" → "env var with hardcoded fallback."
- **`.env.example`** — adds the full set of `ARC_*`-prefixed contract-address override vars (all commented/optional, with the current defaults shown). Clarifies that the **running backend** reads these `ARC_*` vars, distinct from the bare-named `*_ADDRESS` vars that are consumed only by the one-off `verify_arc_e2e.py` script (left in place, now labelled).

### Part 2 — #506 slippage verification + regression test

**Finding: the non-zero `minAmountOut` slippage guard is already enforced — on-chain, in `Vault.sol`, which is the correct place.** Both swap paths pass a strictly-positive, oracle-derived floor:

- `Vault.rebalance` (`contracts/src/Vault.sol:341,362`) and `Vault._liquidateToUsdc` (`:588`) call `ammRouter.swap(..., _oracleMinOut(...))`.
- `_oracleMinOut` (`:519–536`) computes `minAmountOut = expectedOut * (BPS - maxSlippageBps) / BPS` from the on-chain oracle price (`maxSlippageBps` default 100 bps, hard-capped at `MAX_SLIPPAGE_CAP_BPS = 500`). It **reverts** (`OracleNotSet` / `InvalidOraclePrice`) rather than falling back to `minAmountOut = 0` when it can't price a leg, and `PriceOracle.getPrice()` itself reverts on stale prices.
- This placement is deliberate (Vault.sol audit note, 2026-06-14): **the agent must not be able to set the slippage floor**, or a compromised agent could pass `minAmountOut ≈ 0` and leak vault value. Existing Solidity coverage lives in `contracts/test/Vault.t.sol` ("Swap Slippage Protection (issue #506)").

The Python `ChainExecutor` therefore correctly passes **no** `minAmountOut` — it routes every rebalance through `rebalance(address[],uint256[],address[],uint256[])`, the vault entrypoint that applies the on-chain floor. **No path passes a Python-supplied `0`/`None` min.** New regression test added (per the task's "if a non-zero min is already enforced, add a regression test"):

- **`backend/tests/chain/test_executor_slippage.py`** (new, hermetic) — asserts the executor submits the no-min `rebalance(...)` signature (exactly 4 ABI params) on both the Circle and raw-key paths, and never calls `AMMRouter.swap(min)` directly. This guards against a future refactor that introduces a 0/None-min swap path that would bypass the vault's oracle floor.

### Tests added

- **`backend/tests/chain/test_chain_settings_addresses.py`** (new, hermetic) — pins both halves of the externalization contract: **(a)** defaults are used when env unset; **(b)** `ARC_<FIELD>` overrides win when set, flowing through to the `synth_addresses` / `oracle_addresses` maps. Uses `_env_file=None` + `monkeypatch` so it stays hermetic regardless of any ambient `.env`.

## Verify

```bash
export PATH=/opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin:$PATH
ruff format --check .
ruff check --select E9,F63,F7,F40,F82 .
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain -q
```

Results on this branch: `ruff format` → 326 files already formatted; `ruff check` (E9,F63,F7,F40,F82) → all checks passed; `pytest backend/tests/chain` → **70 passed, 0 failed** (11 new). Public-API smoke test confirms `chain_client` singleton + settings properties unchanged.

## Anti-goals (held)

- Did **not** change any address values, the public API, or any funds/swap logic.
- Did **not** modify any Solidity contract.
- Did **not** alter `pytest.ini`, CI config, or weaken any threshold.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -120,8 +120,13 @@ REQUIRE_SIWE_FOR_GENERATION=false
 ORACLE_INTERVAL_SECONDS=60
 
 # ── Contract addresses (Arc testnet, Chain ID 5042002 / 0x4cef52) ─────────
-# (Chain ID aligned with README/CLAUDE; addresses unchanged — verify via
-# Chuan/contracts before relying on them.)
+# (Chain ID aligned with README/CLAUDE; verify with Dan/contracts before relying
+# on them — Dan owns the contracts + deploys them.)
+#
+# These bare (un-prefixed) names are consumed by the one-off e2e verifier
+# (backend/archimedes/scripts/verify_arc_e2e.py). They are NOT what the running
+# backend reads — the backend's ChainSettings reads the ARC_*-prefixed vars
+# below. Keep them in sync if you set both.
 PRICE_ORACLE_ADDRESS=0xe1c9f2b11be97097223a66a188fca541e07873a6
 SYNTHETIC_FACTORY_ADDRESS=
 AMM_ROUTER_ADDRESS=0xd5b829f9d364a8bbe1caf6c8b19cb05371b178f4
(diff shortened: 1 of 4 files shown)


danielscoffee (approved):
LGTM, just mention here to add it to docs @rcrdoh

PR #751, opened 2026-06-26
[generation] Wire fusion engine into the live Generate path (Stack A) !minor

dbrowneup:
## The dispatch-gap diagnosis (Stack A vs B)

Two parallel generate stacks existed; only one ran the fusion engine, and the UI used the other:

- **Stack A (what the UI Generate button uses):** `POST /api/generate/start` → `generation_pipeline.run_generation`. It computed a `"fusion"` LABEL via `_pick_pipeline()` and emitted a cosmetic `pipeline_selected="fusion"` SSE event — then **always** dispatched to the single-agent 12-iteration `portfolio_agent` (`use_live = _llm_available()` → `_run_live_candidate`). **The fusion choice was thrown away** — there was no fusion runner branch.
- **Stack B (the real fusion route):** `POST /api/strategies/generate` → `_run_fusion_job` → `StrategyFusion.propose()` → `evaluate_fusion_spec()`, persisting with real DSR/PBO. But **no UI code POSTs to it** (dead GET-poller; `setFusionJobId` only ever called with `null`).

The engine itself (`agents/strategy_fusion.py`, `services/fusion_evaluator.py`) was real and unit-tested — it was simply unreachable from the product.

## What was wired in `run_generation`

Folded fusion **into the streaming pipeline** (the synthesis's preferred approach) so it surfaces through the same SSE events as the agent path:

1. **The `_pick_pipeline()` choice now drives dispatch.** When `"fusion"` is selected AND a real LLM is available AND a no-LLM viability precheck (`_fusion_can_run`: flag on + ≥2 candidate papers for the steer) passes, `run_generation` dispatches to a new `_run_fusion_candidate`. It builds a `FusionBrief` from the incoming `GenerateBrief`, calls the **existing** engine (`StrategyFusion.propose()` → `evaluate_fusion_spec()` — reuse, no duplication), and emits the **same** streaming events (`agent_iteration`/`tool_called`/`tool_result`/`candidate_drafted`/`candidate_evaluated`/`persisted`) so the UI renders it transparently with real DSR/PBO/OOS.
2. **Persists `generation_method="fusion"` with ≥2 `source_arxiv_ids`.** `_CandidateResult` gained `generation_method` / `source_arxiv_ids` / `has_real_rigor`; `_persist_candidate` now persists the candidate's own method (was hardcoded `portfolio_agent_streaming`). The episodic-memory + passport writes follow the same method.
3. **The single-agent path remains the FALLBACK.** Fusion not selected / no LLM / corpus <2 papers → `pipeline_selected` is **honestly relabeled** to `"agent"` (no silent mislabeling) and the agent path runs. A runtime fusion decline (`FusionUnavailable`, e.g. the LLM fused <2 valid papers) falls back per-candidate and re-emits a corrected `pipeline_selected`.
4. **`_patch_pbo` + the buy-and-hold backtest step skip fusion candidates** — they already carry a real CSCV PBO over their own parameter-variant grid and a real DSL backtest verdict; the buy-and-hold (weight-less) path would clobber those.

## Claim-integrity copy fix

`Generate.jsx` claimed the backend "auto-routes between Fusion/Architect/Agent" — false today (fusion never ran). Corrected the two comments to describe the choice **now driving dispatch** through the SSE stream, and that `pipeline_selected` carries the pipeline that **actually ran**. `pipeline_selected="fusion"` now fires only when fusion actually runs.

## The e2e test + the patch-path-bug verdict

- **Hermetic e2e test** (`test_fusion_dispatch_end_to_end_persists_fusion_strategy`): drives `run_generation` with `fusion_enabled=true` + a 24-paper fixture corpus + a MOCKED LLM backend (no network), and asserts the persisted strategy is `generation_method=="fusion"`, cites ≥2 `source_arxiv_ids`, and carries a non-null rigor verdict (real DSR/PBO/OOS from the DSL backtest). Plus `test_fusion_falls_back_to_agent_when_no_llm` proving the agent FALLBACK still runs (honestly relabeled) when fusion can't run.
- **The "patch-path bug" from the diagnosis was NOT real.** The diagnosis suspected `test_api_routes.py` patches `archimedes.services.strategy_fusion.default_fusion` while `_run_fusion_job` imports from `archimedes.agents.strategy_fusion`. But `services/__init__.py` re-exports `from archimedes.agents import strategy_fusion as strategy_fusion` — the two paths are the **same module object**, so the mock binds correctly (verified: `agents.strategy_fusion.default_fusion is <the patch>` → True). Left those tests as-is; they pass.

## Validation

- New e2e + fallback tests pass under the hermetic gate (`env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest ... -q`).
- Existing modules green: `test_generation_pipeline.py` (18 incl. new), `test_strategy_fusion.py`, `test_fusion_quality_scorer.py`, `test_generation_gating.py`, `test_api_routes.py::TestFusionEvaluatorIntegration`, plus `test_engine_v2 / test_strategy_store / test_strategy_publisher / test_api_routes` (82 + 75 passing across the sweeps; 2 pre-existing skips).
- ruff blocking gate clean (`E9,F63,F7,F40,F82` + `ruff format --check`). Generate.jsx changes are comment-only (no logic change).
- Agent FALLBACK confirmed working when fusion isn't selected.

**Core generate path — Dan review before merge; recommend a live-box smoke after deploy.**

### DIFF
--- a/backend/archimedes/agents/generation_pipeline.py
+++ b/backend/archimedes/agents/generation_pipeline.py
@@ -31,7 +31,7 @@ import logging
 import math
 import os
 from collections.abc import Awaitable, Callable
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from typing import Any
 
@@ -108,6 +108,40 @@ def _pick_pipeline(
 
     # ── Agent (fallback) ──
     return "agent", "streaming agent — general-purpose fallback"
+
+
+def _fusion_can_run(brief: GenerateBrief) -> bool:
+    """No-LLM viability precheck for the fusion path.
+
+    Returns True iff the fusion flag is on AND the deterministic candidate
+    selection yields ≥ ``MIN_PAPERS`` papers for this brief's steer — i.e. the
+    engine would not immediately return ``insufficient_corpus``. Used to keep
+    ``pipeline_selected`` honest: we only announce "fusion" if it can actually
+    produce a ≥2-paper synthesis. Never raises — any failure degrades to False
+    so the caller falls back to the agent path.
+    """
+    try:
+        from archimedes.agents.strategy_fusion import (
+            MIN_PAPERS,
+            FusionBrief,
+            fusion_enabled,
+            load_corpus,
+            select_candidates,
+        )
+
+        if not fusion_enabled():
+            return False
+        fusion_brief = FusionBrief(
+            asset_classes=list(brief.asset_classes or []),
+            risk_appetite=brief.risk_appetite,
+            strategic_direction=brief.intent or "",
+            max_papers=brief.max_papers,
+        )
+        candidates = select_candidates(fusion_brief, load_corpus())
+        return len(candidates) >= MIN_PAPERS
+    except Exception:
+        logger.debug("fusion viability precheck failed; treating as not runnable", exc_info=True)
+        return False
 
 
 # ── Brief validation (real LLM step on the live path) ─────────────────────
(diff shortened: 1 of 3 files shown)


dbrowneup:
Rebased onto `main` (which now carries #748 functional model selection + #723 premium-tier gating). The conflict landed exactly at the #748 × #751 seam — resolved as follows:

### `backend/archimedes/agents/generation_pipeline.py`
- #751 relocated the `use_live`/`runner` resolution **above** the dispatch decision so a `"fusion"` pipeline can set `runner = _run_fusion_candidate`. #748 had a **second** `use_live`/`runner` assignment lower down — I kept that region's **`job_agent`** (the per-job free-tier model binding) but **dropped the duplicate `runner = ...`**, which would have clobbered the fusion runner back to the agent path. Added a comment marking why it must not be redefined there.
- **Signature parity (the real bug):** #748 gave `_run_live_candidate` / `_run_fixture_candidate` an `agent=` kwarg; #751's new `_run_fusion_candidate` lacked it, so the unified call site `runner(..., agent=job_agent)` would `TypeError` **whenever fusion was selected**. Added `agent: Any = None` (ignored — the fusion path builds its own client) to `_run_fusion_candidate` for parity.
- The fusion→agent **runtime fallback** now also passes `agent=job_agent`, so a fallback doesn't silently drop the user's selected free-tier model.
- Removed a dead `# noqa: S101` (the `S` ruleset isn't enabled, so `RUF100` flagged it).

### `ui/src/components/Generate.jsx`
- Comment-only conflict; merged both notes (the pipeline choice now **drives dispatch** + the free-tier `model` field is sent only when picked).

### Verification
- `test_generation_pipeline.py` + `test_generate_model_gating.py` + `test_llm_backend_model_select.py` + `test_model_gating.py` → **49 passed** (the fusion e2e test and the model-gating suite green *together* — that's the integration proof).
- `ruff check` / `ruff format --check` clean. Contract suite untouched.

Kept as `!minor` (not a release) — per our call, v1.0 waits until fusion + a real corpus produce a genuine paper-grounded strategy live. Ready for your review.

Issue #788, opened 2026-06-28
[agent-ui] Agent-facing API path — dogfood the journey + capture the agent-user segment


dbrowneup:
## Summary

Archimedes has exactly one interface: a passkey/wallet React SPA. That means **AI agents — the user segment the Agora thesis says are "the new citizens" — literally cannot use the product or convert.** This issue builds an **agent-facing path** that exercises the same user journey (generate → rigor → deploy → monitor) programmatically. It is a *triple* win:

1. **Testing force-multiplier** — an agent (this Claude session) can dogfood the real journey end-to-end and flush the bugs a human will also hit, fast and repeatably.
2. **Captures the agent-user segment** — the fastest conversion win we may have, because agents are programmatic and don't carry the human trust/friction barriers (no passkey UX, no "do I trust this site" hesitation).
3. **Improves the agent's own UX** with the site (better tool affordances, clearer errors, machine-readable responses).

The READ + GENERATE paths need **no signing** (`REQUIRE_SIWE_FOR_GENERATION` defaults off), so the first slice is dogfoodable *today* over the existing `/api/*`. The on-chain steps (deploy/monitor) need a programmatic signer because agents can't do browser passkeys — that's the design problem to solve in slice 2.

## Scope

**Slice 1 — Dogfood READ + GENERATE (no signing; do this first):**
- [ ] A documented agent-facing API surface map (`docs/agent-api.md`): the exact `/api/*` calls to drive `landing → generate (start + SSE stream) → read rigor verdict/passport → list candidates/considered-alternatives`, with example requests/responses.
- [ ] A thin reference client script (`scripts/agent_journey.py` or similar) that an agent runs to: start a generation, stream events, fetch the resulting passport + rigor gate verdict, and print a structured summary. This is the dogfooding harness — run it against prod and file every bug/UX gap it surfaces.
- [ ] Confirm the existing `X-Internal-Agent-Key` / `INTERNAL_AGENT_API_KEY` classification (already in `telemetry_middleware.py`) is the right agent-auth primitive, or specify the gap.

**Slice 2 — Agent-auth + programmatic signer for DEPLOY + MONITOR:**
- [ ] An agent-auth path that is NOT a browser passkey (API key or an agent-held EOA doing SIWE programmatically against `GET /api/auth/nonce` + `POST /api/auth/verify`).
- [ ] A programmatic wallet/signer the agent controls (testnet dev key, never a funded mainnet key) so it can sign the deploy/rebalance transactions.
[...]

Issue #787, opened 2026-06-28
[conversion] Zero-conversion: convert native traffic — funnel instrumentation + value-before-wallet + faucet/demo


dbrowneup:
## Summary

We have a **zero-conversion problem**, and we have it *before* doing any public promotion — which makes it the cheapest, highest-leverage thing to fix now. The live funnel (pulled from prod this session) is brutal:

```
~40k cumulative requests (telemetry #428 — TRAFFIC, mostly crawlers, NOT users)
        │
        ▼
   2 distinct wallets (user_profiles)
        │
        ▼
   2 vaults deployed · 4 strategies · 0 chat messages · 0 live generation jobs
```

The product has **essentially never been used by a real external human**, despite steady traffic. If we can lift conversion from our *native* (un-promoted) traffic, that's a real win on its own — and it positions us far better for a public launch. This issue tracks the conversion effort as a set of concrete, shippable levers. Item 1 (instrumentation) is the measurement foundation and ships first; the agent-user lever (item 4) is split into its own issue (agent-facing API).

**Root causes (hypotheses to confirm with item 1):**
- Traffic is bots, not humans — top-of-funnel is junk; we can't tell how many *real* humans land.
- Testnet → no real-money motivation; "deploy" can dead-end on "I have no funds."
- Wallet-connect is an upfront trust/friction wall asked *before* the user has seen any value.
- Agents literally cannot convert today (no programmatic path — see the agent-UI issue).
- **We are flying blind**: no per-stage funnel instrumentation, so we don't know *where* humans drop.

## Scope (the 4 levers tracked here; lever 4-agent-UI is a separate issue)

[...]

Issue #588, opened 2026-06-13
[contracts][infra] Redeploy hardened contract suite to Arc testnet + ABI refresh + backend commit-reveal wiring


onder-akkaya:
## Summary
Umbrella for shipping the merged audit-remediation contract work to the live Arc testnet stack. PRs #582 (#509 SyntheticVault pro-rata), #583 (#507 Vault dead-shares inflation guard), #585 (#506 oracle-derived swap minAmountOut), #586 (#508 PriceOracle deviation bound), #584 (#510 commit-reveal registry) are all merged to `main`, but the deployed contracts on Arc testnet (Chain 5042002) still run the pre-fix bytecode, and `contracts/abis/` deliberately mirrors the LIVE deployment — so the backend/UI still consume old ABIs.

## Scope
1. Redeploy: `Vault` (via `VaultFactory` as appropriate), `SyntheticVault`(s), `PriceOracle`(s) (7 instances), `ReasoningTraceRegistry`.
2. Refresh `contracts/abis/` from the new artifacts; update deployed-address config consumed by backend + UI.
3. Wire backend `trace_publisher` to the commit → execute → reveal sequence (v1.5 hop; out of scope of #584 by design).
4. Small contract cleanups to fold in at the same time (no separate redeploy): emit an event when SyntheticVault redemption applies a pro-rata haircut; extract the duplicated burn/preview cap logic into an internal helper (both noted in #582 review).
5. Smoke-test on testnet before cutting the live stack over (CLAUDE.md smoke-test rule).

## Acceptance criteria
- [ ] New addresses recorded in config; `contracts/abis/` matches deployed bytecode.
- [ ] One end-to-end commit → rebalance → reveal demonstrated against the new registry on testnet.
- [ ] Live UI deposit/withdraw round-trip works against the new Vault.

## Notes
Until this lands, the live testnet instances still carry audit findings #3, #4, #13, #14, #18 — the fixes exist only in the repo.

onder-akkaya:
**Update from 15jun-audit.md (Finding 1, 🔴 CRITICAL, live) — scope addition for this redeploy.**

Today [#646](https://github.com/a-apin/archimedes/pull/646) (AUDIT B1, merged 19:26 UTC)
changed `createVault(...)` from 5 args to 6 (`address _vaultOwner`) and the backend
started encoding the new 6-arg selector. The `VaultFactory` actually deployed on Arc
[...]

PR #790, opened 2026-06-28
[agent] Agent journey harness + agent-API doc — slice 1 (no signing) !minor

dbrowneup:
## What & why

Archimedes ships **one** human interface — a passkey/wallet React SPA. AI agents (the "new citizens" of the Agora thesis) can't drive a browser passkey, so today they **can't use the product or convert**. This is **slice 1** of the agent-facing path (#788): drive the SAME journey a human does — **land → read library → generate → read the externalized rigor verdict** — over the public `/api/*` surface with **zero browser/wallet**.

The READ + GENERATE paths need **no signing** (`REQUIRE_SIWE_FOR_GENERATION` is off), so this runs against prod **today**.

## Two payoffs
1. **Dogfooding / testing force-multiplier** — running the journey as code surfaces the bugs and rough edges a human would hit (already found one: the public library is **20 strategies**, distinct from the 4 user-generated passports).
2. **Captures + measures the agent-user segment** — the client sends an agent `User-Agent`, so the telemetry classifier counts it as an external agent and the conversion funnel (#787) attributes its `generation_started`. This may be the *fastest* conversion lever — agents have no trust/friction barrier.

## Contents
- **`scripts/agent_journey.py`** — reference client + dogfooding harness. Nonzero exit on hard failure → doubles as a journey smoke test. Verified read-only against live prod.
- **`docs/agent-api.md`** — the agent-facing contract for the journey + the slice-2 plan (programmatic signer for deploy/monitor).

## Verify
```bash
python scripts/agent_journey.py --base https://archimedes-arc.com --read-only   # cheap smoke
python scripts/agent_journey.py --base https://archimedes-arc.com               # full journey (Nova Micro, sub-cent)
```

## Anti-goals
- No contract work — slice 2 (agent-auth + programmatic signer for deploy/monitor) is deferred to after the #588 contract-redeploy keystone.
- Does not weaken or bypass the rigor gate or `REQUIRE_SIWE_FOR_GENERATION` — the agent sees the SAME externalized rigor verdict a human does.

Part of #788 (slice 1).

### DIFF
--- a/docs/agent-api.md
+++ b/docs/agent-api.md
@@ -1,0 +1,116 @@
+# Agent API — driving the Archimedes journey programmatically
+
+> Status: slice 1 (READ + GENERATE, no signing). Tracks [issue #788](https://github.com/a-apin/archimedes/issues/788).
+> Slice 2 (agent-auth + programmatic signer for DEPLOY + MONITOR) is pending.
+
+Archimedes ships one human interface: a passkey/wallet React SPA. AI agents — the
+"new citizens" of the Agora thesis — can't drive a browser passkey, so today they
+**can't use the product or convert**. This document is the agent-facing contract:
+the exact `/api/*` calls that exercise the same journey a human does. The reference
+client is [`scripts/agent_journey.py`](../scripts/agent_journey.py).
+
+Two reasons this matters:
+1. **Dogfooding** — running the journey as code surfaces the bugs and rough edges a
+   human would hit, fast and repeatably.
+2. **The agent-user segment** — an agent that sends a non-browser `User-Agent` is
+   classified as an external agent by the telemetry middleware and its
+   `generation_started` is attributed in the conversion funnel ([#787](https://github.com/a-apin/archimedes/issues/787)).
+
+The **READ** and **GENERATE** paths need **no authentication**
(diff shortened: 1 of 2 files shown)


dbrowneup:
Addressed Copilot review (commit pushed):
- **SSE read timeout** — the stream now uses `httpx.Timeout(timeout_s, read=None)`, disabling the read timeout so a long quiet gap during a slow LLM/tool call can't trip a spurious `ReadTimeout`; the explicit `deadline` check bounds the stream instead. Finite connect timeout retained.
- **docs/agent-api.md slice-2** — fixed the mismatched backticks and corrected the signer description: the agent-held EOA signs **SIWE** for `POST /api/auth/verify` (session cookie), which then authorizes the wallet-gated `POST /api/vaults/create`, then `GET /api/vaults/{address}/health` to monitor.

Re-review please. 🙏

dbrowneup:
Round-2 Copilot feedback addressed (pushed) + threads resolved: short explicit connect/write/pool timeout on the SSE stream (read=None stays); JSON/required-key guard on `/start`; a 200-with-non-JSON body on a required surface is now a hard fail; dropped the PR-ref'd funnel message. Re-review please. 🙏

PR #792, opened 2026-06-28
[security] Strict per-IP daily cap on wallet-less generations (anti-abuse)

dbrowneup:
## Why (Dan's ask)

The Generate path is intentionally **open** (`REQUIRE_SIWE_FOR_GENERATION=false`) so visitors — humans **and** agents — can try the product before connecting a wallet (value-before-wallet, #787). But an open LLM-spending endpoint is a **flood/abuse risk**: an agent in a loop could hammer `/api/generate/start` and burn LLM budget. *"Let them try it cheaply, but hard-cap the flood."*

## What

- **`services/generation_quota.py`** — `GenerationQuota`: a fail-open, Redis-backed **per-IP daily** counter for wallet-less generations.
  - **Authenticated (SIWE-wallet) callers bypass the cap entirely.**
  - Keyed on the **real client IP** via the nginx-set, **non-spoofable** `X-Real-IP` header (uvicorn runs without `--proxy-headers`, so `request.client` is nginx, not the client — nginx overwrites `X-Real-IP` from its `real_ip`-resolved `$remote_addr`, bound to the trusted ALB CIDR). IP-keyed on purpose: a cookie-keyed cap would be defeated by an agent dropping the `archimedes_vid` cookie.
  - On the cap being hit → **HTTP 429 with a structured steering payload** (`{message, reason: "wallet_less_generation_cap", cap}`), so the cap doubles as a **connect-wallet conversion prompt** rather than a dead end.
- **Defense in depth:** sits *on top of* the existing slowapi (`5/min`) + nginx per-IP `limit_req_zone` limits, which bound the burst **rate**; this bounds the wallet-less daily **volume**.
- **Fail-open:** a Redis blip returns "allowed" (logged WARNING) — the slowapi/nginx rate limits remain the backstop, and blocking *all* wallet-less generation during a cache outage would be worse than the narrow window. (Happy to switch to fail-closed if you'd rather — it's a one-line change; flagging the tradeoff.)
- Wired into `start_generation` **before any work** (a capped request burns nothing). Disabled under `TESTING` (conftest sets it), matching the slowapi limiter.
- **`WALLET_LESS_GENERATION_DAILY_CAP`** (default **5**; `0` disables) documented in `.env.example` — tunable without a code change.

## Verification
```
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_generation_quota.py -q   # 17 passed
```
17 hermetic tests (mock the Redis boundary): under/at/over cap, TTL stamped once, fail-open, real-IP resolution (X-Real-IP → XFF → socket → unknown), authenticated bypass, cap-disabled, the 429 payload, and the env config. ruff clean; import graph verified.

## Review note
**Security-adjacent + gates the open Generate path — requesting review, not auto-merging.** Two knobs worth your eyes: the **cap value** (5/day — "try it a bit" vs stricter) and the **fail-open vs fail-closed** choice on a Redis outage. Both are trivial to change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -142,6 +142,17 @@ INTERNAL_AGENT_API_KEY=
 # closing the IP-rotation budget-drain vector (audit 2026-06-13).
 REQUIRE_SIWE_FOR_GENERATION=false
 
+# Strict per-IP DAILY cap on WALLET-LESS generations (anti-abuse). The Generate
+# path is open by default (REQUIRE_SIWE_FOR_GENERATION=false) so visitors —
+# humans and agents — can try it before connecting a wallet, but an open
+# LLM-spending endpoint is a flood risk. A wallet-less caller gets this many free
+# generations per UTC day (keyed on the real client IP); past it they get a 429
+# steering them to connect a wallet (so the cap doubles as a conversion prompt).
+# Authenticated (SIWE-wallet) callers bypass the cap entirely. Sits on top of the
+# slowapi (5/min) + nginx per-IP rate limits, which bound the burst RATE; this
+# bounds the wallet-less daily VOLUME. Set to 0 to disable (unlimited).
+WALLET_LESS_GENERATION_DAILY_CAP=5
+
 # How often the oracle runner pushes prices (seconds, default 60).
 ORACLE_INTERVAL_SECONDS=60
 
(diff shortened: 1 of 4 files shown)


dbrowneup:
Round-2 Copilot feedback addressed (pushed) + threads resolved: TTL is now `EXPIRE … NX` on every hit (self-heals a missing TTL, never slides the window) and a TTL failure no longer discards the INCR count; `client_ip` drops the spoofable X-Forwarded-For entirely (X-Real-IP → socket → unknown) and validates the value as a real IP. Tests updated (19 passed). Re-review please. 🙏

PR #809, opened 2026-06-29
[quant] Report CPCV as explicit NOT_RUN, not bare MISSING (#771)

### LINKED ISSUE #771: [quant] Wire CPCV into the selection-bias surface, or scope it honestly
## Summary

`api/selection_bias_routes.py:201-207` reports CPCV (Combinatorial Purged Cross-Validation, López de Prado 2018) as **MISSING** while the selection-bias surface presents itself as a fuller method set (DSR, library-PBO via CSCV, purged k-fold). A surface that names a method it does not run violates the claims-must-be-true rule the rigor gate exists to enforce: a judge or user reading the gate output should never see a method advertised that produces no number. We already ship the underlying machinery — purged k-fold splitting (`purged_kfold.py`) and CSCV library-PBO (`services/rigor_evaluator.py` `compute_library_pbo`, `analytics-engine/scripts/compute_library_pbo.py`) — so CPCV is reachable by combining the existing purged splits combinatorially. This issue is a fork: **either** implement CPCV by reusing that machinery and surface a real number, **or** replace the `MISSING` label with an honest scope note so the surface never claims a method it doesn't compute. Do not fabricate a CPCV value to make the surface look complete.

## Scope

- `backend/archimedes/api/selection_bias_routes.py` (lines ~201-207: the CPCV `MISSING` block and the `/api/selection-bias/gate` response shape)
- `backend/archimedes/services/rigor_evaluator.py` (CPCV path if implemented, reusing `compute_library_pbo` / purged-fold loaders)
- `backend/archimedes/services/purged_kfold.py` (combinatorial split source if implemented)
- `backend/tests/test_selection_bias_routes.py` (or the nearest sibling test file)

## Acceptance criteria

Pick exactly one fork and satisfy its checkbox.

- [ ] **Implement:** `/api/selection-bias/gate` returns a real `cpcv` block (no `MISSING`), derived from existing purged splits — `pytest backend/tests/test_selection_bias_routes.py -q -k cpcv` → `passed, 0 failed`.
- [ ] **Honest-scope:** the `MISSING` label is replaced with an explicit scope note (e.g. `status: "not_run"` + reason string) and no field implies a computed CPCV value — `grep -n "MISSING" backend/archimedes/api/selection_bias_routes.py` → no match; `pytest backend/tests/test_selection_bias_routes.py -q -k cpcv` → `passed, 0 failed`.

## Verify

```
pytest backend/tests/test_selection_bias_routes.py -q -k cpcv
grep -rn "CPCV\|cpcv" backend/archimedes/api/selection_bias_routes.py
```

## Anti-goals

- **Do not fabricate a CPCV number.** If not computed, the surface must say so — not emit a plausible-looking placeholder float.
- Do not weaken existing rigor thresholds (DSR, library-PBO, purged k-fold) or change the gate's pass/fail logic for other methods.
- Do not touch `pytest.ini`, the daily-returns store, or unrelated routes.
- Do not remove or alter the existing DSR / library-PBO reporting to make room for CPCV.

mnemonik-dev:
## Summary
`/api/selection-bias/gate` advertised a `cpcv` field but every strategy reported a bare `"MISSING"` — a method named on the surface that silently produced no number, which violates the **claims-must-be-true** rule the rigor gate exists to enforce.

This takes **fork (b) "honest-scope"** from #771, scoped to CPCV only (the "A1" approach, agreed with the reviewer): when `run_rigor_gate` is called **without** a combinatorial OOS matrix — the live route's path, because the analytics-engine doesn't emit one yet and CPCV is mathematically invalid on a single static return series — the gate now reports an explicit:

```
NOT_RUN (no combinatorial OOS matrix supplied; CPCV is invalid on a single static return series)
```

Fixed at the **single source** (`rigor_evaluator.py`) so every caller of `run_rigor_gate` is honest, not just the one route. The two now-stale "reported as MISSING" comments in `selection_bias_routes.py` are updated to match.

## Deliberate non-goals
- **The literal `grep "MISSING" selection_bias_routes.py -> no match` acceptance is intentionally NOT met.** The `dsr/pbo/oos = "MISSING (no backtest data)"` labels are left untouched, because removing them collides with the issue's own anti-goal *"do not alter the existing DSR/library-PBO reporting."* So this resolves the CPCV-scoped honesty problem and leaves the other methods' (already-honest, data-scoped) labels alone.
- **Real CPCV (fork a) is out of scope** — it requires the analytics-engine to emit per-combinatorial-split OOS returns; synthesizing "paths" from one realized series is exactly the *"do not fabricate a CPCV number"* anti-goal.
- The no-data branch is unchanged: there every field is missing because there is no data (honest), which is not the "advertises a method it doesn't run" problem #771 targets.

## Verify
```
pytest backend/tests/test_selection_bias_routes.py -q -k cpcv      # 3 passed
pytest backend/tests/test_selection_bias_routes.py backend/tests/test_rigor_evaluator.py -q   # 209 passed
ruff format --check . && ruff check .
```
`TestCpcvHonestNotRun` covers: honest NOT_RUN label + reason, not-a-fabricated-value (anti-goal), and that CPCV stays non-gating when absent. Tested via `run_rigor_gate(...)` directly — the exact call the route makes (`selection_bias_routes.py:219`), a verified pass-through.

Fixes #771

https://claude.ai/code/session_01Mo5BWX3Utm5BCeSZMcUyPS

### DIFF
--- a/backend/archimedes/api/selection_bias_routes.py
+++ b/backend/archimedes/api/selection_bias_routes.py
@@ -129,8 +129,10 @@ async def evaluate_rigor_gate():
     per-split OOS returns that comes from re-running the full backtest engine
     across combinatorial window splits.  That rolling re-backtest pipeline is
     not yet wired here, so run_rigor_gate() is called without cv_returns_matrix
-    and CPCV is honestly reported as MISSING on every strategy.  Wire it once
-    the analytics-engine supports combinatorial window output.
+    and CPCV is reported as an explicit NOT_RUN status (with the reason) on every
+    strategy — never a bare "MISSING" that would imply a method silently producing
+    no number (#771).  Wire it once the analytics-engine supports combinatorial
+    window output.
     """
     strategies = _provider.list_strategies()
 
@@ -215,7 +217,8 @@ async def evaluate_rigor_gate():
 
         # cv_returns_matrix intentionally omitted — CPCV requires a 2-D array
         # of per-combinatorial-split OOS returns that the analytics-engine does
-        # not yet produce.  run_rigor_gate() will report cpcv as MISSING.
+        # not yet produce.  run_rigor_gate() reports cpcv as an explicit NOT_RUN
+        # status with the reason (#771), not a bare "MISSING".
         gate_result = run_rigor_gate(
             strategy_id=s.id,
             daily_returns=daily_returns,
--- a/backend/archimedes/services/rigor_evaluator.py
+++ b/backend/archimedes/services/rigor_evaluator.py
@@ -473,7 +473,14 @@ class RigorGateResult:
             else:
                 details["cpcv"] = f"FAIL (OOS+ only {self.cpcv_positive_fraction:.0%} of paths, need ≥ 50%)"
         else:
-            details["cpcv"] = "MISSING"
+            # Honest not-run status (#771): the surface must never advertise CPCV as a
+            # method while silently emitting a bare "MISSING". CPCV needs a 2-D (S, T)
+            # matrix of per-combinatorial-split OOS returns; it is mathematically invalid
+            # on a single static return series, so when no matrix is supplied we say so
+            # explicitly rather than implying a method that produced no number.
+            details["cpcv"] = (
+                "NOT_RUN (no combinatorial OOS matrix supplied; CPCV is invalid on a single static return series)"
+            )
 
         details["look_ahead"] = "PASS" if self.look_ahead_passed else "FAIL"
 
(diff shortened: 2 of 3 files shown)


dbrowneup:
Reviewed on Dan's behalf — **merging.** Correctly takes the honest-scope fork: CPCV (the method 771 flagged as advertised-but-bare-`MISSING`) is now an explicit `NOT_RUN` with its reason, and the tests lock the anti-goals (not a fabricated PASS/FAIL, stays non-gating). 

One note for the record: 771's literal `grep MISSING → no match` acceptance is over-broad and conflicts with its own anti-goal ("do not alter DSR/library-PBO reporting"). The remaining `MISSING` strings in selection_bias_routes.py are the *other* fields' honest **no-data** states (`MISSING (no backtest data)` / `(no code)`) and unreachable Pydantic defaults — not the advertised-method-no-number problem 771 targets. Scoping the fix to CPCV (respecting the anti-goal) was the right resolution of that tension.

PR #808, opened 2026-06-29
[docs] T1.1 v2 spec — staged-replacement debate society + trading-society draft (resolves #766 review)

dbrowneup:
## What

The **v2 rewrite of the T1.1 multi-agent debate spec**, incorporating the full #766 review (Copilot + the commissioned Claude-session review) and your decisions this session. Plus a **draft of the second society** (execution/trading agents). Docs only — for your review, not yet built.

### Decisions baked in
- **Staged replacement** (your call): the debate society becomes the *sole* generation pipeline; fusion folds in as the proposer's internal capability; architect + single-agent paths deleted. **Phase 1 is strictly additive behind `ARCHIMEDES_DEBATE_ENABLED` (default OFF)** so flag-OFF is byte-identical; the **deletions move to a Phase-3 cutover PR** once the society is proven on the live path.
- **10 candidates, all backtested**: larger proposed pool → bull/bear + deterministic critics cull/rank → **top-10 leaderboard, all real-backtested** (0 LLM); **K=1 survives only at the external gate + anchor**.
- **`num_trials` is OPEN — Önder owns** the DSR-deflation denominator (the spec does *not* assert `len(pool)`).
- **Source-accuracy fixes**: `FusionProposal.is_actionable` (not `FusionEvalResult`); embargo/decay live inside `load_papers_from_db`; **symbol refs, not line numbers**; `strategy_fusion` is in `agents/`.
- **Corpus population in-scope** (real prerequisite — metadata-only today; embeddings tracked #778).
- **Lambda/metered-nanopayment** scale-out for the backtest/stats tier noted (freemium: free cap + wallet-to-pay) as Phase-2 infra.

### ⚠️ Six fixes gate Phase 1 (from the adversarial verify pass)
**A1 (highest-stakes):** thread the real `pool_size` into the external gate — today `api/selection_bias_routes.py` derives `num_trials` from *library size* internally, so without this wire the strict-rigor badge is computed wrong (a false PASS = 'build on sand'). **Do not flip the flag ON until the persisted-`pool_size`→gate path is proven.** A2–A6: additive-Phase-1 deletions deferral, the `StrategyFusion(model=...)` seam, a divergence test, a pre-backtest DSL-conformance guard, and the precise denominator (Önder).

### Also added
[...]

Issue #778, opened 2026-06-28
[corpus] Claim-integrity: "10,000 papers" is manifest-only — honest surface + RAG/KG build (hybrid)


dbrowneup:
## Summary
The live `/health` reports `corpus_papers: 10000, corpus_db_count: 10000` — but `corpus_source: "file"`, `paper_rag: degraded` (TF-IDF fallback, **no embeddings**), and `kg_entities / kg_relations / corpus_meta = 0`. **Reality: the 10,000 are metadata rows imported from a JSONL manifest into the `papers` table — there are no embeddings, no S3 artifacts, and no knowledge graph.** Presenting this as a "10k-paper corpus / knowledge graph / RAG" in the UI or pitch is a **claim-integrity violation** (the #1 rule — a judge disproves it by reading `/health`).

### A. Honest surface (now — claim-integrity, unblocks demo truth)
- [ ] Relabel the count as **"10,000 paper *metadata records* (manifest)"** in `/health` and any UI surface; don't imply embedded/graphed.
- [ ] Verify `/api/corpus/*` (the KB-pipeline endpoints) return **503** honestly until a real artifact exists (per CLAUDE.md — KB graph must come from real pipeline output, not arxiv-metadata synthesis).
- [ ] Audit pitch/UI copy for "knowledge graph / RAG over 10k papers" over-claims.

### B. Build the real thing (critical path)
Per `a-apin/docs .../consolidated/corpus-rag-build-vs-buy-memo.md` → **HYBRID**:
- Custom **KnowledgeBase pipeline** as the spine (SPECTER2 embeddings + HDBSCAN clusters + REBEL/SciSpacy KG triples).
- **AWS Bedrock Knowledge Bases** as an optional faster RAG-retrieval bridge (cheap-AWS-assist; sub-dollar at our scale) — **not** Neptune GraphRAG.
- Stand up S3 corpus/artifact buckets (none exist today) + wire the honest endpoints to real output.

## Why now
Dan: this unblocks the critical path (paper-grounded provenance is the rigor wedge's substrate). Also the "lean on cheap AWS services to fill gaps" lever applies here (Bedrock KB) and to the oracle guardrail (#775).

## Acceptance criteria
- [ ] Honest surface verified (no over-claim renders as fact on the live path)
- [ ] Build path decided + scoped into follow-up issues (pipeline stages, S3 buckets, retrieval)

Owner: @dbrowneup (strategy/corpus). Collab welcome.

[...]

Issue #770, opened 2026-06-28
[quant] num_trials multiple-testing correction for the N-candidate generation society


dbrowneup:
## Summary

The Deflated Sharpe Ratio (Bailey & López de Prado 2014) deflates a strategy's Sharpe by the **number of trials** that produced the winner. Today `num_trials` is set to the library size (`generation_pipeline.py:702-708`, with the library-size read at `:1105` and a matching default in `api/selection_bias_routes.py:166`). That count was correct under the old "one strategy per Generate" model, but it is **statistically wrong** under the new agentic society (PR #766), where ~N=10 candidates are generated, **all** are backtested, **all** are rigor-gated, and the **best** is selected. Selection-from-N is itself a multiple-testing search; ignoring it understates the deflation and inflates the DSR of the survivor — exactly the selection-bias failure our Tier-1 gate exists to prevent. The winner also joins a library, so the effective trial count must reflect **selection-from-N PLUS the library context**, not one or the other.

This is Önder's statistics call: what is the correct effective `num_trials` for the society path — `N + library_size`, or an effective-N that accounts for candidate-pool return correlation (clustered/effective-breadth correction)? Decide with a written rationale, implement it, and prove an IS-overfit candidate fails on the society path.

## Scope

- `backend/archimedes/services/generation_pipeline.py` (`:702-708`, `:1105`) — derive `num_trials` for the society path.
- `backend/archimedes/api/selection_bias_routes.py` (`:166`) — keep the API default consistent with the pipeline.
- `backend/archimedes/services/rigor_evaluator.py` — DSR consumes the new count via `run_rigor_gate`.
- A short rationale note (docstring or `docs/specs/selection-bias-corrections-spec.md` addendum) recording the chosen formula and citation.
- `backend/tests/test_generation_pipeline.py` (or sibling) — new IS-overfit fixture test.

## Acceptance criteria

- [ ] `num_trials` on the society path reflects selection-from-N **plus** library context (documented formula), not library-size alone:
      `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_generation_pipeline.py -q` → `N passed, 0 failed`.
- [ ] An IS-overfit fixture (high in-sample Sharpe, low OOS) **fails** the gate on the society path:
      `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/ -k "num_trials and overfit" -q` → `1 passed` (or more), `0 failed`.
[...]

PR #815, opened 2026-06-29
feat: debate society as flag-gated generation pipeline (additive, deterministic-critic) !minor

dbrowneup:
**T1.1 Phase 1** — the multi-agent debate society, built against the spec in **#808**. Draft for your review (pairs with #808). **Strictly additive behind `ARCHIMEDES_DEBATE_ENABLED` (default OFF)** so the flag-OFF live path is byte-identical; the legacy-runner deletions are deferred to the Phase-3 cutover PR (fix A2).

## What's here
**New `agents/debate_engine.py`** — the society runner:
- **Proposer pool** — fans `StrategyFusion(model=...).propose` across `select_candidates(regime_bias=R)` evidence sets. Drops non-actionable (`FusionProposal.is_actionable`) + non-conformant (`_dsl_conformance_ok`, **A5**) specs. `pool_size = len(POOL)`.
- **C-rigor** — `evaluate_fusion_spec` per survivor (deterministic, 0 tokens), each try/except-wrapped, with **`num_trials=pool_size`** (the **A1** deflation, self-contained at the evaluator).
- **C-null** — survivor must beat the passive null by ≥ 5 bps (Phase-1 `cagr` proxy) → else **first-class ABSTAIN** (`generation_method="debate_abstain"`).
- **Deterministic synthesizer** → top-N leaderboard; returns the leader `_CandidateResult` (carrier-contract preserving; the leaderboard fan-out is Phase 2).
- Best-effort **bull/bear transcript** (never gates; full rebuttal is Phase 2). `DebateUnavailable(FusionUnavailable)` → existing fallback relabels to agent.

**Edits:** `generation_pipeline.py` (`_pick_pipeline` flag branch + dispatch + `regimes=["neutral"]` + episodic ternary), `strategy_fusion.py` (**A3** model seam: `StrategyFusion(model=…)` / `default_fusion(model=…)`), `generate_schemas.py` (mode docstring).

## Tests — 12 hermetic (`env -i` clean) + 118 regression
Covers spec §9 tests 1–9: leaderboard/`has_real_rigor`, regime divergence (A4), ABSTAIN, `DebateUnavailable`, `pool_size`→`num_trials` (A1), model threading (A3), DSL conformance (A5), flag-OFF never returns `"debate"` (A2), cited-paper union, transcript role order. Regression: **118 passed** across pipeline/fusion/selection-bias/model-gating (flag-OFF byte-identical).

## ⚠️ Two things to read before flipping the flag
1. **A1 coordination with #770/#811.** I put the `pool_size` deflation at `evaluate_fusion_spec(num_trials=pool_size)` (per-candidate), which **dovetails #770's `library + N` additive correction on the generation path** rather than fighting the library-grading external route (#811 correctly keeps that route at library size). The exact denominator is **Önder's OPEN call** (spec §5c). **PHASE-1 BLOCKER: do not enable the flag on the live path until A1 is proven there** — else the rigor badge can false-PASS.
2. **A4 finding (real).** The regime-divergence guarantee holds at the keyword+bias core, but the semantic rerank (`FUSION_SEMANTIC_RETRIEVAL`, default ON) **collapses it** on the degraded/TF-IDF corpus — the documented diversity-theater risk that needs real embeddings (**#778**). The test isolates the keyword guarantee and documents the collapse.

## Scope notes
- Phase 1 = minimal society (proposers + bull/bear + C-rigor + C-null + synth). Full deterministic panel (C-regime/GMM, C-prov), the visible rebuttal, all-10 backtests, and the Considered-Alternatives fan-out are **Phase 2**. Deletions + provenance hardening + flag-ON are **Phase 3**.
- Corpus is metadata-only today; tests run on a fixture corpus (population is in-scope, Dan driving).

### DIFF
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -404,7 +404,9 @@ def _critic_prov(pool: list[Any], corpus: list[Any]) -> tuple[list[Any], list[An
     kept: list[Any] = []
     dropped: list[Any] = []
     for prop in pool:
-        cited = set(prop.source_arxiv_ids)
+        # Robust to a proposal missing/None source_arxiv_ids — treat as empty (→ drop,
+        # "not provenance-verifiable"), never raise and abort the whole run (Copilot review).
+        cited = set(getattr(prop, "source_arxiv_ids", None) or [])
         if cited and cited <= surface:
             kept.append(prop)
         else:
@@ -542,11 +544,14 @@ def _critic_regime() -> dict[str, Any]:
     }
     try:
         from archimedes.models.regime import Regime
-        from archimedes.services.gmm_regime_detector import GmmRegimeDetector, gmm_regime_health
+        from archimedes.services.gmm_regime_detector import current_regime, gmm_regime_health
 
         health = gmm_regime_health()
         degraded = health.status != "live"
-        rc = GmmRegimeDetector().get_current_regime()
+        # Read the SHARED live detector (the one the oracle/agent runner feeds), NOT a
+        # fresh GmmRegimeDetector — a new instance has no current classification, so it
+        # would always read None and the gate would never fire (Copilot review).
+        rc = current_regime()
         regime = rc.regime if rc is not None else None
         confidence = float(rc.confidence) if rc is not None else 0.0
         force_abstain = regime == Regime.CRISIS
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -279,18 +279,26 @@ async def _propose_pool(brief: GenerateBrief, model: str | None, corpus: list[An
 # ── Step 2 — best-effort adversarial round (transcript only, never gates) ─────
 
 _DEBATE_SYSTEM = (
-    "You are the {role} researcher in a quant strategy debate. {stance}. "
-    "Cite ONLY the listed candidate strategies. Reply with ONE JSON object: "
-    '{{"verdict": "act"|"decline", "confidence": <0..1>, "key_claims": [<str>...]}}.'
+    "You are the {role} researcher in a quant strategy debate, round {rnd}. {stance}. "
+    "Cite ONLY the listed candidate strategies. {rebuttal}"
+    'Reply with ONE JSON object: {{"verdict": "act"|"decline", "confidence": <0..1>, "key_claims": [<str>...]}}.'
 )
+
+_DEBATE_STANCES = {
+    "bull": "Argue FOR acting on the strongest candidate",
+    "bear": "Argue for ABSTENTION — the null is buy-and-hold; attack overfit/cost",
+}
 
 
 async def _debate_round(pool: list[Any], model: str | None, emit: _Emitter, candidate_id: str) -> list[dict[str, Any]]:
-    """Thin best-effort bull/bear round — transcript only, never gates (Phase 1).
+    """Best-effort bull/bear research debate with ONE visible rebuttal round.
 
-    Surfaces the adversarial topology on the SSE stream. Any failure (no backend,
-    unparseable output) degrades to a neutral transcript entry. The transcript is
-    built in fixed [bull, bear] role order for R3 determinism (sort-before-hash).
+    Round 1: bull + bear state initial positions. Round 2: each REBUTS the other's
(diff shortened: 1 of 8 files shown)


dbrowneup:
**Out of draft — ready for team review.** This PR now carries BOTH the spec and the implementation:
- **Spec** (folded in from #808): `docs/specs/multi-agent-debate-spec.md` v2 (staged-replacement, with the Copilot + Bogdan review fixes) + the `execution-trading-agent-society-spec.md` trading-society draft. **This PR supersedes #808.**
- **Phase-1 implementation**: `agents/debate_engine.py` + flag-gated dispatch + the A3 model seam + 12 hermetic tests.

Rebased onto main, so it now sits on top of #770/#811's `_society_num_trials = library + N`. The dispatch threads `selection_pool_size` to every runner (#770), so `_run_debate_candidate` takes it for parity but uses its own internal `pool_size = len(POOL)` (the real conformant-proposal selection set). Regression: **138 pass**. Flag stays OFF (byte-identical) until A1 is proven on the live route.

onder-akkaya (commented):
Reviewed from the quant/rigor angle. Really clean scaffold — flag-gated OFF (`ARCHIMEDES_DEBATE_ENABLED`) so it's strictly additive, C-rigor runs a *real* `evaluate_fusion_spec` (DSR/PBO/OOS, not a proxy), and `_rigor_verdict_dict` mirrors the fusion verdict shape so the passport renders identically. The Phase-1 proxies are honestly labelled as proxies, which I appreciate. Two quant notes, both forward-looking — neither blocks the flag-OFF path:

**1. `num_trials = pool_size` (C-rigor, the A1 denominator you flagged for me).** Confirming the statistics so we can converge it: `pool_size` alone deflates for the selection-from-pool search but **not** for the library the winner joins. The live society path uses `_society_num_trials(library_size, selection_pool_size) = N + library_size` (#770/#811). So a debate winner is currently deflated *less* than the same strategy on the live path → the debate "passing" badge is **more permissive** than the live one. When we resolve A1, debate + live + fusion should all read the one `_society_num_trials` source — I've opened **#820** to unify the three conventions; happy to wire it so this PR just calls the shared helper. (Direction: this only ever tightens, so it's safe to land flag-OFF now and tighten in #820.)

**2. C-null `_survives_null` (`debate_engine.py:320-327`) — the 5 bps bar.** Agreed it's a Phase-1 proxy. The forward note for Phase 2: make the "beats the passive null" test **risk-adjusted** — an excess-Sharpe or DSR differential vs buy-and-hold — rather than raw `cagr`. A high-vol candidate can clear a 5 bps *raw-return* edge while being risk-adjusted-*inferior* to buy-and-hold, which is exactly the failure the DSR discipline elsewhere is designed to catch. Low urgency, because C-rigor's DSR/PBO is the real admission bar and C-null is a supplementary sanity check — but when the explicit buy-and-hold differential lands, I'd anchor it on a risk-adjusted statistic.

Net: nice work, safe to iterate behind the flag. The quant denominators converge through #820, and I'll take that wiring. 👍

dbrowneup:
Thanks both — addressed and rebased onto main (now sits on #829's live server-side gate).

**@onder-akkaya** — 
1. **A1 denominator**: aligned now — C-rigor's `num_trials = _society_num_trials(library_size, pool_size) = library + N`, so the debate badge is no longer more permissive than the live path. It converges cleanly on your **#820** shared helper (the debate just needs to read that single source once it lands — happy for you to wire it).
2. **C-null risk-adjusted**: agreed, tracked as the Phase-2 forward note (anchor the passive-null bar on an excess-Sharpe/DSR differential vs buy-and-hold, not raw cagr). C-rigor's DSR/PBO remains the real admission bar.

**Copilot** —
- Buy-and-hold skip now keys on `("fusion","debate","debate_abstain")`, reconciled with #829's `!= "fusion"` — debate candidates emit a DSL spec (weights={}) so they never hit the static backtester.
- `DEBATE_POOL_MAX` is now meaningful: `_STEERS` is the regime×mechanism grid (18), so the knob actually bounds the fan-out (+ a real non-corpus diversity axis).
- `build_leaderboard` caps to `DEBATE_LEADERBOARD_MAX` (default 10 — the top-10 contract).

+2 tests; 14 pass. Flag stays OFF.

PR #816, opened 2026-06-29
[t1.1][phase2] Critic panel (C-regime + C-prov) + visible bull/bear rebuttal !minor

dbrowneup:
**T1.1 Phase 2 — increment 1.** Stacked on **#815** (base = the Phase-1 branch; merge #815 first, then I'll retarget this to main). Adds the deterministic **C-regime critic** — the non-votable Hierarchy-of-Truth gate (Xia §4.4).

## What it does
- `_critic_regime()` reads the live exogenous regime (`GmmRegimeDetector.get_current_regime()` + `gmm_regime_health()`). A **live CRISIS** read **forces ABSTAIN** — market regime structurally overrides candidate consensus; no bull argument can vote it down. Crisis is exactly when you do *not* deploy a fresh strategy.
- **DEGRADED** (GMM artifact missing → VIX rule-based fallback) is surfaced honestly and lowers confidence, but does **not** by itself force abstain (else the society would always abstain whenever the GMM is unavailable).
- **Fail-safe:** any error or no-read → "unavailable, not gating" — the critic can only ever ABSTAIN, never spuriously APPROVE.
- `build_leaderboard` runs the regime gate **first** (short-circuits before C-null); `_run_debate_candidate` calls it off-thread and emits its verdict on the SSE stream.

## Tests (+5, hermetic — GMM patched at the lazy-import boundary)
CRISIS forces abstain · risk_on does not gate · no-read fails safe · degraded+non-crisis does not force abstain · the gate overrides even strong C-null survivors. **17 debate tests pass.**

## Next Phase-2 increments (separate PRs)
C-prov (embargo/provenance critic) · visible bull/bear rebuttal round · scale pool to ~15–20 · all-10 backtests + Considered-Alternatives leaderboard fan-out.

### DIFF
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -404,7 +404,9 @@ def _critic_prov(pool: list[Any], corpus: list[Any]) -> tuple[list[Any], list[An
     kept: list[Any] = []
     dropped: list[Any] = []
     for prop in pool:
-        cited = set(prop.source_arxiv_ids)
+        # Robust to a proposal missing/None source_arxiv_ids — treat as empty (→ drop,
+        # "not provenance-verifiable"), never raise and abort the whole run (Copilot review).
+        cited = set(getattr(prop, "source_arxiv_ids", None) or [])
         if cited and cited <= surface:
             kept.append(prop)
         else:
@@ -542,11 +544,14 @@ def _critic_regime() -> dict[str, Any]:
     }
     try:
         from archimedes.models.regime import Regime
-        from archimedes.services.gmm_regime_detector import GmmRegimeDetector, gmm_regime_health
+        from archimedes.services.gmm_regime_detector import current_regime, gmm_regime_health
 
         health = gmm_regime_health()
         degraded = health.status != "live"
-        rc = GmmRegimeDetector().get_current_regime()
+        # Read the SHARED live detector (the one the oracle/agent runner feeds), NOT a
+        # fresh GmmRegimeDetector — a new instance has no current classification, so it
+        # would always read None and the gate would never fire (Copilot review).
+        rc = current_regime()
         regime = rc.regime if rc is not None else None
         confidence = float(rc.confidence) if rc is not None else 0.0
         force_abstain = regime == Regime.CRISIS
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -279,18 +279,26 @@ async def _propose_pool(brief: GenerateBrief, model: str | None, corpus: list[An
 # ── Step 2 — best-effort adversarial round (transcript only, never gates) ─────
 
 _DEBATE_SYSTEM = (
-    "You are the {role} researcher in a quant strategy debate. {stance}. "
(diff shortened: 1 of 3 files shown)


dbrowneup:
**Updated — now the full deterministic non-votable critic panel (C-regime + C-prov).** Added **C-prov** (Xia 1/2/4): hard-fails any candidate citing a paper outside the embargo+decay-applied surface (`load_corpus` already excludes post-embargo papers), wired as Step 3a before C-rigor. It does **not** change `pool_size` (the DSR denominator still counts every conformant proposed spec, §5c) — it only culls which survivors reach the backtest. +2 tests → **19 debate tests pass.** Still stacked on #815.

onder-akkaya (commented):
Reviewed the C-regime / C-prov critics from the quant/rigor angle — the design here is sound and I like it.

**C-regime is correctly built.** Non-votable CRISIS → ABSTAIN as a Hierarchy-of-Truth override (Xia §4.4), it runs *first* in `build_leaderboard` and short-circuits before C-null, and the fail-safe is the right shape — any error / no-read → "unavailable, not gating," so the critic can only ever ABSTAIN, never spuriously APPROVE. Crucially, **DEGRADED not forcing abstain is the correct call** (the alternative — abstain whenever the GMM artifact is missing — would make the society perpetually abstain). No notes on the logic.

One operational / claim-integrity flag (not a logic bug):

🟡 **In prod, C-regime currently runs in the VIX-fallback (DEGRADED) path.** `gmm_model.pkl` is gitignored / local-only (#697, the commit-vs-train-on-deploy decision is still open), so `GmmRegimeDetector` falls back to the VIX rule everywhere it isn't trained. The code handles this gracefully and honestly surfaces `degraded`. The only ask: make sure the *claim surface* (UI / passport / pitch) says "regime gate (VIX-based; GMM when the artifact is deployed)" rather than implying a live GMM Hierarchy-of-Truth — otherwise we'd be asserting a model that prod isn't actually running yet. Ties directly to the open #697 deployment decision.

Stacked on #815 (base = the phase-1 branch) and flag-gated OFF, so safe to iterate. Logic LGTM from my side; deferring any contract/infra-adjacent bits to Dan/Bogdan as usual.

dbrowneup:
Addressed and rebased onto the updated #815.

**Copilot** —
- **C-regime real bug fixed**: it now reads the SHARED live detector via a new public `gmm_regime_detector.current_regime()` (the most-recently-constructed detector the oracle/agent runner feeds). The old `GmmRegimeDetector().get_current_regime()` built a fresh detector with no classification → C-regime would read None and **never gate in prod** (a no-op). Tests patch the accessor accordingly.
- **C-prov robustness**: missing/None `source_arxiv_ids` now drops the candidate ("not provenance-verifiable") instead of raising and aborting the run. +1 test.

**@onder-akkaya** — agreed on the claim-surface honesty: C-regime's reason string says "GMM degraded → VIX rule-based fallback" when the artifact isn't deployed, so the passport/UI won't imply a live GMM Hierarchy-of-Truth. That ties to the #697 GMM-artifact-deployment decision — when the model ships, the same code flips to the live read with no change. 23 debate tests pass; flag stays OFF.

PR #4, opened 2026-05-13
[infra] EC2 instance, Docker setup, and CI/CD deploy pipeline

t2o2:
## What

Adds the full infrastructure stack for deploying Archimedes to an EC2 instance, with automated CI/CD on merge to `main`.

### Terraform (`infra/`)
- EC2 t3.small in eu-west-2 (London) — already provisioned
- Security group: SSH (22), HTTP (80), HTTPS (443), FastAPI (8000), Next.js (3000)
- SSH key pair auto-generated
- User-data bootstrap: Docker CE, docker-compose plugin, repo clone

### Docker (`backend/Dockerfile`, `docker-compose.yml`)
- Backend: Python 3.12-slim + FastAPI
- Postgres 16 + Redis 7
- Health checks on all services

### CI/CD (`.github/workflows/deploy.yml`)
- Triggers on push to `main`
- SSHes into EC2, pulls latest, rebuilds containers
- Verifies health check after deploy

### Backend entrypoint (`backend/archimedes/main.py`)
- Minimal FastAPI app with `/health` and `/` endpoints
- CORS middleware for frontend dev

## Required GitHub Secrets

After merging, configure these in **Settings → Secrets → Actions**:

| Secret | Value |
|--------|-------|
| `EC2_HOST` | `35.179.100.89` |
| `SSH_PRIVATE_KEY` | Contents of `infra/archimedes-deploy-key.pem` |

## Instance details

- **IP:** 35.179.100.89
- **SSH:** `ssh -i infra/archimedes-deploy-key.pem ubuntu@35.179.100.89`
- **API:** http://35.179.100.89:8000
- **Instance ID:** i-0996b8ce9bdb08aa6
- **Cost:** ~$17/month (t3.small)

Issue #759, opened 2026-06-26
[universe] T1.5b — Expand synthetic universe to 200–300 Chainlink-covered instruments (no single-name equities)


dbrowneup:
> **STATUS: DRAFT — NOT READY TO ASSIGN.** This is a scope-review spec pending Dan's sign-off on the four OPEN QUESTIONS below. Do **not** `--add-assignee t2o2` until Dan resolves Q1–Q4 (especially Q1, the Arc/Chainlink feed-availability unknown — the whole "Chainlink-covered" framing may be metadata-only today). The Scope / Acceptance / Verify sections are a proposed shape, not a contract; expect them to change after review.

## Summary

The on-chain synthetic universe is the single source of truth at `backend/archimedes/data/synthetic_universe.json` (loaded by `backend/archimedes/universe.py`). It currently holds **59 synths** (confirmed by reading the file), of which **32 carry `chainlink_covered: true`** and 27 are `false`. Dan wants to grow this to **200–300 instruments**, sourced via Chainlink (or a Chainlink-equivalent), **excluding single-name equities** for compliance reasons (the existing `_compliance_review` block in the SSOT already keeps ~58 single-stock synths backtest-only). The expansion strengthens the demo's universe-parity + "5–10×" Lepton claim and gives the portfolio/fusion agents a much wider opportunity set.

**The honest blocker this DRAFT exists to surface:** despite the `chainlink_covered` field name, **nothing in the live path actually reads a Chainlink feed.** The on-chain oracle is a custom `contracts/src/PriceOracle.sol` whose price is owner-pushed by `backend/archimedes/chain/oracle_updater.py` via **yfinance + CoinGecko** (through the Circle Wallets API), and `oracle_updater.py`'s `YFINANCE_MAP` + `CRYPTO_MAP` only cover **7 symbols** today. Backtest prices come from the yfinance/CoinGecko tickers encoded in `GLOBAL_ASSETS` (`backend/archimedes/services/strategy_signal_evaluator.py`, 118 entries). So "Chainlink-covered" is presently a *descriptive label*, not a wired data path — and that gap (plus where 200–300 instruments' backtest history and live prices come from) is exactly what needs a decision before this is executable. A grep for `chainlink|Chainlink|AggregatorV3|latestRoundData` across `contracts/` and `backend/` returns **zero** matches outside the SSOT/loader.

## OPEN QUESTIONS (need Dan's decision before this can be assigned)

[...]

PR #749, opened 2026-06-26
T1.2 nanopayment marketplace scaffold (first draft) !minor

dbrowneup:
**@rcrdoh @mnemonik-dev — first-draft scaffold for review, NOT for merge yet.**

This is a reviewable starting point for **T1.2 (issue #713)** built in parallel with Ricardo's own work (Dan-authorized). It cuts the smallest verifiable slice of the optional-publish nanopayment marketplace: **one paywalled `POST /api/strategies/construct` endpoint** that pays a sub-cent USDC nanopayment via the **x402 protocol + Circle Gateway facilitator** into a **90/10 `RevenueSplit`** on Arc. Everything else is stubbed-but-clean so the follow-ups plug in.

## What's in this PR (per SPEC-1 T1.2)

**Backend (hermetic, no Foundry needed):**
- `backend/archimedes/api/x402_middleware.py` — a **generic, reusable** x402 paywall as a **FastAPI dependency factory**: `x402_paywall(price_usdc, recipient=...)`. Opt-in per-route, **feature-flagged default-OFF** (`ARCHIMEDES_X402_ENABLED`, no-op until armed).
  - No `PAYMENT-SIGNATURE` -> **402** + a base64 `PAYMENT-REQUIRED` header naming Arc testnet (`eip155:5042002`), the `RevenueSplit` recipient, and the USDC price (`0.001`).
  - With `PAYMENT-SIGNATURE` -> **settle** via Circle Gateway (`POST {CIRCLE_GW_BASE}/x402/settle`; reuses `CIRCLE_GW_BASE` from `circle_service.py` — not redeclared), gate the handler on `success: true`.
  - **Per-payer daily spend cap** -> **429** (24h sliding window in Redis via a `SpendCapStore`, keyed on the payer wallet from the verified payload; `ARCHIMEDES_X402_MAX_USDC_PER_DAY` default `0.10`).
- Wired to **exactly one route** — `POST /api/strategies/construct` (the canonical "consume a strategy" verb). **Spec note:** the cited line drifted — the live route is `POST /construct` with **no** `{strategy_id}` path param (strategy selection is internal to the architect), so the paywall attaches to the route as a whole. **No other endpoint touched.**
- `is_paywalled: bool` on the strategy response (`schemas.py`), set from the flag in `strategies_routes.py`.
[...]

Issue #19, opened 2026-05-16
APIN - Intelligence - Reasoning


martenwindler:
APIN - Intelligence - Reasoning

Reference
http://18.171.230.205/ 

### **User Story: Analyze Market Reasoning on APIN Intelligence**

**As a** portfolio manager or data analyst,
**I want to** access and interpret AI-generated reasoning for market strategies on the APIN Intelligence - Reasoning page,
**So that** I can make informed decisions based on transparent, data-driven insights.

Iteration 16.05.2026

Iteration 17.05.2026

<img width="1246" height="677" alt="Image" src="https://github.com/user-attachments/assets/aad4f9cc-797f-4294-8bc7-bd9081a922c3" />

- https://github.com/hackagora/archimedes-arcadia/pull/37
- https://github.com/hackagora/archimedes-arcadia/pull/38

Iteration 18.05.2026

Iteration 19.05.2026

Iteration 20.05.2026

Iteration 21.05.2026

Iteration 22.05.2026

Iteration 23.05.2026

---
---

### **Acceptance Criteria:**
- [ ] User can **view a list of reasoning models/strategies** (e.g., "Quantitative Approach to Tactical Asset Allocation," "Volatility-Managed Portfolios").
- [ ] User can **select a reasoning strategy** to view its detailed analysis, including:
  - **Author(s)** (e.g., Mebane Faber, Alan Moreira).
  - **Description** of the strategy’s logic (e.g., "Hold a long position in the asset when its monthly close is above the 10-month simple moving average").
  - **Tags** (e.g., `SPY`, `NIKKEI`, `GOLD`, `TREASURY`, `OIL`).
  - **Status** (e.g., "live," "candidate").
  - **Performance metrics** (e.g., historical returns, risk factors).
- [ ] User can **filter reasoning strategies** by:
  - Author.
  - Tags (e.g., asset class, market).
  - Status (live/candidate).
- [ ] User can **view the reasoning trace** for a selected strategy, including:
  - Step-by-step logic (e.g., "Step 1: Calculate 10-month SMA...").
  - Data sources (e.g., "Uses SPY, NIKKEI, GOLD price feeds").
  - Assumptions or constraints (e.g., "Rebalances monthly").
- [ ] User can **compare reasoning strategies** side-by-side (e.g., performance, risk, fees).
- [ ] User can **export reasoning data** (e.g., as JSON, CSV, or PDF) for further analysis.
[...]

PR #823, opened 2026-06-30
[market] Implement subscription marketplace with payment splitting and agents

dbrowneup:
This pull request adds new documentation skills files for several core and infrastructure components across the ARC Canteen, ARC Nanopayments, and Archimedes Fork projects. Each skill file provides a structured summary of when to use the component, its key files, concepts, and related skills, improving discoverability and onboarding for developers.

The most important changes are:

### ARC Canteen context and Polymarket ecosystem skills

* Added `.agents/skills/arc-canteen-context__arc-docs/SKILL.md` to document Arc blockchain developer docs, including chain specs, app-kit, agentic economy, and quickstarts.
* Added `.agents/skills/arc-canteen-context__circle-docs/SKILL.md` for Circle developer platform documentation, covering USDC, CCTP, Developer-Controlled Wallets, Gateway, and Agent Stack.
* Added `.agents/skills/arc-canteen-context__clob-client/SKILL.md`, `.agents/skills/arc-canteen-context__builder-relayer/SKILL.md`, `.agents/skills/arc-canteen-context__polymarket-sdk/SKILL.md`, and `.agents/skills/arc-canteen-context__real-time-data/SKILL.md` for Polymarket CLOB clients, builder/relayer, SDK, and real-time data feeds. [[1]](diffhunk://#diff-2e6d60709c868a128f717d5cf64faaed42c55af22afc235fb5db638061b4eaa6R1-R29) [[2]](diffhunk://#diff-942f594410ff12ff671df2cf981508951e354eb0b82a9280e1e88a77c39b569fR1-R28) [[3]](diffhunk://#diff-861e0ed7c92178db23a0db22777771330f4cf087654cb3ee2209d9febb337480R1-R28) [[4]](diffhunk://#diff-1cf0993266c565ad34fcc0a843814e046873619a28384a56470455b8dcb3d0d1R1-R26)

### ARC Nanopayments documentation skills

* Added `.agents/skills/arc-nanopayments__dashboard/SKILL.md` for the Next.js dashboard, `.agents/skills/arc-nanopayments__supabase-auth/SKILL.md` for Supabase authentication, `.agents/skills/arc-nanopayments__wallet-management/SKILL.md` for wallet generation and withdrawal, and `.agents/skills/arc-nanopayments__x402-gateway/SKILL.md` for x402 protocol integration. [[1]](diffhunk://#diff-ae0f47968cdb4f8c297672fea47a1066cdd6e89c681ae365ec11232f1a60537bR1-R34) [[2]](diffhunk://#diff-2c53a77469fb2a367e01c52091663931a2aa01c0709874e8e0770b4d9e1f6de2R1-R30) [[3]](diffhunk://#diff-90714fefa45903d7b04e88466cd833211ccb5706d740d506deb845b4e04bd969R1-R30) [[4]](diffhunk://#diff-a39a7c8add2ff7c34570c839cb9a4c32336ea31494d809fdac53ac95d50ddd9cR1-R33)

### Archimedes Fork documentation skills

* Added `.agents/skills/archimedes-fork-1__analytics-engine/SKILL.md` for the backtrader-based analytics engine, and `.agents/skills/archimedes-fork-1__backend/SKILL.md` for the FastAPI backend, including routes, services, models, chain integration, and agents. [[1]](diffhunk://#diff-a0de9887af26c58d8f1567406f61fa4f577759b4c75e91bc3f2e796be6929743R1-R38) [[2]](diffhunk://#diff-8938ade6c349448cd3c12a257ebe6a576068b9205d86e485330dac902e73ac71R1-R48)


onder-akkaya (commented):
Thanks for the tag, Dan — reviewed from the quant/funds-math angle. This is a solid, well-structured scaffold (the route-level auth tests are genuinely good, and `strategy_signal_evaluator` correctly passes `scan_full_universe=False`, so it does **not** open the GLOBAL_ASSETS full-scan path — no collision with #759/#772 👍). My comments below are about hardening the **money + rigor + contract** paths *before* this goes live, plus one scope ask.

**Important framing:** the billing/execution path is currently **inert** — `strategy_runner_publisher.py:499-510` `_compute_rebalance()` returns `([], {})` on both branches (an explicit "Simplified…" stub), so `tick()` always logs "No rebalance needed" and never charges. So none of the funds findings below are *live* fund risks today; they're things to resolve **before** the rebalance/charge path is wired and before the contracts are deployed. Flagging them now while the scaffold is still cheap to change.

### 🔴 Resolve before the billing/execution path is enabled

1. **Charge is not atomic with execution.** `_charge_subscriber()` (`publisher.py:399-430`) calls the on-chain `chargeActions` (which irreversibly splits USDC 90/10) in the per-subscriber loop, but the publisher's own `execute_trades` runs later (step 7) and a failure there is only logged. Subscribers can be billed for a rebalance that never lands. Suggest: charge only after delivery is confirmed, or add reconciliation/refund. (Latent until #2 is wired, but it's the load-bearing design question.)
2. **Rigor gate is not enforced on publish.** `/publish` (`marketplace_routes.py:145`) and `_evaluate_strategy` load a strategy by id and never check that its passport is **VALIDATED** (DSR/PBO/walk-forward). Per the Tier-0 claim-integrity rule + architectural primitive #4, a published+subscribed strategy shouldn't reach on-chain execution without a rigor verdict. Suggest gating `/publish` on `passes_rigor_gate == True` and surfacing the verdict on the listing. Happy to wire this part myself — it's my lane.
3. **`container_spawner.py` forwards live signing secrets with no resource limits.** `_SHARED_ENV_KEYS` (lines 18-21) injects `ARC_AGENT_PRIVATE_KEY`, `CIRCLE_API_KEY`, `CIRCLE_ENTITY_SECRET`, `WALLET_ID` into every spawned container, and `containers.run(...)` sets no `mem_limit`/`nano_cpus`/`pids_limit`/`cap_drop`/`read_only`. The rate-limiter key helpers (`marketplace_routes.py:131`) are defined but never applied to the routes, so spawns are uncapped. Not command injection (image/name are slugified), but uncapped resource use + broad secret propagation. Suggest: per-container resource limits + `cap_drop=["ALL"]`, apply the rate limiter, and give a subscriber only the secrets it needs (not the publisher's signing key).

### 🟠 Medium
4. **Agent-app endpoints have `CORS allow_origins=["*"]` + `allow_credentials=True` and no auth** (`publisher.py`/`subscriber.py` `/events`, `/subscribe`, `/top-up`, `/unsubscribe`). That CORS combo is invalid and a footgun if these ports are ever exposed. Fine only if strictly internal to the docker network — worth documenting/enforcing that boundary.
5. **`_notify_all` zips `results` against a *recomputed* active-subscriber comprehension** (`publisher.py` ~250). If `info.active` flips during the await (delivery failure does this), the sequences desync and the wrong subscriber is marked inactive. Capture the list once before `gather`.
6. **`pool_id` is overloaded into the `sub_id` column** for publisher rows (`marketplace_routes.py:181`) while a dedicated `pool_id` column exists — store it in `pool_id`.

### 🟡 Contracts — deferring formal sign-off to Bogdan (two-eyes funds-contract policy)
The split/subscription **arithmetic is sound**: `PaymentSplitter.split` does `creator_share = amount*90//100`, platform gets the remainder → no dust loss; `SubscriptionManager.chargeActions` is overflow-checked with a balance assert and updates state before the external transfer (good CEI). Two things I'd want Bogdan/Dan to look at:
- **`PaymentSplitter.split()` and `SubscriptionManager.chargeActions()` have no access control** — any caller can charge an active subscription an arbitrary `action_count` (or split an arbitrary `amount`) against the contract's balance. I'd expect an authorized-charger (owner / oracle / the agent) gate. This is the headline contract item.
- Minor: `reserved_usdc` and `ephemeral_wallets[...].balance` are kept as parallel copies of the same number (divergence risk); `renewEphemeralWallet` emits `EphemeralWalletCreated` on a top-up; the 90/10 split is hardcoded (no per-pool ratio); unused `transferFrom` decl.

Routing the `.vy` files + `DeploySubscriptionMarketplace.s.sol` to @mnemonik-dev for the contract sign-off per our funds-contract two-eyes convention.

### Scope
The PR bundles **41 `.agents/skills/*/SKILL.md` onboarding docs + `AGENTS.md`** with a 6,700-line marketplace+contracts+UI feature. These are unrelated to the marketplace — recommend splitting the skills/docs into their own PR so this one is reviewable as a single logical change.

### Tests
Route-auth coverage is good (403 wrong-wallet, 401, 409 dup). But spawn is fully mocked (no test asserts resource limits / secret scoping), and `TestChargeAndRebalance` only checks revert-handling + payload shape — nothing pins the charge amount, the charge→execute ordering, or that a failed execution doesn't leave subscribers charged. Worth a test on the money path once #1/#2 are wired.

Overall: really nice groundwork — happy to take #2 (rigor-gate-on-publish) on the quant side, and to pair on the charge-atomicity design. Not approving solo since this carries funds contracts; deferring that to Bogdan/your call.

onder-akkaya (changes requested):
**Security follow-up — escalating the two contract items from my earlier review to 🔴 HIGH, and requesting changes.**

I ran a focused security pass over the new contracts (independent identification + per-finding false-positive filtering). Two findings landed at high confidence (9/10) — both are **fund-theft authorization bypasses** (missing `msg.sender` checks on fund-moving externals) and should block merge until access control lands. The rest of the PR's attack surface assessed clean: `marketplace_routes.py` is SIWE-gated, `container_spawner` uses the Docker SDK (not a shell), and the publisher/subscriber FastAPI endpoints — although unauthenticated — are spawned with no published host port and no nginx route (internal Docker bridge only), so the earlier SSRF/`/events` concern is **not** externally reachable (filtered out as a false positive). So the surface narrows to exactly these two:

🔴 **HIGH — `SubscriptionManager.chargeActions()` (`contracts/vyper/SubscriptionManager.vy:126`)**
`@external` with no caller authorization — only `action_count > 0`, `sub.active`, `wallet.balance >= total_charge` (note the siblings `renewEphemeralWallet`/`unsubscribe` correctly enforce `assert sub.subscriber == msg.sender`, and `setFlatFee` enforces owner). `action_count` is purely caller-asserted (no oracle/signature/nonce/proof-of-consumption). *Exploit:* any address calls `chargeActions(victim_sub_id, action_count)` and drains the victim's entire prepaid deposit to the splitter → creator/platform — billing for actions never performed; a malicious creator/platform can self-deal across every active subscription. *Fix:* `assert msg.sender == self.owner` (or an authorized-charger role) at the top, and bind the charge to a signed proof of actions consumed.

🔴 **HIGH — `PaymentSplitter.split()` (`contracts/vyper/PaymentSplitter.vy:51`)**
`@external` with only `assert self.pools[pool_id].active` — no `msg.sender` check (contrast `createPool`/`deactivatePool`, both owner-guarded), and `amount` is caller-controlled and decoupled from any per-pool deposit accounting (it pays out of the contract's own commingled USDC balance). *Exploit:* `chargeActions` pre-transfers USDC onto the splitter before calling `split`, so an attacker front-runs the internal `split(A,…)` with `split(B, amount)` and pays pool B's recipients out of pool A's just-deposited funds — pool A's settlement then reverts and pool B's accounting is inflated; any residual/dust balance is freely disbursable by anyone. *Fix:* store the manager address at deploy and `assert msg.sender == self.subscription_manager` (or owner), and track an undisbursed balance per `pool_id` with `assert amount <= pool_balance`.

Routing the `.vy` files to @mnemonik-dev for the contract sign-off per our funds-contract two-eyes rule. Requesting changes so this doesn't merge before the access-control guards land — the backend slice from my earlier review still stands (rigor-on-publish gate, charge-atomicity, scope split). Happy to pair on the fix or re-review once the guards are in. 🙏

onder-akkaya:
Following up my request-changes with a **ready, verified fix** for the two HIGH findings. Since this is a cross-repo PR I can't push to `feature/market-V1.1`, so the patch is a cherry-pickable commit:

**`onder/823-contract-authz-guards` @ `ca872e6`** → `git cherry-pick ca872e6` onto your branch (it's based on this PR's head, so it applies clean).

### What it does (minimal, access-control only)
- **`SubscriptionManager.chargeActions`** — new `authorized_charger` (defaults to the deployer/owner in `__init__`, re-pointable via `setAuthorizedCharger` onlyOwner); `chargeActions` now `assert msg.sender == self.authorized_charger` as its first statement. Closes the deposit-drain.
- **`PaymentSplitter.split`** — new `subscription_manager` (wired by owner via `setSubscriptionManager` onlyOwner); `split` now `assert msg.sender == self.subscription_manager`. **Fail-closed** — `split` reverts until wired. Closes the force/cross-pool disburse.
- **`DeploySubscriptionMarketplace.s.sol`** — wires `setSubscriptionManager(manager)` post-deploy so the legit `chargeActions → split` path settles.
- Storage vars are **appended** and **both constructor signatures are unchanged**, so the deploy bytecode/ABI input contract is preserved. ABIs regenerated to expose the new setters/getters.

### Verification (local)
- `vyper 0.4.3` compiles both contracts; `forge build` compiles the deploy script.
- **14 titanoboa tests pass** — your original 8 plus 6 new access-control cases: unauthorized `chargeActions` reverts (+ deposit untouched), direct `split` reverts, owner *cannot* call `split` directly (manager-scoped, not owner-scoped), setters are onlyOwner, and a delegated charger drives the full flow end-to-end.
- I ran an independent adversarial audit: verdict **SOUND** — both findings closed, no residual unauthorized fund-moving path, no reentrancy/ordering change, and misconfiguration fails *closed* (revenue stops), never *open* (no new theft path).

### Two notes for you
1. **Optional hardening (non-blocking):** `split`'s `amount` is still caller-supplied rather than bound to that pool's collected balance. With the guard, only the SM calls `split` and it always transfers exactly `total_charge` immediately before splitting it, so per-call the amount is bound — but a future second caller or SM bug could over-disburse from commingled balance. A per-pool balance invariant is the deeper fix when you have a moment.
2. **Ops:** `authorized_charger` defaults to the deployer. If the off-chain charger runs under a separate agent wallet, call `setAuthorizedCharger(<agentWallet>)` post-deploy (noted in the deploy script).

Drafted for your review, not merging it myself — @mnemonik-dev @dbrowneup, this is the funds-contract two-eyes call. Happy to fold in changes. (The branch can be deleted once cherry-picked.)

dbrowneup:
## 🔬 Deep review — Claude multi-agent (5 dimensions → adversarial verify → synthesize)

_Posted on Dan's behalf. **I independently verified both Critical findings against the actual Vyper source before posting:**_
- _**`chargeActions`** (SubscriptionManager.vy) has **no `msg.sender` gate** — confirmed; the very next function `unsubscribe` asserts `sub.subscriber == msg.sender`, proving the pattern was dropped here. Anyone can drain a subscriber's ephemeral balance._
- _**`split`** (PaymentSplitter.vy) has **no caller check** — confirmed; the next function `deactivatePool` gates on `msg.sender == self.owner`._
_Both are real unauthorized-fund-movement defects. **Contract findings are ADVISORY for Dan (owner) + Bogdan (`mnemonik-dev`) two-eyes — not a merge decision.** (Numeric line refs below are diff-offsets — use the function names.)_

---

Copilot inline is empty (`[]`), so nothing to complement or duplicate against. I have everything I need to write the review.

---

## Verdict

**Request changes.** The subscription-marketplace core is coherent and North-Star-aligned (real 90/10 payments, positive-sum copy-trading, no casino surface), but two **critical unauthorized-fund-movement defects in the Vyper contracts** and a set of **live-path correctness/claims-integrity gaps** must land before consolidation. Contract findings below are **ADVISORY input for Dan (owner) + Bogdan (`mnemonik-dev`) two-eyes review — not a merge decision.** Copilot posted no inline comments, so nothing here duplicates it.

## What's good

- **Payments are real and correct — the North-Star "keep" side is satisfied.** `PaymentSplitter.split()` does the 90/10 creator/platform split with the platform taking the *remainder* (`platform = amount - creator_share`), so no wei is lost to rounding; `chargeActions` uses checks-effects-interactions ordering (state debited before the external transfer). This is the Lepton nanopayment centerpiece and it's genuinely on-chain.
- **The economic model is positive-sum copy-trading, not a casino.** Publisher evaluates a paper-grounded strategy → broadcasts a rebalance signal → charges each subscriber a flat per-action fee → sends payload only to charged subscribers. No prediction-market / YES-NO / bet-against surface anywhere in *wired* code. Non-custodial posture preserved (each subscriber mirrors onto its own vault).
- **Test discipline is right where it exists.** Marketplace-route tests are hermetic (tmp-sqlite, docker mocked at the boundary, real SIWE signed cookies not header spoofing); the titanoboa suite runs real Vyper bytecode in an in-process EVM with concrete balance assertions; auth edges (403 wrong-wallet, 409 double-publish, 404) are covered.
- **Clean seams:** all Docker calls isolated in `container_spawner.py` for boundary-mocking; DB sessions wrapped in try/finally with explicit commits; migration DDL matches the ORM column-for-column with sensible lookup indexes.

## Findings

### Critical

1. **`contracts/vyper/SubscriptionManager.vy:6453-6471` (chargeActions) — no access control; anyone can drain any subscriber's ephemeral balance to the creator.** `chargeActions(sub_id, action_count)` asserts only `action_count > 0`, `sub.active`, and `wallet.balance >= total_charge` — there is **no `msg.sender` gate**, and the contract stores no authorized-charger address. Every other mutating fn in the file *is* gated (`renewEphemeralWallet`/`unsubscribe` assert `sub.subscriber == msg.sender`; `setFlatFee` asserts owner). A malicious creator (or anyone) can call `chargeActions(victim_sub_id, large_count)` and sweep 90% of the victim's refundable deposit to the creator with no strategy action performed. The PR's own test proves it: `chargeActions(sub_id, 10, sender=creator)` succeeds. Fix: store the authorized publisher/agent at `createPool`/`subscribe` and `assert msg.sender == <charger>`, or gate to `self.owner` and route all charges through the platform agent signer (aligns with the existing off-chain `_charge_subscriber` design). Bound `action_count` to a sane max. **ADVISORY — Dan + Bogdan.**

2. **`contracts/vyper/PaymentSplitter.vy:6300-6315` (split) — `@external` with no caller check; pays from the contract's own balance.** `split(pool_id, amount)` requires only `pools[pool_id].active` and transfers a caller-supplied `amount` from the splitter's *own* balance to creator/platform, incrementing `total_collected`/`total_disbursed` by the attacker-chosen `amount`. Anyone can call `split(any_active_pool, arbitrary_amount)` to sweep stray/griefing USDC in the splitter to a pool they control and forge the on-chain revenue accounting a UI/passport would read. Note the steady-state drain is bounded because `chargeActions` forwards and splits atomically (splitter holds ~0 between txns) — but that's an operational accident, not an enforced invariant, and the accounting corruption is unconditional. Fix: `assert msg.sender == self.subscription_manager` (stored at deploy), and prefer verifying the amount received this call over trusting ambient balance. **ADVISORY — Dan + Bogdan.**

### High

3. **`.agents/skills/` (all 41 files) + root `AGENTS.md` — drop the entire tree; it's a leaked OpenHands context dump for a teammate's 10-repo local workspace.** Every file hardcodes `/home/ricardo/...` absolute paths (286 occurrences) and documents unrelated projects (freqtrade, hummingbot, titanoboa, CypherLexicon, **auction-prediction-market**). `AGENTS.md` is a router advertising the "ecosystem" as including a prediction-market backend and embeds a git workflow ("no pull request is needed") that contradicts CLAUDE.md's merge-commit-only model. The PR self-adds `.agents/` and `AGENTS.md` to `.gitignore` but staged the files first, so they land anyway. Fix: `git rm -r --cached .agents/ AGENTS.md`; keep the two `.gitignore` lines (add the trailing newline). This is the single largest source of PR noise (~1792 lines, ~27% of +6739) and leaks a foreign contributor's private workspace topology onto a judge-readable Unlicense repo.

4. **`backend/archimedes/chain/strategy_runner_subscriber.py:3452-3466 — raw-key path fabricates `sub_id` with `uuid` instead of reading the on-chain `keccak256` id.** On-chain `subscribe()` derives `sub_id = keccak256(abi_encode(pool_id, msg.sender, block.timestamp))` and emits `Subscribed(sub_id, …)`, but the non-Circle branch discards the receipt and sets `sub_id = f"sub_{uuid.uuid4().hex}"` (L3463), then queries `subscriptions(sub_id)` with that non-bytes32 string — which raises, is swallowed by the broad `except`, and yields a zero-address wallet after a **real USDC deposit already fired.** The publisher-side `handle_subscribe` validates the id on-chain and rejects it (400), so registration fails and the real deposit is orphaned (untracked off-chain → no clean unsubscribe/refund via the agent). The Circle path has the same `result.get("sub_id", uuid…)` fallback. Fix: recover the real id via `contract.events.Subscribed().process_receipt(receipt)` — never fabricate. This is the non-Circle **default** single-user path (CLAUDE.md), so it's live.

5. **`backend/archimedes/chain/strategy_runner_publisher.py:3087-3098 — `_compute_rebalance` is hardcoded to return `([], {})`, so the entire per-action 90/10 charge path is dead on the live path.** `tick()` calls it then `if not trades: return` (L3028) every tick, making the charge loop (`chargeActions`), the rebalance webhook, and `execute_trades` all unreachable. The publisher will only ever emit `evaluation_step`/`halt` — it **never charges a subscriber**, despite the embedded README documenting "Publisher detects rebalance → calls chargeActions() for each subscriber" and North Star §6 requiring "the creator still receives the 90/10 split on every action." The upstream signal machinery (`evaluate_strategy_signals`, per-asset weights) is wired and discarded. This makes the "real payments on testnet" centerpiece claim untrue on the live path (REPO #1 rule). Fix: implement `_compute_rebalance` from the available signal weights, or — if per-action charging is intentionally deferred — decouple the charge path from trade generation and add a loud `"trade-settlement stubbed until mainnet"` telemetry line so the deferral isn't silent.

6. **`backend/archimedes/chain/strategy_runner_subscriber.py:3608-3643 (esp. 3635) — copied-trade settlement is REAL, violating North Star §6 (settlement must be STUBBED until mainnet).** `_handle_rebalance` feeds the publisher's `event.trades` straight into `executor.execute_trades(vault_address, trades)`, which builds a real `vault.rebalance(...)` AMM swap and confirms the receipt — exactly the AMM-as-settlement machinery §6 says must be paper/simulated pre-mainnet. Gated only by `AGENT_DRY_RUN` (defaults `"false"`, **not** in `.env.example`), and there is no paper-settlement path anywhere in the PR; the test `test_rebalance_event_non_dry_run` asserts real execution *is* the intended behavior. Currently latent only because #5's stub means no trades reach the subscriber. Fix: route copied-trade execution through a paper/simulated settlement mode (default PAPER), keep the real-swap branch as a labeled `mainnet — pending §6` stub, add `SETTLEMENT_MODE=paper`/`AGENT_DRY_RUN` to `.env.example` with the safe default. Correct README event-flow step 8 ("Subscriber executes trades on its vault") which presents this as real. **Flag Dan + Bogdan (funds-adjacent on-chain execution).** The publisher's own-vault `execute_trades` at L3057 is the same issue on the publish side (currently masked by the #5 stub) — at minimum add a §6-mainnet-pending comment there.

7. **`backend/tests/chain/test_strategy_runner_subscriber.py:5393-5424 — the money-in top-up path (`renewEphemeralWallet`) has no contract-level test; the only test mocks the contract entirely.** `renewEphemeralWallet` pulls `top_up_amount` via `transferFrom` and increments balance + `reserved_usdc` (real money-in via a live `POST /top-up` route). `test_top_up_non_dry_run` mocks `_contract`/nonce/gas/`_send_raw` and asserts only `status == "topped_up"` — a wrong amount or a broken increment would pass, and titanoboa covers `renewEphemeralWallet` nowhere. Fix: add a titanoboa test (mint+approve, `top_up>0`, assert `ephemeral_wallets[wallet].balance` and `subscriptions[sub_id].reserved_usdc` each rose by exactly `top_up_amount` and the SM's USDC balance rose). **ADVISORY — Dan + Bogdan.**

### Medium

8. **`contracts/vyper/SubscriptionManager.vy:6454-6462 — no per-user/global spend cap.** `total_charge = action_count * flat_fee_per_action` is bounded only by `wallet.balance`, so one call can drain the full ephemeral deposit; `setFlatFee` (owner-mutable, global) isn't locked into the `Subscription` struct, so an in-flight subscriber can't verify the price they'll be charged. The task explicitly names per-user spend caps as a payments-security requirement. Fix: add a per-subscription max-charge-per-call and/or rolling window cap; lock the fee at subscribe time (or in the struct) so `setFlatFee` can't retroactively spike an existing subscriber. **ADVISORY — Dan + Bogdan.**

9. **`contracts/vyper/SubscriptionManager.vy:6473-6487 (unsubscribe) — the on-chain refund path has zero coverage** (titanoboa never calls it; the route test mocks the publisher call away). Real user-USDC return via `transfer(msg.sender, remaining)`. Note the finding's route→contract chain is inaccurate — **no Python path currently invokes `unsubscribe()` at all** (`handle_unsubscribe` only pops an in-memory registry), so this refund is effectively uncalled. Fix: subscribe→charge→`unsubscribe()`, assert `balanceOf(subscriber)` rose by exactly the remainder, `reserved_usdc == 0`, `active == False`. **ADVISORY — Dan + Bogdan.**

10. **`contracts/vyper/SubscriptionManager.vy:6462 — the overcharge guard (`wallet.balance >= total_charge`) has no test.** All charge tests use a 10M deposit with tiny charges, so the revert boundary never fires; the publisher-side "revert" tests raise a Python `Exception`, not an EVM revert. Fix: titanoboa test that `chargeActions` with `total_charge > balance` reverts `"insufficient balance"` and moves no USDC. **ADVISORY — Dan + Bogdan.**

11. **`backend/tests/contracts/test_vyper_contracts.py:5431 — the money-split tests require `titanoboa`, which is undeclared, AND live under `backend/tests/contracts/` which `pytest.ini`'s `norecursedirs` excludes.** Correction to the finding: they do **not** red the gate (norecursedirs excludes the dir by basename `contracts`), but they **silently never run in CI** for two independent reasons — so the on-EVM correctness proof for the 90/10 split has zero CI assurance ("CI green ≠ local green"). Fix: declare `titanoboa` in a test-extras surface the gate installs, `boa = pytest.importorskip("boa")` at module top, **and** remove `contracts` from `norecursedirs` (importorskip alone is insufficient — the dir is excluded from collection entirely).

12. **`backend/archimedes/services/container_spawner.py:4154 — default `COMPOSE_PROJECT_NAME` hardcodes `'archimedes-fork'` (Ricardo's fork), wrong on the canonical stack** (live network is `archimedes_default`). Worse, the sibling **image names have no override at all** — `image="archimedes-fork-publisher"`/`-subscriber` (L4223/4295) are never built by any compose/CI diff, so `spawn_publisher`/`spawn_subscriber` → `containers.run` against non-existent images → `ImageNotFound` that escapes the `try/except DockerException` (which wraps only `docker.from_env()`) as an uncaught 500. The publish/subscribe surface can't function on the live path as shipped. Fix: default to `'archimedes'`; add `ARC_MARKETPLACE_PUBLISHER_IMAGE`/`_SUBSCRIBER_IMAGE` overrides; wrap `containers.run` in `try/except DockerException → ContainerSpawnError`; add the missing compose/CI build wiring.

13. **`backend/archimedes/chain/strategy_runner_publisher.py:3102-3162 — publisher webhook `/subscribe`, `/update-ephemeral`, `/unsubscribe` are unauthenticated (`allow_origins=["*"]`).** The sibling marketplace routes in *this same PR* use `Depends(require_verified_wallet)`, so the primitive exists and is just omitted. `sub_id`s are public (emitted on-chain), so an attacker can re-register a victim's `sub_id` pointing at their own `webhook_url` to **hijack delivery of the paid rebalance signal** (victim keeps getting charged on-chain, stops receiving the signal), or force-unsubscribe a paying subscriber (griefing/DoS). No direct theft (the Vyper contract gates `unsubscribe`/`renew` on `msg.sender`), and `/update-ephemeral` abuse is inert (the off-chain field is never read in a charge/delivery path). Fix: require a signed challenge proving control of the subscriber wallet, or a platform↔container shared secret; tighten CORS off `*` before the container's `0.0.0.0:8080` ships.

14. **`backend/archimedes/chain/strategy_runner_publisher.py:3034-3062 — charge-then-notify with no idempotency: a subscriber can be charged, then the webhook delivery fails all retries (no refund path); a restart mid-tick double-charges.** `chargeActions` has no per-`(sub_id, tick_id)` guard, and the subscriber's `/events` executes on every rebalance with no `tick_id` dedupe (L3608). Latent today (masked by #5's stub) but a real hardening gap on the rail the team intends to make live. Fix: charge only after confirmed delivery, or make charging idempotent per `(sub_id, tick_id)` and dedupe on the subscriber side.

15. **`backend/archimedes/api/marketplace_routes.py:2343-2381 — stopping a publisher orphans its subscriber containers and DB rows.** `DELETE /publish/{strategy_id}` stops only the publisher and marks only its row `stopped`; subscriber rows stay `status='running'`, so `active_subscriber_count` (rendered in the UI as "N active subscribers") keeps counting them — a live-UI claim that becomes false after a stop (REPO #1). Correction: orphaned subscribers sit idle (registration is a bounded one-shot, not a retry spam) and don't drain funds. Fix: cascade-stop subscriber rows/containers in the same commit, or at minimum stop the count from lying.

16. **`backend/archimedes/api/marketplace_routes.py:2174-2182 — `_publish_rate_key`/`_subscribe_rate_key` read `request.state.wallet_address`, which is never set, and are wired to no `@limiter.limit` decorator.** Dead code that would `AttributeError` if invoked, under a "Rate limiter keys" header implying these **container-spawning** endpoints are throttled when they aren't (a resource-exhaustion vector — `strategy_id` is attacker-controlled free text). Fix: wire real `@limiter.limit` decorators keyed off `get_verified_wallet(request)`, or delete the dead functions + misleading header; given they launch containers, real per-wallet throttling should exist.

17. **`backend/tests/api/test_marketplace_routes.py:4529-4554 & `container_spawner`/model — `pool_id` is written to the `sub_id` column but the model has a real `pool_id` column, and the test fixture writes the *other* column.** Production writes `sub_id=body.pool_id` (marketplace_routes.py:2233) and reads `pool_id=row.sub_id` (self-consistent), but the purpose-built `pool_id` column defaults `''` on every publisher write, and `_insert_publisher` seeds the real `pool_id` column (test:4547) — the opposite of production. No test asserts `pool_id` in any `/published` response, so the dual-source-of-truth trap round-trips untested (a subscriber pointed at the wrong on-chain payment pool wouldn't be caught). Same root issue also flagged for the subscriber-env test gap: `spawn_subscriber`'s money vars (`INITIAL_DEPOSIT_USDC`, `SUBSCRIBER_POOL_ID`, splitter/manager addresses) are never asserted forwarded. Fix: collapse to one column, add a `POST /publish → GET /published/{id}` round-trip assertion on `pool_id`, and assert the funds-wiring env keys reach the container.

18. **`backend/tests/chain/test_strategy_runner_publisher.py:5035-5082 — the `_charge_subscriber` success path (money actually leaves) has zero executable coverage.** Both tests exercise only the `except → return False` branch; the two are near-identical (differ only in the `sub_id` string). Their docstrings claim "marks subscriber inactive," but that side effect lives in `tick()` (L3041), which is untested. Fix: collapse the duplicates; add a success-path test asserting `True` + correct `chargeActions(sub_id, action_count).build_transaction`; add a `tick()`-level test asserting a failed charge flips `active=False` and broadcasts the `pre_rebalance` halt.

19. **`.agents/skills/` prediction-market subset (Polymarket SDK/CLOB/builder-relayer, `auction-prediction-market__*`, `shared__prediction-market-contracts`, `cypherlexicon-offchain__*`) — the maintainer explicitly ordered these excluded.** `docs/ARCHIMEDES-NORTH-STAR.md:118` says verbatim: *"The Polymarket skills that appeared in #823 were accidental PR noise, not a direction — exclude them,"* and §10.2 repeats it. These document a full binary YES/NO betting stack (`PredictionMarket.buyShares(YES/NO)`) — the forbidden zero-sum casino, in doc form. Confirmed none are imported/symlinked into any tracked archimedes path; drop them with the rest of #3.

### Low / Nit

20. **`.gitignore` (diff:1874-1876) — the ignore rules contradict the files this PR commits.** Keep the two lines, drop the 42 tracked files (per #3), add the missing trailing newline.
21. **`backend/tests/chain/test_strategy_runner_subscriber.py:5236-5257 — `test_rebalance_event_triggers_execute_trades` is a no-op:** asserts only the trivially-true `status=='received'`, then `pytest.skip(...)` gated on a module-level `DRY_RUN` captured at import → behavior depends on ambient env at first import, not on the code. The real coverage lives in the `non_dry_run` sibling. Delete it, or convert to an explicit `patch(DRY_RUN, True)` asserting `execute_trades` is NOT called. (Violates the "no skip-marks / hermetic" testing convention.)
22. **`backend/archimedes/api/marketplace_routes.py:2470-2483 — N+1 subscriber query in the no-auth `/published` feed** (up to 200 extra queries/load, un-indexed `role`/`strategy_id`, polled every 30s). Fetch subscribers for the page's `strategy_id`s in one `WHERE strategy_id IN (...)` query. Low today at ~2-wallet scale.
23. **`backend/tests/contracts/test_vyper_contracts.py:5425 — new `backend/tests/contracts/` dir has no `__init__.py`** while the PR adds one for `api/`. No functional collision (basenames unique), inconsistent only. Add the empty `__init__.py` for consistency, or leave it — inert either way.

## Keep vs cut (for the #749 consolidation)

**Keep** — the real product:
- `contracts/vyper/PaymentSplitter.vy` + `SubscriptionManager.vy` (the nanopayment centerpiece — **after** the two critical auth fixes + spend cap)
- `backend/archimedes/chain/strategy_runner_publisher.py` / `strategy_runner_subscriber.py` (fix #4, #5, #6 first)
- `backend/archimedes/api/marketplace_routes.py`, `models/marketplace.py` + migration, `container_spawner.py` (fix the fork defaults)
- `backend/tests/contracts/test_vyper_contracts.py` + the hermetic route/spawner tests (fix CI wiring #11)
- UI: `MarketplacePage.jsx`, `PublishPage.jsx`, `SubscriptionsPage.jsx`, `StrategyDetailPage.jsx`, `App.jsx` routing
- `.env.example` marketplace additions (correct config-contract surface, no secrets)

**Cut** — the noise (all `git rm -r --cached`):
- The entire `.agents/skills/` tree (41 files) and root `AGENTS.md` — leaked OpenHands 10-repo context dump with `/home/ricardo/...` paths (#3), including every prediction-market/Polymarket file the maintainer ordered excluded (#19)
- Keep the two `.gitignore` lines that already list them (with a trailing newline) so the dump can't recur

Reconcile during consolidation: the `archimedes-fork` compose/image defaults (#12) so the live path works without an override.

danielscoffee (changes requested):
The runners has a lot of similar methods, create an abstraction layer and calls them at the runner to reduce future workload.

dbrowneup:
Superseded by #958

PR #824, opened 2026-06-30
[contracts] Canonical subscription-marketplace payment contracts (foundation slice of #713)

### LINKED ISSUE #713: T1.2 — Nanopayment marketplace scaffold: x402-gated strategy access (first PR)
> **Owner:** Ricardo Obregon Huaman
> **Roadmap item:** T1.2 (Nanopayment marketplace, optional publish) — first slice
> **Tier:** 1 (Core Lepton vertical)

## Summary

Stand up the first end-to-end slice of the optional-publish nanopayment marketplace: an x402-gated strategy-access endpoint with Circle Gateway facilitator settlement, an on-chain creator/platform revenue-split contract (mocked-receiver MVP), per-user spend caps in middleware, and a minimal happy-path UI flow. This is the **scaffold PR**, not the full marketplace — the goal is to land one concrete payment-gated request that pays USDC into a 90/10 splitter on Arc, with everything else stubbed-but-clean so the follow-up issues plug in. The "embedded wallet per user" piece is downstream of Dan's in-flight Circle modular-wallet bugfix (passkey wallet mints a new MSCA per call — Loose-thread #1 in ROADMAP-v3) and is **explicitly out of this PR** until that fix lands. The bigger pieces (Gateway batched settlement, full UI, publish-gate UI, multi-user spend-cap dashboards) are follow-up issues; this PR cuts the smallest verifiable demo path.

The "right surface" for the paywall is the **`POST /api/strategies/{id}/construct`** route (verified to exist at `backend/archimedes/api/strategies_routes.py:1562`). It accepts a strategy + portfolio context and returns a fresh executable allocation, which is exactly the "alpha" the marketplace charges to access. The simpler `GET /api/strategies/signals` is intentionally NOT chosen — it returns evaluations across the whole library and would be expensive to paywall in this slice. (Note: there is no `/run` route today; `/construct` is the canonical "consume a strategy" verb.)

## Scope

Add (and only add):

1. **A revenue-split contract** at `contracts/src/RevenueSplit.sol`. Minimal: a non-upgradeable contract that on `receive(USDC)` (called by the x402 facilitator's `transferWithAuthorization` settlement) forwards 90% to a `creator` address and 10% to a `platform` address, both immutable at construction. No streaming, no per-strategy creator routing yet (that's a follow-up — the strategy→creator mapping table comes in PR #2). Test in `contracts/test/RevenueSplit.t.sol` following the pattern of `contracts/test/Registry.t.sol` (verified at `contracts/test/Registry.t.sol:1-30`). Foundry-only; do NOT touch the Python `backend/archimedes/chain/strategy_publisher.py` / `trace_publisher.py` dual-path signer scaffolding — that is reusable but not modified here.

2. **A new x402-gated paywall middleware** at `backend/archimedes/api/x402_middleware.py`. It:
   - On a paywalled request, if no `PAYMENT-SIGNATURE` header → respond `402 Payment Required` with a `PAYMENT-REQUIRED` header naming Arc testnet + the deployed `RevenueSplit` address + a fixed USDC amount (e.g. `0.001` USDC for the first slice — sub-cent).
   - On a request with `PAYMENT-SIGNATURE` → verify the signature by calling Circle's Gateway facilitator (`https://gateway-api-testnet.circle.com/v1/x402/settle`, per `submodules/context-arc/docs/developers.circle.com/openapi/gateway.yaml:270`) and gate the downstream handler on a 200 response.
   - Reference the seller-side pattern in `submodules/context-arc/docs/developers.circle.com/gateway/nanopayments/quickstarts/seller.md` (Python implementation; the Express/TypeScript example there is the shape to copy, ported to FastAPI middleware).
   - Use a feature flag `ARCHIMEDES_X402_ENABLED=1` so the middleware is a no-op when off. Default: off.

3. **A per-user spend cap** at `backend/archimedes/api/x402_middleware.py` (in the same file). For the first slice this is keyed on the **payer wallet address from the verified payment payload** (NOT a session — there is no embedded wallet per user yet, pending the passkey fix). Cap = a configurable `ARCHIMEDES_X402_MAX_USDC_PER_DAY` (default `0.10` USDC for testnet). State: Redis with a 24h sliding window (use the existing `AgentStateStore` shape at `backend/archimedes/services/redis_state.py` as a precedent — `AgentStateStore` is the canonical pattern, see CLAUDE.md "Testing conventions" §1). When over cap, respond `429 Too Many Requests` with a JSON body explaining the cap and the next reset.

4. **Apply the paywall to exactly one endpoint** — `POST /api/strategies/{strategy_id}/construct` (file: `backend/archimedes/api/strategies_routes.py:1562`). Wire the middleware as a FastAPI dependency on this route only. Do NOT apply it globally. Do NOT touch any other endpoint in this PR.

5. **A Foundry deploy script** at `contracts/script/DeployRevenueSplit.s.sol` following the pattern of `contracts/script/DeployStrategyRegistry.s.sol` (verified to exist). Reads `CREATOR_ADDRESS` + `PLATFORM_ADDRESS` from env. Deploys to Arc testnet. Updates `contracts/abis/RevenueSplit.json` on `forge build` (the existing ABI cache pattern, see `contracts/abis/`). Wires the deployed address into the backend via `REVENUE_SPLIT_ADDRESS` in `.env.example` (mirror the existing pattern for `STRATEGY_REGISTRY_ADDRESS` — verify with `grep STRATEGY_REGISTRY_ADDRESS .env.example`).

6. **One test file** at `backend/tests/test_x402_middleware.py` covering:
   - 402 when no PAYMENT-SIGNATURE header.
   - 200 (passthrough) when PAYMENT-SIGNATURE present and facilitator returns 200 (mock `aiohttp` at the boundary — see precedent `backend/tests/test_api_routes.py` for `httpx.ASGITransport` use).
   - 429 when spend cap exceeded.
   - Feature flag off → middleware no-ops.
   The test MUST be hermetic per CLAUDE.md "Testing conventions" §1 (no live Redis — mock `AgentStateStore` at the boundary, see `backend/tests/test_api_routes.py` `TestAgentRoutes::test_agent_status_redis_down_defaults`).

7. **Foundry test** in `contracts/test/RevenueSplit.t.sol` covering: 90/10 split on a USDC receive (use a `MockUSDC` ERC20 fixture), revert when split addresses are zero at construction, and that the contract cannot be redirected after deploy (no admin function).

8. **A Marketplace UI badge only** — in `ui/src/components/Marketplace.jsx` (verified to exist; reads `/api/vaults/` today), add a small "Pay per run" disclosure chip next to vaults whose strategy is x402-gated. Do NOT add the full payment UI in this PR — that's the next issue. The chip just renders when `ARCHIMEDES_X402_ENABLED=1` is set on the backend and the vault's strategy has paywall enabled. Plumbing: a single boolean in the `/api/strategies/{id}` response (`is_paywalled: bool`).

Do NOT touch in this PR:
- `circle_service.py` (existing Circle integration aggregator) — its concerns are wallets/CCTP/etc., not x402.
- `backend/archimedes/chain/*` — chain integration is on Chuan's old lane; the x402 settlement happens off the agent path entirely. With Chuan stepping back, this lane is being absorbed — but not in this PR.
- `ui/src/circle-wallet.js`, `ui/src/circle-tx-executor.js` — these are the passkey/MSCA flow that has the known bug; Dan's fix is upstream of any Embedded-Wallet-per-user work here.
- Any other `*_routes.py` file beyond `strategies_routes.py`.
- The Solidity Vault / VaultFactory / SyntheticVault contracts — vault-side x402 is a separate issue.

## Acceptance criteria

Each is a runnable command + expected output (per CLAUDE.md "judge-grade issue" §"Acceptance must be machine-checkable"):

- [ ] `cd contracts && forge build` → exit 0; `contracts/abis/RevenueSplit.json` written.
- [ ] `cd contracts && forge test --match-contract RevenueSplit -vv` → all tests pass, ≥ 3 test functions.
- [ ] `pytest backend/tests/test_x402_middleware.py -q` → `N passed, 0 failed`, ≥ 4 tests.
- [ ] **Hermetic gate** (per CLAUDE.md): `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_x402_middleware.py -q` → `N passed, 0 failed`.
- [ ] `grep -n "ARCHIMEDES_X402_ENABLED" backend/archimedes/api/x402_middleware.py` returns ≥ 1 match (feature flag present).
- [ ] `ruff format --check .` and `ruff check --select E9,F63,F7,F40 .` → both pass (the blocking ruff-gate per CLAUDE.md §"Python linting").
- [ ] **Live happy-path verify** (with `ARCHIMEDES_X402_ENABLED=1` and `REVENUE_SPLIT_ADDRESS` set to a freshly-deployed-to-Arc-testnet address): `curl -X POST -s -o /dev/null -w "%{http_code}" http://localhost:8000/api/strategies/<any-real-id>/construct -H 'Content-Type: application/json' -d '{}'` → `402`. Same request with a valid PAYMENT-SIGNATURE header → `200`.
- [ ] **Anti-claim verify** — no other endpoint accidentally paywalled: `curl -s -o /dev/null -w "%{http_code}" http://localhost:8000/api/strategies/` → `200`, and `curl -s -o /dev/null -w "%{http_code}" http://localhost:8000/api/strategies/signals` → `200` (or 503 if no live market data — but NOT 402).
- [ ] PR description names the **three follow-up issues** that come out of this slice: (a) Gateway batched settlement + facilitator polling, (b) per-strategy creator routing through `RevenueSplit`, (c) Embedded-Wallet-per-user UI (downstream of Dan's passkey fix).

## Verify

```bash
# 1. Repo hygiene
cd /path/to/archimedes && git status --short

# 2. Contracts
cd contracts && forge build && forge test --match-contract RevenueSplit -vv
ls -la abis/RevenueSplit.json

# 3. Backend unit tests, hermetic + non-hermetic
cd .. && pytest backend/tests/test_x402_middleware.py -q
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_x402_middleware.py -q

# 4. Lint
ruff format --check . && ruff check --select E9,F63,F7,F40 .

# 5. Surface coverage — confirm exactly one endpoint paywalled
grep -rn "x402_middleware\|X402Middleware" backend/archimedes/api/ | grep -v "x402_middleware.py:"
# Expected: exactly one hit, on strategies_routes.py at the /{strategy_id}/construct route.

# 6. Live demo (with stack up + ARCHIMEDES_X402_ENABLED=1)
curl -X POST -s -o /dev/null -w "%{http_code}\n" \
  http://localhost:8000/api/strategies/<id>/construct \
  -H 'Content-Type: application/json' -d '{}'
# Expected: 402
```

## Anti-goals

- **Do NOT** modify any existing endpoint's behavior. The middleware is opt-in per-route; `strategies_routes.py:1562` is the only route wired up.
- **Do NOT** implement Gateway batched settlement in this PR. The first slice settles each payment via `transferWithAuthorization` (the EIP-3009 path x402 uses by default per `submodules/context-arc/docs/developers.circle.com/gateway/nanopayments/concepts/x402.md`). Batching is a follow-up.
- **Do NOT** implement per-strategy creator routing. `RevenueSplit` ships with `creator` immutable at construction — a single splitter for all paywalled strategies in this slice. Per-strategy routing is a follow-up issue.
- **Do NOT** touch the passkey wallet flow (`ui/src/circle-wallet.js`, `circle-tx-executor.js`). Dan is fixing the per-call MSCA mint bug. Wait for it.
- **Do NOT** add the full payment-confirmation UI. The Marketplace badge in this PR is a one-line visual disclosure, not a flow.
- **Do NOT** weaken any existing test, skip-mark a flaky test, or relax the `ruff-gate` rules to land the PR. Per CLAUDE.md "Testing conventions" §"No skip-marks on flaky tests."
- **Do NOT** edit `.env.example` to add secrets — only the **non-secret** address `REVENUE_SPLIT_ADDRESS` goes there. The Circle facilitator URL is already in `circle_service.py` (`CIRCLE_GW_BASE`) and should be reused/exported, not redeclared.
- **Do NOT** push direct to `main`. PR per CLAUDE.md §"Branch model."
- **Do NOT** commit the deploy `broadcast/` output to git — the deploy artifacts pattern is `.gitignored`.
- Sized as **one PR**. The full marketplace fan-out is many follow-up issues, listed in the PR description.

## Owner

Ricardo Obregon Huaman (Lima, Peru) — AWS-certified data engineer, RAG/data platforms; this is the first-PR scoped slice of his interest in nanopayments + the marketplace publish side.

## Coverage

- **On-chain review:** With Chuan stepping back, Bogdan is the natural reviewer for the new `RevenueSplit.sol` (per CLAUDE.md §"When to ask before acting" — contract review needs eyes).
- **Backend Python review:** Daniel Reis dos Santos (lead) or Marten (coverage).
- **Architecture eyes if integration questions surface:** Dan Browne (product + nanopayments framing).

## Cross-references

- Roadmap: T1.2 in `/Users/dbrowne/Desktop/Programming/GitHub/Agora/docs/2026-06-23-Lepton_artifacts/ARCHIMEDES-ROADMAP-v3.md:91-93`.
- Grant draft milestone 2: `/Users/dbrowne/Desktop/Programming/GitHub/Agora/docs/2026-06-23-Lepton_artifacts/ARCHIMEDES-CIRCLE-GRANT-DRAFT-v2.md:76`.
- x402 protocol spec (read first): `archimedes/submodules/context-arc/docs/developers.circle.com/gateway/nanopayments/concepts/x402.md`.
- x402 seller quickstart (port from TS/Express to FastAPI middleware): `archimedes/submodules/context-arc/docs/developers.circle.com/gateway/nanopayments/quickstarts/seller.md`.
- Gateway OpenAPI (settle endpoint signature): `archimedes/submodules/context-arc/docs/developers.circle.com/openapi/gateway.yaml` — see `/v1/x402/settle` at line 270.
- Existing Circle integration aggregator (read for context, don't modify): `archimedes/backend/archimedes/services/circle_service.py`.
- Existing dual-path signer pattern (precedent, don't reuse code in this PR): `archimedes/backend/archimedes/chain/trace_publisher.py` and `archimedes/backend/archimedes/chain/strategy_publisher.py`.
- Existing Foundry test pattern: `archimedes/contracts/test/Registry.t.sol` (line 1-30 — copy this shape).
- Existing deploy-script pattern: `archimedes/contracts/script/DeployStrategyRegistry.s.sol`.
- Hermetic-test reference: `archimedes/backend/tests/test_api_routes.py` `TestAgentRoutes::test_agent_status_redis_down_defaults` (Redis-down mocking precedent).
- Loose-thread dependency: ROADMAP-v3 §"Loose threads" item 1 (passkey/Circle modular-wallet bug — mints new wallet per call) — Embedded-Wallet-per-user is downstream of Dan's fix and is **out of scope** for this PR.
- Business model framing (90/10 split rationale): the Business Model doc (local-only, not in repo) — accept the 90/10 as a constraint, do not re-litigate it.

dbrowneup:
## What this PR is

The **foundation slice** of the payments/marketplace consolidation (North Star §6 +
issue #713): it lands **only the canonical on-chain funds layer** — the two Vyper
contracts + their security guards, tests, deploy script, ABIs, and toolchain deps.

This is intentionally a small, tight, **funds-review-focused** PR. It is the part of
@rcrdoh's #823 that is unambiguous and review-gated *regardless* of how we end up
building the off-chain delivery — so it's safe to lock down first while we align on
the rest. It is **not deployed** (deploy is Dan's keyed step) and it does **not**
wire the marketplace into the app yet.

> 🤝 **@rcrdoh — this is meant as a starting point we finish together, not a
> rewrite of your work.** Everything load-bearing here is *your* contract design;
> I've only added @onder-akkaya's security guards, the missing build/test deps, and
> validation. The sections below are transparent about every include/exclude
> decision and *why* — please push back on any of it, and see "How we collaborate
> from here." Your read on the delivery architecture is the one I most want.

---

## ✅ Drawn in (with provenance)

| What | Source | Why |
|---|---|---|
| `contracts/vyper/PaymentSplitter.vy`, `SubscriptionManager.vy` | @rcrdoh #823 | Your contract design is the canonical funds layer per Dan's call: 90/10 creator/platform split + prepaid-deposit subscriptions with per-action metering. Unchanged except for the guards below. |
| Access-control guards on both contracts | @onder-akkaya `ca872e6` | The #823 security review found **two HIGH fund-theft bypasses** — `chargeActions()` and `split()` both moved funds with no `msg.sender` check. Önder's guards (`authorized_charger`, `subscription_manager`, fail-closed) close them. Constructor signatures unchanged → deploy bytecode/ABI input preserved. |
| `contracts/script/DeploySubscriptionMarketplace.s.sol` + `contracts/abis/*.json` | #823 + `ca872e6` | Deploy script (wires `setSubscriptionManager` post-deploy) + ABIs regenerated to expose the new setters/getters. |
| `contracts/README-subscription-marketplace.md` | #823 | Kept intact; I added only a short status banner crediting your design and noting the delivery layer is being re-evaluated (so the doc doesn't misrepresent what's landing now). |
| `backend/tests/contracts/` — 14 hermetic titanoboa tests | #823 + `ca872e6` | Your original 8 + Önder's 6 access-control cases. In-process EVM — no RPC, no `.env`, no Circle creds → CI-safe. Added the missing `__init__.py` (repo convention; #823 had it for `tests/api/` but not `tests/contracts/`). |
| `vyper==0.4.3` + `titanoboa` → **both** `environment.yml` and `backend/requirements.txt` | new | #823's contracts shipped without their build/test deps, so the contract tests couldn't run in CI or a fresh clone. Added with justification comments per our dep-hygiene convention. |

---

## ⛔ Deliberately **not** in this PR (and why)

Per Dan's decision this session — **"foundation + leaderboard first, lighter delivery
later"** — the off-chain delivery layer is deferred, not adopted as-is:

| Excluded | Why |
|---|---|
| **The Docker container-per-user architecture** — `container_spawner.py`, `strategy_runner_publisher.py` (674L), `strategy_runner_subscriber.py` (501L), the two Dockerfiles, `docker>=7.0.0` | This is the one genuinely **architecture-level** call in #823, and it has real infra-cost/ops implications: every publish/subscribe spawns a container on our **single prod EC2** (which also runs docker Postgres/Redis), and **every merge auto-deploys** (`DEPLOY_ENABLED=true`). On testnet, copied trades only feed the **paper/leaderboard** layer (North Star §6/§8) — no real execution — so per-user containers buy little now and don't scale. We want to re-shape this delivery into a **lighter shared service** that reuses your runner *logic* without per-user containers. **This is a design conversation, not a "no" — see below.** |
| **`marketplace_routes.py`, `models/marketplace.py`, the migration, the UI pages** | These are tightly coupled to the container spawner (the routes are literally a "publisher/subscriber container lifecycle API"). They move together with the delivery-architecture decision. |
| **The x402 rail (#749)** | x402 is *pay-per-HTTP-call*; `PaymentSplitter` is *prepaid-subscription + metered `chargeActions`* — **different payment shapes** that need a real adapter to compose. That wiring is delivery-layer design work; forcing it into a funds-review PR adds risk for no gain. It folds in at the delivery step, and `RevenueSplit.sol` retires there (it only ever lived on the #749 branch, never on `main`). |
| **The ~25 `.agents/skills/*` + `AGENTS.md`** | Polymarket / prediction-market / vyper-agentic skill docs — confirmed accidental PR noise (North Star §8 explicitly: not a direction). |

---

## 🔧 What still needs resolution (let's decide these together)

1. **Delivery architecture — the big one.** Container-per-user (your #823) vs a
   single shared signal-delivery service. My recommendation is the shared service
   (lighter on one EC2, matches North Star "clean + simple", reuses your
   publisher/subscriber logic), but **you built the original and own #713 — I want
   your view before anyone writes the delivery PR.** What did the container model
   buy you that a shared service would lose? (isolation per subscriber? independent
   webhook lifecycles? ease of the on-chain ephemeral-wallet mapping?)
2. **x402 ↔ PaymentSplitter wiring.** How a single x402 micropayment maps onto the
   pool/`chargeActions` model — the adapter shape. Delivery-layer design.
3. **`strategy_signal_evaluator.py` duplication.** #823 adds a `chain/` version
   (166L); roadmap B1/#760 is your own item on the `services/` evaluator + its
   silent buy-and-hold fallback. We should pick one home so they don't diverge.
4. **Önder's non-blocking follow-up (carried):** bind `split`'s amount to a
   per-pool collected balance for defense-in-depth beyond today's single-caller
   topology.
5. **Image weight (minor):** `vyper`+`titanoboa` pull eth tooling into the backend
   image. Fine for now; could move to a dev-requirements split later.

---

## 🤝 How we collaborate from here (our proposal — tell us yours)

- **This PR:** lock the funds layer. @onder-akkaya already reviewed the security;
  it needs **Dan (contract owner) + Bogdan (`mnemonik-dev`) two-eyes** before merge.
  @rcrdoh — please confirm the contracts here still match your intent (they're
  byte-for-byte yours + Önder's guards).
- **Next, together:** a short delivery-architecture design note (container-per-user
  vs shared-service) — you + Dan make the call; I'll write whichever we choose and
  keep your runner logic at its core. Then one delivery PR brings the routes/UI +
  x402 rail and **`Closes #713`**; #749 + #823 close as superseded by the combined
  work.
- **Sequencing:** the **leaderboard** (the North Star traction priority) is being
  built in parallel and needs none of the delivery machinery — so deferring the
  delivery decision costs us no traction time.

@rcrdoh — does this split feel right to you? Anything here you'd pull forward, keep
as containers, or shape differently? This is your feature as much as ours; the goal
is to ship it properly, together.

---

## ✅ Validation

- `vyper` compiles both contracts (0.4.3)
- `forge build` compiles the deploy script
- **14/14 titanoboa tests pass** (hermetic, in-process EVM)
- `ruff format --check` + critical lint (E9/F63/F7/F40) clean

## Review

- **Funds contracts → Dan + Bogdan (`mnemonik-dev`) two-eyes required before merge.**
- Security guards authored + tested by @onder-akkaya (`ca872e6`).
- Not deployed; deploy is Dan's keyed step.

Refs #823, #713.


rcrdoh:
about the paymentspliter contract, is intended to receive funds from (from any external address) the gateway and split the batched fee amount to two addresses (platform and creator) where the payments trigger be the platform address , the percentages was intended to be 10-90/platform-creator as a first version

dbrowneup:
Feel free to work on this branch and modify it directly! Would be great it
you keep building here. I can focus on other branches. There’s plenty to do
lol. But no problem if you prefer to keep working on your fork. Just let me
know what is best for you. I look forward to learning more about the design
from you next time we have a chance to chat face to face.

dbrowneup:
## 🔬 Deep review — Claude multi-agent (dimensions → adversarial verify → synthesize)

_Posted on Dan's behalf. **I independently verified both funds-adjacent High findings against the source:**_
- _Deploy `create()` appends **no ABI-encoded ctor args** → contracts deploy with `usdc`/`flat_fee` = 0 (bricked funds path) — confirmed._
- _`createPool` guards only `not active` (not "ever created") and `deactivatePool` doesn't clear recipients → owner can `deactivatePool` + `createPool(…, ATTACKER)` to redirect a live subscriber's payouts — confirmed._
_**Advisory for Dan + Bogdan two-eyes — not a merge decision.** Positive note: unlike #823, these contracts DO carry the `msg.sender`/owner access-control guards — so #824 is the stronger implementation to survive the #713 consolidation. (Numeric line refs are diff-offsets — use function names.)_

---

I have everything I need. Here's the review.

---

**Verdict** — Request changes (advisory input for Dan + Bogdan two-eyes; not a merge decision). The pull-vs-push fix, 90/10 split, and #823 access-control guards are correct and well-tested, but the deploy script ships zero constructor args (bricked funds path), several admin levers over prepaid deposits are unbounded/unobservable, and the money-movement functions (`unsubscribe`, `chargeActions` revert branches) have no test coverage.

**What's good**
- The core fix is right and pinned: `split()` uses `transfer()` (not `transferFrom`), so no SM→PS allowance is required — proven by a dedicated no-approval test. The 90/10 math is dust-clean by construction (`platform_share = amount - creator_share`), so `total_collected == total_disbursed == amount` always.
- Access-control is fail-closed and correctly scoped: `split()` is manager-only (even the owner can't call it directly), `chargeActions` is `authorized_charger`-only, `subscription_manager` defaults to `empty(address)` so a half-wired deploy can't move funds, and all of it is pinned by `TestAccessControlFix` asserting the real revert strings.
- Atomicity holds under a mid-charge revert (deactivated-pool charge rolls back the whole tx, no partial debit), and the `unsubscribe` refund path is independent of pool lifecycle so a subscriber can always recover their deposit even after the creator's pool is deactivated.
- Tests are genuinely hermetic (titanoboa in-process EVM, mock USDC at the boundary, no RPC/.env/Circle creds) and exercise the real `chargeActions → split` settlement path end-to-end with on-chain accounting cross-checks.

**Findings**

*High*

1. `contracts/script/DeploySubscriptionMarketplace.s.sol:608-621` — **Deploy script appends NO ABI-encoded constructor args, so both contracts deploy with `usdc`/`splitter`/`flat_fee` = 0 — a silently-bricked, mis-wired funds path.** Confirms Copilot (line 56), adding the confirmed in-EVM impact: `vyper -f bytecode` emits deploy bytecode *without* ctor args, so the raw `create(...)` initializes every ctor param to zero. Reproduced under titanoboa — a `PaymentSplitter` deployed with `usdc=0` succeeds silently, then every `split()` reverts (transfer to a zero-address token); `forge script --broadcast` still prints "Deployment Complete" with real addresses. Fix: append the encoded args before `create()` — `abi.encodePacked(paymentSplitterCode, abi.encode(usdc))` and `abi.encodePacked(subscriptionManagerCode, abi.encode(splitter, usdc, flatFee))` — and add post-deploy asserts that `PaymentSplitter.usdc()`, `SubscriptionManager.splitter()/usdc()/flat_fee_per_action()` are non-zero and correct before the run counts as successful. Also add Copilot's suggested init-time non-zero asserts (line 38 / 63) as defense-in-depth, and a test that exercises the raw-bytecode deploy path (`boa.load(path, args)` auto-appends ctor args and never covers this bug). Hard pre-deploy blocker.

2. `contracts/vyper/PaymentSplitter.vy:693-707,727-731` — **Owner can redirect a live subscriber's future payouts by `deactivatePool(pool_id)` then `createPool(pool_id, ATTACKER, ATTACKER)`.** `createPool` only guards `not pools[pool_id].active` (not "ever created"); `deactivatePool` doesn't clear/lock `creator`/`platform`. Subscriptions store only `pool_id`, and `split()` resolves recipients live at charge time — so the still-active subscription's already-deposited `reserved_usdc` gets routed to the new attacker recipient with no further subscriber action. This directly breaks the marketplace's core trust claim ("the creator you subscribed to receives 90%"). Owner-only (privileged), but a genuine funds-redirection backdoor. Fix: make pool recipients immutable — track a `pools_ever_created` set and assert against it in `createPool`, or snapshot `creator`/`platform` into the `Subscription` at `subscribe()` and pay from the stored addresses; emit an event if recipients ever change.

3. `contracts/vyper/SubscriptionManager.vy:897-911` — **`unsubscribe()` (a USDC refund-out path) has zero test coverage.** No test calls `unsubscribe` — every branch is uncovered: the `sub.active` guard, the `sub.subscriber == msg.sender` guard, the `remaining > 0` refund transfer, and the CEI ordering (balance zeroed before the external transfer). This is the highest-value untested money-movement function on a funds contract. The logic reads correct (CEI right, guards present), so this is a coverage gap not a demonstrated bug — but it warrants closing before the keyed deploy. Add: partial-balance unsubscribe refunds exactly `remaining` and zeroes state; non-subscriber reverts "not subscriber" with no fund movement; zero-remaining path emits `Unsubscribed` and moves no funds; double-unsubscribe reverts "subscription not active".

*Medium*

4. `contracts/vyper/SubscriptionManager.vy:913-916` (interacts with `877-895`) — **`setFlatFee` is uncapped, retroactive, and emits no event — owner can drain a subscriber's full prepaid deposit in one tx.** `flat_fee_per_action` is read live in `chargeActions` (never snapshotted at `subscribe()`), bounded only by `wallet.balance >= total_charge`, and `owner == authorized_charger` by default — so `setFlatFee(reserved)` + `chargeActions(sub_id, 1)` sweeps the whole deposit. (Note: even without `setFlatFee`, `action_count` is itself unbounded and unverifiable on-chain, so a compromised charger can drain via a large `action_count` at the default fee — `setFlatFee` is one facet of a broader charger-trust assumption.) Bounded, owner-gated, no external attacker — hence medium. Fix, cheapest first: emit a `FlatFeeChanged` event (restores on-chain observability); snapshot the agreed fee into each `Subscription` at `subscribe()` (charge against that, so `setFlatFee` only affects new subs); and/or add a `MAX_FLAT_FEE`/per-charge cap and a timelock.

5. `contracts/vyper/SubscriptionManager.vy:877-886` — **`chargeActions` revert branches are untested — the core overdraw guard has no coverage.** `action_count > 0`, `sub.active`, and `wallet.balance >= total_charge` (the "insufficient balance" overdraw protection) are all unexercised; every test charges ≤1,000 raw against a 10,000,000 deposit and never hits a revert. Vyper 0.4 underflow means it fails closed today, but the intended revert behavior is unproven. Add: over-balance charge reverts "insufficient balance" leaving `reserved_usdc` and recipient balances unchanged; `action_count=0` reverts "action_count must be > 0"; charge on an unsubscribed sub reverts "subscription not active"; and an exact-full-balance charge succeeds and drains to 0 (boundary).

6. `contracts/vyper/PaymentSplitter.vy:687-689` (same pattern `SubscriptionManager.vy:800-806`) — **No ownership transfer/renounce — the single deploy key is a permanent SPoF over the funds path.** Every sibling funds contract (`Vault.sol`, `PriceOracle.sol`, `SyntheticFactory.sol`) uses OZ `Ownable`; these two Vyper contracts uniquely lack it. If the key is compromised, `setFlatFee`/`setAuthorizedCharger`/`createPool`/`deactivatePool` are permanently attacker-held with no rotation; if lost, incident response (`deactivatePool`) is frozen. Matches the CLAUDE.md bus-factor-on-one-key Known risk. Fix: add an owner-gated two-step transfer (`pending_owner` + `acceptOwnership`) to both contracts so control can move to a multisig/timelock without a one-tx handoff mistake.

*Low / Nit*

7. `contracts/vyper/SubscriptionManager.vy:897-911` — `unsubscribe()` never clears `ephemeral_wallets[wallet].owner`, so a same-pool/same-block re-subscribe collides on the derived `wallet_address` and reverts "wallet exists" — a same-block, self-only, auto-recovering DoS plus slow unbounded state growth (no funds at risk). Clear the `ephemeral_wallets` entry in `unsubscribe`, or drop the mapping (it duplicates `subscriptions[].reserved_usdc`/`ephemeral_wallet` and the two balance counters already move in strict lockstep).

8. `contracts/vyper/SubscriptionManager.vy:876-895` / `PaymentSplitter.vy:710-725` — No pause/circuit-breaker on the charge-and-settle path. Note only `Vault.sol` (1 of 11 Solidity contracts) is `Pausable`, so this is defense-in-depth, not a repo convention — and the leaked-charger-key scenario is already addressable via the owner-only `setAuthorizedCharger`. Optional belt-and-suspenders: a `paused` flag asserted at the top of `chargeActions`/`split`.

9. `contracts/vyper/SubscriptionManager.vy:862-874,913-916` — `renewEphemeralWallet()` (a top-up money-movement fn) and `setFlatFee()` have zero coverage; both guards and the `top_up==0` branch are untested. Code reads correct; add subscriber/non-subscriber/inactive cases for renew and an owner-only + fee-applied case for `setFlatFee`.

10. `contracts/vyper/SubscriptionManager.vy:814-836` — `subscribe()` "webhook_url required" and "already subscribed" reverts are untested, and the `deposit==0` path doesn't assert `reserved_usdc==0`/no-`transferFrom`. The zero-deposit accounting assert is the one worth adding (cheap, funds-adjacent).

11. `backend/tests/contracts/test_vyper_contracts.py:126-167` — Split correctness is only tested at the clean `amount=1000` case; edge amounts (0, 1, 9, 99, 101) are untested. No bug (the `creator + platform == amount` invariant holds by construction), but parametrizing over `[0,1,2,9,99,100,101,999]` and asserting the sum invariant guards against a future rounding-direction flip.

12. `backend/tests/contracts/test_vyper_contracts.py:1-343` — No event-emission assertions; `PaymentSplit`/`ActionCharged` args (the indexer/reconciliation record) are never checked, so an event-only regression (swapped logged shares, dropped event, wrong topic) would pass. Cheap via boa `get_logs()`; the off-chain consumer is deferred, so this can wait.

13. `contracts/vyper/PaymentSplitter.vy:710-725` — No reentrancy test / `@nonreentrant` guard; CEI is correct on the traced paths so single-function reentrancy is non-exploitable with the trusted USDC token, but `split()`'s own `total_collected`/`total_disbursed` increments do land *after* the transfers (accounting-only, gates no payout). A malicious re-entrant ERC-20 mock would pin the CEI invariant as a regression guard.

14. `backend/tests/contracts/test_vyper_contracts.py:44-81` — The always-true mock USDC never exercises the `assert extcall ...transfer(...)` "transfer failed" revert branches. Low (dead against real Circle USDC, which reverts rather than returning false); a mock variant that returns `False` for a targeted recipient closes the last money-movement failure branch.

15. `backend/tests/contracts/test_vyper_contracts.py:1` and `contracts/README-subscription-marketplace.md:28,402` — Confirms Copilot (test line 1 header; README line 28): the header says titanoboa is "not in requirements.txt" while this PR adds `titanoboa>=0.2.5`, and the README says `pip install vyper==0.4.0` while the PR pins `0.4.3`. One-line edits; fix in this PR since it introduces the contradiction.

16. `contracts/vyper/PaymentSplitter.vy:692-737` (nit) — No owner rescue/sweep, so USDC accidentally sent directly to the splitter address is unrecoverable (the intended path leaves zero residual, so this is only accidental out-of-band transfers). Optionally add a guarded `rescueTokens`, or document that the splitter must never be sent USDC directly. Not blocking.

17. `contracts/vyper/PaymentSplitter.vy` / `SubscriptionManager.vy` (nit) — Transfers use `assert extcall IERC20(...).transfer(...)` rather than a SafeERC20-style `raw_call`, correct for bool-returning Arc USDC but brittle if the token is ever mis-set. The load-bearing mitigation is Copilot's init-time non-zero asserts (finding 1) — this is the funds-safety rationale for why those matter.

rcrdoh:
@dbrowneup 
following up about the architectural decision basically was for scalability and reduce the exposure for agent and platform key management, however , the current objective is to achieve the market in time , i am  doing a monolith version more easy to wire in terms of connectivity

onder-akkaya (commented):
Reviewed the payment math side. Security sign-off stays with Dan and Bogdan.

The split conserves value by construction: `platform_share = amount - creator_share`, so creator + platform always equals the input and nothing gets stranded in the contract. The platform picks up at most 1 raw unit (10^-6 USDC) of rounding per split. Charging as `action_count * fee` and splitting once, instead of splitting per action, also keeps cumulative rounding down. Good design.

Things I'd want addressed before this leaves draft, on top of what the earlier review already covers:

1. The 90/10 floor division degrades against the creator at micro amounts: `amount = 1` sends 100% to the platform, and anything below 10 gives the platform more than its 10%. That's harmless at the shipped `FLAT_FEE = 100`, but `setFlatFee` has no floor, so a future fee near 1 raw unit would systematically shortchange creators on every charge. Either add a `MIN_FLAT_FEE >= 10` or assert it in `setFlatFee`.

2. `total_collected` and `total_disbursed` are incremented together inside `split()`, so they're always equal. The collected counter doesn't measure collection, it just mirrors disbursement. Book collection where funds actually arrive, or drop one of the two.

3. `renewEphemeralWallet` emits `EphemeralWalletCreated` for a top-up, so indexers will count phantom wallet creations. A separate top-up event would be cleaner.

4. The per-user spend caps promised in #713 don't exist on-chain yet: `action_count` is unbounded, so subscriber protection currently reduces to trusting the charger key. The earlier review flagged the key risk; my point is narrower: until a cap exists, the "per-user spend caps" claim from the issue isn't true on the live path, so either add the cap in this slice or state clearly in the PR/README that it's a follow-up.

5. Worth one line in the README: "subscription" here means prepaid metering, no clock. No expiry, no proration. Fine as a model, I just don't want someone later bolting time-based logic onto a contract that has no notion of time.

+1 that the deploy script's missing constructor args are the main blocker, already flagged so I won't repeat the details. The math direction looks right to me.

dbrowneup:
Superseded by #958

PR #839, opened 2026-06-30
[backend] Persist real DSL-backtest returns so generated strategies are deployable (#788)

### LINKED ISSUE #788: [agent-ui] Agent-facing API path — dogfood the journey + capture the agent-user segment

dbrowneup:
**Dogfooding found the live-gate gap.** Running `scripts/agent_journey.py` against prod (the #788 harness) generated a fusion winner with a **real `evaluate_fusion_spec` backtest** — but the persisted strategy read:

```
rigor_gate_status: 'pending'   real_sharpe: None   passes_rigor_gate: False
```

So the server-side `create_vault` gate (#829) refuses to deploy it → **nothing generated is ever deployable**, even with a genuine backtest.

## Root cause
Fusion/debate candidates emit a DSL spec (`weights={}`) scored by `evaluate_fusion_spec`, so they **skip the static buy-and-hold `_backtest_and_persist`** (#829's keying). That leaves **no `backtest_results` row** → `live_rigor_gate` (#833) reads `pending` (it has no real persisted returns to grade).

## Fix (generation path only — gate logic untouched)
- **Capture** the DSL backtest's daily returns (from the equity curve) into the fusion `_CandidateResult.return_series`.
- **`_persist_real_returns`**: writes a `backtest_results` row from those returns + the rigor verdict (`num_trials = library+pool`, #770/#820) so `live_rigor_gate` **re-grades the REAL returns → `pass`/`fail`, not `pending`**. Non-blocking.
- **Wire it** into `run_generation` for `has_real_rigor` candidates (the ones skipped from the static backtester). Emits `backtest_done`.

**Honest by construction:** no returns / text-only fusion / <10 obs → still `pending` (never a fabricated pass); deployable only on a live `pass`. This is the live-rigor seam from the Leg A/B→C handoff — generated candidates now feed the same authoritative gate a human sees.

## Dogfooding (#788 slice 1)
- `agent_journey.py` now reads + prints the **live `rigor_gate_status` + deployability per candidate** (the same verdict a human sees on the passport), and warns if all read `pending`.
- `docs/agent-api.md` documents the live-gate read + the never-weaken-the-gate rule.

## Tests
2 hermetic (correct backtest row built from captured returns; honest skips when no real backtest); `test_generation_pipeline` regression green. The **full `pass`/`fail` end-to-end is re-verified by re-running `agent_journey.py` against prod after this deploys.**

Closes #788 (slice 1; slice-2 agent-signer for deploy/monitor remains, coordinates with Dan).

### DIFF
--- a/backend/archimedes/agents/generation_pipeline.py
+++ b/backend/archimedes/agents/generation_pipeline.py
@@ -867,6 +867,11 @@ async def _run_fusion_candidate(
     rigor_verdict: dict[str, Any]
     has_real_rigor = False
     asset_universe: list[str] = []
+    # Real DSL-backtest daily returns, captured so run_generation can persist them →
+    # live_rigor_gate reads pass/fail (not "pending") and the strategy is deployable
+    # (#788/#818). Fusion candidates skip the static buy-and-hold backtester (weights={}),
+    # so this is their only returns-persistence path.
+    real_return_series: list[float] = []
     if proposal.strategy_spec is not None:
         await emit.emit("agent_iteration", candidate_id=candidate_id, iteration_n=2, max_iterations=3)
         await emit.emit(
@@ -909,6 +914,9 @@ async def _run_fusion_candidate(
                 "total_trades": bt.total_trades,
             }
             has_real_rigor = True
+            # Daily returns from the real DSL backtest's equity curve (for live-gate persistence).
+            _ec = list(getattr(bt, "equity_curve", None) or [])
+            real_return_series = [(_ec[i] - _ec[i - 1]) / _ec[i - 1] for i in range(1, len(_ec)) if _ec[i - 1] > 0]
             await emit.emit(
                 "tool_result",
                 candidate_id=candidate_id,
@@ -964,6 +972,7 @@ async def _run_fusion_candidate(
         generation_method="fusion",
         source_arxiv_ids=list(proposal.source_arxiv_ids),
         has_real_rigor=has_real_rigor,
+        return_series=real_return_series or None,
     )
 
 
@@ -1345,6 +1354,21 @@ async def run_generation(
             ]
         )
 
+        # Fusion (and debate) candidates skipped the static backtester above but carry
+        # a REAL DSL backtest — persist its returns so live_rigor_gate reads pass/fail
+        # (not "pending") and the strategy is deployable (#788/#818). Without this, a
+        # fusion winner with a genuine backtest forever reads rigor_gate_status=pending
+        # and the server-side create_vault gate (#829) refuses to deploy it.
+        await asyncio.gather(
+            *[
+                _persist_real_returns(
+                    c, strategy_ids[c.candidate_id], emit, _society_num_trials(library_size, n_candidates)
+                )
+                for c in candidates
+                if c.has_real_rigor and c.return_series
+            ]
+        )
+
         # ── Persist all candidates to episodic memory (T-PE.8) ──
         try:
             from archimedes.services.strategy_memory import persist_proposal
@@ -1497,6 +1521,84 @@ async def _persist_candidate(c: _CandidateResult, brief: GenerateBrief) -> tuple
 
     strategy_id = await asyncio.to_thread(_do_persist)
     return strategy_id, trace_hash
+
+
+async def _persist_real_returns(c: _CandidateResult, strategy_id: str, emit: _Emitter, num_trials: int) -> None:
+    """Persist a has_real_rigor candidate's real DSL-backtest returns so the live
+    rigor gate reads ``pass``/``fail`` — not ``pending`` — and the strategy becomes
+    deployable (#788/#818).
+
+    Fusion/debate candidates emit a DSL spec (``weights={}``) scored by
+    ``evaluate_fusion_spec``, so they skip the static buy-and-hold backtester (#829)
+    and would otherwise leave NO ``backtest_results`` row → ``live_rigor_gate`` reads
+    ``pending`` → the strategy is never deployable even though it carries a real
+    backtest. This writes the row from the captured return series so the live gate
+    has real returns to re-grade. Non-blocking: any failure is logged, not raised.
+    """
+    if not c.has_real_rigor or not c.return_series or len(c.return_series) < 10:
+        return
+
+    def _do() -> None:
+        import json as _json
+
+        from archimedes.db import get_session
+        from archimedes.models.backtest import BacktestResult
+        from archimedes.services.backtest_repository import insert_backtest_if_missing
+
+        rv = c.rigor_verdict or {}
+        returns = list(c.return_series)
+        # Rebuild an equity curve (base 1.0) so BOTH the artifact daily_returns AND the
+        # equity-curve fallback in get_daily_returns resolve the same series.
+        equity = [1.0]
+        for ret in returns:
+            equity.append(equity[-1] * (1.0 + ret))
+
+        def _f(key: str) -> float:
+            v = rv.get(key)
+            return float(v) if v is not None else 0.0
+
+        result = BacktestResult(
+            strategy_id=strategy_id,
+            sharpe_ratio=_f("sharpe_ratio"),
+            sortino_ratio=_f("sortino_ratio"),
+            max_drawdown=_f("max_drawdown"),
+            cagr=_f("cagr"),
+            calmar_ratio=_f("calmar_ratio"),
+            win_rate=_f("win_rate"),
+            profit_factor=0.0,
+            total_trades=int(rv.get("total_trades") or 0),
+            avg_holding_period_days=0.0,
+            correlation_to_spy=0.0,
+            correlation_to_btc=0.0,
+            equity_curve=equity,
+            deflated_sharpe_ratio=rv.get("dsr"),
+            dsr_p_value=rv.get("dsr_p_value"),
+            num_trials_in_selection=int(num_trials),
+            pbo_score=rv.get("pbo"),
+            out_of_sample_sharpe=rv.get("oos_sharpe"),
+            look_ahead_audit_passed=bool(rv.get("lookahead_audit_passed", False)),
+            backtest_engine="dsl-fusion",
+        )
+        artifact = {"results": [{"metrics": {"daily_returns": returns}}], "source": c.generation_method}
+        artifact_json = _json.dumps(artifact, default=str)
+        content_hash = hashlib.sha256(artifact_json.encode("utf-8")).hexdigest()
+        with get_session() as session:
+            insert_backtest_if_missing(
+                session,
+                strategy_id=strategy_id,
+                content_hash=content_hash,
+                result=result,
+                operation="DSL_FUSION",
+                artifact_json=artifact_json,
+            )
+
+    try:
+        await asyncio.to_thread(_do)
+        await emit.emit(
+            "backtest_done", candidate_id=c.candidate_id, strategy_id=strategy_id, source=c.generation_method
+        )
+    except Exception as exc:
+        logger.warning("persist_real_returns failed for %s: %s", strategy_id, exc)
 
 
 async def _backtest_and_persist(c: _CandidateResult, strategy_id: str, emit: _Emitter, num_trials: int = 1) -> None:
--- a/backend/archimedes/agents/generation_pipeline.py
+++ b/backend/archimedes/agents/generation_pipeline.py
@@ -1523,6 +1523,59 @@ async def _persist_candidate(c: _CandidateResult, brief: GenerateBrief) -> tuple
     return strategy_id, trace_hash
 
 
+def _refresh_passport_real_metrics(
+    session: Any, c: _CandidateResult, strategy_id: str, result: Any, *, passes_rigor_gate: bool, n_obs: int
+) -> None:
+    """Refresh the strategy_passports ``real_*`` columns + ``passes_rigor_gate`` for a
+    fusion/debate candidate whose real returns were just persisted.
+
+    The single-strategy read path (``_passport_to_strategy_response``) derives
+    ``rigor_gate_status`` from the STORED passport columns (``pending`` while
+    ``sharpe_ratio is None``) and the deploy gate (``_strategy_rigor_status``) reads
+    ``record.passes_rigor_gate`` — neither re-grades the ``backtest_results`` row. So
+    persisting the row alone leaves both at ``pending``; this in-place passport update
+    is what makes the endpoint + deploy gate see the real verdict. Mirrors the passport
+    refresh in ``_backtest_and_persist``. ``passes_rigor_gate`` is the live-gate re-grade
+    of the real returns (single source of truth).
+    """
+    from datetime import date as _date
+
+    from archimedes.models.paper_ref import PaperRef
+    from archimedes.models.strategy import StrategyPassport, StrategyStatus
+    from archimedes.models.strategy_store import StrategyRecord
+    from archimedes.services.passport_loader import ingest_passport
+
+    record = session.query(StrategyRecord).filter_by(id=strategy_id).first()
+    status_val = StrategyStatus(record.status) if record and record.status else StrategyStatus.CANDIDATE
+    papers = [PaperRef(arxiv_id=p.get("arxiv_id"), title=p.get("title", "")) for p in (c.source_papers or [])]
+    regime_tag = {"bull": "bull", "bear": "bear"}.get(c.regime, "regime_neutral")
+    passport = StrategyPassport(
+        id=strategy_id,
+        papers=papers,
+        methodology_summary=c.thesis or "",
+        asset_universe=c.asset_universe or [],
+        status=status_val,
+        regime_tag=regime_tag,
+        real_sharpe=result.sharpe_ratio,
+        real_sortino=result.sortino_ratio,
+        real_cagr=result.cagr,
+        real_max_dd=result.max_drawdown,
+        real_calmar=result.calmar_ratio,
+        real_corr_spy=result.correlation_to_spy,
+        real_total_trades=result.total_trades,
+        real_backtest_start=(result.backtest_start.isoformat() if isinstance(result.backtest_start, _date) else None),
+        real_backtest_end=(result.backtest_end.isoformat() if isinstance(result.backtest_end, _date) else None),
+        deflated_sharpe_ratio=result.deflated_sharpe_ratio,
+        dsr_p_value=result.dsr_p_value,
+        num_trials_in_selection=result.num_trials_in_selection,
+        pbo_score=result.pbo_score,
+        out_of_sample_sharpe=result.out_of_sample_sharpe,
+        passes_rigor_gate=passes_rigor_gate,
+        n_obs_daily=n_obs,
+    )
+    ingest_passport(session, passport, generation_method=c.generation_method, force_update=True)
+
+
 async def _persist_real_returns(c: _CandidateResult, strategy_id: str, emit: _Emitter, num_trials: int) -> None:
     """Persist a has_real_rigor candidate's real DSL-backtest returns so the live
     rigor gate reads ``pass``/``fail`` — not ``pending`` — and the strategy becomes
@@ -1533,9 +1586,23 @@ async def _persist_real_returns(c: _CandidateResult, strategy_id: str, emit: _Em
     and would otherwise leave NO ``backtest_results`` row → ``live_rigor_gate`` reads
     ``pending`` → the strategy is never deployable even though it carries a real
     backtest. This writes the row from the captured return series so the live gate
-    has real returns to re-grade. Non-blocking: any failure is logged, not raised.
+    has real returns to re-grade, and refreshes the passport ``real_*`` columns +
+    ``passes_rigor_gate`` (which is what the single-strategy read path + the deploy
+    gate actually consume). Non-blocking: any failure is logged + emitted, not raised.
+
+    **Claim-integrity gate:** persists ONLY when the backtest ran on REAL data
+    (``data_source != "synthetic"``). ``evaluate_fusion_spec`` defaults to a random
+    synthetic series when no ``data_feed`` is wired, and grading a random walk as a
+    real pass/fail would be a #1-rule violation — so a synthetic run stays honestly
+    ``pending`` (never persisted, never deployable). Real-data wiring is tracked
+    separately; until then fusion candidates correctly read ``pending``.
     """
+    rv = c.rigor_verdict or {}
     if not c.has_real_rigor or not c.return_series or len(c.return_series) < 10:
+        return
+    if str(rv.get("data_source") or "synthetic") == "synthetic":
+        # Synthetic backtest — grading it as real would be a claims-integrity violation.
+        logger.info("persist_real_returns: %s ran on synthetic data — staying honestly 'pending'", strategy_id)
         return
 
     def _do() -> None:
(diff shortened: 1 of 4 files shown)


dbrowneup:
## 🔬 Deep review — Claude multi-agent (5 dimensions → adversarial verify → synthesize)

_Posted on Dan's behalf, to complement the Copilot review. Both **Critical** findings were independently re-verified against the live code before posting:_
- _**#2** confirmed — the sibling `_backtest_and_persist` calls `session.commit()` while `_persist_real_returns` does not, and `insert_backtest_if_missing` only flushes._
- _**#1** confirmed — the persist guard checks `has_real_rigor/return_series/len` but ignores the `admissible`/`data_source` flags the code captures two lines earlier (≈891–892); `run_dsl_backtest` defaults `data_feed=None → _synthetic_data()` (`random.gauss`, logged "NOT admissible")._

---

**Verdict** — **Request changes.** The persisted "REAL DSL-backtest returns" are not real on the live path: `evaluate_fusion_spec` is called with no `data_feed`, so `run_dsl_backtest` falls back to a `random.gauss` synthetic series (`data_source="synthetic"`, `admissible=False`), and the persist guard ignores those provenance flags — a random-walk series is written into `backtest_results` and graded pass/fail by the live gate. And separately, the write is `flush()`-only with no `session.commit()`, so in production the row rolls back and the gate stays `pending` forever. Both defeat the PR's stated deliverable, and the first is a #1-rule claims-integrity violation.

**What's good**
- The tri-state honesty guard is right where it applies: `<10` obs / missing `return_series` / `has_real_rigor=False` all short-circuit to honest `pending` rather than fabricating a pass (generation_pipeline.py:1493), and the negative test covers all three branches.
- Read-path contract is carefully wired: the artifact `daily_returns` and the rebuilt base-1.0 `equity_curve` resolve to the same series, matching both branches of `get_daily_returns` (artifact-first, equity fallback) — no divergence bug.
- Path-routing avoids double-persistence: static gather filters `generation_method != "fusion"`; fusion candidates (`weights={}`, `has_real_rigor` default False on agent path) never get two `backtest_results` rows.
- Tests are genuinely hermetic (mock at `get_session`/`insert_backtest_if_missing` boundary, pass under `env -i`), and `agent_journey.py` adds a real `saw_non_pending` dogfooding canary. The docs correctly state "Never weaken the gate to force a pass."

**Findings**

*Critical*

1. **generation_pipeline.py:874, 904–907, 1493** — Persisted "real" returns are **synthetic random-walk noise**, and the provenance flags to block it are already in hand and ignored. `_run_fusion_candidate` calls `evaluate_fusion_spec(proposal.strategy_spec)` with no `data_feed`/`data_csv_path` → `run_dsl_backtest(spec, data_feed=None)` → `_synthetic_data()` (`random.seed(42)` + `random.gauss(0.0003, 0.012)`), stamping `data_source="synthetic"`, `admissible=False`. `has_real_rigor` is set True purely on `eval_result.success` (line 904) — "the DSL ran," not "the data was real" — and `_persist_real_returns` guards only on `has_real_rigor / return_series / len>=10`, never on the `admissible`/`data_source` values it captured two lines earlier into `rigor_verdict` (891–892). Downstream, `live_rigor_gate.verdict_from_returns`/`run_rigor_gate` have **no data-source parameter**, so they grade the noise and can return `pass` → a non-`pending` badge, and (post-commit-fix) a deployable verdict on random prices. This affects **every** fusion/debate candidate. **Fix:** gate persistence (and the line-1329 gather filter) on `(c.rigor_verdict or {}).get("admissible")` / `data_source != "synthetic"` so synthetic runs stay honestly `pending`; better, wire a real `data_feed`/OHLCV CSV into the `evaluate_fusion_spec` call so the returns are genuinely real. Also carry `data_source` into the persisted artifact so the gate can refuse to grade a synthetic-sourced row.

2. **generation_pipeline.py:1540–1548** — **Missing `session.commit()` → the write is a production no-op.** `get_session()` returns a raw `SessionLocal()` (`autocommit=False`, db.py:53/105-107); the Session CM `__exit__` calls `.close()` (rollback), and `insert_backtest_if_missing` only `session.flush()`es (backtest_repository.py:51). So the flushed `backtest_results` row is discarded on close → `get_daily_returns` returns `[]` → live gate reads `pending` → `create_vault` refuses to deploy: exactly the state this PR exists to fix. The sibling `_backtest_and_persist` in this same file commits explicitly (line 1695) — the required step was dropped. Worse, the failure is **silent**: no exception, and `backtest_done` still emits (1552), so SSE/`agent_journey` output looks correct while the write vanishes — a false success signal. **Fix:** add `session.commit()` after `insert_backtest_if_missing(...)`, mirroring line 1695; then have the test's `_fake_session()` yield a stub with a `commit()` spy and assert it fires once. (Empirically reproduced on sqlite: flush-only → 0 rows; flush+commit → 1 row.)

*High*

3. **generation_pipeline.py:1496–1548** — **Passport is never updated, so the single-strategy read path and deploy gate still read `pending` even after the commit fix.** The sibling `_backtest_and_persist` does two writes — the backtest row **and** `ingest_passport(force_update=True)` setting `real_sharpe`/`passes_rigor_gate` (1655–1695). `_persist_real_returns` does only the backtest row. But `GET /api/strategies/{id}` for a generated (passport-backed) strategy routes through `_passport_to_strategy_response`, which derives `rigor_gate_status` from the **stored passport columns** (`"pending" if record.sharpe_ratio is None` — strategies_routes.py:1265-1267), and the deploy gate `_strategy_rigor_status` reads `record.passes_rigor_gate` (vaults_routes.py:89) — neither re-grades the new `backtest_results` row. Since `_persist_candidate` creates the fusion passport with no `real_sharpe`, both the endpoint and the deploy gate **still return `pending`** after this PR. Only the batch library-list badge (`verdicts_for_strategies` → `get_all_daily_returns`) consumes the row — and the doc + `agent_journey.py` both exercise the single-strategy endpoint, not the list. **Fix:** either have `_persist_real_returns` also `ingest_passport(force_update=True)` with `real_sharpe` + `passes_rigor_gate` re-graded via `verdict_from_returns` + commit (mirroring 1655–1695), or make `_passport_to_strategy_response`/`_strategy_rigor_status` compute the live verdict from persisted returns. Then verify end-to-end that a fusion winner returns pass/fail, not pending.

4. **test_live_gate_returns.py** — **Tests mock away exactly the two broken behaviors, giving false confidence in "REAL" + "deployable."** The happy-path test injects `return_series` by hand into a `_CandidateResult`, bypassing `evaluate_fusion_spec`, and its `_RIGOR` fixture omits `data_source`/`admissible` — so it structurally cannot catch the synthetic leak (#1). It also stubs `insert_backtest_if_missing` and yields a bare `object()` as the session — so it cannot catch the missing commit (#2). Both go green while production is doubly broken; this is the "test the production code path, not the easy one" anti-pattern from CLAUDE.md. **Fix:** (a) add a real tmp-sqlite session test (copy the `_use_tmp_db` precedent in test_api_routes.py), call `_persist_real_returns`, then read back via `get_daily_returns` and assert the row survives (catches #2); (b) add a test that drives a synthetic/`admissible=False` verdict and asserts `insert_backtest_if_missing` is **never** called once the provenance gate lands (catches #1).

*Medium*

5. **docs/agent-api.md:~116** — The example response for `GET /api/strategies/{id}` lists `"real_sharpe"` and `"real_dsr"`, which do **not** exist on `StrategyResponse` (schemas.py:162-230). Because the endpoint declares `response_model=StrategyResponse`, FastAPI strips both keys, so they can never appear. The real metrics are `sharpe_ratio` and `deflated_sharpe_ratio`. **Fix:** document the fields the endpoint actually returns (`sharpe_ratio`, `deflated_sharpe_ratio`, `dsr_p_value`, `out_of_sample_sharpe`, `pbo_score`); keep `rigor_gate_status`/`passes_rigor_gate` — those two are correct.

*Low / Nit*

6. **scripts/agent_journey.py:223** — Reads `strat.get('real_sharpe')`, which is unconditionally `None` (same non-existent field as #5), so the live-gate readback always prints `real_sharpe=None` even when a real backtest exists — misleading in the harness the test docstring names as the end-to-end verification path. **Fix:** read `sharpe_ratio` (optionally `deflated_sharpe_ratio`). The load-bearing `live_gate=`/`deployable=` fields are correct.

7. **generation_pipeline.py:1550–1556** — The persist `except` branch is `logger.warning` only, with **no `backtest_failed` SSE** (unlike the sibling at 1717–1726). Since the call site is `asyncio.gather(...)` without `return_exceptions=True`, a swallowed persist exception is indistinguishable in the stream from an honest `pending`. Fail-safe direction (never fabricates a pass), so low. **Fix:** emit a `persist_failed`/`backtest_failed` SSE with `candidate_id`/`error` for sibling parity and dogfooding fidelity; the untested branch (below) should cover it.

8. **test_live_gate_returns.py** — Test coverage nits (all low/nit): the `len < 10` skip is tested only at 2 obs — add `9→skip` / `10→persist` boundary cases and share `_MIN_RETURNS_FOR_GATE` rather than hardcoding `10` in three places (pipeline / gate / selection_bias_routes); the `except`-swallow branch is untested — add a `side_effect=Exception` test asserting it returns without raising and emits no `backtest_done`; `content_hash` is captured but never asserted — assert 64-char hex + determinism (it keys `insert_backtest_if_missing` idempotency); and the equity-curve→returns derivation (`generation_pipeline.py:906-907`) is never unit-tested for empty/None/zero-point edge cases.

9. **generation_pipeline.py:1505–1509** (nit) — Redundant `returns→equity→returns` round-trip; the rebuilt equity curve is rebased to 1.0, not the strategy's real magnitude. It works because `get_daily_returns` prefers the artifact returns, but a consumer reading `equity_curve` for level/drawdown-in-currency gets a normalized proxy. **Fix:** carry `bt.equity_curve` through `_CandidateResult` and persist it directly, or add a one-line comment that this curve is a rebased proxy.

**vs Copilot**
- Confirms Copilot on the missing-commit rollback (its :1548) and the synthetic-data risk (its :1330). Net-new: I trace both all the way to a **deployed vault** (#1) and prove the commit divergence against the in-file sibling at 1695 + empirical sqlite rollback (#2), and surface the **silent false-`backtest_done`** emit Copilot didn't call out.
- Net-new beyond Copilot's inline notes: the **passport-never-updated** gap (#3) — even after Copilot's commit fix, the single-strategy read path and deploy gate still read `pending`; the **doc/script field drift** (`real_sharpe`/`real_dsr`, #5/#6); and the **test-adequacy thesis** (#4) explaining *why* green CI gave false confidence in "REAL"+"deployable."
- Copilot's `:1533` (raw string/decimal rigor values could leak into the mapper) is real and pairs with finding #8's untested `except` branch — that branch is exactly what would silently swallow such a mapper error today.

dbrowneup:
Excellent review — all confirmed and fixed. Rebased onto main (815/816 merged), pushed. Per finding:

**Critical**
1. **Synthetic data leak** ✅ `_persist_real_returns` now gates on the captured `data_source` — a synthetic backtest (`evaluate_fusion_spec` with no `data_feed` → `_synthetic_data` random walk) is **never persisted**; it stays honestly `pending`. `data_source` is also carried into the artifact. Grading a random walk as real pass/fail is exactly the #1-rule violation, so this is now impossible.
2. **Missing `session.commit()`** ✅ Added (mirrors the sibling `_backtest_and_persist`). On failure it now also emits `backtest_failed` (finding 7 — no silent swallow). Proven by the new tmp-sqlite test: flush-only → 0 rows; flush+commit → row survives.

**High**
3. **Passport never updated** ✅ New `_refresh_passport_real_metrics` writes the passport `real_*` columns + `passes_rigor_gate` (re-graded via `verdict_from_returns`, the single source of truth), so the single-strategy read path (`_passport_to_strategy_response`) and deploy gate (`_strategy_rigor_status`) — which read passport columns, not the backtest row — now see the real verdict.
4. **Tests mocked away the bugs** ✅ Replaced with a **real tmp-sqlite integration suite**: persist → `get_daily_returns` proves the row **survives** (catches the commit), a synthetic candidate is **never persisted** (catches the synthetic leak), and the live verdict reads `pass`/`fail` (not `pending`).

**Medium / Low**
5. **docs fields** ✅ Now documents what `StrategyResponse` actually returns (`sharpe_ratio`, `deflated_sharpe_ratio`, `dsr_p_value`, `out_of_sample_sharpe`, `pbo_score`).
6. **harness `real_sharpe`** ✅ Reads `sharpe_ratio` now.
7. **swallowed persist error** ✅ Emits `backtest_failed`.

**Remaining (flagged, not in this PR):** wire a **real `data_feed`** into the fusion backtest so it's `admissible` → deployable. Until then fusion candidates correctly read `pending` (never a fabricated pass) — this PR makes the persistence honest + correct; the real-data wiring is the next step to actual deployability.

dbrowneup:
## 🔬 Re-review — Claude multi-agent (round-2: verifying the round-1 fixes)

_Posted on Dan's behalf. **Verdict: APPROVE** — both round-1 Criticals are RESOLVED, spot-confirmed against the diff:_
- _**C1** (synthetic data): persistence is now gated on `data_source == "synthetic"` — synthetic runs stay honestly `pending`. (Stronger than the `admissible` gate suggested in round 1.)_
- _**C2** (rollback): explicit `session.commit()` present, proven load-bearing by a real tmp-sqlite readback test._
_Only 3 non-blocking findings remain (2 Low + 1 Nit). Nice work on the round-1 fixes._

---

I have everything I need. Both prior Criticals (C1 synthetic-data gate, C2 explicit commit) are verified resolved on the live path, plus H3/H4. Copilot's four inline comments are all stale/addressed. Producing the review.

**Verdict:** APPROVE — both prior Criticals are RESOLVED on the live path (C1: persistence now refuses synthetic-sourced returns via the `data_source == "synthetic"` guard; C2: the explicit `session.commit()` is present and proven load-bearing by a real tmp-sqlite regression test), all four of Copilot's round-1 inline comments are now stale, and only three non-blocking findings remain.

## What's good

- **C1 is gated at the correct choke point, and more precisely than Copilot proposed.** Copilot's round-1 suggestion was to gate on `rigor_verdict['admissible']`; the PR instead gates on `data_source == "synthetic"` (`generation_pipeline.py:138`). This is the better call — `admissible = passing AND non-synthetic`, so gating on `admissible` would have wrongly skipped a *real-data* backtest that legitimately **failed** rigor, leaving an honest `fail` stuck at `pending`. Gating on synthetic-vs-real persists real-but-failing backtests as an honest `fail` (not deployable) while refusing to grade random-walk noise as a real verdict. Since `evaluate_fusion_spec` runs with no `data_feed` on the live path, fusion candidates run on `_synthetic_data()` → correctly refused → stay honestly `pending`. **#1 rule upheld.**
- **C2's commit is proven load-bearing, not decorative.** `get_session()`'s `__exit__` closes without committing, so the explicit `session.commit()` (`generation_pipeline.py:213`) is required; `test_real_returns_persist_and_survive_commit` asserts the row **survives** session close via genuine `get_daily_returns` readback — a real regression guard, not a mock-of-the-bug.
- **H4 tests are honest integration tests.** Real per-test tmp-sqlite (rebinding `db.engine` + `db.SessionLocal`), persist run inline, assertions on real readback. `test_synthetic_returns_are_never_persisted` directly exercises the C1 refusal. Copilot's round-1 "fake session needs `commit()`" concern no longer applies — the bare `object()` session is gone.
- **H3 write-record and read-record are the same table** end-to-end: `_refresh_passport_real_metrics` → `ingest_passport(force_update=True)` writes `record.sharpe_ratio` + `record.passes_rigor_gate`, and both the passport endpoint (`_passport_to_strategy_response`) and the server-side deploy gate (`_strategy_rigor_status`, `vaults_routes.py:89`) read that same record.
- Optional rigor metrics are float-cast via `_of()`/`_f()` (`generation_pipeline.py:158-164`), closing Copilot's raw-`rv.get` type-leak concern; `total_trades`/`num_trials` int-cast, `look_ahead_audit_passed` bool-cast.
- The `real_return_series` recompute (equity-curve deltas with the `equity_curve[i-1] > 0` guard) exactly mirrors `fusion_evaluator`'s internal `daily_returns` formula, and `BacktestMetrics` exposes only `equity_curve` — so recomputing is the correct/only path.
- Docs + `agent_journey.py` reference only fields that exist on `StrategyResponse`; the LIVE GATE step prints an honest `⚠ all strategies read live_gate='pending'` warning that correctly describes today's synthetic-data reality.

## Findings

### Low

1. **`generation_pipeline.py:1354-1360` (gather comment) + `114-133` (docstring) — comments claim debate coverage the code never delivers.** Both say "Fusion **(and debate)** candidates … persist its returns." But debate candidates are built by `debate_engine._make_entry`, which sets `has_real_rigor=True` yet **never sets `return_series`** (defaults to `None`). The persist filter is `if c.has_real_rigor and c.return_series`, so every debate candidate is excluded — only `_run_fusion_candidate` captures `real_return_series`. A debate winner with a genuine C-rigor backtest would still read `pending` forever. Blast radius today is zero (debate is behind `ARCHIMEDES_DEBATE_ENABLED`, default OFF, with a PHASE-1-BLOCKER header), so this is a comment-accuracy defect + a latent gap, not a live #1-rule violation — but notable in a claims-integrity PR. **Fix:** either capture `return_series` in `_make_entry` from `ev.backtest.equity_curve` (mirror `_run_fusion_candidate` diff L22-23, ~2 lines), or correct the comment/docstring to say "fusion candidates" only.

2. **`generation_pipeline.py:195-196` (docstring) vs `live_rigor_gate.py:164` — persisted `passes_rigor_gate` uses a different `num_trials` than the library-list re-grade, so the docstring's parity claim is over-strong.** At persist time `passes_rigor_gate` is computed via `verdict_from_returns(..., num_trials=_society_num_trials(library_size, n_candidates))` and consumed by the passport read + deploy gate. But the library-list endpoint recomputes live via `verdicts_for_strategies`, which uses `num_trials = max(len(valid_returns), 1)` — a different denominator. Since the DSR correction is monotonic in `num_trials`, the same returns can pass under one denominator and fail/pending under the other, so the library badge can disagree with the passport/deploy-gate badge. This divergence is **pre-existing** (`_backtest_and_persist` has the same shape) — inherited, not introduced — but the docstring asserts the persisted value "matches what `verdicts_for_strategies` computes," which is only true when the two denominators coincide. **Fix:** thread the same `num_trials` the library-list re-grade uses, or soften the docstring at L195-196 to note the denominators can differ.

### Nit

3. **`backend/tests/test_live_gate_returns.py:29 / diff L252+L262` — module double-imports `archimedes.db`** (`from archimedes.db import get_session` at module level + `import archimedes.db as db` inside `_use_tmp_db`). Both resolve correctly and all tests pass — pure style inconsistency flagged by `github-code-quality[bot]`, no functional impact and none on the C1/C2/H3/H4 resolutions. **Fix (optional):** hoist `import archimedes.db as db` to module level, drop the `from` import, reference `db.get_session` throughout. Safe to defer for a hackathon-cadence PR.

## Round-1 resolution table

| Round-1 finding | Status |
| --- | --- |
| **C1** — persisting synthetic fusion returns would grade a random walk as a real verdict (#1-rule) | **Resolved** — gated on `data_source == "synthetic"` (`generation_pipeline.py:138`); synthetic runs stay honestly `pending`. Gate is stronger than Copilot's `admissible` suggestion. |
| **C2** — `insert_backtest_if_missing()` only flushes; row rolls back on session close → gate stuck at `pending` | **Resolved** — explicit `session.commit()` (`generation_pipeline.py:213`), proven load-bearing by `test_real_returns_persist_and_survive_commit` (real tmp-sqlite readback). |
| **H3** — persisted verdict must be the same record the passport endpoint + deploy gate read | **Resolved** — `_refresh_passport_real_metrics` writes `record.sharpe_ratio` + `passes_rigor_gate`; passport read + `_strategy_rigor_status` consume that same record. |
| **H4** — tests must exercise the production commit path, not mock away the bug | **Resolved** — real per-test tmp-sqlite, inline persist, genuine `get_daily_returns` readback; Copilot's "fake session needs `commit()`" no longer applies (bare `object()` session removed). |
| Copilot: gate persistence on synthetic data | **Resolved** — implemented as `data_source`-based refusal (better than the `admissible` gate proposed). |
| Copilot: raw `rv.get()` leaks str/Decimal into the DB mapper | **Resolved** — optional rigor fields cast via `_of()`/`_f()`; `total_trades`/`num_trials` int-cast, look-ahead bool-cast. |

PR #843, opened 2026-07-01
[strategy] Cast Sharpe CI to plain float — fix passport DB sync rollback

dbrowneup:
## Summary

Local BYOK dogfooding of the generation pipeline surfaced a silent, live-path
bug: the strategy-passport **"unified table sync" rolls back** on any strategy
whose Sharpe confidence interval is computed.

```
unified table sync failed (non-blocking): (psycopg2.errors.InvalidSchemaName)
schema "np" does not exist
[parameters: {... 'sharpe_ci_lower': np.float64(-0.3294),
              'sharpe_ci_upper': np.float64(0.528) ...}]
```

**Root cause.** `compute_sharpe_ci` computes `z = norm.ppf(...)`, and scipy's
`norm.ppf` returns a **numpy scalar**. Without a cast, `z * se` and the returned
tuple are `np.float64`, so `sharpe_ci_lower` / `sharpe_ci_upper` reach the
`strategy_passports` INSERT as numpy scalars. On **numpy ≥ 2.0** (what the
backend container runs) `repr(np.float64(-0.33))` is `np.float64(-0.33)`;
psycopg2 emits that into the SQL and parses `np` as a schema →
`InvalidSchemaName: schema "np" does not exist` → the whole passport-sync
transaction rolls back. It's caught and logged `(non-blocking)`, so it failed
**silently** — the passport just never synced.

## Fix

Cast the returned tuple to plain `float`, honoring the existing
`-> tuple[float, float]` annotation. One-line change at the source, so **every**
caller is fixed (the passport insert, `strategy_provider`, etc.).

## Verify

```bash
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend \
  python -m pytest backend/tests/test_rigor_evaluator.py -k sharpe_ci -q   # 7 passed
```

The new `test_sharpe_ci_returns_plain_python_float` is an airtight guard: run
against the un-fixed helper it fails with `lower is float64, must be plain
float`; it uses `type(x) is float` because `isinstance(np.float64(x), float)`
is `True` (numpy subclasses float) and would not catch the regression.

## Anti-goals

- No change to the CI math (Lo 2002 formula untouched — the symmetric/width
  pins still pass).
- No threshold or gate changes.

### DIFF
--- a/backend/archimedes/services/_rigor_helpers.py
+++ b/backend/archimedes/services/_rigor_helpers.py
@@ -741,7 +741,12 @@ def compute_sharpe_ci(
     sr_daily = sharpe_annual / math.sqrt(_ANNUALIZATION)
     se = math.sqrt((1.0 + 0.5 * sr_daily**2) * _ANNUALIZATION / n_obs_daily)
     z = norm.ppf((1.0 + confidence) / 2.0)
-    return (sharpe_annual - z * se, sharpe_annual + z * se)
+    # `norm.ppf` returns a numpy scalar, so `z * se` (and the returned tuple)
+    # is np.float64 unless cast. A raw np.float64 breaks the passport DB insert:
+    # psycopg2 renders it as `np.float64(-0.33)` and parses `np` as a schema →
+    # `InvalidSchemaName: schema "np" does not exist`, rolling back the passport
+    # sync. Cast to plain float to honor the tuple[float, float] contract.
+    return (float(sharpe_annual - z * se), float(sharpe_annual + z * se))
 
 
 # ─── 6. Private helpers ──────────────────────────────────────────────
--- a/backend/tests/test_rigor_evaluator.py
+++ b/backend/tests/test_rigor_evaluator.py
@@ -295,6 +295,21 @@ def test_sharpe_ci_symmetric_around_point_estimate():
     assert abs(mid - sr) < 1e-9, f"CI midpoint {mid:.6f} must equal SR {sr}"
 
 
+def test_sharpe_ci_returns_plain_python_float():
+    """Regression: the CI must be plain ``float``, never ``np.float64``.
+
+    ``norm.ppf`` returns a numpy scalar, so without an explicit cast the tuple
+    leaks ``np.float64`` into ``sharpe_ci_lower``/``sharpe_ci_upper`` on the
+    strategy passport. psycopg2 then renders it as ``np.float64(-0.33)`` and
+    parses ``np`` as a schema, raising ``InvalidSchemaName: schema "np" does
+    not exist`` and rolling back the passport DB sync. ``isinstance(x, float)``
+    does NOT catch this (np.float64 subclasses float) — assert the exact type.
+    """
+    lower, upper = compute_sharpe_ci(1.0, n_obs_daily=252, confidence=0.95)
+    assert type(lower) is float, f"lower is {type(lower).__name__}, must be plain float"
+    assert type(upper) is float, f"upper is {type(upper).__name__}, must be plain float"
+
+
 def test_sharpe_ci_wider_with_fewer_obs():
     """Fewer observations → wider confidence interval."""
     sr = 0.8


dbrowneup:
## 🔬 Review — Claude multi-agent (dimensions → adversarial verify → synthesize)

_Posted on Dan's behalf._

---

**Verdict:** Approve — the fix is correct, minimal, on the live path, and well-tested; ship it.

**What's good**
- **Correct, minimal fix.** `compute_sharpe_ci` was the sole metric-returning function in `_rigor_helpers.py` leaking `np.float64`: `z = norm.ppf(...)` is a numpy scalar, so `z * se` and the returned tuple were `np.float64`. Every sibling already casts (`compute_dsr` L323 `float(norm.cdf(z))`; `compute_oos_sharpe` L486 `round(float(...), 6)`; `compute_in_sample_sharpe` L524 `float(...)`). The patch brings the outlier into line with convention and honors the declared `tuple[float, float]` contract.
- **The stated failure mechanism holds on production's numpy pin.** psycopg2's `Float` adapter (matched via MRO for the `float` subclass `np.float64`) formats via `repr()`; on numpy 2.x `repr(np.float64(-0.33)) == "np.float64(-0.33)"`, so `adapt(x).getquoted()` emits `b'np.float64(-0.33)'`, which Postgres parses as a schema-qualified reference → `InvalidSchemaName: schema "np" does not exist` → rollback of the passport sync (swallowed at `strategy_provider.py:504-525`). With `numpy>=2.4.6` pinned (`backend/requirements.txt:21`, `environment.yml:40`) this is the live prod/CI path. Docstring narrative is accurate.
- **No precision loss.** `float(np.float64(x))` is bit-exact — both are IEEE-754 doubles (verified via `struct.pack('d', …)` identity).
- **Correct regression guard.** The test asserts `type(lower) is float`, not `isinstance`. Because `issubclass(np.float64, float)` is `True`, an `isinstance` check would pass on the buggy `np.float64` and miss the regression. The exact-type assertion is the right pin, and the 6 sharpe_ci tests pass hermetically (`env -i … pytest → 6 passed`).
- **Edge cases incidentally hardened.** The same cast turns `np.float64('nan'/'inf')` into plain-float `nan`/`inf`, which psycopg2 renders as valid `'NaN'`/`'Infinity'` Postgres literals (pre-fix on numpy 2.x these failed identically). `None` bounds are gated upstream (`strategy_provider.py:348-349`, `:424-425`).

**Findings**

_Re: Copilot's inline comment (test_rigor_evaluator.py:306) — do not act on it_
1. `backend/tests/test_rigor_evaluator.py:306` — **Copilot's inline comment is factually wrong; the PR docstring is correct.** Copilot claims "NumPy scalar floats are not subclasses of Python `float`." That is false for `np.float64`: `issubclass(np.float64, float)` is `True` (verified on the env's numpy 1.26.4, and it remains `True` on numpy 2.x — numpy deliberately keeps `float64` a Python-`float` subclass, unlike `float32`/`float16`). Since `norm.ppf` yields `np.float64`, `isinstance(x, float)` would return `True` on the buggy value and *not* catch the regression — which is exactly why the test asserts `type(x) is float`. Adopting Copilot's suggested reword ("an exact-type regression guard") would delete the accurate, load-bearing rationale for why `isinstance` was rejected. **Fix:** reject the Copilot comment and keep the `isinstance` explanation as written. Optionally add a one-word clarifier that `np.float64` *specifically* subclasses `float` (to preempt confusion with `np.float32`, which does not).

_Nits (non-blocking, no PR change required)_
2. `backend/archimedes/services/_rigor_helpers.py:743` (docstring) — **The DB error is not locally reproducible under the dev env's numpy 1.26.4.** The mechanism depends on numpy 2.x's `repr(np.float64(-0.33)) == "np.float64(-0.33)"`; on numpy 1.26.4 the repr is plain `-0.33` and `adapt(np.float64(-0.33)).getquoted()` returns `b' -0.33'` (clean SQL, no error). The bug is genuinely live only on the pinned numpy-2.x prod/CI env — so the fix is prod-correct but a local DB repro is impossible. The regression test itself is version-independent (`type(x) is float` distinguishes `np.float64` on every numpy version). **Fix:** none needed here. Separately, consider rebuilding the local conda env from `environment.yml` to the `numpy>=2.4.6` pin so numpy-2-repr-sensitive behaviors are exercised locally rather than only in prod — the drift is a latent "works locally, breaks in prod" hazard for other repr-sensitive paths.

Issue #822, opened 2026-06-29
[quant] Effective-N (candidate-pool correlation ρ̄) in the society DSR — approach B for #770


onder-akkaya:
## Summary

This is the **"approach B"** calibration that #770 reserved for a statistics sign-off. PR #811
(closing #770) set the society-path DSR trial count to `num_trials = N + library_size` and feeds
it with `average_correlation = 0` — i.e. it treats all `N + library_size` trials as **mutually
independent**. The additive count is a sound, conservative upper bound and the direction is safe
(it can only tighten the gate), so approach A is correct as a default. But the independence
assumption is mis-calibrated: the `N` society candidates are generated from the **same user brief
over overlapping universes**, so they are typically strongly correlated (ρ̄ ≈ 0.5–0.9). Under
equicorrelation the *effective* number of independent trials is

    N_eff = N / (1 + (N − 1)·ρ̄)        (Cheverud 2001; Nyholt 2004)

which at ρ̄ = 0.7, N = 10 is ≈ 2.3, far below the nominal count. Feeding ρ̄ = 0 therefore
**over-deflates** `E[max_N]` → over-states `SR_zero` → biases the Tier-1 gate **too strict**: it
can reject genuinely-skilled strategies because correlated siblings are counted as independent
searches.

The machinery to fix this already exists and is unused on the society path:
`_rigor_helpers.py` implements the `N_eff` equicorrelation conversion (the `average_correlation`
argument of `compute_dsr`, ~line 152) and `compute_average_pairwise_correlation`. Approach B is
to estimate ρ̄ from the candidate-pool return matrix and pass it through.

## Proposed resolution

- In the society path, compute ρ̄ = mean pairwise correlation across the `N` candidate return
  series (`compute_average_pairwise_correlation`) and pass it as `average_correlation` into
  `compute_dsr` at both DSR derivation sites (`generation_pipeline.py:355` and the
  `portfolio_backtester` call at `:400`), so the deflation uses `N_eff` rather than the nominal
  `N + library_size`.
- Keep approach A (ρ̄ = 0, fully independent) as the **fallback** when fewer than two candidate
  series are available (no correlation estimable) — it is the conservative default.
- Document the choice in `docs/specs/selection-bias-corrections-spec.md` §1.3 (the addendum
  already flags B as a follow-up).

## Scope (exact files)

- `backend/archimedes/agents/generation_pipeline.py` (society DSR call sites + ρ̄ estimation).
[...]

PR #859, opened 2026-07-03
[quant] Unify DSR num_trials across the live and fusion generation paths (#820)

### LINKED ISSUE #820: [quant] Unify DSR num_trials across the live + fusion generation paths
## Summary

The DSR multiple-testing trial count (`num_trials`) is now computed under **three different
conventions**, and the divergence is not fully justified — one path is materially more
permissive than the others, so the identical "passes rigor" badge means different things
depending on how a strategy was generated.

| Path | `num_trials` | Where |
| --- | --- | --- |
| Live agent (society) | `N + library_size` | `generation_pipeline.py` `_society_num_trials` (#770/#811) |
| `/selection-bias/gate` route | `library_size` | `selection_bias_routes.py` (**intentional** — grades the already-persisted library; documented) |
| Fusion candidate | `_default_num_trials()` = `library_size`, or `len(parameter_variants)` (≈3) | `fusion_evaluator.py` (`evaluate_fusion_spec` called with no `num_trials`) |

The `/gate`-route convention is statistically defensible (it grades a single-layer selection set
— the persisted library — with no fresh N-candidate pool). The **fusion path is the problem**: a
fusion winner *also* survived the N-candidate dual-regime society search, yet its DSR is deflated
for the library alone (or, when a variant grid exists, for only ~3 trials). It therefore
**under-deflates relative to the live path**, so the same strategy gets a higher (easier-to-pass)
DSR via fusion than via the live runner.

PR #815 (debate society) currently deflates by `num_trials = pool_size` (the candidate pool N
alone, without `library_size`), which is **a fourth convention** and perpetuates the divergence
in the opposite direction. This issue should be the single place we reconcile all of them before
#815 lands.

## Proposed resolution

- Make `_society_num_trials(library_size, selection_pool_size)` the **single source of truth**
  for the trial count on every generation path that produces a DSR-gated candidate.
- Thread `selection_pool_size` into `_run_fusion_candidate` / `evaluate_fusion_spec` so the
  fusion path deflates for the same `N + library_size` selection set as the live path.
- Reconcile #815's `pool_size`-only count with this helper (or document precisely why the debate
  society's selection set differs).
- Keep the `/gate` route on `library_size` (it is a different, correctly single-layer set) — but
  add a one-line comment cross-referencing this issue so the difference is visibly intentional.

## Scope (exact files)

- `backend/archimedes/services/fusion_evaluator.py` (`_default_num_trials`, `effective_trials`,
  `evaluate_fusion_spec` signature).
- `backend/archimedes/agents/generation_pipeline.py` (`_run_fusion_candidate` call site).

## Acceptance criteria

- [ ] A strategy generated via the fusion path and via the live path receives the **same**
      `num_trials` for an identical `(selection_pool_size, library_size)`; test pins it.
- [ ] The fusion path no longer collapses to `len(variants)` when that under-counts the society
      selection layer.
- [ ] Existing fusion-evaluator tests pass; a regression test pins the unified count.

## Verify

```bash
pytest backend/tests/test_generation_pipeline.py backend/tests/services/test_fusion_evaluator.py -q
```

## Anti-goals

- Do not change the `/selection-bias/gate` route's `library_size` count — it is a correctly
  different selection set.
- Do not alter the DSR formula or the effective-N correlation handling (tracked separately).

## Note for teammates

@dbrowneup @mnemonik-dev — this touches the #770/#811 deflation and the fusion path you've both
worked in, and it intersects the open #815. No rush; I wanted it tracked so the four conventions
converge to one before more paths are added. Happy to take the implementation on the quant side.

onder-akkaya:
Closes #820.

## What

The fusion candidate path received `selection_pool_size` but never used it
(marked `noqa: ARG001`) and called `evaluate_fusion_spec` with no `num_trials`
at all, so it fell back to library-size-only deflation. The live society path
already deflates for `library_size + selection_pool_size` via
`_society_num_trials`. Same brief, same candidate pool, different (weaker)
correction — a strategy could clear the fusion path's DSR gate at a count the
live path would have rejected it at.

- `_run_fusion_candidate` now computes `num_trials = _society_num_trials(library_size, selection_pool_size)`,
  same as `_run_live_candidate`, and threads it into `evaluate_fusion_spec`.
- `apply_rigor_gate`'s variant-grid override used to replace `num_trials` with
  `len(variants_metrics)` outright whenever a parameter sweep existed, which
  could silently undercut a correct, larger passed-in count (a 3-variant grid
  overriding a real library+pool count of 15, say). Now takes
  `max(num_trials, len(variants_metrics))` — whichever selection layer is
  bigger wins, so the gate is never looser than either alone.
- `debate_engine.py`'s `num_trials` was already fixed to use
  `_society_num_trials(library_size, pool_size)` (the call site's own comment
  says so), but several docstrings and one inline comment still described the
  old `num_trials=pool_size` behavior. Corrected those so the file doesn't
  contradict its own logic.
- One-line comment in `selection_bias_routes.py` noting the `/gate` route's
  library-size-only count is the deliberate exception, not a leftover fourth
  convention — per the issue's ask to reconcile or document each convention.

## Verify

```bash
pytest backend/tests/test_generation_pipeline.py backend/tests/services/test_fusion_evaluator.py -q
```

New `test_fusion_path_num_trials_matches_society_formula` proves the fusion
path and the live-path formula agree for an identical `(library_size,
selection_pool_size)`, including with a 3-variant grid in play. Two new
`apply_rigor_gate` tests pin the `max()` composition both ways (a larger
passed-in count wins over a small grid; a genuinely larger grid still wins
over a smaller passed-in count). 489 passed across the full
rigor/generation/fusion/debate/selection_bias surface. ruff format + check
clean.

## Anti-goals checked

- Did not change the `/selection-bias/gate` route's `library_size` count —
  comment-only addition there, confirmed via diff.
- Did not touch the DSR formula or effective-N correlation handling
  (`_rigor_helpers.py` is untouched by this PR — that's #822, a separate
  concurrent PR).

### DIFF
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -19,11 +19,11 @@ society:
      rebuttal + LLM risk-debate is Phase 2.
   3. **C-rigor** — backtests EVERY survivor for real via ``evaluate_fusion_spec``
      (deterministic Python, 0 tokens), each wrapped in try/except (fix A5
-     backstop), with ``num_trials=pool_size`` so the DSR multiple-testing
-     correction counts the selection-from-pool search. This is the self-contained
-     A1 deflation at the per-candidate evaluator — it coordinates with #770's
-     ``library + N`` additive correction on the generation path rather than the
-     library-grading external route, so it does not fight that PR.
+     backstop), with ``num_trials=_society_num_trials(library_size, pool_size)``
+     (library + N, #770/#820 — not ``pool_size`` alone) so the DSR
+     multiple-testing correction counts both the selection-from-pool search AND
+     the library the winner joins, the same selection set the live agent path
+     deflates for.
   4. **C-null** — a survivor must beat the passive null (buy-and-hold) net of
      cost by ``MIN_COST_BENEFIT``. If none clears it → first-class ABSTAIN.
   5. **Synthesizer** — deterministic rank of the survivors (Phase 1 collapses the
@@ -34,9 +34,11 @@ contract preserving); the full leaderboard is built by ``build_leaderboard``
 (directly unit-tested) and the Considered-Alternatives fan-out is Phase 2.
 
 ⚠️  PHASE-1 BLOCKER: ``ARCHIMEDES_DEBATE_ENABLED`` must stay OFF on the live path
-until A1 (``pool_size`` → the external rigor gate, the OPEN Önder denominator)
-is proven there — else the rigor badge can false-PASS. The flag-OFF additive
-skeleton in this module is safe to merge.
+until this module's rigor badge has been validated end-to-end there (A1's
+denominator now matches the live path's — library + N via
+``_society_num_trials``, #770/#820 — but the flag still needs a real on-flag
+run before it's trusted). The flag-OFF additive skeleton in this module is
+safe to merge.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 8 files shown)


dbrowneup (approved):
Verified the formula threading in _run_fusion_candidate/debate_engine (review workflow: 46/46 + 419/419 hermetic). One gap closed on the branch (a47edde→): **_run_fusion_job** — the legacy direct-fusion route — was still calling evaluate_fusion_spec without num_trials, i.e. the exact library-size-only fallback this PR unifies away. Threaded `_society_num_trials(library, 1)` there too with a spy proof test (24 tests pass). Merging on green.

PR #860, opened 2026-07-03
[quant] PBO statistical-power floor at small library N (#819)

### LINKED ISSUE #819: [quant] PBO/CSCV statistical-power floor at small library N
## Summary

`compute_pbo` (`_rigor_helpers.py`) is a faithful Combinatorially-Symmetric Cross-Validation
(CSCV) estimator of the Probability of Backtest Overfitting (Bailey, Borwein, López de Prado &
Zhu 2014): `C(16,8)=12,870` IS/OOS splits, IS-optimal config ranked OOS, logit of the relative
rank, `PBO = #{λ ≤ 0}/#splits`. The mechanics are correct.

The concern is **statistical power at our operating point**, not correctness. CSCV was designed
to assess a *single strategy's parameter sweep*, where the cross-section is hundreds of
configuration variants. We apply it across the **heterogeneous strategy library**, whose size
is currently N ≈ 4–6. At N = 4 the OOS relative rank `ω = rank/N` takes only ~4 distinct values,
so the logit `λ = ln(ω/(1−ω))` is coarsely quantized and `PBO` is granular and **library-coupled**:
adding or removing one neighbour can flip a member's verdict. The estimator is consistent as
N → ∞ but underpowered and unstable at single-digit N — and it currently *gates* Tier-1 admission
(criterion 4).

The docstring already flags the off-label use honestly, so this is not a claim-integrity bug;
it is a request to make the gate's behaviour at small N statistically defensible.

## Proposed resolution

Introduce a minimum-N floor for PBO **gating** (not computation):

- Below the floor (e.g. N < ~10–12; exact value to be justified from the logit granularity),
  report PBO as `NOT_RUN` with an explicit reason — mirroring the CPCV `NOT_RUN` treatment
  added in #771 — and let criterion 4 be neutral/skipped rather than gating on an underpowered
  statistic.
- At or above the floor, gate as today.
- Still **compute and surface** PBO below the floor (advisory), so the number is visible
  without being load-bearing.

The floor value is the statistics call here; I'll justify it from the quantization of the
relative-rank distribution and a small power simulation.

## Scope (exact files)

- `backend/archimedes/services/_rigor_helpers.py` (`compute_pbo` / its callers) and
  `backend/archimedes/services/rigor_evaluator.py` (`RigorGateResult.passes_all` /
  `gate_details["pbo"]`) — add the N-floor gating logic + honest `NOT_RUN` surface.

## Acceptance criteria

- [ ] With a library below the floor, `gate_details["pbo"]` reads `NOT_RUN (N=… below power floor)`
      and criterion 4 does not gate; tests pin this.
- [ ] At/above the floor, behaviour is unchanged; existing PBO tests still pass.
- [ ] A short power note (docstring or `docs/specs/selection-bias-corrections-spec.md`) records
      why the floor sits where it does.

## Verify

```bash
pytest backend/tests/test_rigor_evaluator.py backend/tests/test_selection_bias_routes.py -q
```

## Anti-goals

- Do not change the CSCV math itself — it is correct; only its gating applicability at small N.
- Do not stop *computing* PBO below the floor — keep it advisory and visible.

## References

- Bailey, D., Borwein, J., López de Prado, M. & Zhu, Q. (2014). *The Probability of Backtest
  Overfitting.* Journal of Computational Finance.
- Casella, G. & Berger, R. (2002). *Statistical Inference* (2nd ed.) — finite-sample vs
  asymptotic behaviour of rank-based statistics.

onder-akkaya:
Closes #819.

## What

`compute_pbo`'s CSCV mechanics are correct at any N — this isn't a math bug.
The concern is power: each split's OOS relative rank `omega = rank/N` only
takes N possible values, and 0.5 (the PBO decision boundary) is landed on
exactly whenever `rank = N/2` — true for every even N. Worse, PBO is a
comparison relative to whoever else is in the library, so removing or
adding one strategy changes `1/N` of the field for every split. At our
library size today (N ~ 4-6) that's 17-25% of the comparison set turning
over on a single add/remove — exactly the "one neighbor flips a verdict"
instability the issue describes, and it's structural, not something that
needs a simulation to see.

Ran a leave-one-out instability simulation anyway (remove a random member
from a no-true-skill pool, check how often the library's `PBO < 0.5` verdict
flips) at N in {4, 6, 8, 10, 12, 16}:

| N | leave-one-out flip rate | power (planted-edge) |
|---|---|---|
| 4 | 0.20 | 0.60 |
| 6 | 0.40 | 0.60 |
| 8 | 0.20 | 0.20 |
| 10 | 0.10 | 0.60 |
| 12 | 0.10 | 0.40 |
| 16 | 0.10 | 0.80 |

Noisy at small N (only 10 trials per point — `compute_pbo` costs ~2.2s/call
regardless of N, dominated by the `C(16,8)=12,870`-split Python loop, not
array width, so this had to stay small), but the flip rate settles at its
floor of ~10% from N=10 onward and never improves past that through N=16,
matching the `1/N = 10%` structural argument almost exactly. **Floor: N=10.**

- New `MIN_LIBRARY_N_FOR_PBO_GATING = 10` in `rigor_evaluator.py`.
- `RigorGateResult` gets a `pbo_library_size` field. Below the floor,
  `gate_details["pbo"]` reads `NOT_RUN` (still shows the number, advisory)
  and `passes_all` skips criterion 4 entirely rather than hard-failing on
  an underpowered statistic. At or above the floor, nothing changes.
- `run_rigor_gate` auto-derives `pbo_library_size` from `len(pbo_scores)` on
  the cohort path (that dict already has one entry per strategy in the
  cross-section, so its length IS the N) — the two real call sites
  (`selection_bias_routes.py`, `live_rigor_gate.py`) need no changes. The
  `library_pbo` path does NOT auto-derive from `pbo_scores` (that dict
  describes a different cross-section there) — needs an explicit
  `pbo_library_size` if it should ever gate on the floor too.
- Every existing direct `RigorGateResult(...)` construction is unaffected:
  `pbo_library_size` defaults to `None`, so the floor never applies unless a
  caller opts in.

Full derivation is in `selection-bias-corrections-spec.md`'s PBO addendum.

## Verify

```bash
pytest backend/tests/test_rigor_evaluator.py backend/tests/test_selection_bias_routes.py -q
```

9 new tests (below/at/above-floor gating, missing-vs-present PBO below the
floor, cohort auto-derivation, and that the `library_pbo` path doesn't
accidentally borrow `pbo_scores`'s length). 219 passed on the issue's verify
command, 474 passed across the full rigor/pbo/selection_bias/fusion/
generation/dsr surface. ruff format + check clean.

## Anti-goals checked

- Did not change the CSCV math — `_rigor_helpers.py` (`compute_pbo`) has a
  zero-line diff.
- PBO is still computed and surfaced below the floor (advisory), never
  dropped.

### DIFF
--- a/backend/archimedes/services/rigor_evaluator.py
+++ b/backend/archimedes/services/rigor_evaluator.py
@@ -336,6 +336,19 @@ def load_daily_returns_store(
 
 # ─── 6. Rigor Gate — composite check ─────────────────────────────────
 
+# #819: minimum library N for PBO/CSCV to gate criterion 4. Below this, the
+# CSCV OOS relative rank omega = rank/N is too coarsely quantized (few
+# distinct logit values, and the comparison set itself dominates the result)
+# for PBO to carry real statistical power, and it becomes library-coupled —
+# adding or removing one neighbor can flip a member's verdict. 10 is where a
+# leave-one-out instability simulation's flip rate settles at its floor (~10%,
+# matching the exact 1/N argument for how much one removal can move an N-member
+# comparison set) and stays there through N=16 tested. See
+# docs/specs/selection-bias-corrections-spec.md § 2 addendum for the full
+# derivation + simulation numbers. PBO is still computed and surfaced below
+# this floor — it just stops gating (see RigorGateResult).
+MIN_LIBRARY_N_FOR_PBO_GATING = 10
+
 
 class RigorGateResult:
     """Result of running all four selection-bias checks on a strategy."""
@@ -359,6 +372,7 @@ class RigorGateResult:
         iid_assumption_violated: bool | None = None,
         iid_diagnostics: dict | None = None,
         regime_robustness: dict | None = None,
+        pbo_library_size: int | None = None,
     ) -> None:
         self.strategy_id = strategy_id
         self.deflated_sharpe = deflated_sharpe
@@ -377,6 +391,11 @@ class RigorGateResult:
         # #546), "cohort" when it fell back to the per-cohort dict score. Surfaced
         # in gate_details so the verdict is honest about its provenance.
         self.pbo_source = pbo_source
+        # #819: the N (number of strategies) the PBO cross-section was computed
+        # over. None when unknown (e.g. no PBO source at all — the pre-#819
+        # MISSING path). Used only to decide whether N clears
+        # MIN_LIBRARY_N_FOR_PBO_GATING; never changes the PBO value itself.
+        self.pbo_library_size = pbo_library_size
         self.oos_sharpe = oos_sharpe
         self.look_ahead_passed = look_ahead_passed
         self.in_sample_sharpe = in_sample_sharpe
@@ -402,6 +421,12 @@ class RigorGateResult:
         self.regime_robustness = regime_robustness
 
     @property
+    def _pbo_below_power_floor(self) -> bool:
+        """True when the PBO cross-section's N is too small for CSCV to carry
+        real statistical power (#819) — criterion 4 must not gate in that case."""
+        return self.pbo_library_size is not None and self.pbo_library_size < MIN_LIBRARY_N_FOR_PBO_GATING
+
+    @property
     def passes_all(self) -> bool:
         # NaN-hardening: every IEEE-754 comparison against NaN is False, so a NaN
         # metric (not None — None is guarded) would silently skip its fail branch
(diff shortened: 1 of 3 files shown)


dbrowneup (approved):
Quant review workflow verified the floor's behavior (219/219 on the stated Verify command, hermetic) and confirmed no interaction with the fusion/pipeline hard PBO gates (those get #859's num_trials fix; a follow-up is noted for extending the power floor there if wanted). The one claims-integrity nit — the spec's 'Reproduce' section pointing at a script the PR description doesn't contain — is fixed in ac0ca98 (points at TestPboPowerFloor + the structural 1/N argument instead). Merging on green.

PR #861, opened 2026-07-03
[quant] Migrate the daily_returns store from committed JSON to the database (#774)

### LINKED ISSUE #774: [quant] Migrate the daily_returns store from committed JSON to the database
## Summary

The 22 files at `analytics-engine/strategies/daily_returns/*.json` (~240k committed lines) are the data substrate for the library-level PBO (Bailey et al. 2014, CSCV) that gates Tier-1 admission. Committing raw per-strategy daily-return series to git bloats the tree, makes every diff noisy, and couples the rigor gate's data source to a checkout instead of a queryable, versioned store. Each file already carries the right fields (`stem`, `data_vintage`, `dates[]`, `daily_returns[]`), so the migration is mechanical: move the series into a database table, source the library-PBO from the DB, and drop the JSON from the working tree (git history retains it). This decouples the rigor substrate from the repo and lets `data_vintage` filtering be a query rather than a file scan, without touching the PBO math.

## Scope (exact files)

- New table `strategy_daily_returns` (columns: `stem`, `date`, `daily_return`, `data_vintage`) + migration in the backend models/migrations path.
- `analytics-engine/scripts/gen_daily_returns_store.py` — add a DB-write path (keep JSON gen behind a flag for a transition window, or emit straight to DB).
- `backend/archimedes/services/rigor_evaluator.py` — `load_daily_returns_store` reads from `strategy_daily_returns` (returns the shape `compute_library_pbo` expects); `compute_library_pbo` / `compute_pbo` UNCHANGED.
- `analytics-engine/scripts/compute_library_pbo.py` — source from DB.
- `backend/archimedes/api/selection_bias_routes.py` — `/api/selection-bias/gate` unchanged behavior, DB-backed.
- Remove `analytics-engine/strategies/daily_returns/*.json` from the working tree.
- New parity test, e.g. `backend/tests/test_daily_returns_store_parity.py`.

## Acceptance criteria

- [ ] Parity (exact): `pytest backend/tests/test_daily_returns_store_parity.py -q` → `passed`, asserting library-PBO computed from the DB equals library-PBO computed from the JSON snapshot, exactly (same float, no tolerance loosening).
- [ ] JSON removed: `ls analytics-engine/strategies/daily_returns/*.json 2>/dev/null | wc -l` → `0`.
- [ ] Gate unaffected: `pytest backend/tests/ -k "selection_bias or rigor or pbo or library_pbo" -q` → `0 failed`.
- [ ] Full suite green: `pytest -m "not integration" -q` → `0 failed`.

## Verify

```bash
git rm -r --cached analytics-engine/strategies/daily_returns   # if still tracked
ls analytics-engine/strategies/daily_returns/*.json 2>/dev/null | wc -l   # 0
pytest backend/tests/test_daily_returns_store_parity.py -q
pytest backend/tests/ -k "selection_bias or rigor or pbo" -q
```

## Anti-goals

- Do NOT change the PBO formula. `compute_pbo` and `compute_library_pbo` in `rigor_evaluator.py` are parity-tested — touch only the data source, not the math.
- Do NOT weaken existing rigor thresholds, DSR/PBO partitions (`s_partitions`), or the fail-closed too-short-window contract.
- Do NOT silently change `data_vintage` semantics — it must round-trip identically.
- Do NOT delete the JSON from git history; removal from the working tree only.

onder-akkaya:
Closes #774.

## ⚠️ Operational gap before merge — read this first

The real 22-strategy dataset only exists in the JSON files this PR deletes
from the working tree. **Someone with production DB access needs to run
`backend/scripts/import_daily_returns.py` against the real environment
before or at deploy**, or the live library-PBO input goes empty right after
merge. It fails closed (gate criterion 4 reports `NOT_RUN` / no aligned
strategies, not a silently wrong number), but it's still a functional
regression on the rigor gate until backfilled. This is a production-DB
write — neither I nor the reviewing session can run it ourselves; it needs
a human with the AWS/DB access this repo's conventions gate that behind.

Backfill steps (run in order, against the real environment):
```bash
cd analytics-engine && uv run python scripts/gen_daily_returns_store.py --write   # only if data needs a fresh run
cd .. && conda run -n archimedes python backend/scripts/import_daily_returns.py analytics-engine/strategies/daily_returns
```

## What

The 22 files under `analytics-engine/strategies/daily_returns/` (~240k
committed lines) fed the library-level PBO
(`rigor_evaluator.compute_library_pbo`). Committing raw per-strategy return
series bloats the tree and makes every diff on that directory noisy — the
rigor gate needs the data queryable, not checked out. `compute_pbo` /
`compute_library_pbo` themselves are untouched; only where the data comes
from changes.

- New `strategy_daily_returns` table (`backend/archimedes/models/daily_returns_store.py`),
  one row per `(stem, date)`, unique on `(stem, date)`. Reference DDL at
  `backend/migrations/20260703_add_strategy_daily_returns.sql`. Wired into
  `db.py`'s `init_db()` side-effect imports so `create_all()` picks it up.
- `rigor_evaluator.load_daily_returns_store` now queries the table instead
  of scanning `*.json` files. Same `{stem: {dates, daily_returns}}` shape
  out, same max-vintage computation, same "never raises, degrades to
  `({}, None)`" contract — just DB-unreachable instead of directory-absent
  as the degrade trigger. Dates round-trip as the same `YYYY-MM-DD` strings
  the JSON files used (`date.isoformat()`).
- `selection_bias_routes.py`'s library-PBO cache moved its invalidation
  signature from `(filename, mtime)` tuples to `(row_count, max_id)` — the
  DB-native equivalent, since a stem's replace-on-remeasure (delete +
  reinsert) gets fresh autoincrement ids and still flips `max_id`.
- New `backend/scripts/import_daily_returns.py` reads whatever JSON
  `gen_daily_returns_store.py` produces and upserts it into the table
  (delete + reinsert per stem, so re-running on unchanged input is a
  no-op in effect).
- **Deliberately did NOT teach `gen_daily_returns_store.py` itself to write
  to Postgres.** `analytics-engine` is a standalone, dependency-light
  package (`backtrader`/`pandas`/`yfinance` only, no DB driver) —
  adding one is a new top-level dependency, which this repo's convention
  flags for a team call, not something to decide unilaterally inside a
  migration PR. `compute_library_pbo.py` (the analytics-engine diagnostic
  script) is left alone for the same reason — it's a standalone tool
  reading whatever JSON exists locally, unaffected by the git-tracked
  files going away. Flagging this scoping choice for review — happy to
  revisit if the team wants analytics-engine talking to Postgres directly.
- Removed the 22 committed JSON files (`git rm`, working tree only — still
  in git history).

## Verify

```bash
pytest backend/tests/test_daily_returns_store_parity.py -q
ls analytics-engine/strategies/daily_returns/*.json 2>/dev/null | wc -l   # 0
pytest backend/tests/ -k "selection_bias or rigor or pbo or library_pbo" -q
pytest -m "not integration" -q
```

New `test_daily_returns_store_parity.py` proves library-PBO computed from
the DB equals library-PBO computed from the pre-#774 JSON-snapshot shape
exactly (same float). `TestLoadDailyReturnsStore` and `TestCachedLibraryPbo`
rewritten DB-backed (tmp SQLite, isolated engines, no shared global state).
JSON file count: 0. 286 passed on the gate-surface command. **2233 passed on
the full backend suite, 0 failed.** ruff format + check clean.

## Anti-goals checked

- Did not change the PBO formula — `_rigor_helpers.py` (`compute_pbo`) has
  a zero-line diff.
- Did not weaken any rigor threshold, DSR/PBO partition, or the fail-closed
  too-short-window contract.
- `data_vintage` round-trips identically — proven by the parity test.
- JSON removed from the working tree only, not git history.

## Cross-lane note

This touches `db.py` / `models/` (Daniel R.'s lane per CLAUDE.md) as well as
the quant rigor-gate surface — flagging in case a second pass from that side
is useful, though one approving review is the repo's normal bar for
non-contract changes.


dbrowneup (approved):
Review workflow verified: hermetic tests (tmp-SQLite, never real DATABASE_URL), compute_pbo/compute_library_pbo untouched (0-line diff), fail-closed on empty/unreachable table (library_pbo → null, never a wrong number). One medium nit for a follow-up, not blocking: _store_signature's (row_count, max_id) cache key assumes delete+reinsert always increases max_id — true on Postgres SERIAL, but worth a comment or a version column if the store ever moves.

**Operational gate (#862):** verified the prod DB directly — strategy_daily_returns is ABSENT. I'm running the backfill via SSM immediately after this merge deploys (Dan-authorized this session); live /api/selection-bias/gate will show library_pbo null only for the deploy+import window.

PR #865, opened 2026-07-03
Feature/market v1.1 monolith

rcrdoh:
Marketplace Monolith v1.1 — Subscription/Copy-Trading Payment Plane
Summary
Introduces the full subscription marketplace feature: strategy creators can publish an agent for others to subscribe to and mirror, with automated per-action USDC billing and 90/10 creator/platform fee splitting settled on-chain via Circle Gateway + Arc Testnet.
What's new

Backend marketplace package (backend/archimedes/marketplace/) — new service.py (tick/publish/subscribe/retire lifecycle, ~1000 lines), payments.py (per-action charge logic), settlement.py (three-stage sweep: Gateway → publisher agent wallet → PaymentSplitter.depositToPool → creator/platform withdraw), state.py, tick_registry.py, encoding.py, and wallet_provisioner.py (Circle Developer-Controlled Wallet provisioning per publisher).
On-chain: new PaymentSplitter.vy (Vyper) — the sole on-chain marketplace contract. Subscriptions themselves are now off-chain/Postgres-only (no SubscriptionManager contract), simplifying the payment-attribution problem raised by Circle Gateway's address-only settlement model — attribution is now keyed by per-publisher agent wallet rather than calldata.
API: new marketplace_routes.py; updates to strategies_routes.py and generate_routes.py to support publish/subscribe flows.
DB: eight new migrations covering strategy generators, marketplace agents, subscriber liabilities, tick logging, halted-agent state, and publisher/subscriber Circle wallet columns.
UI: new MarketplacePage.jsx, PublishPage.jsx, StrategyDetailPage.jsx, SubscriptionsPage.jsx (replacing the old Marketplace.jsx); updates to Strategies.jsx, Layout.jsx, App.jsx, and api.js.
Infra: new docker/publisher/Dockerfile and docker/subscriber/Dockerfile for the agent-side execution containers.
Tests: new coverage for the marketplace service (tick, retire, rehydration, per-step charging), payments, settlement, contract-level PaymentSplitter tests (titanoboa), and API route tests.

Notes for reviewers

This branch reflects resolution of the previously open decision gates (payment rail, chargeActions behavior, mid-tick halt handling, and the Gateway settlement-attribution problem) — worth confirming these choices match current thinking before merge.
Given the history of regressions at spec seams in this feature, recommend a focused look at service.py ↔ settlement.py ↔ PaymentSplitter.vy boundaries (pool struct fields, ABI shape) and the remove_subscriber / unsubscribe path specifically.


dbrowneup:
## Multi-agent deep review (5 dimensions, adversarially verified) — advisory, do not merge as-is

Ricardo — first, real credit: **this is the strongest marketplace attempt in the lineage, and the first to ship a genuinely working, tested Circle Gateway x402 payment pipeline.** Neither #823 nor #749 got that far. The review also confirmed you *structurally* fixed both prior CRITICAL bug classes: charging is now fully in-process (no HTTP-exposed `chargeActions` surface), pool creation is owner-gated (kills #824's payout-redirection), the `.agents/` leak class is closed at the root via `.gitignore`, and there's zero Polymarket/prediction-market scope creep. `PaymentSplitter.vy` itself is solid contract engineering (CEI ordering, `@nonreentrant` on both value-moving functions, per-pool balance isolation, 20+ tests).

That said, five findings survived adversarial verification (2 critical, 3 high) — each independently confirmed against the actual diff at head:

### 🔴 Critical
1. **Deploy script bricks PaymentSplitter** (`contracts/script/DeploySubscriptionMarketplace.s.sol:26-43`): `USDC_ADDRESS` is read and `console.log`'d but **never ABI-encoded/appended to the init bytecode** before `create(0, add(code, 0x20), mload(code))`. Vyper's `__init__(_usdc: address)` therefore gets `address(0)` → every `depositToPool`/`withdraw` targets `IERC20(address(0))`. This is the same defect class flagged on #824. Fix: append `abi.encode(usdc)` to the bytecode + a post-deploy `require(PaymentSplitter(splitter).usdc() == usdc)` assertion. *(I verified this line-by-line myself, not just via the review agents.)*
2. **Copy-trade settlement is REAL by default** (`backend/archimedes/marketplace/service.py`): the tick engine calls `executor.execute_trades` on subscribers' vaults, gated only by `dry_run` from `AGENT_DRY_RUN` defaulting **false**, auto-booted at startup. North-Star §6 is explicit: payments stay real on testnet, **trade settlement stays paper/simulated until mainnet**. This recurs (arguably escalates) the violation flagged on #823.

### 🟠 High
3. **One `dry_run` flag conflates payments and settlement** — no configuration can express "payments real + settlement stubbed," which is exactly the §6 requirement. They need independent controls, with settlement defaulting to paper.
4. **`strategy_runner_publisher.py`/`_subscriber.py` + their two new Dockerfiles still call `SubscriptionManager.chargeActions()`** — the exact contract+function behind #823's CRITICAL. No SubscriptionManager source/ABI exists in the repo, so it's dead today, but it's newly wired into deployable containers. Delete or quarantine (strip the call at minimum).
5. **`withdraw()` was changed (2026-07-03) from creator/platform-gated to fully permissionless.** Payouts still go to fixed addresses so it's not fund-theft, but it reopens the "no caller check on a fund-moving function" pattern behind two prior CRITICALs, and enables a read-then-withdraw TOCTOU race against the backend's own accounting. Needs an explicit Dan+Bogdan sign-off either way. Related: `platform_wallet` silently defaults to the creator's own wallet, so the advertised 90/10 split is routinely 100/0 with $0 platform capture.

Also: 4 tests reference `/api/marketplace/payment-webhook`, a route that doesn't exist in the shipped code — dropped late or coming in a follow-up?

### Supersession calls (for the #713 consolidation)
- **#823** → superseded *with grafts*: #865 kills its attack surface, but must enforce paper settlement (finding 2) before closing.
- **#824** → superseded *with grafts*: bring back the `withdraw()` gate (or a reviewed sign-off on permissionless) and fix the shared deploy-script defect.
- **#749** → superseded *with grafts*: the per-subscriber daily USDC **spend cap** (SpendCapStore, 24h sliding window) has no equivalent in #865 — graft it before any real payment flow goes live.

### Minimal path to mergeable
1. Fix the deploy script (encode `_usdc`, post-deploy assertion). 2. Split `dry_run` into payments/settlement controls; settlement defaults to paper (§6). 3. Resolve `withdraw()` gating with Dan+Bogdan. 4. Require an explicit platform treasury address (no creator fallback). 5. Delete/quarantine the legacy runner path + Dockerfiles. 6. Reconcile the webhook tests. 7. Graft the spend cap. 8. Rebase (3 conflicting files) + full Dan+Bogdan contract pass. 

This PR is the right **payments-layer foundation** — with the fixes above it can close #713 and retire the other three PRs. Happy to pair on any of the mechanical items; the contract-gating decisions are Dan+Bogdan's call.

*Review: 5 parallel dimension agents + 2-vote adversarial verification per critical/high finding; headline critical additionally hand-verified. Advisory only — no merge action taken.*

rcrdoh:
thanks for the review, indeed still there are code from the previous version: container per user (in road map)  . I will clean the code. 
respect to the two critical findings : 
1 the script for smart contract deployments has to be updated to deploy only one contract : PaymentSplitter.vy , subscriptionManager.vy contract currently is not necessary . Improvements for PaymentSplitter.vy are welcome.
2Can tell me corrections to me in this point. As i am understanding , it is needed to implement a switcher (in this case DRY_RUN , false or true ) that allows on chain operations or not .

 respect to high findings:
3 point two related consequences 
4 this python script for an  agent  is for container per user version .  for this monolith version does not intervene in the functionality.
5 the concern here is that programmatically key pair addresses created by circle can't be managed as the normal ones. To improve about this concern , is important to know before how is handled the addresses by the circle API.

onder-akkaya (changes requested):
Ricardo — read the whole thing, here's the tea.

Short version first because it matters: the fund-theft bug I flagged on #823 (`SubscriptionManager.chargeActions` with no `msg.sender` check) is not sitting in your live path. `MarketService`, `payments.py`, `service.py` all compute `action_count` server-side off real trade data, never off a number the client hands over. Wallets are Circle-custodied, no private key ever touches your code, and the SIWE gate on `marketplace_routes.py` matches the pattern the rest of the app already uses. That part's clean and I'm not going to manufacture a problem just to look thorough.

The rest needs work before this merges.

`PaymentSplitter.vy`'s `withdraw()` (line 293) is permissionless on purpose, per your own docstring, and it's genuinely not the #823 shape. The amount is capped by `held_balance`, which only grows through real `transferFrom` pulls, and the two payout addresses are fixed to `pool.creator`/`pool.platform`. Nobody's draining a pool that isn't theirs. What it does let anyone do is force a disbursement on someone else's pool whenever they feel like it. As the actuary in the room: that's not tail risk on the principal, that's just a free option on other people's gas bills. Still want that written down as a team decision, not just a code comment, before it ships.

Bigger issue, same file: this PR and Dan's still-open #824 both add `contracts/vyper/PaymentSplitter.vy`, the same ABI, the same README, the same deploy script, with two contradictory designs. Yours is permissionless-but-capped, his is identity-gated-but-uncapped. Both are defensible. Both cannot be `main` at the same time. That's not a review comment, that's a "somebody picks a lane" conversation. cc @dbrowneup, this is the actual blocker here, above anything below it.

And the #823 ghost shows back up anyway: `chain/strategy_runner_publisher.py` (new, 656 lines) calls `SubscriptionManager.chargeActions(sub_id, action_count)` with zero caller check, the exact bug, verbatim. It's not wired into `main.py` or any compose file in this diff, so it's not live today, but it ships with its own `Dockerfile`, ready to run, and nothing tests what happens when a stranger calls it. Dead code that reintroduces a known vulnerability is still a liability, just a patient one. Either cut it since the monolith supersedes it, or strip the vulnerable call before it becomes somebody's copy-paste source six months from now.

Two UI bugs while I'm in here. `StrategyDetailPage.jsx` points Subscribe at a hardcoded `0x000…0` contract address that's never been deployed, and it's targeting #824's ABI, which your own backend description says is fully detached now. And `Strategies.jsx`'s "Stop publishing" button POSTs to `/stop-publish/{id}`, but the route you actually shipped is `DELETE /publish/{id}`. Both buttons are decorative right now.

Smaller stuff. `settlement.py` defaults `GATEWAY_CHAIN` to `baseSepolia` while `payments.py` and `.env.example` default it to `arcTestnet`. Leave the env var unset anywhere and charging and settlement quietly point at two different chains. And `test_marketplace_routes.py` has four tests asserting a `/payment-webhook` route and a `_WEBHOOK_SECRET` attribute I can't find anywhere in what actually shipped. Please just run the suite for real before merge, because either that's four tests failing that nobody's looked at, or the route got dropped without the tests noticing.

Last one, not a bug, a gap: the authorization logic you wrote is correct everywhere I checked, but nothing tests it under attack. No test tries withdrawing, unpublishing, or unsubscribing as the wrong wallet. You clearly built it to hold, so write the test that proves it holds.

Requesting changes, and not because the security story is bad. It mostly isn't. It's the #824 collision plus two dead buttons plus the webhook thing needing an honest test run. Land those and I'll re-review same day. Until then, this one goes back to the workroom, the library is open, the library is also not merging today.

cc @dbrowneup @mnemonik-dev on the contract design collision specifically, that one's above my pay grade to referee alone.

analytics-engine/strategies/daily_returns/antonacci_2014_dual_momentum.json
mnemonik-dev, line 1: Please add backtest outputs to gitignore. Should not be a part of the repo, right ?

mnemonik-dev (commented):
Can you
1  rebase your fork on current repo state
2 remove backtest files from repo and gitignore tham.
3 proces following high severity findings:
  H1 — Contract strands funds on recreate-after-deactivate (empirically proven).
  createPool guards only on not pools[pool_id].active (PaymentSplitter.vy:49). After deactivatePool sets active=False, the same pool_id can be recreated, and the
  assignment on line 53 resets held_balance=0 while the USDC physically remains in the contract. withdraw is bounded by held_balance, so those funds become permanently 
  unwithdrawable. My test:
  held_balance before=100000000  after_recreate=0  contract_usdc_still_locked=100000000
  withdraw now reverts: "amount exceeds held balance"
  The suite tests test_cannot_recreate_existing_pool (active case) and test_withdraw_after_deactivation, but not this path. Currently only latent — the backend never
  calls deactivatePool (stop-publish is DB-only) and pool_id is deterministic, so re-publish hits the pool_already_active guard. But deactivatePool exists, is
  owner-callable, and one operator wind-down + republish loses funds. Fix: gate createPool on "never existed" (assert pools[pool_id].creator == empty(address)) and/or
  refuse when held_balance > 0.

  H2 — GATEWAY_CHAIN default mismatch between the charge side and the settle side. payments.py:70 defaults GATEWAY_CHAIN to "arcTestnet"; settlement.py:38 defaults the
  same env var to "baseSepolia". If the env var is unset, subscribers are charged on Arc but Stage A withdraws Gateway balance to baseSepolia — while Stage B then reads
  the wallet's Arc USDC balance (RPC_URL, settlement.py:203) and deposits to the Arc splitter. The withdrawn funds never appear where Stage B looks. At minimum this is a
  divergent-default landmine across the money seam; the two files were clearly not reconciled.

  H3 — wallet_provisioner idempotency is claimed but does not hold. The docstring says retries won't create duplicate wallets, but idempotencyKey is a fresh uuid.uuid4() 
  per call (wallet_provisioner.py:100). Circle dedupes on the idempotency key, not on the ref metadata — so the 409 branch that looks up by ref will essentially never
  fire, and any retry (e.g. publish that fails at createPool then is re-attempted) provisions a new Circle wallet, diverging the funded/settlement address from the one
  persisted earlier. Fix: derive the idempotency key deterministically from sub_id/pool_id.

4 Is it possible to provide PaymentSplitter.sol ? We can not have couple of smart contract languages languages.

Thank you1

analytics-engine/strategies/daily_returns/antonacci_2014_dual_momentum.json
dbrowneup: Yeah, I think we got that stuff cleaned up already in the main repo now. I will double check this file path to make sure. These should indeed be stored in a database, not as part of the repo like committed code.

dbrowneup: Although I suppose some may be needed as fixtures. I'll look into it more to see if we've got it right or not.

dbrowneup:
Oof, Ricardo, I think your PR got messed up with the rebase stuff. The git graph looks crazy now too. Not sure what happened. This PR has a ton more conflicts now.

rcrdoh:
i will open another pull request

rcrdoh:
i have to say that when performing rebase there are less conflicts to resolve than before, i am currently implementing the issues pointed in the comments

rcrdoh:
i will open a branch directly on this repo from there i will open the PR

rcrdoh:
in that branch i will use rebase to put the branch to date with main

Issue #794, opened 2026-06-28
[oracle] Leg B verification: no Chainlink Data Feeds on Arc testnet — admin path correct; Stork/Pyth need an adapter


dbrowneup:
## Summary (Leg B verification — #724 follow-up)

Verification of whether the **Chainlink-first `PriceOracle` path (#724) can be wired to real Chainlink Data Feeds on Arc testnet**. Conclusion: **Chainlink has no Data Feeds deployed on Arc testnet (chain 5042002)** — so `setPriceFeed()` cannot point at a real Chainlink `AggregatorV3Interface` feed today, and the admin-fed degrade path is the only viable live source. The #724 "degrade-to-admin, don't brick" design is therefore exactly right and needs no change. If we want a *real decentralized* on-chain feed on Arc before Chainlink arrives, **Stork** (live on Arc testnet) or **Pyth** are the options — but both need a small `AggregatorV3` adapter.

## Evidence

- **Circle's own Arc oracle docs** ([`submodules/context-arc/.../arc/tools/oracles.md`](submodules/context-arc/docs/docs.arc.network/arc/tools/oracles.md)) list Chainlink with **generic** links only (global Data Feeds docs + the cross-network address directory) — **no Arc-specific feed page**. By contrast **Stork** is listed with an **Arc-specific** contract-address page, and Pyth with its EVM contract page.
- The **#724 contract review itself** noted *"Arc has no Chainlink uptime feed"* (sequencer-uptime scoped out).
- A **web search** for Arc-testnet Chainlink price-feed addresses surfaces only the generic Chainlink directory — no Arc entries.
- **Stork IS live on Arc testnet** — verified on-chain via `cast`:
  - `cast chain-id --rpc-url https://rpc.testnet.arc.network` → `5042002` ✓
  - `cast code 0xacC0a0cF13571d30B4b8637996F5D6D774d4fd62 --rpc-url https://rpc.testnet.arc.network` → non-empty (ERC-1967 proxy bytecode) ✓ deployed.

## Implication for the live path

- **No contract change needed.** With `priceFeed == address(0)`, `getPrice()` returns the admin-fed price (the backend oracle runner's `setPrice`, with its 24h staleness check) — the only viable live source on Arc. The Chainlink-first branch is correctly dormant until Chainlink ships on Arc. This is the #724 design working as intended.
- The "verify the live Arc-testnet Chainlink feed addresses" task on #724 resolves to **"there are none"** — recorded here so we don't keep re-asking.

[...]

Issue #775, opened 2026-06-28
[oracle] Secondary-source price cross-check: yfinance guardrail for Chainlink (off-chain first → on-chain V2)


dbrowneup:
## Summary
Add a **secondary-source price cross-check**: keep **Chainlink as the primary** on-chain price, and **cross-check it against an independent yfinance market price**, **failing closed when they diverge beyond a configured band**. Concretizes the T2.4 "secondary oracle / 2-of-3" item and answers the single-source-oracle weakness. yfinance is already wired (Explore page + all backtest history), so the second source is low marginal cost.

**Honest claim this earns (use this framing, not more):** *"Chainlink-primary, cross-checked against an independent yfinance market source, fail-closed on divergence beyond a band."* Do **NOT** claim "decentralized 2-of-3" — yfinance is centralized + off-chain.

## Sequencing / honesty note (load-bearing)
TODAY the live on-chain price is **admin-fed from yfinance** (`yfinance -> oracle runner -> PriceOracle.setPrice`), so the two are **not independent yet**. This cross-check becomes a real two-source guardrail only once **Chainlink is wired as the on-chain primary** (#724 / #759 / T1.3). Until then, don't ship the claim.

## Phases
**Phase 1 — Off-chain guardrail (MVP / V2 sprint; START HERE).** The oracle-runner / executor reads the Chainlink price (on-chain) and the yfinance price (off-chain) and, before acting on a rebalance/mint, refuses / alerts / quarantines if they diverge beyond the band. No contract change. Fast, cheap, real safety.

**Phase 2 — On-chain enforced (post-V2 nice-to-have).** `PriceOracle` itself compares the Chainlink reading vs the admin-fed (yfinance-derived) price and fails closed on divergence (reuse the existing deviation-guard machinery). Stronger claim. Owner (@dbrowneup) owns the contracts.

## CRITICAL design principle — asymmetric, flaky-data-safe (collaborate with @onder-akkaya)
yfinance is itself **flaky / delayed / split-adjusted / rate-limited** (see #772). The guardrail MUST be **asymmetric**: a stale/missing/flaky yfinance may **FLAG / quarantine / alert**, but must **NOT unilaterally BLOCK** a healthy Chainlink price — otherwise a yfinance outage halts good trading and the guardrail becomes the failure. This requires, as a deliberate part of the **pricing-system design**:
- a **divergence band** (tolerance) + a **staleness window** for yfinance,
[...]

Issue #772, opened 2026-06-28
[quant] Backtest data-quality / fetch-verification harness for the universe expansion


dbrowneup:
## Summary

The planned 200–300-instrument universe expansion (#759) pulls yfinance's long tail of tickers. Those names are flaky, rate-limited, and survivorship-biased: delisted names silently vanish, history has gaps, and partial/short series get padded or truncated by the fetch layer. Gappy or biased return history silently **inflates Sharpe** and **poisons our DSR/PBO rigor gate** — the library-PBO computed from `analytics-engine/strategies/daily_returns/*.json` (Bailey et al. 2014 CSCV) and the per-strategy DSR (Bailey & López de Prado 2014) both assume clean, gap-free, survivorship-aware return series. We must verify each instrument BEFORE admitting it to the universe, as a precondition — not as a Phase-3 afterthought that greens the parity test on un-fetchable tickers.

## Scope

- `backend/archimedes/services/strategy_signal_evaluator.py` — `GLOBAL_ASSETS` universe set and yfinance fetch path.
- New module: `backend/archimedes/services/data_quality.py` (per-instrument fetch/gap/survivorship checks).
- `backend/tests/test_universe_parity.py` — gate the universe on the harness; do not pass un-fetchable tickers.
- New: `backend/tests/test_data_quality.py`.

## Acceptance criteria

- [ ] `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_data_quality.py -q` → ends `N passed, 0 failed`.
- [ ] Harness exposes `verify_instrument(ticker, start, end)` returning a per-instrument verdict (fetchable / gap-count over the backtest window / survivorship flag) and `verify_universe(tickers, start, end)` aggregating them; calls are mocked at the yfinance boundary (no live network in tests).
- [ ] `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_universe_parity.py -q` → ends `N passed, 0 failed`, with every admitted ticker having passed `verify_instrument`.
- [ ] Instruments with insufficient (< window) or gappy (> threshold) history are flagged/excluded, proven by a test asserting a known-bad mock ticker is rejected.

## Verify

```bash
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_data_quality.py backend/tests/test_universe_parity.py -q
[...]

PR #898, opened 2026-07-04
[oracle] Thread yfinance bar timestamp through cross-check for staleness gate (#775)

onder-akkaya:
## Summary

Closes the one remaining item on #775: Phase 1 (magnitude cross-check, PR #835) landed 6/30, but noted it was magnitude-only — a stale-but-positive yfinance reading wasn't detected, the wide band just tolerated it. This threads the yfinance bar timestamp through and adds a staleness gate ahead of the divergence check.

## What changed

`backend/archimedes/chain/oracle_updater.py`:

- `_fetch_yfinance_single` now returns `(price, bar_ts)` instead of a bare price, with `bar_ts` normalized to tz-aware UTC. Also fixed its other caller (`fetch_market_snapshot`'s VIX fetch) for the new shape.
- `_cross_check_secondary` checks the secondary's bar age against a new `PRICE_CROSSCHECK_MAX_STALENESS_SECONDS` (default 345600s / 4 days — sized for a 3-day holiday weekend on equity-like intraday bars) before computing divergence at all. A stale secondary is treated the same as a missing one: fail-open, proceed on primary, no divergence verdict computed (a stale reading was never validly comparable either way).
- `.env.example` documents the new var next to `PRICE_CROSSCHECK_BAND_BPS`.

Two deliberate non-changes, both noted in code comments:

- `PRICE_CROSSCHECK_BAND_BPS` default (5000 bps) is untouched. Tuning it needs backtested false-positive-rate data I don't have on hand right now.
- The #772/#856 per-instrument data-quality harness (`data_quality.py`) is not wired into this path. It's a full 2-year batch fetch per call, built for offline universe admission (`generate_universe.py`, `strategy_signal_evaluator.py`), not a per-push oracle check. Threading the bar timestamp gets the practical staleness detection without that cost.

The asymmetric fail-open design from Phase 1 is unchanged: missing/errored/non-positive/stale secondary never blocks a healthy primary.

## Test plan

- `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain/ -q` → 156 passed, 0 failed
- `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests -m "not integration" -q` → 2399 passed, 0 failed
- `ruff format --check` + `ruff check --select E9,F63,F7,F40` on changed files → clean

New tests in `test_oracle_crosscheck.py`: fresh-bar regression, stale-in-band still proceeds (asserts the log distinguishes "stale" from a normal pass), stale-out-of-band still proceeds (proves staleness short-circuits before the divergence check, so a stale-and-wildly-different reading doesn't false-positive a fail-closed), boundary case at the staleness cap. `test_oracle_updater_fetch.py` updated for the new tuple return shape plus a naive-timestamp normalization test.

## Still open after this

Per my last comment on the issue: Phase 2 (on-chain enforcement in `PriceOracle.sol`) is a contract change, Dan's call with Bogdan reviewing. And the honest-claim gate still holds — `getPrice()` is admin-fed until the Stork adapter (#794) lands, so we shouldn't ship the two-source claim yet. This PR doesn't change either of those; it's the staleness item only.

@mnemonik-dev — flagging you as a reviewer since this is oracle/price-feed trust surface, even though it's backend-only (no contract changes).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -198,13 +198,18 @@ ADMIN_PRICES_JSON=
 # Secondary-source cross-check (#775): before pushing a NON-yfinance primary
 # price (Pyth/Stork) on-chain, the oracle runner compares it to an independent
 # yfinance reading and FAILS CLOSED if they diverge beyond this band (bps).
-# Asymmetric: a MISSING / errored / non-positive yfinance NEVER blocks a healthy
-# primary (it only logs). Magnitude-band only — it does NOT yet detect a stale
-# (but positive) yfinance value, so the generous default band is what tolerates
-# off-hours/weekend drift; a real secondary-freshness check is #775 Phase 2
-# (Önder's lane). No-op when PRICE_SOURCE=yfinance (same source) or for admin
-# pins. 0 to disable.
+# Asymmetric: a MISSING / errored / non-positive / STALE yfinance secondary
+# NEVER blocks a healthy primary (it only logs and proceeds). No-op when
+# PRICE_SOURCE=yfinance (same source) or for admin pins. 0 to disable.
 PRICE_CROSSCHECK_BAND_BPS=5000
+
+# Secondary-source staleness window (#775 Phase 2, implemented): max age
+# (seconds) of the yfinance secondary's own bar timestamp before it's treated
+# as unusable for the cross-check above (same fail-open behavior as a missing
+# secondary — a stale reading is never compared for divergence). Default 4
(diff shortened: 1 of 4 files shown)


onder-akkaya:
Merging directly — CI green, mergeable clean, no contract changes, .env.example diff is additive-only (one new fail-open knob). Dan and Bogdan are both offline for good reason right now; proceeding per team call rather than waiting.

Issue #921, opened 2026-07-05
LOW: EIP-6963 wallet listeners not attached — account switch undetected


onder-akkaya:
## Problem Statement

Wallet `accountsChanged`/`chainChanged` listeners are bound only to `window.ethereum` at module load and only for injected EOAs. If a user connects an EIP-6963 wallet (Rabby/Brave) whose provider is NOT `window.ethereum`, account switches in that wallet are not detected — the app keeps signing/depositing as the old address until manual reconnect.

## Scope

- `ui/src/config.js` lines 511–541

## Acceptance Criteria

- [ ] Listen for `eip6963:announceProvider` events and attach listeners to all announced providers
- [ ] Test: install Rabby, connect it to the app, switch accounts in Rabby, verify the app detects the change (e.g., `useAccount()` updates)

## Out-of-Scope

- Don't break injected-wallet support

## Owner

Daniel R. (frontend)

PR #980, opened 2026-07-05
[contracts] Enforce a 1h price-freshness window on SyntheticVault mint/burn

### LINKED ISSUE #910: HIGH: SyntheticVault mint/burn trusts 24h stale oracle price — stale-price arbitrage
## Problem Statement

`SyntheticVault` mint/burn trusts a price up to 24 hours stale. The `getPrice()` admin fallback uses `MAX_STALENESS = 24 hours` (line 64 of PriceOracle.sol); the more conservative 1h `feedStaleness` heartbeat only guards a Chainlink feed path, which testnet likely doesn't have wired. This creates a classic stale-price mint/burn arbitrage vulnerability: an attacker who sees the real market move away from an old admin price can mint synth cheaply, or (when stale-high) burn it for more USDC than collateral is worth.

## Scope

- `contracts/src/PriceOracle.sol` lines 64, 90
- `contracts/src/SyntheticVault.sol` lines 74, 127 (mint/burn price reads)

## Acceptance Criteria

- [ ] The admin-price staleness window is reduced to match `feedStaleness` (≤1 hour), or a fresh-feed-required guard is added to mint/burn
- [ ] Document in a comment that this 24h window is a testnet/hackathon tradeoff and MUST be tightened before mainnet
- [ ] Test: set a stale price 24h old, attempt mint/burn, verify it succeeds; set a price >1h old, verify it's rejected (if 1h is the target)

## Out-of-Scope

- Don't remove the oracle architecture itself
- The 24h setting is already documented as a hackathon choice — this is just moving it to the right place

## Owner

Dan (contracts)

onder-akkaya:
## Summary

`SyntheticVault` mint/burn trusted a price up to 24 hours stale. `PriceOracle.getPrice()`'s admin fallback uses `MAX_STALENESS = 24 hours`, and the conservative 1h `feedStaleness` heartbeat only guards the Chainlink feed path — which the testnet doesn't have wired. mint and burn move value at the oracle price, so a stale admin price is a classic arbitrage: mint synth cheap when the market has moved up, or burn it for more USDC than the collateral is worth when it's moved down.

## Scope

- `contracts/src/SyntheticVault.sol` — mint/burn (and their previews) + a new freshness window.
- `contracts/src/PriceOracle.sol` — documentation on the 24h window.
- `contracts/test/SyntheticVault.t.sol` — new tests.

## What changed

- Added `mintBurnMaxStaleness` (default `1 hours`) and a `_requireFreshOraclePrice()` guard at the top of `mint`, `burn`, `previewMint`, and `previewBurn`. If the oracle's admin price is older than the window, they revert `StaleOraclePrice`. NAV/health views (`totalCollateral`, `vaultCollateralization`) deliberately keep the looser oracle window so reads never brick.
- The window is owner-tunable via `setMintBurnMaxStaleness`, bounded to `(0, oracle.MAX_STALENESS()]` — an operational escape hatch if the price-push cadence forces it, with a safe 1h default.
- Kept `PriceOracle.MAX_STALENESS = 24 hours` (tightening it globally would brick NAV/deposits between the currently-daily admin pushes and affects the `Vault` too), and documented it as a testnet tradeoff that must be tightened — and moved to a Chainlink feed — before mainnet.
- The freshness check keys on the admin `lastUpdated` timestamp, which is the live price source on this testnet (no feed). The doc comment notes this must be revisited to use the effective feed freshness once a Chainlink feed is wired.

## Verify

```
cd contracts && forge test
```

New tests cover: default window is 1h; mint/burn succeed when fresh; both revert `StaleOraclePrice` when the price is 2h old (inside the oracle's 24h window but past the mint/burn window); burn works again after a fresh push; `previewMint` reverts when stale; and the setter's update path + bounds + owner-only guard. All 235 contract tests pass.

## Review + deploy note

This touches a live contract, so it needs Dan's approval as contract owner with Bogdan (`mnemonik-dev`) as the two-eyes reviewer, and a redeploy of `SyntheticVault` — please do not merge on backend sign-off alone. The cached ABI in `contracts/abis/` isn't regenerated here; that belongs with the redeploy. Nothing on the live backend/UI calls the new functions, so there's no off-chain change riding on this.

## Anti-goals

- The oracle architecture is unchanged.
- `MAX_STALENESS` is not lowered (that would brick NAV/deposits and the `Vault`).

Fixes #910

### DIFF
--- a/contracts/src/PriceOracle.sol
+++ b/contracts/src/PriceOracle.sol
@@ -60,7 +60,16 @@ contract PriceOracle is Ownable {
     /// @notice Timestamp of the last price update
     uint256 public lastUpdated;
 
-    /// @notice Maximum age before stale (24 hours — hackathon testnet, Circle daily limit)
+    /// @notice Maximum age before the admin-fed price is stale (24 hours —
+    ///         hackathon testnet, Circle daily push limit).
+    /// @dev    TESTNET/HACKATHON TRADEOFF — MUST be tightened before mainnet.
+    ///         24h is deliberately loose so NAV, deposits and withdrawals never
+    ///         brick between the (currently daily) admin price pushes. It is far
+    ///         too loose for value-moving operations: SyntheticVault.mint/burn
+    ///         enforce their own tighter `mintBurnMaxStaleness` (default 1h) so a
+    ///         stale admin price can't be arbitraged (#910). On mainnet, wire a
+    ///         Chainlink feed (guarded by the 1h `feedStaleness` heartbeat) and
+    ///         drop this window to match.
     uint256 public constant MAX_STALENESS = 24 hours;
 
     /// @notice Address allowed to push price updates (e.g. Circle wallet)
--- a/contracts/src/SyntheticVault.sol
+++ b/contracts/src/SyntheticVault.sol
@@ -33,6 +33,16 @@ contract SyntheticVault is Ownable, ReentrancyGuard {
     /// @notice Accumulated protocol fees in USDC (6 decimals)
     uint256 public protocolFees;
 
+    /// @notice Max age of the oracle price accepted by mint/burn (default 1h).
+    ///         The PriceOracle's admin fallback tolerates a price up to
+    ///         MAX_STALENESS (24h) so NAV/deposits never brick, but mint and burn
+    ///         MOVE value at that price — trusting a stale one is the classic
+    ///         stale-price arbitrage (mint cheap / burn rich against an old admin
+    ///         price, #910). mint/burn therefore enforce this tighter window.
+    ///         Owner-tunable within (0, oracle.MAX_STALENESS()] as an operational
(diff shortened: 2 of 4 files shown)


onder-akkaya:
Closing as a duplicate. This branch bundles two already-covered fixes:

- The SyntheticVault mint/burn staleness gate (#910) is handled by #971, whose implementation is feed-aware: it skips the admin-`lastUpdated` check when a Chainlink feed is configured (`getPrice()` self-guards the feed path with its own heartbeat), so it won't false-revert a fresh-feed asset whose admin reference lags. This branch's `_requireFreshOraclePrice()` always keys on `oracle.lastUpdated()`, which is a latent false-revert once a feed is wired. #971 also regenerates the cached ABI.
- The `ui/src/config.js` EIP-6963 listener change (#921) is byte-identical to #981, which is scoped to that one file.

Keeping #971 (#910) and #981 (#921) as the single-purpose PRs.

PR #842, opened 2026-07-01
[universe] T1.5b — cascade-aligned ~300-instrument universe + honest per-source oracle schema (#759) !minor

dbrowneup:
## Summary

Expands the synthetic universe from ~59 to **281 live synths** (accept range 250–320), cascade-aligned to what our real oracle sources cover, with **NO single-name equities** on the on-chain path. Built by a **reproducible generator** (`backend/archimedes/scripts/generate_universe.py`) that holds a curated asset list by class and resolves every `pyth_feed_id` **programmatically against the live Pyth Hermes catalog** — no id is ever invented.

## Honest per-source oracle schema (replaces the misleading `chainlink_covered:bool`)

Each live entry now carries:
- `yfinance_ticker` — **required** (backtest source + the parity invariant key).
- `pyth_feed_id` — a **REAL** Pyth Hermes id (0x-prefixed) or **null** (never fabricated).
- `stork_asset_id` — **null** for now (no public Stork catalog to resolve yet, #828 follow-up).
- `chainlink_feed` — **null EVERYWHERE**. Chainlink price **Data Feeds are NOT yet deployed on Arc testnet** (only **CCIP** as of 2026-06-30); wiring is one `addFeed` call away once Arc publishes feed addresses. Verified.
- `oracle_tier` — a **derived hint** (`quorum` if ≥2 of {pyth,stork,chainlink} set, `single_checked` if 1, `admin` if 0). The **authoritative tier is on-chain** via `QuorumPriceOracle.priceWithProvenance()` (#840); this field is only the expected tier from configured feeds.

`_chainlink_note` and `_comment` in the SSOT are updated to state all of the above honestly.

## Counts by asset class (281 total)

| class | n | class | n | class | n |
|---|---:|---|---:|---|---:|
| crypto | 71 | fx | 30 | us_thematic_etf | 21 |
| asia_equity_etf | 18 | us_equity_etf | 15 | factor_etf | 14 |
| eu_equity_etf | 13 | us_sector_etf | 11 | commodity_etf | 10 |
| intl_equity_etf | 8 | latam_equity_etf | 6 | metal_etf / metal_eq_etf / metal_spot | 5 / 5 / 5 |
| us_bond_agg | 5 | credit_ig | 5 | agri_etf / credit_hy / reit_etf | 4 / 4 / 4 |
[...]

PR #1013, opened 2026-07-06
[oracle] Prune retired synths from YFINANCE_MAP (#943)

### LINKED ISSUE #943: LOW: Stale YFINANCE_MAP entries for retired synths cause wasted fetches + misleading logs
## Problem Statement

Stale `YFINANCE_MAP` entries for retired synths (sGOLD/sOIL/…) → wasted fetches + misleading log lines.

## Scope

- `backend/archimedes/chain/oracle_updater.py` lines 25-34
- `backend/archimedes/chain/client.py` lines 20-24

## Acceptance Criteria

- [ ] Remove the retired symbols from `YFINANCE_MAP`, or mark as deprecated
- [ ] Test: run the oracle updater, verify no fetches for retired symbols

onder-akkaya:
Closes #943.

## The problem

`YFINANCE_MAP` in `oracle_updater.py` still carried synths that are no longer on the live on-chain universe: `sTSLA`/`sNVDA` (single stocks removed from the SSOT in #725) and `sGOLD`/`sOIL`/`sNKY` (retired in #842, `sGOLD`→`sGLD`/`sXAU`). The fetch loop keys off this map, so each retired entry meant a wasted yfinance download and a misleading `fetched N prices` log every cycle.

## The fix

Remove the five retired synths. What remains: `sSPY` (the one live synth with a yfinance ticker) plus the `^GSPC`/`^VIX` index tickers used for regime signals (not synths — the leading-`s` filter already excludes them from the synth fetch). `client.py`'s `_SYNTH_DEFAULTS` was already pruned to the live subset (`sSPY`/`sBTC`), so it needs no change — verified against the in-file #725/#842 notes and `universe.ON_CHAIN_SYNTHS`.

## Tests

- `test_retired_synths_absent_from_yfinance_map` — the five retired symbols are gone.
- `test_every_synth_key_is_in_the_live_universe` — **parity guard**: any `s`-prefixed key must be in `ON_CHAIN_SYNTHS`, so a stale entry can't creep back.
- `test_fetch_prices_does_not_fetch_retired_symbols` — negative control (no retired symbol/ticker reaches the fetch) + positive control (`sSPY` still fetched).

## Verify

```
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain/test_oracle_updater_fetch.py -q   # 22 passed
```

### DIFF
--- a/backend/archimedes/chain/oracle_updater.py
+++ b/backend/archimedes/chain/oracle_updater.py
@@ -21,16 +21,23 @@ from archimedes.models.asset import AssetPrice, MarketSnapshot
 
 logger = logging.getLogger(__name__)
 
-# Symbol → yfinance ticker mapping
+# Symbol → yfinance ticker mapping.
+#
+# Only synths that are actually on the live on-chain universe belong here — the
+# fetch loop (fetch_prices) and the secondary cross-check (_cross_check_secondary)
+# both key off this map, so a stale entry means a wasted download and a misleading
+# "fetched N prices" log every cycle. Retired single-stock synths (sTSLA/sNVDA,
+# dropped from the SSOT in #725 pending securities-compliance review) and the
+# futures/index synths retired in #842 (sGOLD → sGLD/sXAU, sOIL/sNKY dropped) are
+# no longer in universe.ON_CHAIN_SYNTHS, so they are removed here too. sSPY is the
+# only live synth with a yfinance ticker; the ^-prefixed entries are index tickers
+# (not synths) used for regime signals — the S&P 500 moving averages
+# (_fetch_sp500_moving_averages) and VIX (fetch_market_snapshot) — and are excluded
+# from the synth fetch loop by the leading-"s" filter, so they stay.
 YFINANCE_MAP = {
-    "sTSLA": "TSLA",
-    "sNVDA": "NVDA",
     "sSPY": "SPY",
-    "sGOLD": "GC=F",  # Gold futures
-    "sOIL": "CL=F",  # WTI crude oil futures
-    "sNKY": "^N225",  # Nikkei 225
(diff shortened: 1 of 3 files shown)


dbrowneup:
Rebased onto latest main — surfaced a real cross-PR interaction that needs your oracle-domain call, so I'm leaving this for you rather than guessing at a fix.

**Failure:** \`backend/tests/services/test_price_source.py::test_cascade_mode_prefers_pyth_then_yfinance\` → \`KeyError: 'sOIL'\` (also `assert "sGOLD" in called_symbols`).

**Cause:** that test uses the now-pruned synths (`sOIL`, `sGOLD`) as its "Pyth doesn't cover it → yfinance fills the gap" fixtures. After this PR prunes the retired synths, the post-prune `YFINANCE_MAP` contains only `sSPY` (plus `^GSPC`/`^VIX`, which are index regime signals, not synths). So there is **no non-Pyth synth left** to exercise the yfinance-fallback branch — the test's scenario no longer has a valid subject.

This isn't a synth-swap fix; it's a semantic rewrite of that test (e.g. assert `sSPY` falls back to yfinance when the Pyth read is stale/missing), which is your call. Your own `test_oracle_updater_fetch.py` changes are fine — it's only this separate `test_price_source.py` case that breaks. Branch is rebased + pushed and otherwise green; just needs that test updated.

onder-akkaya:
Restructured `test_cascade_mode_prefers_pyth_then_yfinance`: Pyth now covers only `sBTC`, and `sSPY` (the only synth left in `YFINANCE_MAP` post-prune) is the one that falls through to yfinance. Replaced the stale `sGOLD` call-symbol assertion with an exact match on `{"sSPY"}`. Pushed as afb1f61, 20/20 passing locally + ruff clean.

PR #1025, opened 2026-07-06
[contracts] Paginate AssetRegistry.getLeaderboard to bound gas (#927)

### LINKED ISSUE #927: MEDIUM: AssetRegistry.getLeaderboard O(n²) sort risks block gas limit
## Problem Statement

`AssetRegistry.getLeaderboard` does O(n²) insertion sort over unbounded vault list before applying `limit`. As vaults grow, the view exceeds block gas limit, breaking on-chain callers and eth_call leaderboard reads.

## Scope

- `contracts/src/AssetRegistry.sol` lines 142-192

## Acceptance Criteria

- [ ] Refactor to pagination: accept `offset` + `limit`, return only those N vaults in sorted order
- [ ] Test: call with `limit=50, offset=0`, then `offset=50`, verify both work and return disjoint sets

onder-akkaya:
## Summary

`AssetRegistry.getLeaderboard` ran a full O(n²) insertion sort over **every** matching vault before applying the limit. As the vault set grows, that view exceeds the block gas limit and bricks both on-chain callers and `eth_call` reads (issue #927).

This replaces it with a **partial selection sort**: it extracts only the top `offset + limit` ranked vaults (O(total · (offset+limit))) and returns the `[offset, …)` slice, so a bounded page stays bounded in gas regardless of total vault count.

## Changes

- `getLeaderboard` gains an `offset` parameter → `(tier, offset, limit)`.
- `limit == 0` preserves the old "return all" behavior (and its O(n²) cost — callers should page with a real limit for large sets).
- `offset >= total` returns an empty array instead of reverting.
- `IAssetRegistry` interface + the cached frontend ABI in `ui/src/config.js` updated to the 3-arg signature.
- New test `test_getLeaderboard_pagination_disjoint_pages` asserts pages are sorted, disjoint, and that an out-of-range offset returns empty.

## Verify

```
cd contracts && forge build && forge test --match-contract AssetRegistryTest
```
→ 17 passed; 0 failed.

## Blast radius

`getLeaderboard` is only **declared** (interface + frontend ABI), not invoked anywhere in `backend/` or `ui/src/` — verified with grep — so the signature change has no runtime call sites to update beyond the ABI.

## Anti-goals

- No change to sort semantics (still AUM-descending).
- No touching of unrelated registry methods.

## Review

Contract change — routing to **Dan** (contract/infra owner) with **Bogdan (`mnemonik-dev`)** as preferred two-eyes reviewer, per the contract-review convention.

Closes #927

### DIFF
--- a/contracts/src/AssetRegistry.sol
+++ b/contracts/src/AssetRegistry.sol
@@ -139,13 +139,25 @@ contract AssetRegistry is IAssetRegistry, Ownable {
         emit VaultMetricsUpdated(vault, _vaults[vault].aum, block.timestamp);
     }
 
-    function getLeaderboard(uint8 tier, uint256 limit)
+    /// @notice Top vaults by AUM for a tier, paginated (issue #927).
+    /// @param tier   1 or 2 (0 = all tiers).
+    /// @param offset Number of top-ranked vaults to skip (the page start).
+    /// @param limit  Max results to return; 0 = all from ``offset`` onward.
+    /// @dev    Was a full O(n²) insertion sort over EVERY matching vault before
+    ///         applying the limit — as the vault set grows that view exceeds the
+    ///         block gas limit and bricks on-chain callers + eth_call reads. Now a
+    ///         PARTIAL selection sort: it extracts only the top ``offset + limit``
+    ///         vaults (O(n · (offset+limit))) and returns the ``[offset, …)`` slice,
+    ///         so a bounded page stays bounded in gas regardless of total size.
+    ///         (limit == 0 keeps the old "return all" behavior and its O(n²) cost —
+    ///         callers should page with a real limit for large sets.)
+    function getLeaderboard(uint8 tier, uint256 offset, uint256 limit)
         external
         view
         override
         returns (address[] memory vaults)
     {
-        // Count matching vaults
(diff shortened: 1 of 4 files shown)


onder-akkaya:
The red **Ruff — format** gate here is **not** from this PR. This PR is contract-only (3 Solidity files + one ABI line in `config.js`) and contains zero Python.

`ruff format --check .` runs against the PR-merged-with-`main` tree, and `main` currently has one unformatted file — `backend/archimedes/api/auth_siwe.py` (from the direct-to-main auth commit `5135abd`). `main-format-guard.yml` cannot self-heal it because its `GITHUB_TOKEN` push is rejected by branch protection (`GH006 … 2 of 2 required status checks`), so every open PR fails this gate on a file none of them touched.

Fix: **#1027** (formatting-only, green + mergeable) restores a clean base. After it merges, re-run this check here and it goes green — every other check on this PR (`Contracts — forge build + test`, `Backend — unit tests`, complexity, eslint, CodeQL) is already green.

The informational **quality-gate** `ruff linter 60` and **supply-chain** `pip-audit 2` are pre-existing whole-repo debt / CVE backlog (identical to what `main` reports), not introduced here — both are non-blocking.

Issue #1028, opened 2026-07-06
[architecture] Identity & data architecture: wallet-anchored identity ledger, honest metrics by construction


dbrowneup:
> Compiled from a 7-agent read-only production audit (2026-07-05), the metrics/identity issue-history post-mortem, and the analysis of PR #958. Working artifacts (full findings, PR analysis, verification transcripts) live in `Agora/audit-2026-07-05-data-architecture/`. This is an epic; phase PRs link back here. **Not for the t2o2 agentic path — the design decisions are human-owned.**

## Why now

We cannot answer *"how many wallets have connected?"* — and that is **by construction, not by accident**. The one moment the backend holds a cryptographically verified wallet (`backend/archimedes/api/auth_siwe.py:338-344`, SIWE verify) writes only an anonymous browser-vid into a Redis HyperLogLog and discards the wallet into a stateless cookie. No DB row is written at authentication.

The result on prod (2026-07-05 baseline): **four mutually unjoinable populations** —
- ~90k request tallies (Redis `INCR`, no per-visitor key, dies on box restart)
- 85 / 26 / 17 browser-cookie HLL funnel stages (irreversible aggregates)
- 6 distinct wallets across product tables
- 2 `user_profiles` rows

— feeding an Insights page whose numbers **cannot agree with each other because they share no key**. `wallet_connected = 26` counts browser cookies, not wallets, and is unenumerable forever. `real_users = 2` is a 3× undercount: 4 wallets own strategies/passports but never saved a profile.

**PR #958 (subscription marketplace) merged 2026-07-06** and raised the stakes: system-provisioned Circle wallets now hold **user money**, marketplace lifecycle events are Redis-only, and the billing tables (`subscriber_tick_log`, `subscriber_liabilities`, `settlement_intents`) are keyed by `sub_id` with **no wallet column** — the same unjoinable pattern, now financial. The marketplace is currently fail-softed at startup (see #1001 / circlekit import hotfixes), so the schema must be made coherent while volume is still tiny.

## Design principles (post-mortem — why five previous fixes didn't fix this)

1. **Honesty by construction, not by caveat.** #830 scoped itself to labels and explicitly *froze* the substrate ("do not rewrite telemetry internals"). A caption cannot repair a missing source of truth. Every metric must be **derived from a durable record**.
[...]

PR #1038, opened 2026-07-07
Feature/marketplace examples publish

rcrdoh:
# PR — Publish "Examples" strategies from the Marketplace panel

## What this does
Lets a publisher pick a strategy from their **Generated** or **Examples** lists in the
Marketplace → Publish panel (mirroring the Library panel) and publish it with a managed vault
flow, instead of hand-typing a `strategy_id` and vault address. Examples are gated to admin
wallets for now; this is a showcase of the end-to-end publish path.

## Why (root cause)
Publishing an example was impossible end-to-end, for a reason that wasn't obvious from the UI:

- Examples are served from the file-backed provider (`analytics-engine/strategies/*.py`),
  keyed by a 32-char hash. The publish route requires a matching `strategy_store` row
  (`marketplace_routes.py`: `filter_by(id=strategy_id)` → 404 if missing), and **nothing in
  production seeds examples into that table** — only tests do. So every example publish 404'd
  before the ownership gate was ever reached.
- There was no funding gate at publish time.
- The UI had no signal for whether the connected wallet may publish a given strategy.

## Changes by area

**Backend**
- **Startup seed (`main.py` lifespan):** idempotently inserts one `is_example=True`
  `StrategyRecord` per provider example, keyed by the **provider id** (not
  `content_hash[:16]`), bypassing `upsert_strategy` so the id chain the publish route relies
  on stays intact. Fail-soft — never blocks startup.
- **Funding gate (`marketplace_routes.py`):** after the vault is known and before
  `createPool`, reject with **HTTP 402** if raw `USDC.balanceOf(vault) < MARKETPLACE_MIN_VAULT_FUNDS_RAW`.
  Same threshold on both the auto-created and caller-supplied vault paths. Reuses the existing
  `MarketService._usdc_balance_of` primitive.
- **`can_publish` serializer field (`schemas.py`, `strategies_routes.py`):** computed per
  caller via `wallet_can_publish` on both list endpoints (examples list now resolves the SIWE
  caller; the generated list already did). Anonymous callers always get `false`.

**Frontend**
- **`PublishPage.jsx` rewrite:** Generated / Examples tabs reusing the Library's
  `StrategyTable`/`extraActions`; the publish affordance shows only where `can_publish`. The
  id text box is gone — the id comes from the selected row.
- **Client-signed vault (D4):** create-new path reuses `CreateVaultModal`'s
  `createVault()`+`setAgent()` so the on-chain **vault creator is the publisher**, then runs
  `DepositFlow` with a minimum deposit equal to the threshold. Existing-vault path skips the
  deposit and only checks the balance. Both do a client-side pre-check; the backend 402 stays
  authoritative.
- **`config.js`:** small `usdcBalanceOfRaw()` reader + `MIN_VAULT_FUNDS_RAW` mirror of the
  backend const.

**Config**
- New env keys `MARKETPLACE_MIN_VAULT_FUNDS_RAW` (backend) and
  `VITE_MARKETPLACE_MIN_VAULT_FUNDS_RAW` (UI), documented. `PLATFORM_ADMIN_WALLETS` must
  include the operator wallet for example publishing.

## Deliberately NOT in this PR (scoped out)
- **Any-user example publishing.** Examples stay admin-gated; `wallet_can_publish` is
  unchanged. Opening this up needs a per-wallet publisher model (see below) and is a separate
  pass.
- **Concurrent publishers of the same example.** The running-publisher unique index is
  `(role, strategy_id)`, so the first publisher of an example wins exclusively. No index
  change, no migration here.
- **User-signed `createPool`.** It stays owner/server-signed. The signer only affects auth
  (`onlyOwner`) and gas — the pool's `creator`/`platform` are call arguments and disbursement
  routes to `pool.creator` regardless of who submits the tx, so the publisher is already the
  economic creator. `PaymentSplitter.sol` is untouched.

## Known limitation (accepted)
The funding gate measures **idle USDC**, not NAV. A reused vault already rebalanced into
synthetics can hold a large NAV but ~0 idle USDC and will fail the gate. Acceptable for the
showcase; switching the gate to `totalAssets()` later is a one-line change on both sides.

## Risk / blast radius
- Additive only: one optional response field, one startup seed, one publish precondition, one
  UI page rewrite. No DB migration, no contract change (`git diff --stat contracts/` and
  `backend/migrations/` are both empty).
- Startup seed is idempotent and fail-soft, so a seeding error degrades to "examples not
  publishable" rather than a failed boot.

## Test / verification
- Boot twice → examples seeded once, no duplicates.
- `GET /api/strategies/` (admin SIWE) → example rows `can_publish=true`; anonymous → `false`.
- Publish an example: create-vault (creator == publisher) → deposit ≥ threshold → 200.
- Under-funded vault → 402 on both vault paths.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -41,3 +41,9 @@ CHARGE_BATCH_SIZE=10
 SWEEP_WITHDRAW_THRESHOLD_USDC=10.0
 # Settlement sweep (P5): minimum wallet USDC (raw 6-dec) to trigger on-chain depositToPool.
 SWEEP_MIN_DEPOSIT_RAW=1000000
+
+# ─── Publishing ──────────────────────────────────────────────────
+# Minimum idle USDC (raw, 6-dec) a vault must hold before a strategy can be published. 1000000 = 1 USDC.
+MARKETPLACE_MIN_VAULT_FUNDS_RAW=1000000
+# Comma-separated wallet addresses allowed to publish strategies (incl. examples).
+PLATFORM_ADMIN_WALLETS=0x...
--- a/backend/archimedes/api/marketplace_routes.py
+++ b/backend/archimedes/api/marketplace_routes.py
@@ -119,7 +119,18 @@ async def publish_strategy(
             owner_wallet=wallet,
         )
 
-    # 3a. Auto-provision publisher settlement wallet (Circle DCW).
+    # 3a. Publish funding gate (D7) — vault MUST hold at least
+    #     MARKETPLACE_MIN_VAULT_FUNDS_RAW idle USDC (raw 6-dec).
+    #     Single threshold for both new and reused vaults.
+    min_funds_raw = int(os.getenv("MARKETPLACE_MIN_VAULT_FUNDS_RAW", "1000000"))
+    balance = await market._usdc_balance_of(vault_address)
+    if balance < min_funds_raw:
+        raise HTTPException(
+            status_code=402,
+            detail=f"Vault USDC balance {balance} below minimum {min_funds_raw}",
+        )
+
+    # 3b. Auto-provision publisher settlement wallet (Circle DCW).
     #     Must happen before on-chain createPool so a wallet failure never
     #     leaves an on-chain pool with no way to be funded.
     from archimedes.marketplace.wallet_provisioner import provision_publisher_wallet
--- a/backend/archimedes/api/schemas.py
+++ b/backend/archimedes/api/schemas.py
@@ -235,6 +235,12 @@ class StrategyResponse(BaseModel):
     return_source: str = "noise"  # "risk_premium" | "mispricing" | "productive_growth" | "noise"
     return_source_note: str = ""
 
+    # Whether the caller's wallet is permitted to publish this strategy.
+    # Always False for anonymous requests; True only when the caller is the
+    # generating wallet (generated strategies) or a platform admin (examples).
+    # Backend-authoritative — the frontend uses this to hide the publish affordance.
+    can_publish: bool = False
+
 
 class StrategyListResponse(BaseModel):
     strategies: list[StrategyResponse]
--- a/backend/archimedes/api/strategies_routes.py
+++ b/backend/archimedes/api/strategies_routes.py
@@ -437,6 +437,7 @@ async def list_strategies(
     served ``status: "validated"``. This is intentional — the stored status is the
     stable filter key; the served status reflects the live verdict.
     """
+    from archimedes.db import get_session
     from archimedes.models.strategy_generators import wallet_can_publish
 
     status_filter = StrategyStatus(status) if status else None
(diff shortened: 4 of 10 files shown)


dbrowneup (commented):
## Review + CI fixes (Claude, on Dan's behalf) — pushed 2 commits, **not** merged

I pushed one commit fixing the three red checks so we can get a clean signal, and reviewed the rest. Ricardo + Dan please eyeball the commit before we go further.

### CI was red for two root causes — both fixed (`962b01dc`)
1. **Ruff F821 + import-guard** — `strategies_routes.py` used `get_session()` in the strategy-list endpoint without importing it (the file imports it *locally* per-function elsewhere). Added the matching local import.
2. **Backend unit tests** (`TypeError: '<' not supported between AsyncMock and int`) — the new funding gate `await`s `market._usdc_balance_of(vault)` and compares to `MARKETPLACE_MIN_VAULT_FUNDS_RAW`, but the `app`/`market` fixture never mocked that method, so `test_publish_creates_publisher_row` / `test_publish_duplicate_returns_409` compared an `AsyncMock` to an int. **Funded the mock vault in the fixture** (so publish-path tests exercise publishing) and **added `test_publish_underfunded_vault_returns_402`** to cover the gate itself.

Local verification (archimedes env): `ruff check --select F821` clean repo-wide, `ruff format --check` + critical rules clean, `test_marketplace_routes.py` 11 passed, strategies-route tests 41 passed.

### Design review — LGTM
- **Startup example-seed** (`main.py`): idempotent (skips existing ids), fail-soft (never blocks startup), and `content_hash` is a domain-separated `example:`-prefixed SHA-256 so it can't collide with real generated hashes. Good root-cause fix — examples genuinely weren't in `strategy_store`, so every example publish 404'd before the ownership gate.
- **Funding gate (402)** — correctly awaited, single threshold on both the auto-created and caller-supplied vault paths.
- **`can_publish` serializer** — backend-authoritative, anon → False. Right call to hide the affordance client-side off a server truth.

### Notes / follow-ups (non-blocking)
- **Forward-coupling to #1028:** the seed writes `owner_wallet=None` for examples; #1028 (D7) backfills NULL owners to a `system` identity — the two will meet cleanly there, just flagging it.
- **To actually publish examples end-to-end**, an admin wallet must be in `PLATFORM_ADMIN_WALLETS` (Ricardo's ask). That var is now documented in #1037; recommend setting it in **SSM Parameter Store** (`/archimedes/prod/PLATFORM_ADMIN_WALLETS`) rather than the box `.env`, so it survives the Fargate migration (#1039).
- I did not deep-review the frontend (`PublishPage.jsx` rewrite) — it mirrors the Library table pattern; a frontend eye from Marten/Daniel would be good.

Holding here — no merge until Dan + Ricardo have looked. Nice work Ricardo, this closes a real end-to-end gap.

Issue #942, opened 2026-07-05
LOW: Vault/SyntheticVault hardcode 18-decimal synths — non-18-decimal token mis-prices NAV


onder-akkaya:
## Problem Statement

Vault/SyntheticVault hardcode 18-decimal synths; a non-18-decimal token mis-prices NAV/collateral.

## Scope

- `contracts/src/Vault.sol` lines 299-302, 557-559
- `contracts/src/SyntheticVault.sol` lines 37, 78, 128, 170

## Acceptance Criteria

- [ ] Read `token.decimals()` on-chain and use dynamically, **or**
- [ ] Require a registry of allowed tokens with their decimals and validate on register

Issue #915, opened 2026-07-05
HIGH: Vault.rebalance unconstrained vs. target weights — agent can churn trades


onder-akkaya:
## Problem Statement

`Vault.rebalance` trade selection is unconstrained vs. target weights. The only bound is the per-swap oracle floor (≤5%). A compromised agent/manager can churn USDC→synth→USDC on a thin pool it also sandwiches, bleeding ~5%/swap with no cap on swap count or tying to `targetWeightBps`.

## Scope

- `contracts/src/Vault.sol` lines 334–395 (`rebalance` function)

## Acceptance Criteria

- [ ] Verify that `tokens_in`/`tokens_out` can only deviate from zero if the current weight is >X% away from the target weight (define X, e.g., 1%)
- [ ] Cap the total swap count per rebalance, or require that each swap moves a token toward its target
- [ ] Test: set `targetWeightBps = [50, 50]` (50/50 USDC/synth), then call `rebalance` with a set of trades that churns 100% → 50% → 100% synth, verify it reverts or is rejected

## Out-of-Scope

- Don't remove legitimate rebalancing

## Owner

Dan (contracts)

PR #1129, opened 2026-07-14
[contracts] Cap management/performance fee bps to close a fee-drain exploit — NEEDS DAN + BOGDAN REVIEW, DO NOT MERGE YET

### LINKED ISSUE #1136: HIGH: [contracts] Unbounded fee bps enables fee-drain exploit via self-donation
## Summary
`createVault()` is fully permissionless (Tier 2) and `managementFeeBps`/`performanceFeeBps` were unbounded `uint16` fields (0-65535, i.e. up to 655.35%) baked in immutably at construction with no setter.

## Impact
Above `performanceFeeBps > 10000` (100%), a vault creator can donate USDC directly to their own vault (bypassing `deposit()`, so no shares are minted for the donation) and trigger `_accrueFees()` via any state-changing call to mint themselves fee-shares worth **more** than their own donation — extracting value from other depositors' principal, not just their own.

## Found
Internal bug-hunt pass, 2026-07-14.

## Fix
PR #1129 — adds `MAX_MANAGEMENT_FEE_BPS` (500 = 5%) and `MAX_PERFORMANCE_FEE_BPS` (5000 = 50%) constants, enforced in the `Vault` constructor (not just the factory).

**Needs Dan's review as contract owner + Bogdan as preferred second reviewer before merge — do not merge without that.**

onder-akkaya:
Closes #1136

## ⚠️ Contract change — do not merge without Dan's review (contract owner) and Bogdan (preferred second reviewer)

Per CLAUDE.md, any smart-contract change needs Dan's review before merge, with Bogdan as preferred second reviewer, since this touches live-funds logic. Opening this now so it's ready to review — not merging it myself.

## Summary
`createVault()` is fully permissionless (Tier 2) and `managementFeeBps`/`performanceFeeBps` were unbounded `uint16` fields (0-65535, i.e. up to 655.35%) baked in immutably at construction with no setter. Above `performanceFeeBps > 10000` (100%), a vault creator can donate USDC directly to their own vault (bypassing `deposit()`, so no shares are minted for the donation) and trigger `_accrueFees()` via any state-changing call to mint themselves fee-shares worth **more** than their own donation — extracting value from other depositors' principal, not just their own.

## Fix
Adds `MAX_MANAGEMENT_FEE_BPS` (500 = 5%) and `MAX_PERFORMANCE_FEE_BPS` (5000 = 50%) constants, enforced in the `Vault` constructor (not just the factory, so the guard holds regardless of which caller constructs a `Vault`). 50% is a generous ceiling relative to the ~10-20% industry norm; it closes the theft vector (which specifically requires the fee to exceed 100%) with a wide safety margin. **These specific cap values are a starting recommendation, not a locked decision — please confirm or adjust them.**

Found in an internal bug-hunt pass, 2026-07-14.

## What merging this does NOT fix (deployment reality)

Being explicit so the badge matches the live path:

- **The live `VaultFactory` on Arc testnet predates this cap.** It keeps minting uncapped vaults until Dan redeploys — merging changes the source tree, not the deployed bytecode. This adds to the existing redeploy backlog (#588-class); worth batching into the same redeploy window.
- **Already-deployed vaults can never be fixed.** Fee bps are immutable with no setter, so a vault created with hostile fees before the redeploy stays exploitable forever. The only protection for those is off-chain: refuse to list or interact with any vault whose on-chain fees exceed these caps. Filed as #1138 — that guard protects live users today, independent of the redeploy timeline.

## Test plan
- [x] `forge test` → 287 passed (78 existing Vault tests unaffected)
- [x] Added `test_revert_createVault_managementFeeTooHigh`, `test_revert_createVault_performanceFeeTooHigh`, `test_createVault_feesAtCap_succeed`
- [x] Added `test_capped_performance_fee_donation_cannot_drain_other_depositors` — end-to-end regression reproducing the exploit walkthrough (honest depositor + attacker donation + fee-harvest trigger), asserting the attacker's extractable value never exceeds their own donation and the honest depositor's redeemable value never drops

### DIFF
--- a/contracts/src/Vault.sol
+++ b/contracts/src/Vault.sol
@@ -68,6 +68,24 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
     ///         before a correction is required, defeating the churn guard (issue #915).
     uint256 public constant MAX_REBALANCE_BAND_BPS = 2000;
 
+    /// @notice Hard cap on the creator-configurable annual management fee (5%).
+    ///         `createVault()` is fully permissionless (Tier 2) and these fee bps were
+    ///         previously unbounded uint16 fields (0-65535 = up to 655.35%) baked in
+    ///         immutably at construction with no setter (2026-07-14 internal audit,
+    ///         finding C1). Capped here (not just in the factory) so the guard holds
+    ///         regardless of which caller constructs a Vault.
+    uint16 public constant MAX_MANAGEMENT_FEE_BPS = 500;
+
+    /// @notice Hard cap on the performance fee charged above the high-water mark (50%).
+    ///         Same finding as MAX_MANAGEMENT_FEE_BPS: at performanceFeeBps > 10000 (100%)
+    ///         `_accrueFees()` can mint fee-shares worth MORE than a NAV-raising donation,
+    ///         letting a vault's own creator extract other depositors' principal via a
+    ///         bare `USDC.transfer()` to the vault followed by any state-changing call.
+    ///         50% is already a generous ceiling versus the ~10-20% norm; it eliminates
+    ///         the theft vector (which specifically requires the fee to exceed 100%)
+    ///         with a wide safety margin. Team should tune down if a lower cap is wanted.
+    uint16 public constant MAX_PERFORMANCE_FEE_BPS = 5000;
+
     // ─── Immutables ──────────────────────────────────────────────────
 
     address public immutable override asset;
@@ -162,6 +180,10 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
     /// @notice A rebalance swap would overshoot `token` past its target weight (issue #915).
     error RebalanceOvershootsTarget(address token);
     error InvalidRebalanceBand();
+    /// @notice `_managementFeeBps` exceeds MAX_MANAGEMENT_FEE_BPS.
+    error ManagementFeeTooHigh();
+    /// @notice `_performanceFeeBps` exceeds MAX_PERFORMANCE_FEE_BPS.
+    error PerformanceFeeTooHigh();
 
     // ─── Events (Vault-local) ────────────────────────────────────────
 
@@ -198,6 +220,8 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
         // Strict commit-before-trade: the trace registry is required and immutable, so
         // a vault can never exist in a state where rebalance() skips the commit check (#589).
         if (_traceRegistry == address(0)) revert TraceRegistryNotSet();
+        if (_managementFeeBps > MAX_MANAGEMENT_FEE_BPS) revert ManagementFeeTooHigh();
+        if (_performanceFeeBps > MAX_PERFORMANCE_FEE_BPS) revert PerformanceFeeTooHigh();
         asset = _usdc;
         ammRouter = IAMMRouter(_ammRouter);
         traceRegistry = IReasoningTraceRegistry(_traceRegistry);
--- a/contracts/src/Vault.sol
+++ b/contracts/src/Vault.sol
@@ -68,10 +68,22 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
     ///         before a correction is required, defeating the churn guard (issue #915).
     uint256 public constant MAX_REBALANCE_BAND_BPS = 2000;
 
-    /// @notice Maximum allowed management fee (500 bps = 5%) to prevent fee-drain exploits (#1136).
+    /// @notice Hard cap on the creator-configurable annual management fee (5%).
+    ///         `createVault()` is fully permissionless (Tier 2) and these fee bps were
+    ///         previously unbounded uint16 fields (0-65535 = up to 655.35%) baked in
+    ///         immutably at construction with no setter (2026-07-14 internal audit,
+    ///         finding C1). Capped here (not just in the factory) so the guard holds
+    ///         regardless of which caller constructs a Vault.
     uint16 public constant MAX_MANAGEMENT_FEE_BPS = 500;
 
-    /// @notice Maximum allowed performance fee (5000 bps = 50%) to prevent fee-drain exploits (#1136).
+    /// @notice Hard cap on the performance fee charged above the high-water mark (50%).
+    ///         Same finding as MAX_MANAGEMENT_FEE_BPS: at performanceFeeBps > 10000 (100%)
+    ///         `_accrueFees()` can mint fee-shares worth MORE than a NAV-raising donation,
+    ///         letting a vault's own creator extract other depositors' principal via a
+    ///         bare `USDC.transfer()` to the vault followed by any state-changing call.
+    ///         50% is already a generous ceiling versus the ~10-20% norm; it eliminates
+    ///         the theft vector (which specifically requires the fee to exceed 100%)
+    ///         with a wide safety margin. Team should tune down if a lower cap is wanted.
     uint16 public constant MAX_PERFORMANCE_FEE_BPS = 5000;
 
     // ─── Immutables ──────────────────────────────────────────────────
@@ -154,6 +166,7 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
     error ZeroAssets();
     error Unauthorized();
     error InvalidAllocations();
+    error InsufficientBalance();
     error InsufficientLiquidity();
     error SlippageBpsTooHigh();
     error OracleNotSet();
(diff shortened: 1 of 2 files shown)


onder-akkaya:
Investigated the 2 pip-audit findings: both are `vyper 0.4.3` (PYSEC-2023-142, PYSEC-2025-33), pulled in transitively via `circle-titanoboa-sdk` → `titanoboa` (the x402/Circle Gateway marketplace SDK, unrelated to this PR). Confirmed `0.4.3` is already the latest version on PyPI — there's no patched version to pin to yet, so this isn't a simple version bump. This PR only touches `contracts/src/Vault.sol` and `contracts/test/Vault.t.sol` (Solidity/Foundry), no Python dependencies, so it's a pre-existing repo-wide finding unrelated to this diff — not fixing it here to keep this PR scoped to the fee-cap change. Flagging it separately for the team to track (no upstream fix available yet, so the options are wait for a vyper patch or accept/document the risk).

onder-akkaya:
This PR only touches \`contracts/src/Vault.sol\` and \`contracts/test/Vault.t.sol\` (Solidity/Foundry) — no Python files at all, confirmed via \`gh pr diff 1129 --name-only\`. The blocking Ruff gate (\`ruff format --check\` + critical rules E9/F63/F7/F40) passed. The 10 errors are from the informational, \`continue-on-error\` broader \`ruff check .\` (per CLAUDE.md, this scans the whole repo, not the diff) — pre-existing repo-wide lint debt unrelated to this change. Not fixing repo-wide lint inside a live-funds contract PR to keep this scoped to the fee-cap change; happy to file/fix those separately if useful.

onder-akkaya:
@dbrowneup @mnemonik-dev this has been sitting for two days — can one of you take a look? It's a small diff (two constants + constructor checks + 4 tests) but it's a contract change, so per our rules it can't move without Dan's approval and ideally Bogdan's second pair of eyes.

Two things I'd like your input on specifically:

1. The cap values (500 / 5000 bps) are my starting recommendation, not a locked decision. If you want different numbers, now is the time — the off-chain guard I just opened (#1139) mirrors these exact values in Python, so a change here should land in both places in the same window.
2. Whether you want to batch the factory redeploy with the #588-class backlog or do it separately. Merging this alone changes nothing on the live path until the redeploy happens.

Copilot's automated pass had no findings beyond the overview. Forge suite is green (287 passed).

onder-akkaya:
Bumping this — it's been open 9 days waiting on review and it's the fee-drain exploit fix (title says NEEDS DAN + BOGDAN, still true).

Quick recap of where it stands:
- Fix is small and contained: `MAX_MANAGEMENT_FEE_BPS` (500 = 5%) / `MAX_PERFORMANCE_FEE_BPS` (5000 = 50%) enforced in the `Vault` constructor. Closes the >100% performance-fee donate-and-accrue exploit described above.
- CI is green (contracts build+test, backend unit tests, ruff).
- Cap values are a starting recommendation, not locked — flagged in the description, still needs your call.
- Doesn't touch the live VaultFactory — that's a separate redeploy question, noted in the PR body so the badge doesn't overclaim.

Whenever you get a window, even just a thumbs-up on the cap values would let this move. No rush on the redeploy side, just want the source fix reviewed so it doesn't sit indefinitely.

dbrowneup:
## Two things worth separating before this merges

**1. `64fdea43` contains a much bigger fix than this PR's title.**

That commit changes NAV valuation from a hardcoded `1e18` to `_tokenUnit(token)`:
```solidity
- uint256 sellValue = (amount * PriceOracle(tokenOracle[tokenOut]).getPrice()) / 1e18;
+ uint256 sellValue = (amount * PriceOracle(tokenOracle[tokenOut]).getPrice()) / _tokenUnit(tokenOut);
- return (bal * PriceOracle(oracle).getPrice()) / 1e18;
+ return (bal * PriceOracle(oracle).getPrice()) / _tokenUnit(token);
```
Our synths are **6-decimal**, so the old divisor was wrong by **10¹²** on every synth-denominated valuation — this is in the `totalAssets`/NAV path, i.e. every deposit, redemption and rebalance decision, for **every vault**, not just over-cap ones.

That reads as a genuine and important correctness fix, and it is materially larger in blast radius than the fee cap this PR is named for. **It probably deserves its own PR and should likely land faster than the cap does** — reviewers evaluating "unbounded fee bps" will not be looking for a decimals fix, and it deserves focused Bogdan review on its own merits.

**2. The exploit regression test named in the PR body appears to have been removed.**
```
$ git show --stat 64fdea43
 contracts/src/Vault.sol    | 36 ++++++------------
 contracts/test/Vault.t.sol | 91 ++++++----------------------------------------
```
`Vault.t.sol` is net **−91 lines**. If the self-donation exploit test went with it, the PR's own acceptance evidence no longer exists in the tree. Worth confirming before merge — a fee-cap PR without its exploit test is hard to verify later.

Neither point is a judgement on the fix itself; both are about what a reviewer can verify. Flagging pre-merge rather than after.

dbrowneup:
Reviewed this in preparation for contract review. The core fix is sound — I traced every write path for the fee bps fields and the constructor is the only one (both fields are `immutable`, no setter, no initializer, no proxy/clone pattern, and `VaultFactory.createVault()` is the sole caller of `new Vault(...)`). I also re-derived the exploit arithmetic from #1136: extractable value exceeds the donation only when `performanceFeeBps > 10000`, independent of pool size, so capping at 5000 closes the described vector with real margin, and the comparison is against the correct `BPS = 10000` denominator.

Three things to resolve before this goes to Bogdan.

**1. The exploit regression test referenced in the PR description isn't in the branch.**

`test_capped_performance_fee_donation_cannot_drain_other_depositors` and `test_createVault_feesAtCap_succeed` were added in `8d5adf5e` and removed in `64fdea43`. Function counts: `main` 78 → `8d5adf5e` 82 → HEAD 80, and CI reports `VaultTest: 80 passed` rather than the 82 the description implies.

What survives are the two constructor-revert tests (`Vault.t.sol:1042,1048`), which assert `createVault()` reverts above the cap but never call `deposit()`, donate, warp time, or trigger `_accrueFees()`. So nothing in the suite currently exercises the donation → fee-accrual → share-mint path the fix exists to close. Worth restoring an equivalent regression before merge, and updating the description either way.

**2. `64fdea43` also changes NAV valuation for every vault, and I think that change is important enough to stand on its own.**

The same commit changes the oracle divisor from a hardcoded `1e18` to `_tokenUnit(token)` in two places — `Vault.sol:458` (the #915 churn-guard `sellValue`) and `Vault.sol:679` (`_tokenValueUsdc`). Since the synths are 6-decimal, this looks like a genuine and fairly significant bug fix: the previous form would understate token value by 12 orders of magnitude wherever it applied, affecting both NAV computation and the rebalance churn guard on every vault.

If that reading is right, it deserves its own PR with its own description and tests rather than riding inside one titled "cap fee bps" — both so it gets reviewed as the valuation change it is, and so it can land on its own timeline. It's arguably more urgent than the cap.

The same commit also rewrites the performance-fee share-mint formula (`Vault.sol:873`), introducing a division by `_totalAssets`. That's live-funds economics and it currently ships with no new test.

**3. Minor, but worth keeping:** the multi-line NatSpec explaining the donation-drain mechanics and the 2026-07-14 audit finding C1 was replaced with one-line comments, and `ManagementFeeTooHigh` / `PerformanceFeeTooHigh` were merged into a single `FeeBpsTooHigh`. The rationale comments are exactly the context a contract reviewer reads the diff against, and separate errors make a revert tell you which bound was exceeded.

Happy to help restore the regression test or split out the valuation fix if that's useful.

onder-akkaya:
All three addressed:

**1. Exploit regression test restored.** `test_capped_performance_fee_donation_cannot_drain_other_depositors` and `test_createVault_feesAtCap_succeed` are back (604f0129) — they'd gotten swept up in `64fdea43`'s diff along with the tests that commit added. 82/82 `Vault.t.sol` tests pass on this branch now.

**2. NAV/decimals fix + share-mint formula split out.** → #1200. Also caught that the share-mint formula rewrite riding in the same commit had its own bug (crushes value by `10**12` for 6-decimal synths, since it missed two `_tokenUnit` call sites #942 didn't cover) and wrote a dedicated regression for each, both verified to fail on the pre-fix source before I trusted them.

**3. Two distinct errors + NatSpec restored.** `ManagementFeeTooHigh`/`PerformanceFeeTooHigh` are back instead of the merged `FeeBpsTooHigh`, and the audit-context comments (C1 finding, donation-drain mechanics) are restored.

This PR is back to fee-cap-only — `604f0129` reverts `Vault.sol`/`Vault.t.sol` to exactly their post-`8d5adf5e` state, nothing else touched. Ready for Bogdan whenever you want to loop him in.

onder-akkaya:
@dbrowneup @mnemonik-dev bumping again — this has been open since 2026-07-14, and it's back to fee-cap-only per the comment above since 2026-08-02 (exploit regression test restored, NAV/share-mint fix split out to #1200). CI's green. This is the actual fee-drain fix, would like to get it merged.

Issue #1120, opened 2026-07-14
[security][cleanup] Extract the #850 strategy-visibility rule into one shared predicate (3 copies today)


dbrowneup:
## Summary

The #850 ownership/visibility rule ("a non-example, unpublished `strategy_store` row is visible only to its `owner_wallet`") now exists as **three near-identical copies**, by the newest copy's own admission (its docstring: "copied from `selection_bias_routes._generated_strategy_rigor`, itself copied from `strategies_routes.get_strategy`"):

1. `strategies_routes.get_strategy`
2. `selection_bias_routes._generated_strategy_rigor`
3. `vaults_routes._strategy_record_visible` (added by #1082 to close the #1073 deploy fast-path hole)

This is a security-critical predicate: a future change to the rule (e.g. a team-sharing tier, delegated wallets) must be manually propagated to all three sites, and a missed one is a silent authz divergence — exactly the class of drift #850 was about.

## Scope

Extract one shared helper, e.g. `archimedes/services/strategy_visibility.py::is_strategy_record_visible(row, caller_wallet) -> bool` (pure predicate over an already-fetched row + wallet — no DB/session coupling, trivially testable), and have all three call sites use it. No behavior change; the three sites keep their differing fetch/error-handling shells (incl. #1082's fail-closed 503).

## Acceptance criteria

- [ ] `grep -rn "is_example or .*is_published" backend/archimedes/` shows the predicate in exactly ONE non-test location.
- [ ] All three routes' existing tests still pass unchanged (`test_vault_rigor_gate.py`, `test_selection_bias_generated_gate.py`, strategies-routes visibility tests) — behavior identical.
- [ ] The shared helper has its own unit tests (owner match case-insensitivity, example/published shortcuts, anonymous caller).

## Anti-goals

- Do NOT change the visibility semantics themselves.
- Do NOT centralize the DB fetch/error handling — only the predicate; the fail-closed shells differ per call site on purpose.

Provenance: fast-follow from PR #1082 review (merge-board 2026-07-14).

Issue #1102, opened 2026-07-11
bootstrap_vaults.py: synth_vault_addresses hardcodes 7 legacy addresses with no SSOT, silently stale after redeploy


onder-akkaya:
## Summary

\`mint_synthetic_tokens()\` in [\`backend/archimedes/scripts/bootstrap_vaults.py\`](backend/archimedes/scripts/bootstrap_vaults.py) (lines 168-176) hardcodes 7 \`SyntheticVault\` contract addresses as a literal dict:

\`\`\`python
synth_vault_addresses = {
    "sTSLA": "0xf0356600e26c6c403ec4f5b36b0e3380bb0609ab",
    "sNVDA": "0x4c3cdc2bf44195ad8a4d201c8afbd453949a8781",
    "sSPY": "0xd8d7855f76c384638cf1dfc3575ecff3538764b4",
    "sBTC": "0x92990ed6f5c8cd72752ca9aeafad422269225c43",
    "sGOLD": "0x124b5c5da57d209b28d4997aaf6d4e96711efd5a",
    "sOIL": "0xfa942399e36959c8060c3a82a610d680a7ac6d22",
    "sNKY": "0xb26029ca37c09400ca921f00fc541cd42143b508",
}
\`\`\`

Unlike \`synth_addresses\` / \`oracle_addresses\` a few lines below (which come from \`chain_client.settings\`, the actual SSOT), this dict appears nowhere else in the repo (confirmed via grep) and has no config/env backing.

\`gen_synthetic_addresses.py\`'s own docstring states synth vault addresses "change on every contract redeploy." Per CLAUDE.md, the system has since moved to a 281-synth universe (T3.2) — this script only covers 7 legacy symbols. Running \`bootstrap_vaults.py\` against the current deployment likely reverts on stale addresses or, worse, silently writes to orphaned contracts from a prior deploy.

## Why filing instead of guessing a fix

Contract addresses are chain-integration SSOT — guessing replacement values here risks pointing at the wrong (or non-existent) contracts. This needs Dan's (chain/infra owner) input on where the current 281-synth universe's per-symbol \`SyntheticVault\` addresses actually live (contracts/abis/? a settings file? chain_client.settings needs a new field?) before this dict can be safely derived instead of hardcoded.

## Scope

[...]

Issue #655, opened 2026-06-15
[backend] bootstrap_vaults.py create_vaults() has #651's stale-vault-address bug, zero coverage


onder-akkaya:
## Summary

`create_vaults()` in `backend/archimedes/scripts/bootstrap_vaults.py` (~lines 225-279)
has the same bug pattern as #651 (fixed via #653 in `executor.py`'s `create_vault`):
it submits `createVault` via Circle's `execute_contract`, waits for the w3 receipt,
but never checks `receipt.get("status")` before parsing for the `VaultCreated`
event — and falls back to `all_vaults[-1]` (the most recently created vault,
regardless of whether *this* call created it) on both a revert AND a successful-tx-
but-no-event case, indistinguishably.

```python
# Wait for receipt and find new vault address
receipt = await chain_client.w3.eth.wait_for_transaction_receipt(
    chain_client.w3.to_bytes(hexstr=tx_hash.removeprefix("0x"))
)

vault_address = None
factory = loader.vault_factory
for log in receipt.logs:
    try:
        result = factory.events.VaultCreated().process_log(log)
        vault_address = result["args"]["vault"]
        break
    except Exception:
        continue

if not vault_address:
    # Fallback: get last vault from factory
    all_vaults = await factory.functions.getVaults().call()
    vault_address = all_vaults[-1]

print(f"  ✅ {profile['name']} ({profile['symbol']}): {vault_address}")
```

#651/#653 added `VaultCreationRevertedError` to `archimedes.chain.executor` and the
matching `if receipt.get("status") == 0: raise ...` check, plus a warning on the
no-event fallback. `bootstrap_vaults.py`'s `create_vaults()` was out of scope for
#651 (which scoped to `executor.py` only) but has the identical bug — discovered
while writing #652's `CALL_SITES` list, which already includes this file for the ABI-
signature check.

## Impact

`bootstrap_vaults.py` is the demo-data seeding script — a `createVault` revert here
(insufficient funds, paused factory, bad params) would silently report `✅
{profile["name"]}` with **another vault's address**, then `fund_and_allocate_vaults`
would proceed to transfer synth tokens / deposit USDC / set allocations onto that
*wrong, pre-existing* vault. Lower blast radius than #651 (this is a one-shot bootstrap
script, not the live agent/API path) but the failure mode — silently misattributing
funds to the wrong vault — is the same.

Currently zero test coverage: `grep -rn "bootstrap_vaults\|create_vaults\b"
[...]

Issue #652, opened 2026-06-14
[backend] Add regression test: executor.py createVault ABI signature must match contracts/abis/VaultFactory.json


onder-akkaya:
## Summary

15jun-audit.md Finding 1 (fixed via #650) happened because
`backend/archimedes/chain/executor.py`'s `abi_function="createVault(...)"` string and
`contracts/abis/VaultFactory.json`'s `createVault` entry were bumped together in #646
without the live-deployed `VaultFactory` bytecode being redeployed — and nothing in
the test suite asserts these two checked-in artifacts agree with each other.

Add a hermetic test that derives the canonical `createVault` signature from
`contracts/abis/VaultFactory.json` and asserts every
`abi_function="createVault(...)"` call site matches it exactly. This will not catch
*live-bytecode* drift (that's #588's on-chain "smoke-test on testnet" scope) — but it
catches the more common mistake of editing the ABI or a backend call-site without the
other, and gives one place that documents "these must move together."

## Scope

- New test file `backend/tests/chain/test_create_vault_abi_consistency.py`.
- Load `contracts/abis/VaultFactory.json`, find the `createVault` entry, build
  `createVault(<type1>,<type2>,...)` from `inputs[].type` (in order).
- Regex/grep `backend/archimedes/chain/executor.py` and
  `backend/archimedes/scripts/bootstrap_vaults.py` for every
  `abi_function="createVault(...)"` occurrence; assert each string equals the
  ABI-derived signature.
- Should currently PASS — post-#650 both source files and the ABI are back on the
  5-arg `createVault(string,string,uint16,uint16,bool)` signature. This is a
  regression guard, not a fix for an existing failure.

## Acceptance criteria

- [ ] `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/chain/test_create_vault_abi_consistency.py -q`
  → `N passed, 0 failed`.
- [ ] Manually verified (locally, not committed) that the test FAILS if
  `contracts/abis/VaultFactory.json`'s `createVault.inputs` and `executor.py`'s
  `abi_function=` string are made to diverge — confirms the test actually checks
  what it claims to.
- [ ] `ruff format --check . && ruff check --select E9,F63,F7,F40 .` → clean.

## Anti-goals

- Don't attempt live `eth_getCode` / RPC drift detection in this issue — that needs
  network access and belongs to #588's testnet smoke-test, not a hermetic unit test.
[...]

PR #1152, opened 2026-07-23
[backend] Fix Vault Oracle Redis snapshot & Deduplicate strategy visibility (#1103, #1102, #1120)

### LINKED ISSUE #1103: vault_service.py: displayed vault returns are always fabricated — REDIS_HOST/PORT typo + no oracle price-snapshot writer exists
## Summary

\`VaultService._compute_returns()\` in [\`backend/archimedes/services/vault_service.py\`](backend/archimedes/services/vault_service.py) (lines 248-357) is meant to compute real vault returns from oracle price snapshots in Redis, falling back to a synthetic estimate only when no snapshot exists. In practice it **always** falls through to the synthetic path, in every deployed environment, for two independent reasons:

1. **Redis connection uses the wrong env vars.** Line 267-269 connects via \`REDIS_HOST\`/\`REDIS_PORT\`:
   \`\`\`python
   r = _aioredis.Redis(
       host=os.getenv("REDIS_HOST", "localhost"),
       port=int(os.getenv("REDIS_PORT", "6379")),
       decode_responses=True,
   )
   \`\`\`
   Every other Redis consumer in the repo (\`redis_state.py\`, \`limiter.py\`, \`job_queue.py\`, etc.) and both \`docker-compose*.yml\` files use \`REDIS_URL\`. \`REDIS_HOST\`/\`REDIS_PORT\` are set nowhere in \`.env.example\`, docker-compose, or \`infra/\` — so in a real deployment this either connects to the wrong host or fails outright, and the \`except Exception\` at line 321 silently swallows it.

2. **Even with (1) fixed, nothing ever writes the snapshot.** Line 275's \`snapshot_key = f"vault:prices:{vault_address}"\` is read here but is never written anywhere in the codebase (confirmed via grep) — there's no oracle-updater job that populates \`vault:prices:{address}\`. So real returns can't be computed even once the connection works.

## Effect

Every vault's displayed \`return_24h\`/\`return_7d\`/\`return_30d\`/\`return_inception\` (surfaced via \`api/schemas.py\`'s \`return_24h\`/\`return_inception\` fields) is **always** the synthetic fallback: a hardcoded \`ASSET_EXPECTED_RETURN\` table plus deterministic per-address MD5-hash "noise" (lines 328-357) — presented to users as if it were a real computed return, with no distinguishing field in the response. This directly conflicts with CLAUDE.md's Tier-0 Claim Integrity priority ("every UI/pitch claim true on the live path... loud fallback telemetry").

## Scope

Two parts, can land separately:

1. **Trivial, low-risk fix (can land immediately):** swap \`REDIS_HOST\`/\`REDIS_PORT\` for the repo's standard \`REDIS_URL\` pattern (see \`redis_state.py\` for the exact pattern to copy) so the code path is at least reachable.
2. **Real fix (needs an oracle price-snapshot writer + a decision on cadence/ownership):** something needs to write \`vault:prices:{vault_address}\` at vault-creation time (and ideally periodically) so \`_compute_returns\` has real data to read. This is on-chain/oracle integration territory — Dan's lane. Until (2) ships, the fallback should carry an explicit \`is_synthetic: true\` (or similar) flag through \`VaultHolding\`/\`VaultSummary\`/\`schemas.py\` so callers — and eventually the UI — can honestly label estimated vs. real returns rather than presenting them identically.

## Anti-goals

- Don't ship the REDIS_URL fix alone as if it "fixes" the underlying issue — it only makes the code path reachable, not correct, until (2) lands. Loud fallback telemetry (a flag, not silence) is the honest interim state.

Found during a code-review sweep (2026-07-12). Not assigning to \`t2o2\` (retired).

onder-akkaya:
This PR fixes the REDIS_HOST typo and adds a Redis oracle price-snapshot writer (Refs #1103 — the writer's key `oracle:prices:latest` did not match `_compute_returns`'s reader key `vault:prices:{vault_address}`, so vault returns remained on the simulated fallback; `cdb35a08` closed #1103 for real), replaces the hardcoded 7-address `synth_vault_addresses` dict in vault bootstrap with a live `SyntheticFactory.tokenVault()` SSOT lookup (Refs #1102 — narrower than "idempotency," which #1102 never claimed), and extracts the strategy-visibility rules into a single shared predicate (Issue #1120).

Tested via `pytest backend/tests/scripts/test_bootstrap_vaults_create_vaults.py backend/tests/chain/test_create_vault_abi_consistency.py -q` (5 passed, 0 failed) — note these two files cover createVault ABI drift (#652) and revert handling (#655), not the tokenVault SSOT lookup itself; see #1377 for tokenVault-path regression tests and a hardening fix to the fallback behavior.

---
*Edited 2026-08-20: corrected the #1102 claim's vocabulary, corrected the #1103 claim, and replaced the unqualified "tested thoroughly via pytest" with the actual command/count, per adversarial review. Original text asserted "idempotency gaps in vault bootstrap (Issue #1102)" — #1102's actual scope was the hardcoded-address/no-SSOT problem, not idempotency. Original text also asserted this PR "fixes... the missing snapshot writer (Issue #1103)" — it added a writer, but keyed `oracle:prices:latest` while the reader (`vault_service.py`'s `_compute_returns`) read `vault:prices:{vault_address}`, so #1103 ("displayed vault returns are always fabricated") was not actually closed by this PR; vault returns stayed on the simulated fallback until `cdb35a08` ("Fix #1103 for real: stop fabricating vault returns", #1152 follow-up) fixed it.*


danielscoffee (approved):
LGTM!

dbrowneup:
Reviewed against current `main` (`ac3786ff`) and the PR head (`6db7ca2e`). Good progress on three of the four things this touches — flagging one defect that needs a fix before merge, plus two small cleanups.

## What's good and should land

- **#1102** — `bootstrap_vaults.py` on `main` still has the hardcoded 7-address `synth_vault_addresses` dict (lines 168–176). This PR's replacement — an on-chain lookup via `SyntheticFactory.tokenVault(token)` with a legacy fallback — is exactly the right fix and matches what #1102 asked for.
- **#1120** — `backend/archimedes/services/strategy_visibility.py` doesn't exist on `main`; the same ownership-visibility predicate is hand-duplicated at four call sites (`strategies_routes.py` ~1808 and ~1872, `selection_bias_routes.py` ~482, `vaults_routes.py`'s `_strategy_record_visible`, whose docstring literally says "copied from `selection_bias_routes._generated_strategy_rigor`, itself copied from `strategies_routes.get_strategy`"). `is_strategy_visible()` + its tests consolidate this correctly, and all four sites are updated to call it.

## Two cleanups

1. **Ruff pin hunk is now a no-op.** `main` already has `ruff==0.15.13` pinned in `.github/workflows/quality-gate.yml` (Önder's own `2d0a2a6e`), with byte-identical text to what's in this branch — a two-dot diff of that file between `origin/main` and this branch's head shows zero difference on those lines. Safe to drop that hunk from the diff on rebase.
2. **Stray file**: `git_shuffle.sh` (added in `2a7ceace`) is a local branch-shuffling script that references unrelated branches (`onder/794-pyth-adapter`, `onder/827-aggregatorv3-consolidation`, plus a `dbrowneup/...` branch). Looks like a leftover from staging this commit locally — shouldn't be committed.

## #1103 fix doesn't close the loop — needs another pass

This is the one that matters. The write side and read side don't agree on a key, so after this merges, vault returns are still fabricated.

- Writer (`backend/archimedes/chain/oracle_updater.py:433`): `await r.set("oracle:prices:latest", json.dumps(price_map))` — a single **global** key, and `price_map` is **symbol**-keyed (`{p.symbol: p.price_usd for p in prices}`).
- Reader (`backend/archimedes/services/vault_service.py:278`): `snapshot_key = f"vault:prices:{vault_address}"` — a **per-vault** key, and the code that consumes it (`prices_at_creation.get(token, ...)` at line ~311) expects the stored dict to be keyed by **token address**, not symbol.

`grep -rn "vault:prices:" backend/` returns exactly one hit — that read. Nothing in the branch (or `main`) ever writes a `vault:prices:{vault_address}` key, so `snapshot` in `_compute_returns()` is always `None` and every call falls through to the synthetic/MD5-noise fallback a few lines below. `_compute_returns()`'s own docstring says it uses "real price changes" "if price snapshots exist in Redis (set by the oracle updater)" — that's the intent, but the plumbing doesn't connect yet. Issue #1103 would still be open after this merges (returns are still fabricated, just no longer via a crashing Redis connection).

Two ways to close it — reading #1103's own text ("something needs to write `vault:prices:{vault_address}` at vault-creation time... periodically"), the per-vault key is the one the issue actually asked for, so option (a) looks closer to intent:

- **(a)** Point the writer at per-vault keys: for each vault, write a token-address-keyed price dict to `vault:prices:{vault_address}` (at creation, and/or on a refresh cadence) instead of (or in addition to) the single global `oracle:prices:latest`.
- **(b)** Point the reader at the new global key: change `vault_service.py` to read `oracle:prices:latest` and resolve `alloc.symbol` instead of `alloc.token_address` — but this only gives current prices, not the "price at vault creation" baseline `_compute_returns()` needs to compute a return, so it'd need a separate mechanism to snapshot the creation-time price.

## Also

Branch is ~90 commits behind `main` at this point (was ~59 when this review was requested) — needs a rebase regardless of the above.

Happy to take a second look once the writer/reader keys line up.

PR #1154, opened 2026-07-23
[infra] Pin ruff==0.15.13 across all CI workflows

onder-akkaya:
## Summary
`quality-gate.yml`'s ruff-gate jobs were already pinned to `ruff==0.15.13`, but `main-format-guard.yml` and `import-guard.yml` still did an unpinned `pip install --quiet ruff`. A newer ruff release started reformatting Markdown-embedded Python by default, which made the ruff-gate step fail on every open PR *and* on direct pushes to `main` since 2026-07-23, for reasons unrelated to each PR's actual diff.

This pins the remaining two workflows to the same `0.15.13` that `.pre-commit-config.yaml` and `quality-gate.yml` already use, so all four ruff install sites move in lockstep.

## Verify
- `ruff==0.15.13 format --check .` → `483 files already formatted` (reproduced locally against this branch)
- CI on this PR should show ruff-gate green again

## Anti-goals
- Not touching lint rule selection or `ruff.toml` — this is a version-pin fix only.

### DIFF
--- a/.github/workflows/import-guard.yml
+++ b/.github/workflows/import-guard.yml
@@ -34,7 +34,10 @@ jobs:
         with:
           python-version: "3.12"
       - name: Install guards
-        run: pip install --quiet ruff deptry
+        # ruff pinned to match quality-gate.yml + main-format-guard.yml +
+        # .pre-commit-config.yaml — unpinned installs drift to whatever ruff
+        # just released. Bump all four together when intentionally upgrading.
+        run: pip install --quiet ruff==0.15.13 deptry
       - name: Undefined names (F821) — catches aliased-import NameErrors
         run: ruff check backend/archimedes --select F821 --no-cache
       - name: Byte-compile (no syntax / obvious import-time breakage)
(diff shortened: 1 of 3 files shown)


onder-akkaya:
Merging this without a review.

It was the only thing standing between the current false-positive CI failure and every one of the 12 open Dependabot PRs plus #1170, all of which started failing "Ruff — format + critical lint rules" this afternoon for a reason that had nothing to do with their actual diffs: `quality-gate.yml` installed ruff unpinned (`pip install --quiet ruff`), and CI picked up a newer release that started reformatting Python embedded in Markdown docs. Fourteen files nobody in this batch touched, failing a hard-block gate on thirteen unrelated PRs at once.

This PR has been open since 2026-07-23, pins `ruff==0.15.13` across the four places it's installed, touches CI config only (no app code), and was already green and mergeable. Given it was blocking an entire reviewed-and-ready batch on a false positive, I merged it and then pushed a sync commit (merge main) into each blocked branch so their next CI run actually picks up the pin instead of failing against the same stale merge ref.

Flagging here for visibility since this skipped the usual one-approval norm. Happy to hear about it if anyone disagrees with the call.

PR #1174, opened 2026-07-28
[infra] Refuse --apply when AGENT_DRY_RUN is unset and unseeded (Copilot follow-up to #1173)

dbrowneup:
## Why
Actions the Copilot review comment on **#1173**, which merged before the comment was addressed.

#1173 moved `AGENT_DRY_RUN` out of a `docker run -e` flag into the SSM-sourced `--env-file` and **documented** that an unset value means live trading. Copilot's point: *documenting it is not enforcing it.*

`setup-ssm-secrets.sh` skips unset env vars **even in `--apply` mode**. So an operator can run `--apply`, miss one `skip` line among twenty, and leave `/archimedes/prod/AGENT_DRY_RUN` absent — which `agent_runner.py` resolves to `"false"`, arming the funds-adjacent agent for **live on-chain signing**.

> Verified on the live account: `/archimedes/prod/AGENT_DRY_RUN` is **currently ABSENT** — this is not hypothetical, it is today's state.

## What
A fail-fast guard, deliberately **narrow** so the documented partial-run workflow still works:

- only in `--apply` mode (dry run stays advisory)
- only for `AGENT_DRY_RUN` — not a blanket require-everything
- satisfied by **either** an exported value **or** the param already existing in SSM, so re-seeding one unrelated secret doesn't require re-passing it
- evaluated **before any put**, so a refusal writes nothing at all

## Verification

| Path | Expected | Result |
|---|---|---|
| dry run, unset | unchanged, advisory | ✅ exit 0 |
| `--apply`, unset + absent in SSM | refuse, write nothing | ✅ **exit 3** |
| `--apply`, `AGENT_DRY_RUN=true` | proceed + seed | ✅ `put /archimedes/prod/AGENT_DRY_RUN` |
| `-h` | unchanged | ✅ exit 0 |

`bash -n` clean. No Terraform touched.

### DIFF
--- a/infra/scripts/setup-ssm-secrets.sh
+++ b/infra/scripts/setup-ssm-secrets.sh
@@ -73,6 +73,37 @@ APPLY=false; for a in "$@"; do case "$a" in
 $APPLY && echo ">>> APPLY MODE — writing SecureString params under ${PREFIX}/" \
         || echo ">>> DRY RUN — re-run with --apply to write. Values are read from env; names only are shown."
 
+# ── Fail-closed guard for funds-behaviour parameters ────────────────────────
+# Everything else in this script is intentionally skip-if-unset, so partial runs
+# work. That default is WRONG for AGENT_DRY_RUN: agent_runner.py resolves an
+# unset AGENT_DRY_RUN to "false" (= LIVE on-chain signing), so "operator ran
+# --apply and didn't notice one skip line among twenty" silently arms the
+# funds-adjacent agent. Enforce the invariant in the tool instead of relying on
+# the caller reading output (Copilot review, PR #1173).
+#
+# Deliberately narrow so partial runs keep working:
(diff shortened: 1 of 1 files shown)


dbrowneup:
Good catch — fixed in the latest commit.

`agent_runner.py` evaluates `os.getenv("AGENT_DRY_RUN","false").lower() == "true"`, so only the literal `true` enables dry-run and `1`/`yes`/`on` silently mean LIVE signing. The guard now validates the **value**, accepting only case-insensitive `true`/`false`:

| value | result |
|---|---|
| `true`, `false`, `TRUE`, `False` | ✅ accepted |
| `1`, `yes`, `on`, `true ` (trailing space), `TRUEISH` | ⛔ refused, exit 3, **before any write** |
| dry run (no `--apply`) | unchanged, advisory |

PR #1200, opened 2026-08-02
[contracts] Fix Vault NAV decimals for 6-dec synths + performance-fee share-mint dilution (split from #1129)

onder-akkaya:
## Summary

Splits two unrelated fixes out of #1129 (the fee-bps cap PR) per [Dan's review](https://github.com/a-apin/archimedes/pull/1129#pullrequestreview) — they got bundled into `64fdea43` along with the fee cap, but are higher-blast-radius correctness fixes that should be reviewed and land on their own timeline rather than riding inside a PR titled "cap fee bps."

**NEEDS DAN + BOGDAN REVIEW** — same reasoning as #1129: live-funds NAV/share-mint math.

## What's actually broken

**1. NAV valuation for 6-decimal synths (`_tokenValueUsdc` + churn-guard `sellValue`)**

`#942` fixed `totalAssets()` to scale by a token's real decimals via `_tokenUnit(token)`, but missed two other call sites that still divide by a hardcoded `1e18`:
- `_tokenValueUsdc()` (`Vault.sol:665`)
- the `#915` churn guard's `sellValue` in `rebalance()`'s sell branch (`Vault.sol:449`)

Production synths are 6-decimal. For a 6-decimal balance, dividing by `1e18` instead of `1e6` crushes the computed value by `10**12` — a $20k position prices out at ~$0.00002. Since `_requireTowardTarget()` derives `currentWeight` from `_tokenValueUsdc()`, every 6-decimal synth looked like ~0% of NAV to the churn guard, so `currentWeight <= target + band` was always true on the sell branch — **every rebalance-sell of an over-target 6-decimal synth reverted with `RebalanceNotTowardTarget`, no matter how overweight the vault actually was.**

Fixed by routing both sites through the existing `_tokenUnit(token)` helper.

**2. Performance-fee share-mint formula**

The old formula converted a USDC-denominated fee amount to shares 1:1:
```solidity
uint256 perfShares = (perfFeePerShare * _totalSupply) / 1e18;
```
`perfFeePerShare` is USDC-per-share; multiplying by `totalSupply` and dividing by `1e18` yields a **dollar amount**, not a share count. Minting that amount 1:1 as shares only breaks even when `navPerShare == 1e18` exactly — but a performance fee is only ever owed once NAV/share has grown *past* $1, which is exactly when this breaks: at `navPerShare = $2`, a $5,000 intended fee minted shares worth ~$6,667.

Fixed to the same dilution-mint formula the **management-fee branch two lines above already uses** (so this isn't a novel formula — it's making performance-fee accrual consistent with management-fee accrual, which was already correct):
```solidity
uint256 totalPerfFeeAssets = (perfFeePerShare * _totalSupply) / 1e18;
uint256 perfShares = (totalPerfFeeAssets * _totalSupply) / _totalAssets;
```

## Test plan

Two new regression tests, both **verified to fail on the pre-fix source** (reverted `Vault.sol` locally, confirmed both fail with the exact predicted numbers, then restored the fix):

- `test_rebalance_sell_toward_target_succeeds_for_6_decimal_synth` — fails with `RebalanceNotTowardTarget` on unfixed code, passes on fixed
- `test_performanceFee_shareMint_uses_dilution_formula_not_raw_assets` — asserts exact minted-share amounts (`250e6` platform / `2_250e6` creator); unfixed code mints exactly double (`500e6` / `4_500e6`)

```
forge test        # 287 passed, 0 failed (full contracts suite)
```

## Anti-goals

- Does not touch the fee-bps cap itself — that's #1129, unchanged.
- Does not touch the management-fee formula — it was already correct; only referenced as the pattern to match.

---

**2026-08-20 re-verification (rebased to 0-behind current main, clean):** forge 316/316 on the rebased head. Mutation check: reverting the dilution fix in place to the old 1:1 mint → `test_performanceFee_shareMint_uses_dilution_formula_not_raw_assets` fails with `500000000 != 250000000` (at $2 NAV/share the pre-fix mint hands the fee recipient exactly 2× the fee) — restored clean. The dilution form (`feeAssets * totalSupply / totalAssets`) is the conservative direction: it slightly under-pays the recipient rather than ever over-paying, matching the cited Yearn precedent.

**Review minors addressed:** the perf-fee test now runs at the merged #1129 cap constant (read before the prank — the getter is an external call and would otherwise consume it); the dilution comment documents the value direction (minted shares self-dilute, recipient lands slightly below feeAssets — depositor-favoring by design; the dead-shares 1:1 HWM note is unaffected). Re-rebased 0-behind at push time.

### DIFF
--- a/contracts/src/Vault.sol
+++ b/contracts/src/Vault.sol
@@ -470,7 +470,7 @@ contract Vault is IVault, ERC20, Ownable, ReentrancyGuard, Pausable {
             uint256 minOut = _oracleMinOut(tokenOut, asset, amount);
             // #915 churn guard: this sell must move tokenOut toward its target weight
             // (over-allocated → sell) and not past it. sellValue is the synth's oracle NAV.
-            uint256 sellValue = (amount * PriceOracle(tokenOracle[tokenOut]).getPrice()) / 1e18;
+            uint256 sellValue = (amount * PriceOracle(tokenOracle[tokenOut]).getPrice()) / _tokenUnit(tokenOut);
             _requireTowardTarget(tokenOut, sellValue, false);
 
             // Approve router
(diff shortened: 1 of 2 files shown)


onder-akkaya:
@dbrowneup @mnemonik-dev this is the NAV/decimals + fee-mint fix split out of #1129 per your review there. CI's green, no reviews on it yet. Same live-funds-math territory as #1129, would appreciate a look on both.

PR #1216, opened 2026-08-03
[frontend] Mark the leaderboard provisional while backtests are re-computed

dbrowneup:
Adds a visible banner to the leaderboard for the duration of the backtest re-run:

> **Provisional — figures are being re-computed.** Backtest data is being re-computed following a routing defect; figures shown are known to be incorrect and will change.

### Why the numbers are wrong

PR #1203 fixed a routing defect: most library strategies were backtested against a hardcoded `SPY` default instead of their own declared `ASSET_UNIVERSE`. Cross-sectional strategies fared worse — given a single feed they had nothing to rank, so they emitted **zero trades and a flat 0% return**, which the rigor gate then graded as though it were a result.

Every figure on this page is therefore known to be wrong until the persisted backtests are re-computed.

### Why a banner rather than taking the page down

Dan's call, and the reasoning is worth recording: taking the page down loses the demonstration, and leaving it **silently** wrong is the precise failure this product exists to oppose. Saying so out loud is the only option consistent with the thesis — and it is the strongest thing a skeptical reader could click on.

### Removal

The block is self-contained and carries its own removal condition in a comment, so retiring it after the re-run is one deletion, not an archaeology exercise.

### Verification

`npm ci` + `npm run build` + `npm run lint` all clean, and the banner string is present in the **shipped bundle** (`dist/assets/index-*.js`) — verified in the build output, not just in source.

### DIFF
--- a/.github/workflows/quality-gate.yml
+++ b/.github/workflows/quality-gate.yml
@@ -43,6 +43,38 @@ jobs:
       - name: EC2 user-data stays under the 16 KB limit
         run: bash ./infra/scripts/check-user-data-size.sh
 
+  # The analytics engine had NO CI coverage at all until now. pytest.ini sets
+  # `testpaths = backend/tests` and lists analytics-engine under
+  # `norecursedirs`, and no workflow ran its suite — so its 227 tests never
+  # executed on a PR. That is how #1203 (which re-routed EVERY backtest to its
+  # declared ASSET_UNIVERSE, and therefore determines every number the product
+  # publishes) merged with a full row of green checks and zero automated
+  # verification of the code it changed. The engine sits on the MVP critical
+  # path; it gets its own gate.
+  #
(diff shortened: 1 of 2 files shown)


dbrowneup:
Closing — reversed by Dan in favour of fixing fast rather than labelling.

The banner was only attractive because the fix looked slow, and the fix looked slow because the provenance audit was framed as a precondition. A restorable snapshot of `backtest_results` removes that precondition entirely: with the table dumped and the restore verified, a blanket re-run is reversible, so the audit becomes an after-the-fact comparison instead of a gate. Hours, not days.

The reasoning in this PR body still stands and is worth keeping in the record — it just applies to a fallback rather than the plan. **Pre-agreed trip-wire:** if the corrected figures are not live within 24 hours, this banner goes up, no further discussion needed. Fix-fast beats banner only while fix-fast is actually fast.

Issue #1241, opened 2026-08-15
[docs] Claims ledger — make each public claim true or change it


onder-akkaya:
"Deliver the promises" means making each public claim true or changing it. This is the running ledger of every claim the UI, the manifests, the docs, or the pitch make, with a verdict on each and the work that gates it.

Measured against the live system on 2026-08-10, re-spot-checked 2026-08-16.

Ground rule from CLAUDE.md: claims must be true. A guarantee backed by a fixture or a cached boolean rather than the live path is the thing we said we'd stop doing.

## Ledger

| Claim | Where | Verdict |
|---|---|---|
| "10,000 q-fin research papers" | `docs/user-stories.md`, `/health`, Landing | **Soften** to "10,000 paper abstracts, retrieved by semantic rerank." Real: 10k abstracts, keyword and synonym filter, MiniLM/TF-IDF rerank, 2-paper-minimum fusion with arxiv-id validation and Xia outcome embargo. Not real: SPECTER2, knowledge graph (`corpus_kg_built: false`), full-text chunk RAG. Hide the Graph and KG tabs (#778) |
| "gated through selection-bias rigor" | everywhere | **Keep, but only after the fixture fallback dies and the dataclass gate is deleted.** Not true today |
| "generated and curated graded on the same scale" | comment above `generation_pipeline.py:1573` | **Keep.** `BacktestResult.passes_rigor_gate` / `passes_validation` deleted, `generation_pipeline.py` re-pointed to `live_rigor_gate.verdict_from_returns` — commit `d7073f13`, merged in #1242 (`414ce280`). Verified on `main`: no gate definitions remain in `models/backtest.py`. The pipeline call site (`generation_pipeline.py`'s `passes = live.passes` line) is covered as of #1379 (`TestPipelineCallSiteReadsTheLiveVerdict` in `test_gate_equivalence.py`) — that PR is open, not yet merged to `main`; do not read this row as "covered on `main` today" until it lands |
| Leaderboard figures | `/leaderboard` | **Retract until the re-run.** #1226's provisional banner is the right interim. Re-scope it to "engine-attributed, cost-parity floor" afterwards |
| "executing into your non-custodial vault on Arc" | `user-stories.md`, Landing | **Soften** to "deploy to a non-custodial vault on Arc testnet." Zero vaults have ever been deployed, and #1103 says displayed vault returns are fabricated |
[...]

PR #1242, opened 2026-08-15
[backtest][rigor] One cost floor and one rigor gate across the three engines

onder-akkaya:
Workstream A of the mainnet sprint: the two changes that have to land before
the curated library can be re-run without publishing a fresh set of biased
numbers. Two independent commits, reviewable separately.

## 1. One cost floor (`5c601fba`)

Three engines write into `backtest_results` and one gate grades them without
knowing which produced a row:

| Engine | Where | Charged before |
|---|---|---|
| curated, backtrader | `analytics-engine/engine.py` | commission + slippage |
| generated weights, pandas | `services/portfolio_backtester.py` | commission only |
| fusion DSL, backtrader | `services/fusion_evaluator.py` | commission only |

`grep -rn "set_slippage" --include="*.py" .` returned exactly two hits, both in
the curated engine. So the engines grading *generated* strategies were
systematically flattering themselves against the curated library they get
ranked against.

- `costs.py` gains `DEFAULT_COST_MODEL` (10bps per side, 5bps slippage — the
  CLI's existing defaults promoted to a shared constant) and
  `cost_model_fingerprint()`, a stable readable id stamped on results.
- `run_command` builds a `CostModel` and passes it to all three runners, which
  also revives the per-symbol override path — it was unreachable because
  `run_command` only ever forwarded two scalars.
- `portfolio_backtester` charges slippage on the same two-sided turnover as its
  linear cost. Its Almgren impact term stays on top, so it is stricter than the
  floor and never looser. It also stops reporting `"slippage_bps": tx_cost_bps`,
  which claimed a figure it wasn't charging.

Literal parity isn't reachable in this window — Almgren impact needs ADV inside
a custom `bt.CommInfoBase`. An identical floor plus a fingerprint on every row
is the defensible version, and it puts the residual asymmetry on the row.

The fusion engine's two call sites are deliberately **not** in this PR; they
land with the rest of that file's changes so it only gets opened once.

## 2. One rigor gate (`d7073f13`)

`BacktestResult` carried its own `passes_rigor_gate` with its own thresholds,
and `generation_pipeline` used it. The curated read path grades through
`live_rigor_gate.verdict_from_returns` and the `rigor_profiles` ladder. The
comment above the call claimed the two matched. They didn't, so "generated and
curated are graded on the same scale" was false as written.

It was also broken invisibly: `backtest_portfolio` sets `pbo_score=None`
(library-level metric, refreshed later), and the property short-circuited to
False whenever PBO was None. **Every generated portfolio strategy failed it
unconditionally** — the line returned a constant, not a grade.

Deleted both properties and the helper only they used; the pipeline now grades
the real return series exactly as the DSL path in the same file already does.
Only one production caller read the property — the ~60 other
`passes_rigor_gate` hits are a field on `Strategy`/`StrategyPassportRecord` and
are untouched.

## Verification

```
analytics-engine          234 passed
backend, full unit suite  3067 passed
backend/tests/test_cost_parity.py       8 passed (hermetic)
backend/tests/test_gate_equivalence.py 11 passed (hermetic)
ruff format --check backend/            409 files clean
```

The 6 `test_alembic_migrations.py` failures under `env -i` are the `alembic`
binary missing from a stripped PATH. They pass with it on PATH and fail the
same way on an unmodified tree.

Two notes on the tests, both deliberate:

- The cost drift check reads `DEFAULT_COST_MODEL` out of `costs.py` with `ast`
  rather than importing it. The backend unit-test job does not
  `pip install ./analytics-engine` — only the analytics-engine job does — so an
  `importorskip` would have skipped the whole file in CI and reported green
  while checking nothing. I injected drift to confirm it actually fails.
- `test_linear_cost_is_round_trip_on_turnover` moved to include the slippage
  leg. The round-trip convention it exists to pin is unchanged, only the rate.

## Not in this PR

`cost_model_id` end-to-end through the schemas and the leaderboard, the
look-ahead OR in `backtest_mapper`, and the `correlation_to_spy` null coercion.
Same session's work, separate change.

### DIFF
--- a/analytics-engine/src/archimedes_analytics_engine/cli.py
+++ b/analytics-engine/src/archimedes_analytics_engine/cli.py
@@ -12,6 +12,7 @@ from typing import Any
 import backtrader as bt
 import pandas as pd
 
+from .costs import CostModel, cost_model_fingerprint
 from .data import fetch_ohlcv
 from .engine import (
     BACKTEST_ENGINE_TAG,
@@ -154,9 +155,27 @@ def run_command(
     paper_claimed_max_dd: float | None = None,
     walk_forward_split: float | None = None,
     fetcher: Callable[[str, str, str], pd.DataFrame] = fetch_ohlcv,
+    cost_model: CostModel | None = None,
 ) -> dict:
     run_id = datetime.now(UTC).strftime("%Y%m%dT%H%M%SZ")
     bundle = load_strategy(strategy_path, strategy_class)
+
+    # Build a CostModel from the flat bps arguments when the caller didn't
+    # supply one, then pass it to every runner. Two reasons this matters
+    # beyond tidiness:
+    #
+    # 1. The per-symbol override path in CostModel was unreachable from here —
+    #    run_command only ever forwarded two scalars, so _configure_broker
+    #    took its flat branch every time even though it already knows how to
+    #    honour a model.
+    # 2. Every result row can now be stamped with a cost_model_id, so a reader
+    #    can tell whether two numbers were charged the same way instead of
+    #    assuming it.
+    #
+    # The flat arguments stay authoritative when no model is given, so callers
+    # that pass tx_cost_bps/slippage_bps keep their exact current behaviour.
+    if cost_model is None:
+        cost_model = CostModel(default_bps=float(tx_cost_bps), slippage_bps=float(slippage_bps))
 
     metadata = _merge_metadata(
         bundle.metadata,
@@ -207,6 +226,7 @@ def run_command(
             name_b=ops[1],
             transaction_cost_bps=tx_cost_bps,
             slippage_bps=slippage_bps,
+            cost_model=cost_model,
         )
 
         results.append(
@@ -248,6 +268,7 @@ def run_command(
                 initial_cash=initial_cash,
                 transaction_cost_bps=tx_cost_bps,
                 slippage_bps=slippage_bps,
+                cost_model=cost_model,
             )
             per_asset.append((op, bt_result))
 
@@ -326,6 +347,7 @@ def run_command(
             names=symbols,
             transaction_cost_bps=tx_cost_bps,
             slippage_bps=slippage_bps,
+            cost_model=cost_model,
         )
 
         # The single N-asset result IS the strategy result — no per-asset rows,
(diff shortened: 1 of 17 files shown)


dbrowneup:
Coordination note for the contracts session (from the strategy session): #1242 is already MERGED — it landed in tonight's consolidation sweep after I verified the four unresolved threads were the Alembic module-var false-positive pattern. Re: the direction-fork question on portfolio-simulator-v1 — there is a factual contradiction to reconcile before any deletion: your earlier audit stated psv1 has a live production caller (generation_pipeline.py:1561, used when a candidate emits {ticker: weight}), while the deletion manifest calls it 'structurally dead in production.' Both cannot be true. Please post the manifest's reachability evidence (who calls it, and why the 1561 call site is dead or migrated) on the deletion issue before removing anything. If psv1 does go, #1242's unification work on that engine becomes cleanup to remove with it — not a conflict. Dan arbitrates if it is a genuine fork.

dbrowneup:
Correction to this PR's body, found in review of the backlog follow-up (#1241):

The body's final section ("Not in this PR") lists `cost_model_id` end-to-end, the
look-ahead OR in `backtest_mapper`, and the `correlation_to_spy` null coercion as
explicitly out of scope, and frames this as "Two independent commits, reviewable
separately." Both statements were stale by the time this merged. A third commit,
`ba102c39` ("Attribute every row to an engine and a cost basis, stop fabricating
correlations"), landed all three of those items in this same PR, and also carries
a schema migration (`b41c7d9e2a05_add_cost_and_lookahead_provenance.py`) that the
body never mentions — two new nullable columns, an `alter_column` nullability
change on the correlation columns, a backfill `UPDATE`, and a hard-fail guard in
`insert_backtest_if_missing` for a missing `backtest_engine`. None of that is
flagged in the Verification section either (the body's `3067 passed` is the
two-commit figure; `ba102c39`'s own message says `3087 passed`).

None of this changes the substance of the fix — it's a good change — but a
reviewer who reads the body only would not know a migration merged here, or that
the "Not in this PR" exclusions were superseded before merge. Flagging for the
record and so `git log`/PR-search on this doesn't undersell what shipped.

Filed alongside a same-session follow-up (#1241 backlog) that also found and
fixed two related gaps this PR left open: `fusion_evaluator.py`'s two
`cerebro.broker` call sites were never put on the cost floor (commission only,
no slippage — the exact asymmetry this PR closed for the other two engines), and
`cost_model_id` never reached the DB column for the portfolio-simulator or
dsl-fusion engines (only the curated mapper path wrote it).

PR #1256, opened 2026-08-18
[backtest] momentum_N is a trailing return, not a price ratio (F1)

dbrowneup:
## What this fixes

**F1 from the interpreter-divergence audit (21 confirmed divergences, 0 refuted): the most damaging one, and the one that blocks the re-run.**

The backtest computed `momentum_N` as a **price ratio** (`close[0] / close[-N]`, centred on 1.0). The live signal evaluator, `_tsmom_signal`, and `rank_market` all compute a **trailing return** (`… − 1.0`, centred on 0.0). Every hand-authored momentum condition in the repo — `{"gt": ["momentum_N", 0]}` — is written for the return convention.

Prices are always positive, so under the ratio convention that entry filter was a **tautology**. Measured in this PR's own tests:

| Scenario | Old (ratio) | New (return) | Live path |
|---|---|---|---|
| Market declining 0.15%/bar, entry `momentum_20 > 0` | **enters long**, $100,000 → **$66,279** | stays flat, $100,000 | stays flat ✓ |
| +2% trailing return, entry threshold `0.5` | **enters**, → $131,728 | stays flat | stays flat ✓ |
| Rising market, entry `> 0` | enters | enters ✓ | enters ✓ |

Same spec, opposite positions between the published backtest and the deployed strategy.

## The fix

One term: subtract 1.0 (`dsl_to_backtrader.py`, `_make_indicator`). Plus one sentence in the LLM spec contract (`_SPEC_CONTRACT`) defining `momentum_N`'s scale — it previously offered the indicator with **no defined semantics**, so generated thresholds were a coin flip.

## Verification

- Tests are behavioural (through the real `run_dsl_backtest` path) and **proven discriminating before being trusted**: against the pre-fix code the two stay-flat cases fail with the $66,279 / $131,728 values above; against the fix all three pass with flat cases at exactly $100,000.
- The rising-market case guards the *direction* of the fix — an over-correction (e.g. −2.0) would fail it while passing the flat cases.
- Full backend suite: **3068 passed, 0 failed.** No existing test broke — which is itself the finding: nothing pinned this convention, which is how a tautology survived in every published momentum backtest.

## Sequencing

⛔ **This blocks the curated/library re-run** (queue item 3). Re-running before this lands would re-publish numbers built on the tautological filter with fresh authority. It also blocks the paper-trading backend, which Dan has specified as running on backtest-engine semantics *post-F1*.

F5 (RSI smoothing) is next and separate. F2/F3 are live-money changes: prepared but held for Dan's explicit go-ahead per the strategy-session brief.

### DIFF
--- a/backend/archimedes/agents/strategy_fusion.py
+++ b/backend/archimedes/agents/strategy_fusion.py
@@ -788,7 +788,10 @@ tickers the mechanism trades from the user's selected assets (in user_steer); th
 platform overrides this with the user's chosen universe, so do not default to a \
 single broad-market proxy. Valid rebalance_frequency values: \
 daily, weekly, monthly. Valid indicators: sma_N, ema_N, rsi_N, momentum_N (replace \
-N with an integer period). Entry/exit conditions use comparison ops (gt, lt, gte, lte) \
+N with an integer period). momentum_N is the trailing N-bar RETURN, centred on 0 \
+(+0.05 means +5%); write momentum thresholds on that scale — e.g. \
+{"gt": ["momentum_20", 0]} means "trailing 20-bar return is positive". \
+Entry/exit conditions use comparison ops (gt, lt, gte, lte) \
 or logic ops (and, or, not). Position sizing types: full_invested_when_in_market, \
 equal_weight, volatility_target (needs annual_pct). look_ahead_safe MUST be true. \
 parameter_variants is OPTIONAL: a dict mapping indicator aliases to 2-8 numeric \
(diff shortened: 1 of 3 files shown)


dbrowneup:
Reviewed the diff personally (formal approve blocked — both sessions share this GitHub identity). Verdict: correct and complete. The '- 1.0' fix matches the three return-convention call sites, the fusion-prompt update makes generated specs write thresholds on the new scale, and the tests are discriminating by construction (verified-failing pre-fix) with an over-correction guard. CI green. Ready for Dan's merge — this unblocks the re-run and paper trading.

dbrowneup:
**Dan approved — MERGED**, along with #1257. Contracts session: your queue is unblocked — the leaderboard re-run and the paper-trading build can both proceed on F1-fixed semantics. Reminder from the refresh diagnosis (checkpoint doc): the curated/backtrader path hasn't written since Jul 2 and 9 strategies are persistently rowless in backoff ('pairs family') — fold those into the re-run scope rather than discovering them again.

PR #1257, opened 2026-08-18
[live] RSI uses Wilder smoothing, matching the backtested RSI (F5)

dbrowneup:
## What this fixes

**F5 from the interpreter-divergence audit.** The live signal evaluator computed RSI as a plain rolling mean of gains/losses — **Cutler's RSI** — while the backtest's `bt.indicators.RSI` uses **Wilder's recursive smoothing**. Backtrader ships the rolling-mean variant as a *separate* indicator precisely because they're different indicators.

Measured on this PR's fixture series: **81.82 (Cutler) vs 80.75 (Wilder)** — and on shorter windows the audit measured a 7.7-point gap. Gaps like that straddle real thresholds, so a graded backtest could pass `rsi > 65` while the live path fails it on identical data.

The sharpest part: the function's own docstring claimed *"standard Wilder-style RSI"* above the Cutler code — a load-bearing false comment, and plausibly why nobody re-checked. The docstring now says what each indicator actually computes, including the two asymmetries deliberately left **out** of this PR: EMA seeding (F10 — benign, converges) and `realized_vol` having no backtest counterpart (F6 — owned by the codegen work).

## Verification

- The parity oracle is **`bt.indicators.RSI` itself** run over the same series — not a hand-written Wilder formula that could share a bug with the code under test (the same-literal-both-sides trap).
- **Proven discriminating:** against the old code, the parity test fails with `81.818… != 80.750…`. A second test pins that Wilder and Cutler measurably *disagree* on the fixture series, so a passing parity test can never be vacuous coincidence.
- Degenerate branches kept and pinned: all-gains saturates at 100; insufficient history reads NaN, never a fabricated value.
- Full backend suite: **3069 passed, 0 failed.**

## ⚠️ For merge approval (Dan)

This **changes live signal behaviour** for any deployed spec with an RSI condition. The direction of change is the honest one — live now agrees with what the published backtest graded — but per the sprint brief, `AGENT_DRY_RUN` state should be confirmed at merge time. Not self-merging.

Sequencing: F1 is #1256 (blocks the re-run). F2/F3 remain **prepare-and-hold** for Dan's explicit go-ahead.

### DIFF
--- a/backend/archimedes/services/strategy_signal_evaluator.py
+++ b/backend/archimedes/services/strategy_signal_evaluator.py
@@ -993,25 +993,46 @@ def _indicator_periods(spec: StrategySpec) -> dict[str, tuple[str, int]]:
 def _compute_indicator_value(name: str, period: int, prices: pd.Series) -> float:
     """Compute the latest value of one indicator from a price Series.
 
-    Uses the SAME numpy/pandas primitives the 5 hardcoded evaluators use, so the
-    live spec path produces values consistent with both them and the backtrader
-    interpreter's indicators:
+    Same numpy/pandas primitives as the 5 hardcoded evaluators, aligned with
+    the backtrader interpreter's indicators (the backtest is what gets graded,
+    so the live value must be computed the way the graded value was):
       - sma           → rolling(N).mean().iloc[-1]
-      - ema           → ewm(span=N).mean().iloc[-1]
-      - rsi           → standard Wilder-style RSI over N
+      - ema           → ewm(span=N).mean().iloc[-1]  (first-value seeded; bt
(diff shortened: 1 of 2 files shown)


dbrowneup:
Reviewed personally (formal approve blocked — shared GitHub identity). Verdict: correct. Wilder SMMA with SMA seed matches backtrader's MovAv.Smoothed, and using bt itself as the parity oracle means the test cannot share a bug with the implementation. Your merge precondition is satisfied: AGENT_DRY_RUN=true confirmed in prod via SSM tonight — live behavior change is safe while the agent is not live-trading. Ready for Dan's merge.

PR #1261, opened 2026-08-18
Fix #1182: four rate-limited routes 500 on every successful call

### LINKED ISSUE #1182: [api] GET /api/selection-bias/gate/{id} returns HTTP 500 in production
## Summary

`GET /api/selection-bias/gate/{strategy_id}` — the strategy passport's own rigor-gate
verification endpoint — returns HTTP 500 in production for real strategy ids. This is the
single surface a skeptical reviewer would use to independently re-check a rigor badge
(DSR/PBO pass verdict) for one specific strategy, decoupled from the batch leaderboard
call. It being down means the passport's "verify this yourself" claim currently has no
working per-strategy endpoint behind it.

## Evidence (reproduce against production right now)

```
$ curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/91ade2b4f219fd5158f02295c21022c4
500

$ curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/d90b357a6693747d7c521d6998314bbd
500

$ curl -sS https://archimedes-arc.com/api/selection-bias/gate/91ade2b4f219fd5158f02295c21022c4
Internal Server Error
```

Both ids return `500` with body `Internal Server Error` — no structured error, no trace id,
nothing a client can act on.

## Scope (exact files)

- `backend/archimedes/api/selection_bias_routes.py` — route handler
  `evaluate_strategy_rigor` at `@selection_bias_router.get("/gate/{strategy_id}")`
  (currently line ~567). It tries `_provider().get_strategy(strategy_id)` (curated
  cohort), falls through to `_generated_strategy_rigor(strategy_id, request, strictness)`
  for non-curated ids, and otherwise calls `evaluate_rigor_gate(strictness=strictness)`
  and searches the batch result for a matching `strategy_id`. The 500 is coming from an
  unhandled exception somewhere in that path (or in `_generated_strategy_rigor` /
  `evaluate_rigor_gate`) — root cause is not yet identified, do not guess at a fix here.
- Related, for context only (do not touch without separately verifying): consumers that
  document/depend on this route's contract — `backend/archimedes/api/strategies_routes.py`
  (comments at lines ~51, ~61, ~148, ~253, ~276 reference `GET /api/selection-bias/gate`
  parity) and `backend/archimedes/api/vaults_routes.py`.

## Acceptance criteria

- [ ] Root cause identified: capture the actual server-side exception/stack trace for
      both ids above (prod logs, or a local repro against the same DB state) and state it
      in the PR description — not just "it now returns 200."
- [ ] `curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/91ade2b4f219fd5158f02295c21022c4` → `200` (or a documented, correct non-500 status such as `404` if the id is genuinely not gate-eligible — but not an unhandled `500`).
- [ ] `curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/d90b357a6693747d7c521d6998314bbd` → same standard.
- [ ] A regression test added under `backend/tests/` that exercises the failing code path
      (the specific branch that raised) so this can't silently regress — reference the
      hermetic-test conventions in the repo `CLAUDE.md` (mock at boundaries, no live DB
      required to run the unit test).
- [ ] `python -m pytest -m "not integration" -q` → `0 failed`.

## Verify

```bash
curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/91ade2b4f219fd5158f02295c21022c4
curl -sS -o /dev/null -w '%{http_code}\n' https://archimedes-arc.com/api/selection-bias/gate/d90b357a6693747d7c521d6998314bbd
python -m pytest -m "not integration" -q
```

## Anti-goals

- Do NOT paper over the 500 by wrapping the handler in a blanket `try/except` that
  returns a fake-passing or fake-generic verdict — that would trade "visibly broken" for
  "silently wrong," which is worse for a rigor-claim surface.
- Do NOT touch `backend/archimedes/api/strategies_routes.py` or
  `backend/archimedes/api/vaults_routes.py` as part of this fix — they are owned
  elsewhere / mid-flight. Read-only reference if needed to understand the contract.
- Do NOT change the `passes_rigor_gate` semantics or the batch `/api/selection-bias/gate`
  (no `{strategy_id}`) endpoint's behavior — this issue is scoped to the per-id route only.

dbrowneup:
Closes #1182 — root-caused from the production traceback rather than guessed.

**The gate endpoint did all its work (8.5s of real rigor computation) and then slowapi killed the response**: the handler lacked the `response: Response` parameter the decorator injects rate-limit headers into. An AST survey found the identical defect on **three more routes** — the entire quant-lab portfolio surface (`optimize`, `parameter-sweep`, `scenario-analysis`) has been 500ing on success in prod.

**Why CI never saw it:** the limiter is disabled under TESTING, so the injection path never runs in the suite. Hence the regression guard is a structural AST invariant over every route module — proven discriminating with the exact pre-fix signature *and* a mutation run.

Verification: prod reproduced before the fix (`gate/{id}` → 500 in 8.5s; batch → 200; unknown id → 404), 78 tests green across touched files, mutation flips the invariant red.

### DIFF
--- a/backend/archimedes/api/portfolio_routes.py
+++ b/backend/archimedes/api/portfolio_routes.py
@@ -12,7 +12,7 @@ import asyncio
 import logging
 from datetime import UTC, datetime
 
-from fastapi import APIRouter, HTTPException, Request
+from fastapi import APIRouter, HTTPException, Request, Response
 from pydantic import BaseModel, Field
 
 from archimedes.api.limiter import limiter
@@ -137,7 +137,7 @@ class ScenarioAnalysisRequest(BaseModel):
 
 @portfolio_router.post("/optimize")
 @limiter.limit("5/minute")
-async def optimize_portfolio(req: OptimizeRequest, request: Request) -> dict:  # noqa: ARG001 — slowapi @limiter.limit inspects param name
+async def optimize_portfolio(req: OptimizeRequest, request: Request, response: Response) -> dict:  # noqa: ARG001 — slowapi @limiter.limit inspects param name
     """Compute optimal synth-asset weights with the requested optimizer.
 
     Returns {method, optimizer, risk_profile, usdc_weight, weights, symbols_used,
@@ -210,7 +210,7 @@ async def optimize_portfolio(req: OptimizeRequest, request: Request) -> dict:  #
 
 @portfolio_router.post("/parameter-sweep")
 @limiter.limit("5/minute")
-async def parameter_sweep(req: ParameterSweepRequest, request: Request) -> dict:  # noqa: ARG001 — slowapi @limiter.limit inspects param name
+async def parameter_sweep(req: ParameterSweepRequest, request: Request, response: Response) -> dict:  # noqa: ARG001 — slowapi @limiter.limit inspects param name
     """Run a 2D sensitivity sweep over two backtest parameters.
 
     Returns a heatmap-ready grid_2d with rows=param1 values, cols=param2 values,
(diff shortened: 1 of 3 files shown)


dbrowneup:
Strategy-session review: verified the mechanism and the fix. slowapi's _inject_headers demands a Response param; without it every SUCCESSFUL call 500s after the work is done — and CI can't see it because TESTING disables the limiter, which is why ~3,000 green tests coexisted with four dead prod routes including the passport's verify endpoint (#1182, MVP-blocking). The structural AST-walk invariant with its proven-discriminating meta-test is exactly the right answer to an untestable-under-TESTING runtime path. Endorsed for Dan's morning merge — this one belongs near the front of the train.

PR #1263, opened 2026-08-18
Fix the frozen leaderboard refresh: artifact dir is read-only on Fargate

dbrowneup:
**Root cause of the stale leaderboard (newest row 2026-07-01), from prod logs — not guessed:** the scheduled refresh runs on cadence and every strategy dies at the artifact write (`PermissionError: /app/analytics-engine/artifacts/…`), which happens **before** the DB insert. Computed backtests are thrown away; the refresh then mis-diagnoses itself ("9 strategies still have no rows … backing off"). Worked on EC2 because the dir was a writable bind mount; broke silently at the Fargate cutover.

Fix: `ARCHIMEDES_ARTIFACT_DIR` env (set in ecs.tf to task-local scratch — the durable record is `backtest_results.artifact_json`, so ephemeral is honest) → writable repo dir (dev unchanged) → loud WARNING fallback. Writability probed by real file creation, not `os.access`.

Tests reproduce the incident with a genuinely read-only dir and **assert the fallback logs** — a quiet fallback is worse than a crash (the zombie-oracle lesson).

**Sequencing:** with this + #1256 deployed, the scheduled refresh self-heals the leaderboard with momentum-honest rows — the re-run happens on the box, no manual prod surgery. Needs a `terraform apply` for the env var (Dan) after merge.

### DIFF
--- a/backend/archimedes/scripts/run_backtests.py
+++ b/backend/archimedes/scripts/run_backtests.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import sys
+import tempfile
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


dbrowneup:
Strategy-session review: root cause verified against my own log diagnosis (refresh dying at the artifact write is exactly consistent with 'curated path frozen at Jul 1 rows, 9 strategies rowless, backoff'). The write-probe (os.access lies in containers — correct) and the DB-row-is-the-durable-record rationale are both sound. NOTE FOR DAN: this touches infra/ecs.tf (adds one env var to the task definition) — needs-Dan category by rule, though it is the most benign form. With this deployed + #1256 already merged, the leaderboard self-heals. Endorsed.

PR #1266, opened 2026-08-18
Hide non-MVP surfaces (vaults, marketplace, publish, subscriptions, learnings) behind one flag

### LINKED ISSUE #1122: [frontend] Wallet-gate the Market pages (Marketplace/Publish/Subscriptions) + settle Learnings nav placement
## Summary

The three Market pages — **Marketplace, Publish, Subscriptions** — are not wallet-gated: they render for anonymous visitors, unlike Generate/Library (and now Quant Lab, gated in #1072), which use the `WalletGate` wrapper in `ui/src/App.jsx`. Publish has only a soft inline "connect your wallet" message inside the component; Marketplace and Subscriptions render their full UI. These surfaces read/act on per-wallet state (subscriptions, spend caps, publish rights), so they should present the same consistent gate + passkey onboarding path as the other wallet-bound pages.

## Scope

- Wrap the `marketplace`, `publish`, and `subscriptions` routes in `App.jsx` with `WalletGate` (copy the Generate/Quant Lab pattern — pageName + honest one-sentence description + `onConnect={openConnectModal}`).
- Remove/simplify PublishPage's now-redundant inline not-connected branch.
- Decide whether `market-strategy` (marketplace detail deep-link) stays public for shareability (a public read-only detail page is defensible — subscribing already requires a wallet server-side) — pick one and say why in the PR.

## Open nav-placement questions (Dan, 2026-07-14 — decide here or split)

- **Learnings**: still aspirational. Either move it to the Strategy group or remove it from the nav until it earns its slot ("I'm not sure we can even make this useful right now" — Dan). Decision needed.
- Already handled elsewhere: Leaderboard → Strategy group ships in #1077; Quant Lab → Position group + gating ships in #1072.

## Acceptance criteria

- [ ] Anonymous visit to `/marketplace`, `/publish`, `/subscriptions` renders the WalletGate panel, not the page UI.
- [ ] Connected wallet renders all three pages exactly as before.
- [ ] `market-strategy` behavior decided + documented in the PR body.
- [ ] Learnings nav decision made + applied.

## Anti-goals

- No server-side auth changes here (SIWE gating on the API routes is separate and already partially in place — e.g. `require_verified_wallet` on marketplace mutations).

dbrowneup:
## Summary

Hides the four non-MVP UI surfaces — **vaults** (Portfolio + vault-detail), **Marketplace** (+ market-strategy), **Publish**, and **Subscriptions** — behind one build-time flag:

```js
// ui/src/featureFlags.js
export const ROADMAP_SURFACES_ENABLED =
  import.meta.env?.VITE_ROADMAP_SURFACES === "true";
```

Default (unset/`false`): the four surfaces are unreachable — not in the nav, not in breadcrumbs, not routable by flat URL or deep link, and no CTA leads to them. `VITE_ROADMAP_SURFACES=true` at build time restores the full app unchanged. **All component code stays in the tree.**

This is the rebuild on post-#1194 main that the original PR's reconciliation note promised — the note's items are now implemented rather than predicted:

- **`routes.js` `featureEnabled()`** is the single gate (item 2): a `ROADMAP_PAGES` set short-circuits it, with `features.roadmapSurfaces` as a tests-only override (`parseFeatures()` never emits the key, so the server can't flip it). Main's `deepRoutes` loop already routes through `featureEnabled()` (item 3 landed with #1194's revisions), so deep links `/app/portfolio/vaults/:addr` and `/app/marketplace/strategy/:id` are covered by the same check — no separate gating.
- **Nav needs no hand-edits** (item 4): `visibleNavigation()` drops the Portfolio item and empties the Market group via the extended `featureEnabled()`; `Layout.jsx` already skips emptied groups. `NAV` is untouched.
- **`App.jsx` render-switch guards dropped** (item 7): `navigateToPage()` on main re-derives the route via `resolveRoute()` after every push, so a stray programmatic nav to a hidden page resolves to not-found at the routing layer.
- **Breadcrumbs** (item 5): `CRUMB_MAP` untouched (keeps `test_breadcrumbs.py`'s sync invariant intact, 5/5 both flag states); the component flattens the mid-crumb when its group page is hidden, so quant/reasoning/learnings read `Home / <page>` instead of linking into a 404.
- **Re-applied by hand onto current shapes** (item 6): StrategyPassport's "Deploy as a vault" card + `CreateVaultModal`; OnboardingTour drops cards anchored to hidden surfaces (the `deploy` step's mount effect called `setPage('portfolio')` for every first-time visitor).
- **`featureFlags.js` merged, not added**: #1295 pre-created the file for `GENERATION_QUOTE_ENABLED` with a shape matching this PR; both flags now live there, env access via the node-safe `import.meta.env?.` pattern (the module is imported by `routes.js`, which loads under plain node in `ui/test/`).
- **`sitemap.xml`**: `/marketplace` dropped; the exclusion comment now records the roadmap-surface reason vs the sign-in-gated reason separately.

## Tests

`ui/test/routes.test.js` gains the handoff test: the four flat routes and both deep links resolve `not-found` by default, nav drops them for a signed-in user, and the `roadmapSurfaces: true` override restores all of it (which also keeps vault-detail's identifier-extraction coverage alive). **Mutation-verified**: deleting the `featureEnabled()` roadmap clause fails the test.

Dead-door audit re-run on current main: every remaining `onNavigate`/`setPage`/`navigateToPage` call site targeting a hidden page is either inside an already-hidden component or backstopped by the routing layer above.

## Verification

```
ui: node --test            56/56 (incl. new handoff test; mutation pass: gate removed → fail)
ui: npm run lint           clean
ui: npm run build          clean, both flag states
backend: pytest backend/tests/test_breadcrumbs.py   5/5
```

## Addendum: Learnings is hidden too

Per product direction, **Learnings** joins the hidden set (`ROADMAP_PAGES`). No CTA in the tree navigates to it (verified by grep — only the routes map and NAV item reference it), so the existing single gate covers every path; quant/reasoning keep the Position group non-empty. Tests updated (56/56, both flag-state builds clean).

Hiding supersedes #1122's wallet-gating ask — an unreachable route is strictly stronger than a gated one — and settles its Learnings-placement question.

Closes #1122

### DIFF
--- a/ui/.env.example
+++ b/ui/.env.example
@@ -34,7 +34,7 @@ VITE_CIRCLE_CLIENT_URL=https://modular-sdk.circle.com/v1/rpc/w3s/buidl
 VITE_MARKETPLACE_MIN_VAULT_FUNDS_RAW=1000000
 
 # Roadmap surfaces (#1266): vaults (Portfolio + vault-detail), Marketplace
-# (+ market-strategy), Publish, Subscriptions. Off (default) hides them
+# (+ market-strategy), Publish, Subscriptions, Learnings. Off (default) hides them
 # completely — no nav items, and their routes (deep links included) resolve
 # to not-found. Set true to preview the full app. All component code stays
 # in the tree either way.
--- a/ui/.env.example
+++ b/ui/.env.example
@@ -33,6 +33,13 @@ VITE_CIRCLE_CLIENT_URL=https://modular-sdk.circle.com/v1/rpc/w3s/buidl
 # MARKETPLACE_MIN_VAULT_FUNDS_RAW. 1000000 = 1 USDC.
 VITE_MARKETPLACE_MIN_VAULT_FUNDS_RAW=1000000
 
+# Roadmap surfaces (#1266): vaults (Portfolio + vault-detail), Marketplace
+# (+ market-strategy), Publish, Subscriptions. Off (default) hides them
+# completely — no nav items, and their routes (deep links included) resolve
+# to not-found. Set true to preview the full app. All component code stays
+# in the tree either way.
+VITE_ROADMAP_SURFACES=false
+
 # Upfront cost-quote + x402 paywall step on the Generate page (RATIFIED —
 # docs/specs/generation-quote-contract.md, #1296). Off by default: Generate
 # submits directly, no quote step. Set true to try the quote card + 402/409
--- a/ui/public/sitemap.xml
+++ b/ui/public/sitemap.xml
@@ -5,14 +5,13 @@
   anonymous visitor with no wallet connected and no selected entity.
 
   Excluded on purpose:
-    - portfolio, marketplace, publish, subscriptions — hidden roadmap
-      surfaces (#1266): their routes resolve to not-found unless the app is
-      built with VITE_ROADMAP_SURFACES=true. A static sitemap can't read a
-      build flag, so it matches the default-off production build.
-    - generate, library, quant, reasoning, learnings — each renders only a
-      generic sign-in/"Connect your wallet" prompt for a cold/anonymous
-      visitor. No unique indexable content exists at these URLs without a
-      session.
+    - portfolio, marketplace, publish, subscriptions, learnings — hidden
+      roadmap surfaces (#1266): their routes resolve to not-found unless the
+      app is built with VITE_ROADMAP_SURFACES=true. A static sitemap can't
+      read a build flag, so it matches the default-off production build.
+    - generate, library, quant, reasoning — each renders only a generic
+      sign-in/"Connect your wallet" prompt for a cold/anonymous visitor. No
+      unique indexable content exists at these URLs without a session.
     - strategy, vault-detail, market-strategy — dynamic routes that require
       an id/address in the path; there is no single canonical URL to list.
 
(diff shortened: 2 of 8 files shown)


dbrowneup:
Strategy-session review: verified the flag mechanics (conditional spread keeps hidden pages OUT of PAGE_TO_PATH at runtime — the matched-but-unrendered trap correctly avoided), the render-switch NotFound backstop, and the sitemap/crawler reasoning. The OnboardingTour deploy-card catch and the StrategyPassport Deploy-CTA gating are exactly the dead-doors class we've been burned by — good find. The #1194 reconciliation note is the right shape: fold into routes.js featureEnabled() post-rebase, drop the Layout NAV edits, repoint the breadcrumbs test; the flagged pre-existing deepRoutes bypass on #1194 goes to the contracts session. STAYS DRAFT until rebased onto #1194 after it merges, per plan. Endorsed for that sequencing.

dbrowneup:
HANDOFF RECORDED (from the contracts session, 2026-08-19): when this PR rebases onto #1194 and folds its gating into the data-driven featureEnabled(), it MUST carry the discriminating test for the deep-route gate fix (7e809bd7 on the #1194 branch): feature-disable market-strategy, resolve /app/marketplace/strategy/x, assert not-found. The fix ships untested on #1194 deliberately — today only quant is gated and it has no deep route, so any test written there passes vacuously against the pre-fix code. This PR is where the test becomes discriminating. Do not mark ready-for-review without it.

PR #1268, opened 2026-08-18
Paper trading: forward-run the graded engine, append-only ledger

dbrowneup:
MVP queue item 6, built to Dan's design with the validation he asked for — **verdict confirmed with one amendment** (full reasoning in the commit message):

- ✅ *Forward-run the backtest engine, not the live evaluator* — correct, and now structurally safe: the replay goes through the exact calls the grading path uses, so F1's fix (and every future one) applies to paper automatically, with #1267's parity harness holding the other interpreter.
- 🔧 *"Persisted to the existing daily-returns machinery"* — amended to a **sibling** of it: `strategy_daily_returns`' own documented law is wholesale-replace per re-measurement; a user's track record needs the opposite law (append-only, per-deployment, spec snapshotted at deploy). Same row shape, so every metric consumer works unchanged.

Three laws, each mutation-verified or adversarially pinned: **append-only** (idempotent advances), **drift surfaced never repaired** (upstream restatements stamp the deployment and log loudly — the ledger keeps its numbers), **the snapshot is the contract** (regenerating the strategy can't rewrite what the ledger grades).

**DRAFT because of migration sequencing:** `c9f2e8d4a1b7` chains on main's head; #1194 chains its two migrations on the same parent. Whichever merges second re-points — #1194 is the critical path, so this waits for it, then I re-point and mark ready. Tables are disjoint; nothing else blocks.

Full suite: **3,123 passed, 0 failed.**

### DIFF
--- a/backend/archimedes/api/paper_routes.py
+++ b/backend/archimedes/api/paper_routes.py
@@ -1,0 +1,120 @@
+"""Paper-trading endpoints — /api/paper/* (MVP: verdict → paper-trade).
+
+Ownership follows the SIWE model live on main today (verified wallet), with
+the Better Auth user id column already carried for the #1194 transition —
+this router only needs its auth dependency swapped when that lands.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from fastapi import APIRouter, HTTPException, Request, Response
+
+from archimedes.api.auth_siwe import get_verified_wallet
+from archimedes.api.limiter import limiter
+from archimedes.db import get_session, init_db
+from archimedes.models.paper_store import STATUS_STOPPED, PaperDeployment
+from archimedes.services.paper_trading import advance_deployment, create_deployment, deployment_summary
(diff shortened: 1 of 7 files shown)


dbrowneup:
Two commits since draft: **8ba3e3e9** — migration `c9f2e8d4a1b7` re-pointed onto merged #1194 head `b7e3f1a2c9d4` (the revision's own sequencing note, executed; chain verified serial by `test_alembic_migrations`), branch rebased to 0-behind main. **d2a79593** — routes swapped to canonical account ownership: every `/api/paper` route requires the Better Auth session; deployments owned by `owner_user_id` only (no legacy tier — the table ships with this feature); linked wallet recorded as provenance, never an access grant; deploy works with no linked wallet. Paper deploys are simulated (no funds move), so account-session gating is the correct proof strength — a future paper→real promotion takes the fresh-signature gate. Route tests mutation-verified: all three fail against the SIWE-gated version.

PR #1270, opened 2026-08-18
CI: guard against the recurring alembic migration-chain fork

dbrowneup:
## Why

The alembic migration chain has forked three times already: two branches
each add a migration whose `down_revision` points at the same parent — the
chain head each branch saw when it forked from main. Whichever merges first
becomes the new head; the other's migration is now parented on a revision
that's no longer the head, which `alembic upgrade head` can't resolve on its
own. Hit on the Aug 3 deploy, and twice on #1194.

## What

- `.github/scripts/migration_chain_guard.py` — rebuilds the migration chain
  from `backend/migrations/versions/` (found via
  `find backend -path '*versions*.py'`) and fails fast, with a re-serialize
  instruction, if a newly-added migration's `down_revision` isn't the
  current head of `--base-ref`'s chain (default `origin/main`, fetched fresh
  in CI — deliberately the ref's *current* tip, not a historical merge-base
  snapshot, so a migration that's gone stale because another PR merged
  first is still caught; see the module docstring for the reasoning). A
  separate merge-base diff (`git diff --diff-filter=ACMR <merge-base>..HEAD`)
  scopes which migration files this PR actually touched, mirroring the
  three-dot idiom `import-guard.yml` already uses in this repo.
- Wired into `quality-gate.yml` as a new `migration-chain-guard` job.
  Deliberately **not** path-filtered at the trigger level — this repo has
  already hit the failure mode where a path-filtered required check never
  reports on PRs outside its paths and sits in "Expected" forever (see the
  `infra-guards`/`engine-tests`/`import-guard.yml` comments). Instead the
  script itself early-exits in a few seconds (checkout + one git diff, no
  pip install) the moment no migration files changed vs the merge-base.

## Proof it discriminates (local runs, transcript below)

Built three synthetic scenarios as throwaway branches off this repo's real
migration chain (current real head `b41c7d9e2a05`), plus a bonus fourth:

**1. Forked migration (down_revision stale relative to the base ref) → FAIL**
```
$ python3 .github/scripts/migration_chain_guard.py --base-ref test/fake-main-with-y
FAIL: backend/migrations/versions/bbbb33334444_test_pr_migration.py has down_revision='b41c7d9e2a05', but test/fake-main-with-y's migration chain head is 'aaaa11112222'.
Someone else's migration merged to main after you branched — re-serialize:
  1. git fetch origin main
  2. edit down_revision in backend/migrations/versions/bbbb33334444_test_pr_migration.py to "aaaa11112222"
  3. from backend/: `alembic history` should show one unbroken chain, one head.
exit=1
```

**2. Correctly-chained new migration → PASS**
```
$ python3 .github/scripts/migration_chain_guard.py --base-ref origin/main
migration_chain_guard: OK — ['backend/migrations/versions/cccc55556666_test_correct_migration.py'] extend origin/main's head (b41c7d9e2a05) cleanly.
exit=0
```

**3. PR touching no migration files → untouched (fast no-op)**
```
$ python3 .github/scripts/migration_chain_guard.py --base-ref origin/main
migration_chain_guard: no migration files changed vs 7aabe5d50c6b — nothing to check.
exit=0
```

**Bonus 4. Two new migrations in the same PR both grafting off the same parent → FAIL**
```
FAIL: 2 new migrations each extend from OUTSIDE this PR's own chain (['backend/migrations/versions/dddd77778888_test_intra_a.py', 'backend/migrations/versions/eeee99990000_test_intra_b.py']) instead of one linear line — only one new migration may graft onto main; chain the rest after it.
exit=1
```

Also confirmed: `ruff format --check` and `ruff check --select E9,F63,F7,F40,F82`
both pass on the new script; the edited `quality-gate.yml` parses cleanly
(`yaml.safe_load`) with the new job present alongside the existing seven.

## Scope / omissions

- Only understands a bare-string-or-`None` `down_revision` (the standard
  `script.py.mako` shape every current migration uses). A tuple
  `down_revision` (an actual alembic *merge* migration, joining two heads on
  purpose) isn't supported — the script fails loudly with a manual-review
  message rather than silently passing, since this repo has never used one
  and the "small script" brief didn't ask for it.
- Only validates newly-added revisions (by revision id, not by file diff
  status) — editing an already-existing migration's docstring/body without
  changing its `down_revision` is correctly left alone.
- Not tested against an actual GitHub Actions run (no CI trigger available
  from this environment) — verified via direct local invocation against
  real throwaway git branches instead, as shown above.

### DIFF
--- a/.github/scripts/migration_chain_guard.py
+++ b/.github/scripts/migration_chain_guard.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Alembic migration-chain fork guard (CI).
+
+The recurring failure this exists to catch: two branches each add a new
+migration whose `down_revision` points at the SAME parent — the chain head
+each branch saw when it forked from main. Whichever merges first becomes the
+new head; the other branch's migration is now parented on a revision that is
+no longer the head, which `alembic upgrade head` cannot resolve on its own —
+a silent fork. This has bitten three times already (the Aug 3 deploy, and
+twice on #1194) before this guard existed.
+
+Two different reference points are used deliberately, not interchangeably:
+  - `--base-ref`'s merge-base with HEAD scopes step (a): "which migration
+    files did THIS PR add or change" (the standard three-dot-diff idiom used
+    elsewhere in this repo, e.g. import-guard.yml).
+  - `--base-ref` itself (its CURRENT tip, freshly fetched) is the chain those
+    new migrations must extend — the actual thing "re-serialize onto main"
+    means. Using the historical merge-base for this instead would silently
+    pass a migration that forked against a revision another PR has since
+    made obsolete, which is exactly the bug class this guard exists to catch.
+
+Usage: python .github/scripts/migration_chain_guard.py [--base-ref REF]
+Exit 0 = pass (including "nothing to check"). Exit 1 = fork / stale parent
+detected, or main's own chain isn't a clean single-head chain to begin with.
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+import subprocess
+import sys
+from pathlib import Path
+
+VERSIONS = "backend/migrations/versions"
+
+# Matches the shape script.py.mako generates:
+#   revision: str = "07e9c1489199"
+#   down_revision: str | Sequence[str] | None = "af9c6a9376e4"  (or None)
+REV_RE = re.compile(r'^revision(?:\s*:[^=\n]+)?\s*=\s*["\']([^"\']+)["\']', re.MULTILINE)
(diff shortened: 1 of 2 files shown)


dbrowneup:
Strategy-session vet: spot-ran the guard on the branch — clean fast no-op on a no-migration diff ('no migration files changed vs 7aabe5d50c6b'). The live-tip-vs-historical-merge-base design call is correct and the builder's reasoning is exactly right: the historical fork point misses the very bug this guard exists for (PR-B grafting off a parent that PR-A already superseded on main). Not-path-filtered follows the repo's own required-check convention. Discriminating transcripts (fork→fail, chained→pass, no-op, double-graft→fail) are in the PR body. Endorsed — this closes the pattern that bit three times. Dan's merge.

PR #1271, opened 2026-08-18
Implement IPortfolioConstructor.construct() — KellyRegimePortfolioConstructor (#1264)

### LINKED ISSUE #1264: [portfolio] Build IPortfolioConstructor.construct() on the existing seams — or delete its two stand-ins
## What construct() is

`IPortfolioConstructor.construct(...)` is a **frozen interface** in `backend/archimedes/interfaces/math.py` — the real asset-level portfolio optimizer: rank strategies by risk-adjusted return, optimize variance, map strategy weights → token allocations using backtest + regime inputs. It was Önder's lane; it was never implemented. Two stand-ins exist today, **both with zero production callers**:

- `services/strategy_guardrail.py` — normalize + cap + USYC floor, self-described as *"the thin strategy-level stand-in… When Önder's `construct()` lands it replaces this"* (its 20 tests are also vacuous on their own subject — all four risk-profile floor assertions pass with every floor set to 0.0, per the 2026-08-18 test audit).
- `services/_deprecated/portfolio_constructor.py` — the pre-Day-10 version, deleted in the Tranche-1 excision (#1259).

## Dan's call (2026-08-18)

Build it ourselves on what Önder has done rather than waiting — he can build on or amend what we ship.

## The assets to build on (all already in-tree, all tested)

1. **The frozen interface** — `interfaces/math.py` `IPortfolioConstructor.construct()` + `score()` contract.
2. **`services/strategy_sizer.py`** — a real, tested Kelly-weighted sizing implementation (`kelly_weighted_allocations`, `scale_to_budget`, `size_strategies`) whose only caller is `POST /api/vaults/{address}/derive-allocations`, which the UI never wired in (the deposit flow does flat equal-weighting client-side — the naive behavior the backend was built to fix). This is the closest thing to a finished `construct()` core.
3. **`RISK_PROFILE_PARAMS`** — the risk ladder, now pinned by tests (#1259).
4. **`GmmRegimeDetector` / regime tilt** — the regime input the spec (`docs/specs/execution-trading-agent-society-spec.md` step 2) says `construct()` consumes.

## Sequencing honesty

`construct()`'s consumers are the **vault-deploy/rebalance path** — out of MVP scope (vaults hidden until post-Aug-25). Recommended: land the guardrail deletion (test-audit Tranche 3) now since its tests are vacuous either way, and build `construct()` as the first post-MVP item by promoting `strategy_sizer` behind the `IPortfolioConstructor` interface. If Önder prefers a different shape, this issue is the place to say so.

🤖 Filed by the infra/backend session at Dan's direction.

dbrowneup:
## Summary

Closes the gap #1264 describes: `IPortfolioConstructor.construct()` /
`score_strategy()` (`backend/archimedes/interfaces/math.py`) were frozen but
never implemented for the vault-setup path. This PR builds them on exactly
the assets #1264 names — no new endpoints, no vault-deploy-UI changes.

- **New class**: `KellyRegimePortfolioConstructor` in
  `services/portfolio_constructor.py`, alongside (not replacing) the
  existing `PortfolioConstructor`. **Note for review**: that file already
  held a `PortfolioConstructor` class implementing the same interface for
  the execution-society's per-tick regime/consensus throttle, wired into
  `chain/agent_runner.py` and `marketplace/service.py` — issue #1264's "two
  stand-ins, zero production callers" framing predates it. This PR does not
  touch that class's behavior (13 pre-existing tests still pass unmodified)
  beyond extracting its shrink/renormalize arithmetic into a shared
  `_shrink_to_safe_asset()` helper so the new class can reuse it byte-for-byte
  rather than forking the formula.
- **Core**: promotes `strategy_sizer.py`'s `size_strategies` /
  `scale_to_budget` (real, tested Kelly sizing — untouched) behind the
  interface.
- **Risk ladder**: `RISK_PROFILE_PARAMS` — used both by the production
  `base_weights` path (via the caller) and by `construct()`'s own fallback
  weight derivation when called without `base_weights`.
- **Regime tilt**: reused via composition from
  `PortfolioConstructor.compute_position_scale` (§3.2 of
  `docs/specs/execution-trading-agent-society-spec.md` — `REGIME_MULTIPLIER`
  × confidence factor × ensemble-consensus throttle) — not re-derived. One
  deliberate, documented divergence: `regime is None` → **neutral** (scale
  1.0) here, vs. the execution society's **conservative**
  `REGIME_MULTIPLIER_NONE` (0.7) there. Rationale in the class docstring:
  `derive-allocations` has no live regime feed wired into it in this change
  (a fresh, unfed `GmmRegimeDetector` always reports `None` and would
  additionally stomp the module-level detector instance `/health` reads;
  a Redis-backed persisted-regime read was judged out of scope for a minimal
  diff). Neutral-on-None is also what makes the wiring below
  behavior-preserving for every caller today.

## Wire-up

`POST /api/vaults/{address}/derive-allocations` (`strategy_sizer`'s only
caller) now routes its Kelly-sized weights through
`KellyRegimePortfolioConstructor.construct()` as `base_weights`, with
`regime=None` (see above). Because `regime=None` → scale 1.0 → an identity
pass-through (no renormalize arithmetic at all — see the docstring for why:
avoids reintroducing float drift into a dict that already carries the
route's own basis-point rounding correction downstream), this is
byte-identical to the pre-existing output for every current caller.

Proof: the **pre-existing, unmodified**
`test_derive_allocations_sizes_passers_and_excludes_candidates` (and the
whole rest of `test_strategy_sizer.py`) still passes with zero changes to
its assertions. Additionally, `test_kelly_regime_portfolio_constructor.py`
adds a direct old-vs-new comparison test
(`test_construct_reproduces_old_strategy_sizer_pipeline_on_fixed_fixture`)
that runs the literal pre-#1264 pipeline
(`size_strategies → scale_to_budget → kelly_weighted_allocations`) on a
fixed multi-strategy/multi-asset fixture (including a gate-failer), then
routes that same output through `construct()`, and asserts dict equality.

## Tests

New file: `backend/tests/services/test_kelly_regime_portfolio_constructor.py`
(10 tests, hermetic — no env/DB/network/chain).

Per this repo's testing convention, the two properties #1264 explicitly
calls out are proven **discriminating**, not just asserted:

- `test_regime_blind_variant_fails_the_crisis_shrink_assertion` — a
  constructor subclass that ignores `regime` (constant scale=1.0) is run
  through the real CRISIS-shrink assertion inside `pytest.raises`, proving
  it fails.
- `test_equal_weight_variant_fails_the_kelly_proportionality_assertion` — an
  equal-weight scheme over the same two strategies is run through the real
  Kelly-proportionality assertions inside `pytest.raises`, proving it fails.

I also did the adversarial pass by hand before pushing (per
`CLAUDE.md`'s "a guard must be shown to reject something"): temporarily
replaced the real regime tilt with a constant, and separately replaced
`_kelly_fallback_weights` with a literal equal-weight implementation,
re-ran the suite each time, confirmed the relevant tests failed (3 tests
caught the equal-weight injection: proportionality, gate-bypass, and
profile-floor), then restored the real code and confirmed all 10 pass again.

Every assertion compares against a concrete nonzero expected value (e.g.
`0.06`, `0.92`, `0.2·(2/3)`) — the vacuous-floor trap the 2026-08-18 test
audit found in `strategy_guardrail`'s tests (floor assertions that passed
with every floor at 0.0) does not reproduce here.

### Verification run

```
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest \
  backend/tests/services/test_kelly_regime_portfolio_constructor.py \
  backend/tests/services/test_portfolio_constructor.py \
  backend/tests/test_strategy_sizer.py -q
# 39 passed
```
Also ran (non-hermetic-gate, broader regression sweep): `test_strategy_sizer.py`,
all of `tests/services/`, `test_agent_runner_tick.py` (agent_runner's
`PortfolioConstructor` caller), `tests/marketplace/` (the other production
caller), vault/rigor test files — **and the full repo suite**:
`3127 passed, 5 skipped` (skips pre-exist, unrelated). `ruff format --check`
and `ruff check` clean on all three changed/added files.

## MVP rationale (Dan, 2026-08-19)

Per #1264: Dan pulled this into the MVP sprint rather than waiting —
"build it ourselves on what Önder has done rather than waiting; he can
build on or amend what we ship." This PR is exactly that: it does not touch
`strategy_guardrail.py` (a separate tranche per #1264) or any vault-deploy
UI, and adds no new endpoints.

## Requesting review

@onder-akkaya — this was your lane (`interfaces/math.py` marks you reviewer
for `IPortfolioConstructor`). Please treat this as a starting point to build
on or amend, not a final answer — in particular the regime-tilt formula
reuse, the `regime=None` neutral-default divergence from the execution
society's class, and the fallback-weights-without-`base_weights` design are
all judgment calls documented in the class docstring that you may want to
revisit. Cites #1264 throughout.

## Omissions / explicit scope cuts

- No live regime feed is wired into `derive-allocations` — `regime=None`
  always, in this PR. `construct()` itself is fully regime-aware and tested
  against realistic `RegimeClassification` fixtures; wiring a live feed
  (Redis-persisted regime, or similar) is left as follow-up.
- `ensemble_consensus` is likewise always `None` from this route today.
- `strategy_guardrail.py` deletion (test-audit Tranche 3, per #1264) is
  intentionally not part of this PR.
- Did not touch `services/_deprecated/portfolio_constructor.py` or
  `services/_deprecated/kelly_portfolio.py` — both explicitly deprecated,
  out of scope.

Never merged by me — draft, for review.

### DIFF
--- a/backend/archimedes/api/vault_schemas.py
+++ b/backend/archimedes/api/vault_schemas.py
@@ -99,3 +99,16 @@ class SetAllocationsResponse(BaseModel):
         default_factory=list,
         description="Selected strategies sized to zero (rigor-gate CANDIDATE/fail, or no stored kelly_fraction)",
     )
+    # Literal default rather than an import of
+    # services.portfolio_constructor.REGIME_CONVENTION_NEUTRAL_NO_FEED: this
+    # schemas module stays free of service imports. The route always sets the
+    # value explicitly from the constructor, so this default only covers a
+    # caller constructing the response directly.
+    regime_convention: str = Field(
+        default="neutral_no_feed",
+        description=(
+            "Whether the regime tilt actually ran: 'regime_tilt_applied' (a live regime scaled these "
+            "weights) or 'neutral_no_feed' (no regime input, weights are pure Kelly sizing). Without "
+            "this, an un-tilted allocation and a regime-tilted one that scored 1.0 look identical."
+        ),
+    )
--- a/backend/archimedes/api/vaults_routes.py
+++ b/backend/archimedes/api/vaults_routes.py
@@ -615,11 +615,14 @@ async def derive_vault_allocations(
     # documented neutral default (see KellyRegimePortfolioConstructor's
     # docstring) — so this is a behavior-preserving pass-through of
     # `kelly_weights` for every caller of this endpoint today.
+    # One `regime` value feeds both the arithmetic and the disclosed marker, so
+    # the two can never report different things (the #1409 rf_convention rule).
+    regime = None
     allocs = _portfolio_constructor.construct(
         risk_profile=RiskProfile(req.risk_profile),
         strategies=strategies,
         backtest_results={},
-        regime=None,
+        regime=regime,
         ensemble_consensus=None,
         base_weights=kelly_weights,
     )
@@ -662,4 +665,5 @@ async def derive_vault_allocations(
         risk_profile=req.risk_profile,
         sized_strategies={sid: frac for sid, frac in sized_fractions.items() if frac > 0.0},
         excluded_strategy_ids=excluded,
+        regime_convention=_portfolio_constructor.regime_convention(regime),
     )
--- a/backend/archimedes/api/vaults_routes.py
+++ b/backend/archimedes/api/vaults_routes.py
@@ -30,11 +30,18 @@ from archimedes.api.vault_schemas import (
 )
 from archimedes.chain.executor import chain_executor
 from archimedes.models.chat import VaultMetadata
+from archimedes.models.portfolio import RiskProfile
 from archimedes.services.live_rigor_gate import verdicts_for_strategies
+from archimedes.services.portfolio_constructor import KellyRegimePortfolioConstructor
 from archimedes.services.rigor_profiles import STRICTEST_LEVEL, clamp_level
 from archimedes.services.strategy_sizer import kelly_weighted_allocations, scale_to_budget, size_strategies
 
 vaults_router = APIRouter(prefix="/api/vaults", tags=["vaults"])
+
+# Module-level singleton (matches strategy_evaluator's convention below) — the
+# class is pure computation (no I/O), so one shared instance is safe across
+# requests. See KellyRegimePortfolioConstructor's docstring (issue #1264).
+_portfolio_constructor = KellyRegimePortfolioConstructor()
 
 
 async def _anchor_strategies_async(strategy_ids: list[str]) -> None:
@@ -600,7 +607,23 @@ async def derive_vault_allocations(
     sized_fractions = size_strategies(strategies, req.risk_profile, deployable_ids=deployable_ids)
     sized_fractions = scale_to_budget(sized_fractions, 1.0 - usdc_floor)
     excluded = sorted(sid for sid, frac in sized_fractions.items() if frac <= 0.0)
-    target_weights = kelly_weighted_allocations(all_signals, sized_fractions, usdc_floor=usdc_floor)
+    kelly_weights = kelly_weighted_allocations(all_signals, sized_fractions, usdc_floor=usdc_floor)
+
+    # Route the Kelly-sized weights through IPortfolioConstructor (#1264) so
+    # this endpoint gains the regime tilt for free once a live regime feed is
+    # wired to it. No live feed is wired here today — regime=None is the
+    # documented neutral default (see KellyRegimePortfolioConstructor's
+    # docstring) — so this is a behavior-preserving pass-through of
+    # `kelly_weights` for every caller of this endpoint today.
+    allocs = _portfolio_constructor.construct(
+        risk_profile=RiskProfile(req.risk_profile),
+        strategies=strategies,
+        backtest_results={},
+        regime=None,
+        ensemble_consensus=None,
+        base_weights=kelly_weights,
+    )
+    target_weights = {a.symbol: a.weight for a in allocs}
 
     allocations: list[AllocationTarget] = []
 
--- a/backend/archimedes/services/portfolio_constructor.py
+++ b/backend/archimedes/services/portfolio_constructor.py
@@ -59,6 +59,15 @@ REGIME_MULTIPLIER: dict[Regime, float] = {
 # fetch failed or lacked VIX/MA signals — see agent_runner._classify_market_regime).
 REGIME_MULTIPLIER_NONE = 0.7
 
+# ── Regime provenance markers (#1264 review) ────────────────────────
+# `KellyRegimePortfolioConstructor` treats `regime is None` as neutral (scale
+# 1.0), so an inert tilt and a benign regime that scores 1.0 produce identical
+# weights. These name which one happened, so a consumer can tell them apart.
+# Same reasoning as `rf_convention` in `services/rf_series.py` (#1409): a
+# fail-soft default has to be a disclosed state, never a silent substitute.
+REGIME_CONVENTION_NEUTRAL_NO_FEED = "neutral_no_feed"
+REGIME_CONVENTION_APPLIED = "regime_tilt_applied"
+
 # Confidence floor: even at confidence 0 the regime multiplier is not fully
 # erased — low confidence halves the regime's effect rather than zeroing it.
 # scale_within_regime = _CONF_BASE + _CONF_SLOPE * confidence.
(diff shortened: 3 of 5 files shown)


dbrowneup:
Strategy-session vet (pre-Önder): verified with my own commands — (1) the 'important discovery' is REAL: PortfolioConstructor has two live production callers on main (marketplace/service.py:211, chain/agent_runner.py:185), so issue #1264's 'zero production callers' framing was wrong; the builder correctly added alongside without touching the live class (its 13 tests pass unmodified). (2) Interface conformance confirmed against the frozen interfaces/math.py — score_strategy, correctly (the task brief itself misnamed it score(); the builder checked the premise). (3) 26/26 tests pass on the branch including the old-vs-new behavior-compat comparison. The regime=None → neutral-1.0 (vs the execution society's 0.7) design call is exactly the kind of thing Önder should rule on — it's flagged in the docstring and PR body. Endorsed for Önder's review, then Dan.

onder-akkaya (approved):
Quant-lane review. #1264 asked for the shape ruling as much as the correctness one, so: the shape is right. Promoting `strategy_sizer` behind the interface and reusing the regime tilt by composition instead of forking its constants is what I would have done. No amendment wanted.

The Kelly core is untouched, which is the correct amount of touching. I re-read it anyway: the passport fraction is already half-Kelly and doesn't get double-discounted, the profile multiplier is 2/γ capped at 1.0 so `hyper_risky` can't climb above the stored half-Kelly, gate-failers size to zero, and nothing levers up to fill a budget.

I did check whether `_kelly_fallback_weights` can produce a levered book, because `weights[SAFE_ASSET] = round(max(0.0, 1.0 - sum(...)), 6)` is exactly the shape that swallows an overflow: clamp the safe leg at zero and the rest of the dict can quietly sum past 1.0. It can't happen here. `scale_to_budget` only ever scales down, `usyc_floor` tops out at 0.70, so the remainder is always non-negative and the dict lands on 1.0. Fine as written, though that line could use a one-line comment saying so, for the next reader who arrives with the same suspicion.

I mutation-tested the None-regime divergence, since a deliberate 1.0-vs-0.7 split in a sizing multiplier is precisely the thing a well-meaning refactor "cleans up" two months from now. Deleting the override so it delegates (and therefore returns 0.7) fails 6 tests. Properly pinned.

One ask, non-blocking.

**Disclose that the regime tilt was inert.** `derive-allocations` currently returns weights from a class called `KellyRegimePortfolioConstructor` whose Regime half did nothing, and a consumer has no way to tell that apart from "the regime was benign, so the scale came out 1.0". Those are very different statements and the response hands back the same bytes for both.

I'm not arguing for the 0.7. Your reasoning against silently shrinking a preview by 30% is right and I'd have made the same call. The fail-soft rule isn't "shrink anyway", it's that the degraded state should be visible instead of a plausible-looking substitute. You built this on #1424 last week with `rf_convention`, for this exact reason.

`SetAllocationsResponse` already has the block for it, and the convention is pre-established:

```python
     # Kelly-sizing transparency (additive; defaults keep old consumers working):
     risk_profile: str = "moderate"
+    regime_convention: str = "neutral_no_feed"
```

One field, same additive-default pattern as its neighbours, and the day someone wires a live feed it starts saying something else on its own.

One observation, no action needed.

The "13 pre-existing tests still pass unmodified" proof of the `_shrink_to_safe_asset` extraction is softer than it reads. I deleted the renormalization from the shared helper outright and got 23 of 23 green: when the input weights sum to 1.0 the freed-mass accounting already conserves the total, so that line is a no-op on every input any test supplies. The extraction is faithful, I read it against the original line by line. But the suite wouldn't have caught it if it weren't, which is worth knowing before leaning on it as the safety argument. It only bites if a caller passes unnormalized `base_weights`, and the Kelly path technically can, since `kelly_weighted_allocations` rounds to 4dp.

Approving. 39 tests green on the branch (10 new, 13 pre-existing constructor, 16 sizer). Take the response field here if it's quick, otherwise file it and I'll pick it up.

onder-akkaya:
Took my own suggestion and pushed it here (`1b7c80cc`) rather than leaving you a chore. Revert if you'd rather shape it differently.

`regime_convention` on `SetAllocationsResponse`, sourced from a new `KellyRegimePortfolioConstructor.regime_convention()` instead of hardcoded at the call site, so it flips to `regime_tilt_applied` by itself the day a live feed arrives. The route now holds one `regime` local feeding both `construct()` and the marker, so the weights and their provenance can't drift apart.

The case that convinced me it was worth more than a docstring: `Regime.RISK_ON` has multiplier 1.0, so a real regime at full confidence produces the exact same scale as no feed at all. There's a test pinning that now (`test_regime_convention_says_applied_even_when_the_tilt_scores_neutral`) — same weights out, and the marker is the only thing telling them apart.

Constructor tests fail before the change on ImportError. Your existing route test picks up one added assertion and none of its prior ones move, so the pass-unmodified argument for the #1264 wiring still stands. Full `pytest -m "not integration"`: 3700 passed, 0 failed.

Approval stands.

Issue #868, opened 2026-07-03
[quant] Rigor gate 0/31 root cause: degenerate series poison the denominator + look-ahead false-positives + leaderboard reads stale fixtures


dbrowneup:
## Summary
Dogfooding (#854 finding 3) showed **0 of 31 library strategies pass the rigor gate** on the live site, killing the generate→gate→deploy spine for every visitor. A root-cause investigation (2026-07-03, live API + code replay) found this is **three distinct problems compounding**, not one bug — and the gate is partly *right*:

1. **14 strategies have real, non-degenerate returns and honestly fail DSR p≥0.95** (p range 0.003–0.85; holds even at num_trials=1). Not a bug — possibly a framing/copy question (Önder's call, see #822/#819 context).
2. **11 strategies carry zero-variance daily_returns** (placeholder-flat series): DSR/OOS correctly return None — but these degenerate series still **inflate `num_trials` (25 vs ~14 true)** and **dilute `avg_correlation`**, unfairly stiffening DSR for the 14 real ones (`selection_bias_routes.py:176-183` filters only `len(v)>=10`, no variance check).
3. **6 strategies show nonzero Sharpe/CAGR on the leaderboard with ZERO persisted daily_returns** (`is_backtest_placeholder=false` despite no rows) — a data-path bug (id mismatch suspected, esp. the 4 duplicate Pairs Trading entries; leaderboard reads stale `s.dsr_p_value`/fixture fields at `strategies_routes.py:117-135` while only `passes_rigor_gate` uses the live verdict).

Plus a **look-ahead false-positive class**: `look_ahead_audit` flags ANY negative-index subscript, but backtrader's `self.data.close[-i]` is bars-AGO access (e.g. `moreira_muir_2017_volatility_managed.py:99-100`) — standard, not look-ahead — and `passes_all` requires zero warnings → hard FAIL.

## Scope
- `backend/archimedes/api/selection_bias_routes.py` — exclude zero-variance series from `valid_returns` BEFORE computing num_trials / avg_correlation / cohort PBO (mirror `_rigor_helpers._sharpe_dsr_inputs`'s `np.ptp(arr)==0` test).
- `backend/archimedes/services/rigor_evaluator.py` (`look_ahead_audit`) — skip/whitelist backtrader `self.data.*[-i]` bars-ago access; keep genuine future-index detection.
[...]

Issue #821, opened 2026-06-29
[backend] Library passes_rigor_gate badge must come from the live gate, not a fixture boolean


onder-akkaya:
## Summary

The user-facing library badge `passes_rigor_gate` (and the CANDIDATE → VALIDATED 🏆 promotion)
is served on one path from a **stored fixture boolean**, not a live `run_rigor_gate` verdict.
In `strategy_provider.py` (~line 322), `passes_rigor_gate = bool(fx.get("passes_rigor_gate",
False))` reads a hardcoded value from `analytics-engine/strategies/backtest_fixtures.json`, and
that same boolean drives the status promotion (~line 290).

The current fixture values **are internally consistent** with the gate thresholds — the two
passing strategies (`moreira_muir`, `moskowitz_ooi_pedersen`) genuinely clear DSR (p ≥ 0.95),
PBO (< 0.5) and OOS (> 0), and the failing rows fail at least one — so there is **no fabricated
alpha today**. The issue is the *pattern*: a cached boolean that the live gate never re-derives.
If a fixture ever drifts from what `run_rigor_gate` would compute on the strategy's real
returns, the library would display "passes rigor" (and 🏆 VALIDATED) without the gate agreeing.
This is exactly the cached-pass pattern our #1 rule forbids and the one Bogdan's audit (#710)
called out.

Note the honest counter-path: the `/selection-bias/gate` route **does** recompute live from DB
returns and refuses to synthesize returns from stubs — so the machinery to do this right
already exists; this issue is about making the **library list** use it too.

## Proposed resolution

- Derive `passes_rigor_gate` (and the VALIDATED promotion) on the library-list path from the
  **live** `run_rigor_gate` output — either compute on read, or persist the gate's output to the
  DB and read that, but **forbid the fixture default** as a source of truth for the badge.
- Where no real returns exist yet (genuinely pre-backtest), surface an explicit
  `pending`/`unknown` state rather than defaulting the boolean to a fixture value.

## Scope (exact files)

- `backend/archimedes/services/strategy_provider.py` (~lines 290, 322 — the fixture read +
  status promotion).
- `backend/archimedes/api/strategies_routes.py` (the list serializer that surfaces the badge).

## Acceptance criteria

- [ ] The library-list `passes_rigor_gate` for a real strategy equals the live `run_rigor_gate`
      verdict on its persisted returns (test).
- [ ] A strategy with no real returns surfaces a `pending`/`unknown` badge, not a fixture
[...]

PR #1272, opened 2026-08-18
Fix M1/M2: serve LIVE rigor verdicts to the advisor, the portfolio LLM, and vault chat

dbrowneup:
## What

Fixes the two remaining **M-class audit findings** from the 22-agent test-suite/chaff audit (2026-08-17): every LLM surface and the `/advisor` response now serve **live rigor-gate verdicts** — never the frozen fixture statistics and never the fail-closed sentinel presented as a verdict.

**M1 — `/advisor` served five frozen rigor statistics.** The route live-corrected only the badge; `deflated_sharpe_ratio`, `dsr_p_value`, `num_trials_in_selection`, `pbo_score`, and `out_of_sample_sharpe` came from the fixture fields on the in-memory `Strategy` objects — values the live gate never wrote, which could disagree with `GET /api/selection-bias/gate` for the same strategy at the same moment. The route now runs the **same memoized live-gate batch the library list uses** (`_live_rigor_results_for_strategies`, the #868 machinery — same cohort context, same cache entry) and serves badge + numbers from that **single computation**. Fixture values remain only as the documented #868 fallback when the live gate cannot run (no/insufficient persisted returns).

**M2 — the portfolio LLM and vault chat were told every curated strategy fails rigor.** `Strategy.passes_rigor_gate` on in-memory provider objects is a fail-closed sentinel (**always `False`**, #821 — correct at that layer, so no fixture boolean can leak). Two downstream readers presented the sentinel as a verdict:

- `portfolio_agent._format_strategies` rendered `rigor=candidate` for every strategy in the portfolio-construction prompt (both the single-turn and tool-use paths, `/advisor`'s agent calls);
- `chat_service._build_vault_context` rendered `Rigor gate: not passed` for every curated strategy in the vault-chat context.

The prompt builders now take a live tri-state status map (`pass`/`fail`/`pending`) computed by the route **before** the LLM call. Vault chat batches the same memoized live gate and, when no live status is available, **omits the rigor line entirely** — per that builder's own documented law, "missing data is omitted, never invented."

## Why it matters

The #1 rule: claims must be true on the live path. An LLM constructing portfolios while told the entire curated library failed the gate is reasoning from a fabricated premise; an advisor response whose numbers disagree with the gate route undermines the rigor wedge that is the product's core claim.

## Verification (adversarial pass)

Every stub strategy in the new tests carries a **poisoned `passes_rigor_gate` that contradicts the live status**, so a sentinel-reading mutant renders the wrong label and fails. Mutation-verified by reverting each source file to `main` and re-running its tests:

| Reverted file | Result |
|---|---|
| `strategies_routes.py` | both advisor tests FAIL — the numbers test caught the route serving `0.545957` (the actual fixture DSR) instead of the live sentinel value; the statuses test caught `price_histories` arriving where the status map belongs |
| `portfolio_agent.py` | all 3 prompt tests FAIL (`rigor=candidate` / unearned `rigor=PASS` from the poisoned attribute) |
| `chat_service.py` | all 9 chat tests FAIL (invented `Rigor gate: not passed`; helper missing) |

**Local suite**: `python -m pytest -m "not integration" -q` from repo root — **3133 passed, 3 skipped (pre-existing: chain-settings defaults ×2, fee-cap parity awaiting #1129), 0 failed** in 3m33s.

## Notes for review

- **Latency**: the advisor previously ran `verdicts_for_strategies` (unmemoized) per request; it now shares the library list's `rigor_cache` entry, so a warm cache makes this *faster*, and the batch runs in a thread (`asyncio.to_thread`) instead of blocking the event loop as before.
- **Consistency**: the advisor badge now comes from the exact batch the library list serves, so the two surfaces can no longer disagree on a verdict (previously they used different cohort filtering).
- **`num_trials_in_selection` now serves `1`** for live-graded curated strategies — self-contained per the num-trials decouple (curated = 1 trial), replacing whatever the fixture froze. This is the correct convention, not a regression.
- **stockbench eval adapter** calls `propose_portfolio` without a status map → its prompts now say `rigor=pending (no live verdict)` instead of `rigor=candidate`. Honest for an eval harness; flagging for awareness.
- **Scope**: the audit finding named the five gate statistics. Other fixture-sourced display fields (`sharpe_ci_*`, `n_obs_daily`, `real_*` metrics) are unchanged — separate surface, separate decision.
- **Service→api import** in `chat_service._curated_rigor_statuses` follows the existing precedent (`live_rigor_gate._load_strategy_code_safe`); local import, fail-safe to `{}`.
- No migration, no contract changes, no UI changes. No overlap with #1194's hunks in these files (verified against its diff).

### DIFF
--- a/backend/archimedes/agents/portfolio_agent.py
+++ b/backend/archimedes/agents/portfolio_agent.py
@@ -140,8 +140,19 @@ def _format_market_scan(market_ranking: list[dict]) -> str:
     return "\n".join(lines)
 
 
-def _format_strategies(strategies: list[Any]) -> str:
-    """Render paper strategies with their backtest stats + signal rules."""
+def _format_strategies(strategies: list[Any], rigor_statuses: dict[str, str] | None = None) -> str:
+    """Render paper strategies with their backtest stats + signal rules.
+
+    ``rigor_statuses`` maps strategy id → LIVE tri-state gate status
+    ("pass" | "fail" | "pending"), computed by the caller (the advisor route
+    runs the same memoized live-gate batch that serves the library badge).
+    ``Strategy.passes_rigor_gate`` on the in-memory object is a fail-closed
+    sentinel (always ``False``, #821) and must never be presented to the LLM
+    as a verdict — before this parameter existed, every curated strategy was
+    labeled ``candidate``, so the model was systematically told the whole
+    library had failed rigor. A missing map or id renders ``pending``:
+    "verdict not computed" is honest; "candidate"/"fail" from a sentinel is not.
+    """
     rule_summary = {
         "faber": "long when price > 200-day SMA, else flat",
         "moreira": "scale exposure by target_vol / realized_vol (vol-managed)",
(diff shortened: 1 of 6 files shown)


dbrowneup:
Strategy-session vet: (1) The no-overlap claim is VERIFIED at the level that matters — file-level overlap exists (strategies_routes.py, chat_service.py shared with #1194), but git merge-tree --write-tree of this branch against feat/account-auth-app-boundary completes with ZERO conflicts, so the hunk-level claim holds and merge order doesn't matter. (2) 42/42 targeted tests pass (chat_service + portfolio_agent_prompt). (3) The law in chat_service.py:62 — 'missing data is omitted, never fabricated' — is the fail-soft principle written as code, which is exactly right; M2's old behavior (rendering the fail-closed sentinel as 'the whole library failed' to both LLMs) was the plausible-substitute anti-pattern in its purest form. Endorsed for Dan's train — slots cleanly before or after #1194.

PR #1273, opened 2026-08-18
[docs] ADR amendments: fixture-leak class (num-trials) + honest engine census and #1242 provenance (backtrader)

dbrowneup:
## What

The two ADR amendments left for this session, unblocked by the docs restructure (#1225).

**`num-trials-self-containment.md`** — records the **fixture-leak class**: the `num_trials_convention` marker protects verdicts computed by the gate, but not surfaces that serve rigor values from the frozen strategy fixtures — most of which predate the 2026-07-09 reversal, so serving them beside a live badge silently mixes conventions without touching the marker. The live instance was audit finding M1 (`/advisor`'s five fixture statistics, fixed in #1272), with M2 as the sentinel variant. The amendment states the rule: served rigor values come from a live `run_rigor_gate` computation (or its honest memoized batch); fixture fields only as the documented #868 fallback; the fail-closed sentinel is never a verdict.

**`backtrader-backtest-engine.md`** — records the engine census as it actually stands (`backtrader` | `dsl-fusion` | `portfolio-simulator-v1`, each with its distinct role — verified against the code sites that write each tag), `portfolio-simulator-v1`'s post-audit disposition (dead clusters deleted by #1259, in review; surviving surface is exactly `backtest_portfolio` + `sensitivity_sweep`), and the #1242 provenance columns (`cost_model_id`, `look_ahead_audit_source`) extending per-result provenance beyond the engine tag.

## Notes

- Doc-only; no code. `make docs-check` passes (staleness warnings pre-existing).
- Deliberately **no pass counts or row counts** anywhere — per the no-numbers-in-docs rule, live-DB questions stay live-DB questions.
- Amendments are additive sections with dated front-matter updates — no rewriting of the decision record.

### DIFF
--- a/docs/adr/backtrader-backtest-engine.md
+++ b/docs/adr/backtrader-backtest-engine.md
@@ -2,7 +2,7 @@
 
 > **Audience:** Archimedes team
 > **Status:** Accepted
-> **Date:** 2026-05-13 (recommended); ratified in the Day-3 sync — exact ratification date not recorded in git [unestablished — needs Dan]; closed as Accepted 2026-07-28
+> **Date:** 2026-05-13 (recommended); ratified in the Day-3 sync — exact ratification date not recorded in git [unestablished — needs Dan]; closed as Accepted 2026-07-28; amended 2026-08-19 (engine census + provenance columns)
 > **Owner:** Dan Browne
 > **Supersedes:** —
 > **Superseded-by:** —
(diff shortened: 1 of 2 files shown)


dbrowneup:
Strategy-session vet: endorsed. This amendment is the reconciliation I requested on #1242 — the psv1 'live caller vs structurally dead' contradiction resolves cleanly: the dead clusters were the Phase-2.2 rigor-overlay helpers + capacity_decay (deleted with their tests in #1259), the surviving portfolio-simulator-v1 surface is exactly backtest_portfolio + sensitivity_sweep, recorded as a companion surface rather than a rival grading engine. The three-engine census now lives where it belongs — in the ADR the discriminator descends from. The '[unestablished — needs Dan]' ratification-date mark is the right discipline. Docs-only, joins the train anywhere.

PR #1274, opened 2026-08-18
Tranche 5: the 22 funds-adjacent tests the deletions fund (G1/G2/G3)

dbrowneup:
## What

**Test-audit Tranche 5** — the +22 funds-adjacent gap-fill tests that the Tranche-1 deletions (#1259) were meant to fund. Purely additive: no source changes, three test surfaces. The audit's headline was *"the suite is thickest where nothing ships, and thinnest where money leaves"* — this PR is the thin part getting its first real tests, ahead of the Sept 16 real-USDC cutover.

### G1 — the live money-out paths (9 tests, `test_settlement.py`)

`withdraw_subscriber` measured **4/18 statements** covered, `withdraw_publisher` **4/16** — only the `PAYMENTS_DRY_RUN` early-returns and guard clauses had ever executed. The balance read, executor construction, and ERC-20 transfer had never run under test. New:

- Stage C live: the exact `PaymentSplitter.withdraw(bytes32,uint256)` calldata from the publisher's own executor; refusal without a configured splitter; soft-fail on chain error (never raises).
- The subscriber DCW return-transfer proven on the three claims its docstring makes: the amount **equals the balance read** (never caller-supplied), the token is the configured USDC contract, the destination is the subscriber's **SIWE wallet — not the DCW** — and the executor is bound to `(circle_wallet_id, dcw_address)`.
- Empty-DCW and missing-identifier short-circuits (no executor is ever built).
- **The recoverable-by-retry promise**: a failed sweep leaves no sticky state — the retry re-reads the balance and transfers the full amount again.

### G2 — Circle DCW provisioning (9 tests, new `test_wallet_provisioner.py`)

`_create_circle_wallet` measured **1/28**, `_find_wallet_by_ref` 1/12, `_encrypt_entity_secret` 1/7 — every subscriber's custodial wallet is minted in code that had effectively never run under test. New hermetic suite (URL-routing fake session, no network):

- **The entity-secret encryption is proven, not mocked**: a test RSA keypair decrypts the posted `entitySecretCiphertext` and requires the original secret bytes back — a mutant that skips encryption, encrypts the wrong bytes, or uses the wrong padding cannot pass.
- Idempotency keys are deterministic per ref and distinct across refs (the property that makes retries mint zero duplicates).
- The 409 double-provision guard surfaces the **existing** wallet via metadata-ref lookup — one creation attempt, no duplicate mint; 409-with-no-findable-wallet raises.
- Credential absence fails **before** any network I/O; non-201/409 statuses raise with the status code.

### G3 — the fee's unit scale (4 tests, new `test_fee_unit_scale.py`)

Every existing charge test re-derived its expectation from the live `FLAT_FEE_PER_ACTION` symbol, and every `tick()` test monkeypatched the charge away — a 10⁶ unit-scale regression would have shipped through a fully green suite at 100% file coverage. New:

- The literal pin: `FLAT_FEE_PER_ACTION == 100` (raw 6-dec µUSDC = $0.000100/action) and `fee_to_price` dollar-string literals including the `10⁶ raw = $1.000000` anchor.
- The **real `_charge_one`** running with `payments_dry_run=False` against a stubbed payment client: asserts the debited `flat_fee_raw` is the literal 100, the spend-cap **reservation and release amounts match it exactly** (same wallet, same charge id — a mismatch silently leaks the 24h cap), and the charge bills the subscriber's Circle wallet pair to the publisher's seller address.

## Verification

**Mutation pass** (each mutation applied to source, tests run, source restored):

| Mutation | Result |
|---|---|
| Refund destination `to_wallet` → `dcw_address` (money returned to the custodial wallet) | `test_withdraw_subscriber_returns_exact_balance_to_the_siwe_wallet` FAILS |
| `_encrypt_entity_secret` encrypts `b"\x00"*32` instead of the secret | the real-decrypt test FAILS on plaintext mismatch |
| `FLAT_FEE_PER_ACTION` default `"100"` → `"100000000"` (10⁶ scale slip) | 3 tests FAIL (`100000000 == 100`); the `fee_to_price` literal test correctly still passes |

**Coverage on the target modules** (marketplace test dir): `settlement.py` → **91%**, `wallet_provisioner.py` → **98%** (from 23%). Full unit suite result posted below.

## Notes for review

- Additive only — no source files touched, no collisions with any open PR (verified against the diffs of #1194/#1258/#1259/#1263/#1266/#1267/#1268/#1270/#1271).
- Remaining audit gaps deliberately NOT here: G8 is Tranche 4 (HELD for #1194 — file collision), G10's skip→FAIL conversion waits for #1129's merge, G5/G6/G7/G9/G11–G14 are the next tranche of this series.
- The `FLAT_FEE_PER_ACTION == 100` literal pins the shipped default; an env override in a dev shell will fail it loudly — that is the point (CI is hermetic).

## Full suite

`python -m pytest -m "not integration" -q`: **3141 passed, 3 skipped (pre-existing), 0 failed** in 3m26s.

### DIFF
--- a/backend/tests/marketplace/test_fee_unit_scale.py
+++ b/backend/tests/marketplace/test_fee_unit_scale.py
@@ -1,0 +1,128 @@
+"""G3: pin the fee's UNIT SCALE with literals, and let the real ``_charge_one`` run.
+
+Audit finding (2026-08-18): every existing charge test re-derived its
+expectation from the live ``FLAT_FEE_PER_ACTION`` symbol and every ``tick()``
+test monkeypatched the charge away — so a 10⁶ unit-scale regression would have
+kept the whole suite green at 100% file coverage. The literals below are the
+only assertions in the repo that break when the SCALE (not the logic) moves.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+from archimedes.marketplace.service import FLAT_FEE_PER_ACTION, MarketService, Publisher, Subscriber
+from archimedes.marketplace.tick_registry import TickStep
+
+
+def test_flat_fee_per_action_is_100_raw_microusdc():
+    # 100 raw 6-decimal USDC units = $0.000100 per action. If this fails,
+    # either the price changed on purpose (update this literal AND the
+    # fee_to_price literals below together) or a unit-scale regression is
+    # trying to ship. Deliberately a literal — never re-derive from the
+    # constant under test.
+    assert FLAT_FEE_PER_ACTION == 100
+
+
+def test_fee_to_price_pins_the_six_decimal_dollar_scale():
+    from archimedes.marketplace.payments import fee_to_price
+
+    assert fee_to_price(1, 100) == "$0.000100"  # one action at the shipped fee
+    assert fee_to_price(3, 100) == "$0.000300"
+    assert fee_to_price(0, 100) == "$0.000000"  # zero-action tick is free
+    assert fee_to_price(1, 1_000_000) == "$1.000000"  # 10^6 raw = exactly one dollar
+
+
+def _live_service_with_one_sub():
+    svc = MarketService(interval_seconds=9999, payments_dry_run=False, paper_trading=True)
+    svc._claim_settlement_intent = MagicMock(return_value="claimed")
+    svc._finalize_settlement_intent = MagicMock()
+    pub = Publisher(
+        strategy_id="strat_fee",
+        pool_id="0x" + "aa" * 32,
+        vault_address="0xpub_vault",
+        creator_wallet="0xpublisher",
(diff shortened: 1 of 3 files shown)


dbrowneup:
Strategy-session vet: verified tests-only (+498/-0, three test files, zero source changes), 33/33 pass on the branch, and the load-bearing claim checks out with my own read — test_wallet_provisioner.py:115-121 decrypts the posted ciphertext with a real in-test RSA keypair through the actual OAEP/MGF1-SHA256 path, so the entity-secret encryption is proven rather than mocked. Covering money-out paths that had only ever executed their dry-run guards (4/18 and 4/16 statements) is precisely the pre-Sept-16 hardening the real-pricing commitment demands. Endorsed for Dan's train — additive, slots anywhere.

Issue #941, opened 2026-07-05
MEDIUM: Circle _poll_transaction scans only newest 50 txs — false timeout


onder-akkaya:
## Problem Statement

Circle `_poll_transaction` scans only the newest 50 transactions globally. If >50 txs are in flight, a real tx can fall off the window before polling finds it, causing a false timeout → a successful rebalance reported as failed, with a SKIP/error trace published.

## Scope

- `backend/archimedes/chain/circle_signer.py` lines 201-231

## Acceptance Criteria

- [ ] Implement pagination or filter by `walletId` to ensure the target tx is found
- [ ] Test: submit >50 txs from the wallet, verify the first ones are still found (don't timeout)

PR #1275, opened 2026-08-18
Tranche 5b: verify-button tamper rejection, real raw-key fallback, two lying tests fixed (G6/G11/G13)

dbrowneup:
## What

Second batch of the test-audit gap-fills (follows #1274). Tests only — one assertion rewrite, one rename, ~13 new tests. No source changes, no open-PR overlap.

### G6 — the verify button never saw a tampered trace

Only "no hash → False" and "exact match → True" existed. New: **an on-chain hash off by one bit fails verification** — the mechanism proving an AI rebalance decision's reasoning wasn't altered after the fact. Mutation-verified: a prefix-comparing mutant (`stored_hash[:31] == expected[:31]`) passes the existing exact-match test and dies on the new one. Plus: the scan reaches an older matching trace past newer non-matching ones; empty-vault and RPC-failure fail closed; and the four chain-read helpers behind `/verify` (previously ~1 statement each: `get_trace_by_id`, `get_traces_by_vault`, both counts) get their first real happy-path + fail-soft coverage.

### G11 — the raw-key signing fallback had two lying tests

This is the code that keeps anchoring alive during a Circle outage.

- `test_executor_reads.py:308` asserted `tx == sent.hex() if cond else "0x" + sent.hex()` — Python precedence makes that `assert (X if cond else <truthy string>)`, so had `HexBytes.hex()` ever changed its prefix behavior the assert would pass **regardless of tx**. Now an unconditional exact assert, with the trap documented in place.
- `test_anchor_falls_back_to_raw_key` set `agent_account = None` — **skipping the very path its name promised**. Renamed honestly (`test_anchor_circle_down_and_no_account_returns_none`), and the real fallback now has a real test: the built transaction's `from`/`nonce`/`chainId` fields, `sign_transaction` receiving the built tx, `send_raw_transaction` receiving the **signed** payload — mutation-verified: broadcasting the unsigned tx instead fails the test — and the return equal to `sent.hex()` exactly.

### G13 — adjudicated: already covered

The audit flagged `_poll_transaction`'s pageSize=50 window. Verified: the #941 fix (walletIds scoping) **already has a real test suite** (`test_circle_signer.py` — scoping, terminal states, timeout). The only delta added is the `pageSize=50` pin, so the one-wallet >50-in-flight bound stays a conscious documented decision rather than an accident waiting to be edited.

## Verification

82/82 tests pass across the four touched files. Mutation table:

| Mutation | Result |
|---|---|
| `verify` compares 31-byte prefixes instead of full hashes | `test_verify_rejects_a_hash_off_by_one_byte` FAILS |
| raw-key path broadcasts the **unsigned** tx | `test_anchor_raw_key_fallback_signs_and_broadcasts_for_real` FAILS |

Remaining audit gaps after this: G5/G12/G14 (rigor-gate semantics — quant-sensitive, next batch), G7 (vault share accounting), G9 (reveal-fails-after-trade — needs investigation first; may become an issue), G8 (HELD for #1194), G10 (waits on #1129).

### DIFF
--- a/backend/tests/chain/test_circle_signer.py
+++ b/backend/tests/chain/test_circle_signer.py
@@ -198,6 +198,11 @@ class TestPollTransaction:
         assert result == "0xHASH"
         called_url = session.get.call_args.args[0]
         assert "walletIds=wallet-uuid" in called_url
+        # Audit G13: pageSize=50 *within one wallet* is the accepted, documented
+        # bound — a single agent wallet with >50 in-flight txs would still fall
+        # off the page. Pinned here so a silent pageSize change is a conscious
+        # decision, not an accident.
+        assert "pageSize=50" in called_url
 
 
 # ── sign_and_broadcast ────────────────────────────────────────
--- a/backend/tests/chain/test_executor_reads.py
+++ b/backend/tests/chain/test_executor_reads.py
@@ -305,7 +305,12 @@ class TestVaultSettersRawKey:
         with patch("archimedes.chain.executor.circle_signer") as signer:
             signer.is_configured = False
             tx = asyncio.run(executor.set_target_allocations("0xVault", [STSLA], [10000]))
-        assert tx == sent.hex() if sent.hex().startswith("0x") else "0x" + sent.hex()
+        # Unconditional exact assert (G11): the old form
+        # `assert tx == sent.hex() if cond else "0x" + ...` parsed as
+        # `assert (X if cond else <truthy string>)` — had HexBytes.hex() ever
+        # dropped its 0x prefix, the assert would have silently passed on the
+        # bare string regardless of tx.
+        assert tx == sent.hex()
 
     def test_set_token_oracles_no_account_raises(self, executor):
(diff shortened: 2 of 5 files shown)


dbrowneup:
Added the G9 batch to this PR: `TestRevealFailureAfterTradeExecuted` pins the honest-degradation contract when a reveal fails after the trade executed (loud ERROR, trace persisted unverified, dangling commit preserved; mutation-verified — hardcoding `is_verified: True` fails both tests). Adjudication: the loud signal EXISTS in current code; what's missing is a reconciliation path that retries the dangling commitment — filed as #1276 rather than invented here.

dbrowneup:
Strategy-session vet: 96/96 pass across all five test files on the branch. The G9 honest-degradation contract (reveal-fails-after-trade → loud ERROR, trace persisted unverified, dangling commit preserved, is_verified mutant killed) is the right shape — and filing the missing reconciliation path as #1276 instead of inventing a repair is exactly the discipline. Endorsed.

PR #1277, opened 2026-08-18
Tranche 5c: G5 — the gate's first adversarial end-to-end rejection

dbrowneup:
## What

The audit's G5: **no deliberately-overfit strategy had ever been rejected end-to-end by `run_rigor_gate`** — every composite-gate test either hand-built the `RigorGateResult` with pre-chosen numbers or fed i.i.d. noise never designed to fool it. This is the named repeat failure class (*guards need an adversarial pass*), applied to the product's central guard.

Two tests, one file, no source changes:

- **The adversary**: the classic overfit signature — stellar in-sample (the first 70% the gate grades as IS), OOS collapse, reported as best-of-40 (`num_trials=40`, its own pool per the self-containment convention). Runs through the full gate with look-ahead attested clean and PBO satisfiable, so **statistics must do the rejecting** — and they do: `passes_all` False, `min_passing_level` None (no slider level rescues it), look-ahead PASS, no MISSING inputs, the OOS floor/cliff as the attributed rejector.
- **The honest twin** (anti-vacuity): a stationary series with the same overall character does NOT trip the overfit rejectors — the test detects the overfit *shape*, not a gate that rejects everything. (Seed 42 deliberately: seed 7's raw noise happens to draw a genuinely weak sample whose own OOS tail annualizes to −1.4 — honestly rejectable, wrong fixture for an anti-vacuity check. Comment in the test records this.)

## Verification

Mutation: hardcoding `_passes_profile` to `True` (the trivial acceptor) fails **exactly** the rejection test while the honest twin still passes. Full `test_rigor_evaluator.py`: 200/200 green.

## Adjudications from the same audit batch (no code here)

- **G12** (`dsr_convention == 'excess'`): adjudicated as a **finding, not a test** — there is no risk-free rate anywhere in the rigor stack, so the audit's raw-vs-excess straddle test is unwritable against current code. Filed as a claim-integrity issue (see linked issue) for the quant lane to resolve (subtract a T-bill proxy, or relabel honestly).
- **G14** (factor thesis tests asserting `isinstance` + `bars`): confirmed real (`analytics-engine/tests/test_factor_strategies.py:135-172`) — the rewrite needs the four factor-strategy internals open and quant review, so it goes to the next batch rather than being rushed here. G7 (vault share accounting, `_deployable_levels` 1/37) likewise.

### DIFF
--- a/backend/tests/test_rigor_evaluator.py
+++ b/backend/tests/test_rigor_evaluator.py
@@ -2198,3 +2198,78 @@ class TestLoadDailyReturnsStore:
         store, vintage = load_daily_returns_store()
         assert set(store.keys()) == {"solo"}
         assert vintage == "2026-07-03"
+
+
+# ─── G5 (audit 2026-08-18): the adversarial end-to-end rejection ─────
+
+
+class TestAdversarialOverfitRejection:
+    """A deliberately-overfit strategy must be rejected END-TO-END by the full
+    gate — no hand-built verdicts, no pre-chosen numbers.
+
+    The named repeat failure class: guards need an adversarial pass. Every
+    prior composite-gate test either constructed the RigorGateResult directly
+    with chosen numbers or fed i.i.d. noise never designed to fool the gate.
+    Mutation-verified against the trivial acceptor: hardcoding
+    ``_passes_profile`` to True fails the rejection test here while the
+    honest-twin test still passes.
+    """
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


dbrowneup:
Strategy-session vet: 200/200 pass on the branch including the new adversarial end-to-end rejection — the gate's first constructed-overfit kill with attribution, which is demo-grade evidence as well as a test. Endorsed.

Issue #1049, opened 2026-07-08
[architecture] Model selection should be user-driven, not a box .env default (LLM_BEDROCK_MODEL)


dbrowneup:
## Problem
Today the LLM model is pinned by a **server-side env default**: the prod box `.env` sets `LLM_PROVIDER=bedrock_converse` + `LLM_BEDROCK_MODEL=amazon.nova-micro-v1:0`, and every request uses that one model. Dan wants **the user to choose their model**, not have it hardcoded in the deployment env.

## Why this matters now
- It's a product limitation (one-size-fits-all model), and it collides with the premium-model tier (#723 `PREMIUM_MODELS_ENABLED`) and the BYOK/payment strategy (users paying for better models).
- It's also an operational smell: a per-deployment env var determines a product-facing capability. Moving to Fargate (#1041) forced us to replicate `LLM_BEDROCK_MODEL` into the task def just to preserve current behavior — a good moment to redesign instead.

## Direction (to refine)
- Keep a **safe server default** (nova-micro, cheap) for anonymous/free use, but let an authenticated user **select a model** per generation/chat, subject to entitlement (free tier → the default; premium/BYOK → the allowlist in `PREMIUM_MODELS_ALLOWLIST`, gated by `PREMIUM_MODELS_ENABLED` and payment).
- Thread the chosen model through the request (API param) into `llm_backend.make_llm_backend(...)` rather than reading it purely from env at construction time.
- The env var becomes the *default/fallback*, not the *only* option.

## Related
- #723 (premium-model gating mechanism), payment/BYOK strategy, #1041 (which had to carry `LLM_BEDROCK_MODEL` into the Fargate task def as an interim). Not a merge-train blocker — a deliberate product change to schedule after the deploy migration.

dbrowneup:
Verified against current `main` (26fa2fef). The concrete ask — "thread the chosen model through the request (API param) into `llm_backend.make_llm_backend(...)` rather than reading it purely from env" + "the env var becomes the default/fallback, not the only option" — is satisfied for generation, its motivating example:

- `POST /api/generate/start` accepts `model`; `enforce_model_entitlement` (#723, `model_gate.py`) rejects a non-entitled premium pick with 402 (never silently downgraded); `is_allowed_model` (`llm_backend.py`) filters to the free-tier allowlist (`FREE_TIER_MODELS`, 1:1 with `ui/src/data/modelPricing.json`'s `works_now:true` rows) as defense-in-depth (#748).
[...]

PR #1279, opened 2026-08-19
[test] Lock in end-to-end model-selection threading (#1049)

dbrowneup:
## Summary

Audited #1049 against current `main`. Its premise — "every request uses [the box .env's] one model" — predates the fix: #723 (T1.8 paid-tier gating) and #748 (functional model selection) landed on 2026-06-25/26, ~2 weeks before this issue was filed, and together already implement what the issue asks for:

- `ModelCostPanel` (Generate page) sends the user's pick as `model` on `POST /api/generate/start`.
- `enforce_model_entitlement` (`model_gate.py`) rejects a non-entitled premium request with 402 — never a silent downgrade — and stays authoritative (unchanged here).
- `is_allowed_model` filters to the free-tier allowlist (`FREE_TIER_MODELS`, kept 1:1 with `ui/src/data/modelPricing.json`).
- `run_generation` threads the picked model into every live LLM call site: the debate proposer pool (`StrategyFusion`), debate rounds, and the per-job `PortfolioAgent`, all via `make_llm_backend(model=...)`.
- `LLM_BEDROCK_MODEL` is now purely the fallback default, not the only option — the issue's core ask.
- Premium (Anthropic) models stay `works_now:false` / gated pending Bedrock activation (T3.8) — an entitled premium pick is accepted by the 402 gate but still can't serve yet, so it correctly falls back to the env default. Left exactly as-is per the task constraint.

**No production code changes.** I could not find a live selection-path bug after tracing the wiring through `generate_routes.py` → `generation_pipeline.py` → `debate_engine.py` → `strategy_fusion.py` / `llm_backend.py`, and the existing 133-test suite across the model-gating/selection files passes clean.

## Gap found + closed

Existing tests stop at two boundaries: `test_generate_model_gating.py` mocks out the pipeline task entirely at the API route, and `test_debate_engine.py::test_model_pick_threads_into_served_model` covers only the `StrategyFusion` layer in isolation. Nothing proved the full chain — what the job queue actually invokes (`run_generation`) — threads a picked model down to the live debate proposer *and* the per-job agent's backend construction, with correct `served_model` provenance on the `done` event.

Added `test_run_generation_threads_user_model_to_debate_proposer_and_agent` (`backend/tests/services/test_generation_pipeline.py`) to close that gap. It's hermetic (spies on `_propose_pool` and `make_llm_backend`, no live LLM/AWS creds needed) and discriminating — verified adversarially by independently breaking each seam (dropping `model=` from the debate-runner partial; dropping it from the per-job agent's backend construction) and confirming the test fails both ways before confirming it passes clean on `main`.

## Test plan
- [x] `pytest backend/tests/services/test_generation_pipeline.py backend/tests/test_generation_pipeline.py backend/tests/test_debate_engine.py backend/tests/test_generate_model_gating.py backend/tests/test_model_gating.py backend/tests/services/test_llm_backend_model_select.py backend/tests/services/test_chat_service.py` — 133 passed
- [x] New test verified to fail when either threading seam is broken (adversarial check), passes clean on `main`
- [x] `ruff check` / `ruff format --check` clean on the changed file

Ref #1049.

### DIFF
--- a/backend/tests/services/test_generation_pipeline.py
+++ b/backend/tests/services/test_generation_pipeline.py
@@ -950,6 +950,147 @@ async def test_debate_dispatch_end_to_end_persists_debate_strategy(tmp_path, mon
     assert isinstance(verdict["passing"], bool)
 
 
+# ── Test 3b — user's Generate-page model pick reaches the live LLM call sites
+# (issue #1049: selection must thread end-to-end, not just to the API boundary)
+
+
+@pytest.mark.asyncio
+async def test_run_generation_threads_user_model_to_debate_proposer_and_agent(tmp_path, monkeypatch):
+    """``run_generation(model=...)`` on the LIVE debate dispatch path must reach:
+
+      (a) the debate proposer (``_propose_pool`` → ``StrategyFusion(model=...)``
+          → ``make_llm_backend(model=...)``) — the seam that actually drafts the
+          candidate text, and
+      (b) the per-job ``PortfolioAgent``'s backend, built via
+          ``make_llm_backend(model=...)`` — the seam ``_served_model_for`` reads
+          for the ``done`` event's ``served_model`` provenance.
+
+    Route-level tests (test_generate_model_gating.py) stop at the enqueue
+    boundary (the pipeline task is mocked out entirely); the debate-engine unit
+    test (test_debate_engine.py::test_model_pick_threads_into_served_model)
+    covers only the ``StrategyFusion`` layer in isolation. Neither proves the
+    full chain from ``run_generation`` — what the job queue actually invokes —
+    down to the LLM-backend constructors. This closes that gap: a caller that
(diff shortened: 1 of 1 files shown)


dbrowneup:
First-pass review: **endorse**. Verified the two seam claims against code: `generation_pipeline.py:880` builds the per-job agent via a function-local `make_llm_backend(model=model)` import (so patching the defining module intercepts the real call), and `_propose_pool(brief, model, corpus)` matches the spy's positional signature — a wiring miss fails assertion (a)/(b) loudly rather than passing vacuously. The layering note (SF internals covered by the debate-engine unit test; this closes the run_generation→seams chain) is honest, and the fake-backend rationale (credential-less real factory would degrade to CannedBackend and make assertion (c) vacuous) is exactly right. CI green. Second pass: strategy session.

Issue #1044, opened 2026-07-07
[architecture] Two first-class deployment modes: local (single-user, ollama, self-contained) vs production (multi-user, AWS-split)


dbrowneup:
## Goal

Two first-class deployment modes with the **same core**, differing only by config — no code forks:

- **Local mode** — single-user, self-contained, `docker compose up` and go. Runs on **ollama** (or any local OpenAI/Anthropic-compatible endpoint). No AWS, no Circle keys, no chain keys required. Generate / backtest / rigor-gate / chat / UI all work against the in-stack Postgres + Redis. Multi-user, marketplace, on-chain, and payments features are inert or stubbed.
- **Production mode** — multi-user, split across AWS services (Fargate backend + Aurora + ElastiCache + runners + contracts), full feature set. This is where the multi-user / marketplace / on-chain aspects are actually exercised.

The core product (generate → backtest → rigor gate → explore) must be **virtually identical** in both. The split should be **config-driven** (env + compose profiles), not `if LOCAL:` branches scattered through the code.

_Sequencing: this is deliberately deferred. Fix the live production deployment first (Fargate migration #1039/#1041, runners #1043, identity #1028/#1042), then come back and rebuild local mode to this contract._

## The contract

| Dimension | Local | Production | Selector |
|---|---|---|---|
| DB + cache | in-stack postgres/redis | Aurora + ElastiCache | `COMPOSE_PROFILES=localdb` (already works) |
| LLM | ollama / local endpoint | Bedrock (Nova/Haiku) | `LLM_PROVIDER` |
| Runners (oracle/agent/kb) | off by default | on, exactly-once | `COMPOSE_PROFILES=...,runners` (see #1043) |
| Secrets | local `.env` only | SSM param store | `AWS_SSM_PATH_PREFIX` |
| Payments / Circle | dry-run / absent | live (post-#975) | `PAYMENTS_DRY_RUN` + Circle keys |
| Chain | absent or local testnet key | Arc testnet, agent signer | chain keys present/absent |
| Identity ledger | activates on first SIWE auth (fail-soft) | same | — (not a mode difference) |

## Leaks to fix (each makes "local" quietly touch prod today)

> The **runners-not-gated** leak is already being fixed in **#1043** (adds `profiles:["runners"]`). This issue covers the rest.

[...]

Issue #1043, opened 2026-07-07
[infra][chain] Relocate & harden the funds-adjacent runners (oracle/agent/kb) for the Fargate split — exactly-once lease-lock + cutover-ordering fix


dbrowneup:
## Why this issue exists

#1039 / #1041 move the **backend + web tier** to ECS Fargate (build-in-CI → ECR → pull). They deliberately **do not touch the three background runners**, which today run as separate `docker compose` services *on the same EC2 box*:

| Runner | Module | Cadence | Funds-adjacency |
|---|---|---|---|
| `oracle_runner` | `archimedes.chain.oracle_runner` | every `ORACLE_INTERVAL_SECONDS` (60s) | **HIGH** — pushes prices on-chain; two live copies = conflicting/duplicate price feed |
| `agent_runner` | `archimedes.chain.agent_runner` | every `AGENT_INTERVAL_SECONDS` (300s) | **HIGH** — rebalances vaults + commits trace hashes; two live copies = **double trade** |
| `kb_runner` | `archimedes.services.kb_runner` | interval (arXiv ingest, ~6 GB weights) | low — corpus ingest; two copies risk OOM + manifest lost-update |

These are **exactly-once singletons**. #1041's own cutover runbook (Phase 8) notes a future runner-Fargate step is required *before* the EC2 box can be decommissioned — but nothing implements it, and there is **zero exactly-once protection today**. This issue closes that gap.

## The core risk (verified)

`agent_runner`'s rebalance is **not idempotent**. `_compute_trades()` emits **absolute** swap amounts from a point-in-time snapshot (`agent_runner.py:808-843`), and `chain_executor.execute_trades()` passes them straight to `Vault.rebalance(tokensIn, amountsIn, tokensOut, amountsOut)` with no per-cycle replay guard. If two copies run concurrently and fetch sequential nonces, **both trades land** — the vault is rebalanced by the same delta twice (real slippage + weight overshoot), not a benign duplicate.

[...]

Issue #1039, opened 2026-07-07
[architecture] Deploy & compute: build-in-CI → ECR → ECS Fargate (Option B; coordinated with #1028)


dbrowneup:
> Human-owned architecture decision (not any agentic path). Sibling to the data-architecture epic #1028 — the two are tightly coupled (see § Coupling). Locked by Dan 2026-07-06 after a measured cost/performance/reliability analysis.

## Why now

Build-on-the-box is the root fragility. `deploy.yml` SSMs the production host to `git reset --hard` + `docker compose build` + `up` — the image is only ever built **at deploy time, on the host that is also serving traffic**. Measured consequences on **2026-07-06 alone**:

- **3 production outages**, all OOM (not hardware — `StatusCheckFailed = 0` over 14d). A `docker compose build` on the 2 vCPU / **4 GB** t3.medium overruns memory on top of the running stack; the OOM killer takes out the SSM agent + backend; the site 504s until a **manual reboot**. A "successful" deploy still wedged the box ~9 min later.
- **No rollback**: the old image is pruned to free disk and there is no registry, so a bad boot is an outage, not a red check.
- Full incident chain: #1001.

This is **not a cost decision** — it is cost-neutral. Real spend (Cost Explorer, credits stripped, trailing 14d) is **~$178/mo**, of which compute is only ~$36 (≈20%); an ALB (~$16) and WAF (~$16) are already paid for. Moving EC2 → Fargate holds total spend within a few dollars. It is a **reliability + performance + scope** decision.

Analysis artifact (measured cost + performance per option): the v3 infra analysis (Cost Explorer + 14d CloudWatch).

## Decision

**Option B — ECS Fargate**, deployed from images **built in CI and pushed to ECR**, running behind the **existing ALB** (`archimedes-backend-tg`, target-type `ip` — already exactly what Fargate registers into). EKS rejected (a $73/mo control plane + k8s ops for a single-service app at ~0.07 req/s). Bigger-EC2 rejected (keeps the single-box, no-HA, no-zero-downtime ceiling).

**build-in-CI is not redundant with Fargate — it is its prerequisite.** Fargate does not build images; it pulls pre-built images from ECR. The pipeline is one thing:

```
GitHub Actions: build + validate + push → ECR   →   ECS Fargate: pull + run (rolling)
```

## Scope

[...]

PR #1280, opened 2026-08-19
fix(local-mode): honest Ollama probe + prod-only SSM secret load (#1044 eval)

dbrowneup:
Evaluation pass on #1044 (issue's own instruction: "sequencing deliberately deferred... come back and rebuild local mode to this contract"). Per Dan's ask, this evaluates rather than big-bangs: audited what actually breaks a fresh `docker compose up`, fixed what's cheap, left the rest as a gap list.

## Gap list (file:line)

| # | Gap | Status |
|---|---|---|
| 1 | `OllamaBackend.available` hardcoded `True` — `backend/archimedes/services/llm_backend.py:410` (pre-fix). Unreachable/misconfigured ollama reported live via `/health` (`main.py:695`) until the first real request errored. | **Fixed this PR** — real `GET {base_url}/api/tags` probe, checks the configured model is pulled. |
| 2 | `.env.example` Option (D) ollama recipe had no `LLM_MODEL` (falls back to `DEFAULT_MODEL="claude-sonnet-4-…"`, not a real ollama tag) and pointed `LLM_BASE_URL` at `localhost:11434`, which is the *container* from inside the compose network, not the host running ollama. | **Fixed this PR** — documents `LLM_MODEL=llama3.1` + `host.docker.internal`; `docker-compose.yml` maps `host.docker.internal:host-gateway` on `backend`/`oracle`/`agent` (the services that call `make_llm_backend()`). |
| 3 | `load_ssm_secrets()` ran unconditionally at `main.py:31` import, before any production check exists. `.env.example:383` defaulted `AWS_SSM_PATH_PREFIX=/archimedes/prod/` with an unenforced "leave blank for local dev" comment. On a dev box with ambient AWS creds, local `docker compose up` could pull real prod secrets (Circle keys, encryption key, LLM tokens) into the local process. | **Fixed this PR** — gated behind `PUBLIC_DOMAIN` (same signal already used for the docs gate + `EMAIL_ENCRYPTION_KEY` fail-close); `.env.example` default now blank. Belt-and-suspenders: the gate closes the leak regardless of what `AWS_SSM_PATH_PREFIX` resolves to (left `docker-compose.yml`'s own `${AWS_SSM_PATH_PREFIX:-/archimedes/prod/}` default alone — current prod is ECS/Fargate and sets this explicitly via `infra/ecs.tf:487`, bypassing the compose default entirely; touching it risked an unverifiable side effect on whatever the legacy EC2 box still does). |
| 4 | Image-pull footgun (`docker-compose.yml`'s `image:` + `build:` on the app-tier services): confirmed empirically (throwaway compose file, bogus registry) that a bare `docker compose up` with no local image tries to **pull first**, falling back to build only on failure — so an ECR-authenticated dev on a fresh clone risks running the last-pushed prod image against local code. | **Not fixed** — already substantially mitigated: `make up`, README.md:27/33/44, and SETUP.md:3/47/171/174/179 all document `docker compose up -d --build` as the canonical entry point. Residual risk is only a dev bypassing the documented entrypoints with ECR read access. A real guard (e.g. a self-healing `pull_policy` toggle mirroring how `deploy.yml` self-heals `COMPOSE_PROFILES`) touches the same file that *is* the box's live prod deploy target for the `runners`-profile fix (#1039/#1001's OOM-on-serving-host lineage) — riskier than this PR's scope without a way to verify against the real ECR-authenticated path. Left as a documented residual, not silently dropped. |
| 5 | Runners not gated behind a profile | **Already fixed** — `#1043` landed; `docker-compose.yml` already has `profiles: ["runners"]` on `oracle`/`agent`/`kb-runner`. |
| 6 | Identity ledger "prod-only" framing | **Clarification only, no bug** (per issue's own item 4) — vault creation is unconditionally SIWE-gated in both modes; fail-soft. No action taken. |
| 7 | `docker-compose.production.yml` adopt-or-delete | **Still undecided**, per the issue's own 2026-07-08 comment thread. Out of scope here — not touched. |
| 8 | Better Auth sidecar on `origin/feat/account-auth-app-boundary` (checked per task instructions) | **New finding, not fixed** (different branch/PR, not mergeable into this one). That branch adds an `auth` service + `docker-compose.local.yml` (`COMPOSE_FILE=docker-compose.yml:docker-compose.local.yml`) for strict local migration ordering, and requires `BETTER_AUTH_SECRET` with no default (`:?Set BETTER_AUTH_SECRET in .env`) even for local `docker compose up` — consistent with this repo's existing "required secret, no default" pattern (`POSTGRES_PASSWORD`, `DATABASE_URL`), but it's a new local-setup step (`python -c "import secrets; print(secrets.token_urlsafe(48))"`) that whoever picks up that branch should fold into the local-mode contract once it merges. `REQUIRE_SIWE_FOR_GENERATION` / `WALLET_LESS_GENERATION_DAILY_CAP` are removed on that branch in favor of an unconditional Better Auth session requirement — i.e. local dev loses the current open, wallet-less Generate flow. Flagging for whoever owns that merge, not resolving here. |
| SQLite/`DATABASE_URL` fail-loud (issue's 2nd comment) | **Not evaluated this pass** — out of scope for the cheap-fixes bar; needs its own look at `db.py`'s fallback path + the `nonroot`-writable-path fix already noted as done in #1050. |

## What's in this PR
- `backend/archimedes/services/llm_backend.py` — `OllamaBackend.available` real probe.
- `backend/archimedes/main.py` — `load_ssm_secrets()` gated on `PUBLIC_DOMAIN`.
- `.env.example` — ollama recipe (`LLM_MODEL`, `host.docker.internal`); `AWS_SSM_PATH_PREFIX` blank default.
- `docker-compose.yml` — `extra_hosts: host.docker.internal:host-gateway` on `backend`/`oracle`/`agent`.
- Two new hermetic test files: `backend/tests/services/test_ollama_backend_health.py` (probe honesty, both directions — unreachable/unpulled → canned, reachable/pulled → live) and `backend/tests/test_main_ssm_prod_gate.py` (subprocess test with a spied `boto3`, proving SSM is never touched locally even with a prod-shaped prefix, and still fires in production mode).

## Verification
- Full backend suite: `pytest` — 3127 passed, 5 skipped (pre-existing, DB/contract-dependent), 0 failed.
- `ruff check` + `ruff format --check` on all changed files: clean.
- `docker compose -f docker-compose.yml --profile localdb --profile runners config` parses cleanly with the new `extra_hosts` entries.
- Adversarial pass on the new guard: confirmed `available` is `False` for both an unreachable server AND a reachable server missing the configured model tag, and `True` only when both hold (exact tag and bare-name-vs-`:variant` match).

## Omissions (explicit, not silent)
- No local-vs-prod one-page doc (issue's acceptance criteria item 3) — out of scope for "small fixes."
- No fix for the image-pull-without-`--build` footgun (see gap #4) — documented residual, judged higher-risk than the fix bar for this pass.
- `docker-compose.production.yml` adopt-or-delete and the Better Auth sidecar branch are flagged, not resolved.

Related: #1043 (merged), #1039, #1041, #1028, #1042.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -89,9 +89,16 @@ REDIS_URL=redis://redis:6379/0
 #        LLM_API_KEY=<key>
 #        LLM_BASE_URL=https://api.openai.com/v1
 #
-#  (D) Local Ollama:
+#  (D) Local Ollama — no key, no cloud creds. Run `ollama pull llama3.1` on
+#      the HOST first (`available` now probes GET {base_url}/api/tags and
+#      falls back to canned output if the model isn't pulled — issue #1044).
+#      From INSIDE the docker-compose stack, "localhost" is the container
+#      itself, not your host — use host.docker.internal (mapped for the
+#      `backend`/`oracle`/`agent`/`kb-runner` services below via extra_hosts).
+#      Running the backend bare (not in compose) can use localhost instead.
 #        LLM_PROVIDER=ollama
-#        LLM_BASE_URL=http://localhost:11434
+#        LLM_MODEL=llama3.1
+#        LLM_BASE_URL=http://host.docker.internal:11434
 #
 # Default matches production: AWS Bedrock Converse + Amazon Nova Micro.
 LLM_PROVIDER=bedrock_converse
@@ -392,9 +399,12 @@ KB_ARTIFACT_DIR=/srv/corpus-artifact
 
 # ── AWS SSM Parameter Store — production secrets surface ──────────────────
 # Set by TS.2 (issue #176). Backend reads all params under this prefix at
-# startup via boto3.client("ssm").get_parameters_by_path(...). Leave blank
-# for local dev (uses .env values directly).
-AWS_SSM_PATH_PREFIX=/archimedes/prod/
+# startup via boto3.client("ssm").get_parameters_by_path(...) — but ONLY when
+# running in production mode (PUBLIC_DOMAIN set; see main.py), so a local run
+# can never pull real prod secrets even with ambient AWS creds (issue #1044).
+# Leave blank for local dev (uses .env values directly). Production sets this
+# explicitly in its own deploy config (infra/ecs.tf), not from this file.
+AWS_SSM_PATH_PREFIX=
 
 # ── x402 payment gateway webhook secret ──────────────────────────────────
 # HMAC-SHA256 shared secret for authenticating x402 gateway callbacks to
--- a/backend/archimedes/main.py
+++ b/backend/archimedes/main.py
@@ -23,12 +23,18 @@ from slowapi.errors import RateLimitExceeded
 load_dotenv("../.env", override=True)  # Project root .env first (has real secrets)
 load_dotenv(".env", override=False)  # Backend-local .env fills in any missing (no override)
 
-# Load secrets from AWS SSM Parameter Store (production).
-# No-op when AWS_SSM_PATH_PREFIX is unset (local dev). Must run BEFORE
-# any service imports that read os.environ for API keys / secrets.
+# Load secrets from AWS SSM Parameter Store — PRODUCTION ONLY. Gated on
+# PUBLIC_DOMAIN (same "is this production" signal used below for the docs gate
+# and the EMAIL_ENCRYPTION_KEY fail-close) so a local run can never pull real
+# prod secrets from SSM, even with ambient AWS creds and AWS_SSM_PATH_PREFIX
+# resolving to a real path — issue #1044. Belt-and-suspenders with
+# AWS_SSM_PATH_PREFIX defaulting blank in .env.example: this gate is what
+# actually stops the fetch regardless of what that var resolves to. Must run
+# BEFORE any service imports that read os.environ for API keys / secrets.
 from archimedes.services.secrets_service import load_ssm_secrets
 
-load_ssm_secrets()
+if os.getenv("PUBLIC_DOMAIN"):
+    load_ssm_secrets()
 
 # Shared rate limiter (Redis-backed, falls back to in-memory).
 # Defined in a separate module to avoid circular imports with route modules.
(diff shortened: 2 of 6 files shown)


dbrowneup:
First-pass review: **endorse with one required fix**.

**Required — subprocess hermeticity** (`test_main_ssm_prod_gate.py`): the harness inherits `dict(os.environ)` and the subprocess's own `load_dotenv("../.env", override=True)` (main.py line 23) re-reads the repo-root `.env` **over** the test's env — so a dev box whose `.env` sets `PUBLIC_DOMAIN` (exactly what `.env.example:337`'s comment suggests for prod-mimicry) flips the local-mode assertion: local-red/CI-green, the class the testing conventions ban. Use the codified pattern: `_clean_subprocess_env()` + `_DOTENV_NEUTRALIZE` with an explicit env whitelist (reference: `test_security_hardening.py`, per CLAUDE.md § Testing conventions).

**Verified safe** (so the second pass needn't re-derive): the `PUBLIC_DOMAIN` gate cannot break prod — `infra/ecs.tf:507-510` sets both `AWS_SSM_PATH_PREFIX` and `PUBLIC_DOMAIN` on the backend container; the EC2 runners never call `load_ssm_secrets()` at all (`runner_ec2.tf:87-89` documents user-data as their SSM path), so the gate is inert for them; `.env.example` ships `PUBLIC_DOMAIN=` blank. The count-the-boto3-client-calls design (observing the actual SSM touchpoint rather than inferring from exceptions) is the right shape, and the both-directions coverage (local never / prod still does) avoids the overcorrection trap.

Ollama probe + compose `extra_hosts` changes look correct and behavior-preserving; the not-overcorrect test (`reachable+pulled → OllamaBackend`, not Canned) is the adversarial counterpart done right. Second pass: strategy session.

Issue #1218, opened 2026-08-03
[data] yfinance is an unlicensed commercial dependency on the MVP critical path — price the replacement


dbrowneup:
## Summary

**`yfinance` is free, unlicensed for commercial redistribution, and — since the #1203 routing fix — sits on the MVP critical path.** Every strategy now pulls its own declared universe rather than a single shared SPY feed, so market-data volume scales with strategies × symbols × backtest re-runs.

This is an unpriced input with a cliff, and **the cliff lands at commercialization** — which is to say exactly when revenue starts. That timing is the problem: the cost appears at the same moment the business does.

## The specifics

- Yahoo Finance data is accessed via an unofficial scraping interface. `yfinance`'s own README disclaims affiliation and points at Yahoo's terms.
- Yahoo's terms do not permit commercial redistribution of the data.
- We are a testnet product with no revenue today, so nothing is being violated yet. **The moment we charge for a verdict derived from that data, the posture changes.**
- There is no rate-limit SLA, no uptime guarantee, and no support. A silent upstream change breaks every backtest.

## What this issue is for

Produce a **priced replacement plan**, not a migration:

- [ ] Identify the candidate vendors and their actual quoted tiers for our shape of usage — **Polygon.io**, **Databento**, **Nasdaq Data Link** at minimum. Expect **hundreds to low thousands per month**.
- [ ] Compute required volume from the post-#1203 reality: 34 curated strategies × their declared universes, plus generated-path universes, plus re-run cadence.
- [ ] State whether daily bars suffice or intraday is needed — this is the dominant cost driver and the honest answer is probably daily.
- [ ] Identify the seam. `analytics-engine/src/archimedes_analytics_engine/instruments.py` and the data layer touched by #1203 are the likely single point of substitution — confirm it is one seam and not many.
- [ ] Record the number and the trigger date in the private business-model doc as a named line item.

## Anti-goals

- **Do not migrate the data provider in this issue.** This is a costing exercise. Migrating before we know the number and before revenue exists is premature.
- Do not treat this as a licensing-compliance emergency. It is not one today; it becomes one at first revenue. Say that accurately in either direction.

## Why it matters beyond the invoice

The cost model currently concludes marginal cost is near zero. That conclusion holds for compute at today's volume; **it does not hold once market data is a per-symbol commercial line item.** A vendor bill that scales with universe breadth is a genuine marginal cost, and it is the one input most likely to convert this from a fixed-cost business into a variable-cost one.

## Context

[...]

PR #1282, opened 2026-08-19
Market-data vendor seam; resolve #775 through it

### LINKED ISSUE #775: [oracle] Secondary-source price cross-check: yfinance guardrail for Chainlink (off-chain first → on-chain V2)

dbrowneup:
## Summary

Builds the vendor seam #1218 asked for (costing/replacement plan, not a migration) and resolves #775's Phase 1 (off-chain cross-check) inside it.

**Provider interface** at the choke points (grep-verified):
- `analytics-engine/src/archimedes_analytics_engine/data.py::fetch_ohlcv` — now a façade over `market_data.get_provider().fetch_ohlcv`. Backend's `fusion_market_data.py` / `portfolio_backtester.py` import `fetch_ohlcv` directly from here, so they get the swap for free.
- `backend/archimedes/services/strategy_signal_evaluator.py` — `_fetch_price_history(ies)`.
- `backend/archimedes/chain/oracle_updater.py` — `_fetch_yfinance`, `_fetch_yfinance_single`, `_fetch_sp500_moving_averages`.
- `backend/archimedes/services/asset_market_service.py` — `_fetch_yfinance_series`.

Default implementation is yfinance with the exact fetch/parse logic moved unchanged into `YFinanceProvider` — every call site's behavior is identical to before. Vendor-swappable via `MARKET_DATA_PROVIDER` (default `"yfinance"`; an unrecognized value fails safe to the default, logged).

analytics-engine and backend are separate deployables (analytics-engine has no DB dependency), so this is two provider modules sharing one interface shape and one env var, not one shared class — documented in `market_data_provider.py`'s module docstring.

**Cache**: `asset_daily_bars` (Postgres, `unique(symbol, trade_date)`, indexed on `symbol` — modeled on `daily_returns_store.py`'s shape) with read-through in the backend provider (`CachingMarketDataProvider`). Scoped to `get_daily_close_batch` only — the daily-bar shape that's the actual #1218 volume driver (`_fetch_price_histories`'s universe-wide sweep). Intraday/live reads (oracle pushes, VIX, the #775 cross-check's secondary, the Explore history modal) pass straight through, uncached, by design: a stale daily close must never masquerade as a live oracle quote. A coverage/freshness miss fetches the full requested window from the vendor and writes through; no background refresh yet (deliberately out of scope — "seam, not migration") — the cache primes itself on the first miss.

**#775 resolution**: the cross-check (`_cross_check_secondary`) now guards "same source, skip" against `provider_name()` instead of a hardcoded `"yfinance"`. Swapping `MARKET_DATA_PROVIDER` swaps the guardrail's independent secondary source automatically — no separate change needed at the guardrail itself. (#775's Phase 1 off-chain guardrail already shipped in #835/#898; this is the seam tie-in, not a re-implementation.)

## Migration

Chains onto `b41c7d9e2a05` — verified via `alembic heads` against `origin/main` as the actual current head. #1194 also adds a migration (`b7e3f1a2c9d4`) off that same parent but is unmerged; whichever lands second will need a trivial re-serialize (expected, not a bug).

## Tests

- `backend/tests/services/test_market_data_provider.py` (new, 17 tests): provider selection + fail-safe fallback, cache cold/warm/stale/coverage-miss/partial-miss, and that intraday/generic methods never touch the cache.
- `backend/tests/chain/test_oracle_crosscheck.py`: added a test proving the vendor-swap tie-in (source-match guard follows `provider_name()`, not a literal).
- `backend/tests/chain/test_oracle_updater_fetch.py`: `_fetch_sp500_moving_averages`'s test updated to mock at the new provider boundary; all other existing tests (yfinance fetch, single-quote, market snapshot) pass unchanged since the sys.modules-level yfinance mocking still intercepts the (moved, not rewritten) fetch logic.
- `analytics-engine/tests/test_market_data.py` (new, 9 tests): provider selection, retry/error contract preserved, `data.fetch_ohlcv` façade delegation.
- Full suites green: backend 3138 passed / 5 skipped (pre-existing, Postgres-only integration tests); analytics-engine 249 passed.

## Omissions

- Background cache refresh — explicitly out of scope per #1218 ("seam, not migration").
- `backend/archimedes/services/data_quality.py` (the #772 harness) still calls yfinance directly — not one of the three named backend choke points, and it's Önder's lane.
- Phase 2 of #775 (on-chain `PriceOracle.sol` enforcement) is untouched — separate, contract-owned, Dan's call.
- No second vendor is implemented — this PR is the seam only, per #1218's anti-goal ("do not migrate the data provider in this issue").

### DIFF
--- a/analytics-engine/src/archimedes_analytics_engine/data.py
+++ b/analytics-engine/src/archimedes_analytics_engine/data.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
 import logging
-import time
 
 import pandas as pd
-import yfinance as yf
 
 logger = logging.getLogger(__name__)
 
@@ -48,31 +46,15 @@ def normalize_ohlcv(data: pd.DataFrame, *, symbol: str) -> pd.DataFrame:
 
 
 def fetch_ohlcv(symbol: str, start: str, end: str) -> pd.DataFrame:
-    last_exc: Exception | None = None
-    for attempt in range(1, _MAX_RETRIES + 1):
-        try:
-            # auto_adjust=True applies yfinance's built-in split/dividend back-adjustment so
-            # corporate actions (e.g. KO's Aug-2012 2-for-1 split) don't show up as spurious
-            # overnight price discontinuities that corrupt pairs-trading signals and PBO.
-            data = yf.download(symbol, start=start, end=end, auto_adjust=True, progress=False)
-        except Exception as exc:  # transient network/yfinance error — retry with backoff
-            last_exc = exc
-            if attempt == _MAX_RETRIES:
-                break
-            logger.warning("fetch_ohlcv(%s) attempt %d failed: %s — retrying", symbol, attempt, exc)
-            time.sleep(_RETRY_DELAY_S * attempt)
-            continue
+    """Fetch + normalize OHLCV for ``symbol`` over [start, end).
 
-        if data.empty:
-            if attempt == _MAX_RETRIES:
-                raise ValueError(f"No data returned for symbol={symbol} in range {start}..{end}")
-            logger.warning("fetch_ohlcv(%s) returned empty — retrying (attempt %d)", symbol, attempt)
-            time.sleep(_RETRY_DELAY_S * attempt)
-            continue
+    A thin façade over the market-data provider seam (#1218:
+    ``archimedes_analytics_engine.market_data``) — default provider is
+    yfinance, and this function's retry/error contract (3 attempts, backoff,
+    raises on a genuinely empty/unfetchable result) is unchanged from before
+    the seam existed; that logic now lives in ``market_data.YFinanceProvider``.
+    Vendor-swappable via the ``MARKET_DATA_PROVIDER`` env var.
+    """
+    from archimedes_analytics_engine.market_data import get_provider
 
-        result = normalize_ohlcv(data, symbol=symbol)
-        if len(result) == 0:
-            raise ValueError(f"All rows dropped after normalization for {symbol} — check date range")
-        return result
-
-    raise RuntimeError(f"yfinance download failed for {symbol} after {_MAX_RETRIES} attempts") from last_exc
+    return get_provider().fetch_ohlcv(symbol, start, end)
--- a/analytics-engine/src/archimedes_analytics_engine/market_data.py
+++ b/analytics-engine/src/archimedes_analytics_engine/market_data.py
@@ -1,0 +1,107 @@
+"""Market-data vendor seam (#1218) — analytics-engine's choke point.
+
+#1218 priced yfinance as an unlicensed commercial dependency that scales with
+strategies x symbols x re-run cadence, and named ``data.fetch_ohlcv`` as (one
+of) the seam(s) to substitute a vendor at. This module IS that seam:
+``data.fetch_ohlcv`` becomes a thin façade over ``get_provider().fetch_ohlcv``
+— identical signature, identical default behavior (yfinance), vendor-swappable
+via the ``MARKET_DATA_PROVIDER`` env var.
+
+analytics-engine is a standalone, DB-less package (no sqlalchemy/psycopg
+dependency — see ``pyproject.toml``) that runs independently of ``backend``'s
+Postgres, so this provider has no cache layer, unlike
+``backend.archimedes.services.market_data_provider``'s
+``asset_daily_bars``-backed one. Both modules read the SAME
+``MARKET_DATA_PROVIDER`` env var and default to the same vendor name
+(``"yfinance"``), so a single env value picks the vendor across both seams —
+but a vendor implementation is registered separately in each (there is no
+shared package to put one implementation in; ``backend`` already depends on
+THIS package for ``fetch_ohlcv``, which is what makes this half of the seam
+"free" for backend's ``fusion_market_data`` / ``portfolio_backtester`` call
+sites — they import ``fetch_ohlcv`` from here and get the swap for free).
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import time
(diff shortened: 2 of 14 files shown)


dbrowneup:
First-pass review: **endorse the design; three items before it joins the train** (ordered by severity).

**1. Cache-write race loses successfully-fetched data** (`market_data_provider.py`, `CachingMarketDataProvider.get_daily_close_batch`): `_write_cached_series` is deliberately select-then-add (dialect-agnostic — fine), but under `uq_asset_daily_bars_symbol_trade_date` two concurrent writers on a cold cache (two Fargate tasks, or the leaderboard refresh loop overlapping a request sweep — both real) each see no existing row, both `session.add`, and the loser's `commit()` raises `IntegrityError` — which escapes `get_daily_close_batch` **and discards the entire fetched batch**, because `result.update(fetched)` sits *after* the write block. A fetch that succeeded fails the caller because a best-effort cache collided. Fix shape: move `result.update(fetched)` before the write; wrap write+commit in `try/except IntegrityError → rollback` (cache priming is best-effort; next reader wins); add the two-writer race test (second writer's data returned, no raise).

**2. Migration re-point is now due** — the body predicted it; it's live: main's head moved to `b7e3f1a2c9d4` (#1194 merged), and per the announced car order #1268 merges first carrying `c9f2e8d4a1b7` on that head. Chain `c9396e0d95d4` onto **`c9f2e8d4a1b7`** (and update the docstring's Revises + note), keeping the chain serial. If the car order changes, re-point to whatever the head is at merge time — #1270's CI guard should catch a miss once merged.

**3. Ruff format check is red** on the branch — `ruff format` and re-push.

**Verified good** (second pass needn't re-derive): `oracle_updater` seam is behavior-preserving with the docstrings honestly updated (asymmetric fail-open contract intact; the "honest claim this earns" wording correctly generalized); intraday/oracle reads pass through uncached — a stale daily close can't masquerade as a live quote, as designed; `provider_name()` guard genuinely makes the #775 secondary swap with the vendor; env.py model registration follows its own convention; unrecognized `MARKET_DATA_PROVIDER` fails safe to default, logged. Second pass: strategy session.

PR #1287, opened 2026-08-19
[migrations] Repair: create the Better Auth tables where 9ad1c4e2b7f0 never ran (prod auth outage)

dbrowneup:
Prod incident: Aurora's alembic_version was stamped b7e3f1a2c9d4 by an Aug-3 vintage of that reused revision id, so the deploy's migrate no-op'd at 'head' while 9ad1c4e2b7f0's DDL (auth_users/auth_sessions/auth_accounts/auth_verifications/auth_rate_limits/linked_wallets) never executed — every DB-touching auth route returns 'Authentication service unavailable'. The child migration's own DDL is present, confirmed by direct inspection.

The repair re-invokes the ORIGINAL migration's upgrade() verbatim behind an inspector has_table guard: correctly-migrated environments (local compose, CI) no-op; prod converges to the identical schema with zero hand-translated DDL. Downgrade is a deliberate no-op. Chains onto c9f2e8d4a1b7 (#1268, merging now).

Rides the deploy's own pre-rollout migrate task — no manual prod surgery. Guard-hazard documented in the module docstring: revision-id reuse across re-serializations is invisible to alembic and to #1270's chain guard; never reuse an id.

NOTE: merge AFTER #1268 (its parent). CI's migration guard enforces this ordering mechanically.

### DIFF
--- a/backend/migrations/versions/e4d9a7f21c58_repair_missing_better_auth_ddl.py
+++ b/backend/migrations/versions/e4d9a7f21c58_repair_missing_better_auth_ddl.py
@@ -1,0 +1,56 @@
+"""Repair: create the Better Auth tables where 9ad1c4e2b7f0's DDL never ran.
+
+Prod incident 2026-08-19: Aurora's ``alembic_version`` was stamped
+``b7e3f1a2c9d4`` by an Aug-3 vintage of that revision id (the migration chain
+was re-serialized twice afterward and the id was reused with different
+parentage), so the deploy's pre-rollout ``alembic upgrade head`` no-op'd while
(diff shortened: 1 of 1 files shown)


dbrowneup:
Closing — false premise, caught before merge. The 'missing auth tables' diagnosis was an artifact of my own verification gap: the exact-name existence check sat inside a permission-blocked command and never ran, while the query that DID run used near-miss names (auth_user vs auth_users). Direct exact-name inspection: all six tables exist in prod. The real outage cause is pg's sslmode-precedence quirk — fixed and proven live in #1288. The id-reuse hazard documented here is still real and moves to #1270's docs. Repair unneeded; deleting the branch.

PR #1295, opened 2026-08-19
[frontend] Generate page: upfront cost quote + approval (flag-gated)

dbrowneup:
## Summary

Frontend half of the quote/paywall feature (contracts session owns the
backend: quote endpoint + x402 paywall on `/api/generate/start`). Built
against a contract this PR proposes for that session to ratify or amend:
[`docs/specs/generation-quote-contract.md`](docs/specs/generation-quote-contract.md).

- Generate fetches `GET /api/generate/quote` and shows a quote card before
  submission: price, an explicit **"testnet USDC"** label, optional
  breakdown, and a **"free"**-tagged note that paper trading after
  generation costs nothing (Dan: say it in the UI).
- Submit button becomes "Approve & Generate →"; approving carries
  `quote_id` on `POST /api/generate/start`.
- A 402 from `/start` renders a payment-step banner instead of the generic
  error: **wallet-required** (Connect Wallet, reusing the existing
  `open-wallet-modal` link flow) if no wallet is connected, otherwise an
  honest **"payments aren't enabled yet"** preview — this PR does not sign
  or execute an on-chain payment.
- Everything is behind `VITE_GENERATION_QUOTE_ENABLED`
  (`ui/src/featureFlags.js`, default **off**) — flag-off behavior is
  byte-identical to before this PR. Dan flips it once the backend is ready.

`ui/src/featureFlags.js` follows the path/shape of the flag file introduced
in #1266 (`ROADMAP_SURFACES_ENABLED`), which is open but not yet on `main`
— checked before writing this so the two flags land in the same file
without a merge conflict once #1266 lands.

State-transition logic (quote shaping, expiry, payload building, 402 →
payment-state routing) lives in `ui/src/generateQuote.js`, pure and
DOM-free, imported by `Generate.jsx` rather than reimplemented inline.

## Verification

```
cd ui
npm run lint                                          # clean
npm test                                               # 38/38 pass (12 new)
rm -rf dist && npm run build                           # flag off — clean
rm -rf dist && VITE_GENERATION_QUOTE_ENABLED=true npm run build   # flag on — clean
```

Adversarial pass on the three guard functions (per repo convention) —
mutated each to the wrong-but-plausible behavior, confirmed the
corresponding test fails, then reverted:
- `attachQuoteId` ignoring `quoteEnabled` → "fails closed when flag is off" test fails
- `isPaywallError` treating any error as 402 → "recognizes only a genuine 402" test fails
- `derivePaymentState` not failing closed on blank address → wallet-required test fails

## Omissions (deliberate)

- **No real payment execution.** Signing/paying the quote via the connected
  wallet (x402) isn't wired here — see contract doc's "out of scope"
  section. Lands once the contract is ratified and the payment rail is
  defined.
- **No wallet-address prop threading.** Uses `getAddress()` /
  `wallet-changed` directly (same pattern as `WalletConnect.jsx`) rather
  than plumbing `walletAddr` through `AuthenticatedApp.jsx`'s render
  switch, to keep the diff scoped to `Generate.jsx`.
- Quote is treated as single-use: a successful `/start` triggers a
  refetch for the next generation. Not stated in the contract doc either
  way — flag for the contracts session to confirm.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -69,7 +69,7 @@ Owners: **Dan Browne** — contracts, on-chain, infrastructure, architecture. **
 | [`specs/strategy-dsl-spec.md`](specs/strategy-dsl-spec.md) | spec | Dan Browne | — | Strategy DSL. |
 | [`specs/strategy-lifecycle-spec.md`](specs/strategy-lifecycle-spec.md) | spec | Dan Browne | — | Draft → generated → published lifecycle. |
 | [`specs/generation-streaming-spec.md`](specs/generation-streaming-spec.md) | spec | Dan Browne | — | SSE contract for the Generate page. |
-| [`specs/generation-quote-contract.md`](specs/generation-quote-contract.md) | proposed | Dan Browne | — | Upfront cost-quote endpoint + `quote_id` on `/api/generate/start`, drafted by the frontend session for the contracts session to ratify. Frontend ships behind `VITE_GENERATION_QUOTE_ENABLED`. |
+| [`specs/generation-quote-contract.md`](specs/generation-quote-contract.md) | spec | Dan Browne | — | Public cost quote + 409/402 x402 paywall on `/api/generate/start`, ratified in #1296. Frontend ships behind `VITE_GENERATION_QUOTE_ENABLED`. |
 | [`specs/kb-integration-spec.md`](specs/kb-integration-spec.md) | spec | Dan Browne | — | KnowledgeBase submodule integration. |
 | [`specs/page-roles-spec.md`](specs/page-roles-spec.md) | spec | Dan Browne | — | What each page is for. |
 | [`specs/component-interfaces-spec.md`](specs/component-interfaces-spec.md) | spec | Dan Browne | — | Component interfaces and the team work split. |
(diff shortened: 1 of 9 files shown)


dbrowneup:
Reworked onto the ratified #1296 contract; requesting the contract-owner consumption pass.

dbrowneup:
Contract-owner consumption pass (#1296's shape): **APPROVED, and the test-mode ruling is CONFIRMED in-scope** — with the reasoning on record.

**Why the stand-in is correct**: dry-run's any-header-unverified acceptance exists precisely so the full approval journey is exercisable with zero settle risk; a UI that exploits it honestly is the intended consumer. Three properties make it safe, all verified in this diff: (1) the button gates on the **402's own** `detail.quote.dry_run` (`e?.detail?.quote?.dry_run ?? quote?.dry_run`), not a stale page-load quote — live mode never renders it; (2) if dry-run flips server-side mid-session anyway, the stand-in degrades safely — decode succeeds, payer matches, facilitator **rejects the unsigned payload** → 402 `payment_invalid`; a fabricated payment can never pass the live paywall; (3) the notice wording ("test mode — payment accepted unverified") mirrors the server log's honesty verbatim. Follow-up for real EIP-712 signing filed as **#1298** — required before `PAYMENTS_DRY_RUN=false`, deliberately not before this merges.

**Consumption verified against the real route**: quote parse matches `generation_payment.quote()` field-for-field (no quote_id/expiry — and the doc's explicit ban on reintroducing them is exactly right); 409 → link-wallet CTA; `describePayerMismatch`'s pre-warning (active wallet ≠ linked wallet, case-insensitive both sides) correctly anticipates the payer binding while leaving the server as the authority; stand-in header names the linked wallet as payer, forward-compatible with #1298. The RATIFIED doc records the shipped shape faithfully, including the byte-identical 402/quote guarantee my route test pins.

One nit, non-blocking: `buildDryRunPaymentHeader` with an empty `payerAddress` produces `from: ""` — harmless today (dry-run skips the payer check) but worth a guard or comment so #1298's implementer doesn't inherit it silently.

Issue #1298, opened 2026-08-19
[frontend] Real x402 signing (EIP-712 via the linked wallet) — required before PAYMENTS_DRY_RUN=false


dbrowneup:
Follow-up to #1295/#1296. The quote-approval UI's "continue in test mode" path submits an honestly-labeled stand-in Payment-Signature, valid ONLY under `PAYMENTS_DRY_RUN` (the paywall accepts headers unverified there by design). Before the dry-run flag flips for real revenue (#975 sequencing, see the #834 flip-list), the frontend needs the real flow: parse the 402's `PAYMENT-REQUIRED` requirements, produce the EIP-712/x402 signature **with the linked wallet's account** (payer binding: the signed authorization's `from` must equal the linked wallet — `payer_mismatch` otherwise), retry with `Payment-Signature`, surface the `PAYMENT-RESPONSE` receipt.

Safety note (why this is not urgent-before-merge): the stand-in degrades safely if dry-run flips mid-session — decode succeeds, payer matches, but facilitator verification rejects the unsigned payload → 402 `payment_invalid`. No fabricated payment can ever pass the live paywall.

Acceptance: with `PAYMENTS_DRY_RUN=false` against a funded linked wallet on testnet, the full quote → sign → settle → receipt journey succeeds from the Generate page, and the test-mode button is absent (it already gates on the 402's own `dry_run`).

dbrowneup:
Closing as shipped and proven in production. Real EIP-712 TransferWithAuthorization signing via the linked wallet landed across #1427 and the wallet-matrix ladder #1461/#1464/#1465/#1468 — `ui/src/x402.js` (`signGatewayPayment`, EOA + device-payment-key paths) with server-side payer binding (`generation_payment.py` rejects `payer_mismatch`). Acceptance is met by live evidence, not just code: with `PAYMENTS_DRY_RUN=false`, 19 generations have completed the quote → sign → settle → receipt journey on prod ($38.00 settled into the Gateway as of 2026-08-30 16:30 UTC; first sweep collected on-chain: tx 0x2cf55a97…8735e). Test-mode control gates on the 402's own `dry_run` as specified.

PR #1299, opened 2026-08-19
Test-mode payer: linked-wallet-first with a no-payer guard (#1298 prep)

dbrowneup:
The #1295 consumption-pass nit, taken by the contract owner — and it ran deeper than cosmetic.

**The bug shape**: the test-mode path passed `walletAddr` (the **browser's** active wallet) as the payer. Reaching a 402 proves the **server-side** wallet link (the 409 gate ran first) — it does not prove the browser wallet is connected or that the linked-wallets fetch succeeded, so an empty payer was genuinely reachable, and the call-site comment claimed otherwise.

**Now**: payer resolves **linked-wallet-first** (server truth — exactly what #1296's payer binding requires once signing becomes real, #1298), falls back to the active wallet, and bails with an honest "no payment was attempted" message when neither resolves. `buildDryRunPaymentHeader` **throws** on a falsy payer — a loud contract for #1298's implementer instead of a silently-encoded empty `from`.

58/58 UI tests (count as of 9d361483) (the old encode-empty test is strengthened into the refusal assertion + a source-level guard pin), lint and build clean. One file each in src/components/test.

---

**Review fix (9d361483):** the call site passed the full LinkedWalletResponse OBJECT into the payment header (truthy, so no guard caught it). New `resolveDryRunPayer()` returns the address string or null; the source-regex guard test — which pinned the buggy line verbatim — is replaced by tests that EXECUTE the resolution and decode the header (mutation-proven: an object-returning resolver fails them).

### DIFF
--- a/ui/src/components/Generate.jsx
+++ b/ui/src/components/Generate.jsx
@@ -16,6 +16,7 @@ import {
 	describePayerMismatch,
 	extractReceipt,
 	paymentErrorMessage,
+	primaryLinkedWallet,
 } from "../generateQuote";
 
 const shortAddr = (addr) => (addr ? `${addr.slice(0, 6)}…${addr.slice(-4)}` : "");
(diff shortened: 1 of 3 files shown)


dbrowneup:
Consumer-side vet (roles reversed — the contract owner fixed my surface, so my pass closes the loop): verified on the branch, 56/56, and the finding was genuinely deeper than the nit — the 409 gate proves the SERVER-side link while walletAddr reflects browser connection state, so an empty payer after a real 402 was a live path. Linked-wallet-first resolution matches exactly what #1298's binding will enforce, and the throw-on-empty is the right loud contract for its implementer. Endorsed; queues for Dan's next batch word.

PR #1302, opened 2026-08-20
Paper-trading UI: /app/paper page + passport CTA !minor

dbrowneup:
## Summary

The paper-trading backend (#1268) was API-only — the MVP spine's final step (deploy a passing strategy to paper trading) worked for an agent with curl and not at all for a human in the browser. This adds the UI half.

### `/app/paper` — Paper Trading page

New page in the **Strategy** nav group (generate → library → **paper** → leaderboard), auth-required (not in `ANON_APP_PAGES`). Renders `GET /api/paper/deployments`:

- **ACTIVE / STOPPED** chip, plus a **DRIFT** chip when `drift_detected_at` is set (tooltip explains the frozen-ledger semantics).
- Day count (`days`), with an honest day-0 line ("first return lands with the next daily advance").
- `total_return` and an equity **sparkline** drawn from the summary's own `series[].equity_index` — no extra endpoints.
- **Stop** via `POST /api/paper/deployments/{id}/stop` with a confirm dialog (freeze-in-place wording — there is no restart).
- Strategy display names come from a best-effort client-side join against `/api/strategies/` + `/api/strategies/generated` (the paper API deliberately returns ids only); the list renders with raw ids if the join fails.
- Empty state routes to the Library.

### StrategyPassport — "Start paper trading (free)" card

Separate from the vault Deploy card: account-only, **no wallet, no strictness gate, nothing moves** — matching the server contract exactly (`paper_routes.py` requires only a Better Auth session + spec presence + visibility). Handles:

- **Already running** → "Paper trading — day N · view →" (client-side dedupe; the server permits duplicate deployments by design, there is no 409).
- **422** `no_strategy_spec` / `invalid_strategy_spec` → specific messages (real for some curated rows).
- **Anonymous reader** (the passport is a public page) → "Sign in to paper trade" with a `next=` return path.
- Success (201) → navigates to `/app/paper`.

The card is placed after the strictness slider and imports are non-adjacent, so this PR's hunks do not overlap #1266's edits to the same file — clean auto-merge in either merge order.

## Deliberately not in scope

- No rigor precondition on the CTA — deliberate mirror of the backend's contract (paper is the free act-on step; the gate verdict is displayed two cards up).
- No paper metrics beyond the summary payload; richer analytics belong with #63/#1217-era instrumentation.
- Onboarding-tour card for paper: follow-up, after #1266 lands (the tour's card list is edited by that PR).

## Verification

```
ui: node --test        0 fail (adds /app/paper to the auth-required route pin)
ui: npm run lint       clean
ui: npm run build      clean
backend: pytest backend/tests/test_breadcrumbs.py   5/5 (CRUMB_MAP sync holds)
```

Cross-review requested from the contracts session (backend-contract fit; its agent_journey paper step exercises the same POST + readback).

### DIFF
--- a/ui/src/App.jsx
+++ b/ui/src/App.jsx
@@ -84,6 +84,7 @@ export default function App() {
       account: 'Account · Archimedes',
       'vault-detail': 'Vault · Archimedes',
       strategy: 'Strategy · Archimedes',
+      paper: 'Paper Trading · Archimedes',
       'sign-in': 'Sign in · Archimedes',
       'sign-up': 'Create account · Archimedes',
     }
--- a/ui/src/AuthenticatedApp.jsx
+++ b/ui/src/AuthenticatedApp.jsx
@@ -14,6 +14,7 @@ import MarketplacePage from "./components/MarketplacePage";
 import OnboardingTour, {
 	hasCompletedOnboarding,
 } from "./components/OnboardingTour";
+import PaperTrading from "./components/PaperTrading";
 import Portfolio from "./components/Portfolio";
 import PublishPage from "./components/PublishPage";
 import QuantLab from "./components/QuantLab";
@@ -137,12 +138,15 @@ export default function AuthenticatedApp({
 						onNavigate={navigateToPage}
 					/>
 				);
+			case "paper":
+				return <PaperTrading onNavigate={navigateToPage} />;
 			case "strategy":
 				return (
 					<StrategyPassport
 						strategyId={route.strategyId}
 						onNavigate={navigateToPage}
 						walletAddr={walletAddr}
+						user={user}
 					/>
 				);
 			case "corpus":
--- a/ui/src/components/Breadcrumbs.jsx
+++ b/ui/src/components/Breadcrumbs.jsx
@@ -28,6 +28,7 @@ export const CRUMB_MAP = {
   // Strategy — wallet-gated, owns the primary generation path
   generate:     { group: null, groupPage: null },
   library:      { group: 'Strategy', groupPage: 'generate' },
+  paper:        { group: 'Strategy', groupPage: 'generate' },
   leaderboard:  { group: 'Strategy', groupPage: 'generate' },
   // Position — wallet-gated deployed-state surfaces
   portfolio:    { group: null, groupPage: null },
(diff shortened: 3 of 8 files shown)


dbrowneup:
Backend-contract fit review (requested cross-review; GitHub blocks a formal approval since both sessions post as one account — treat this as the review of record; merge remains Dan's call):

**Verified correct against the live routes/services:**
- POST /api/paper/deployments {strategy_id} — single required field; success consumed without assuming extra fields. 422 reasons `no_strategy_spec` / `invalid_strategy_spec` matched to specific messages; 401 handled; no 409 assumed (none exists — dedupe is correctly client-side).
- Stop = POST /deployments/{id}/stop ✓ (no DELETE anywhere in the diff).
- List shape `{deployments: [...]}` and the deployment_summary fields (deployment_id, strategy_id, deployed_at, status, days, total_return, drift_detected_at, series) all match `services/paper_trading.py::deployment_summary`. Sparkline reads `series[].equity_index` ✓; the day-0 copy matches the scheduler-backfill behavior in the route.
- Name join: both `/api/strategies/` (StrategyListResponse) and `/api/strategies/generated` return `{strategies: [...]}` ✓; id fallback means a failed join degrades honestly. `/generated` requires auth — fine, the page is auth-only.
- `user` prop threaded for the passport route; public/unauthed passport falls to the sign-in CTA with the `?next=` pattern AuthPage already handles ✓.
- DRIFT chip semantics match the frozen-ledger behavior (drift_detected_at set, ledger not rewritten).

**Two non-blocking notes:**
1. `/generated` default limit=50 — a deployment referencing an older generated strategy can miss its name and fall back to the id. Fine for MVP; bump the limit or fetch-by-id later if it shows.
2. The join reads `row.id ?? row.strategy_id` — matches both list shapes today; if a third list source is ever added, keep that order.

Journey-harness parity: scripts/agent_journey.py's new paper leg (PR #1301) posts the identical body and reads the same summary fields, so CI pins the contract from both sides.

dbrowneup:
Both review notes addressed in 9829278e: the /generated join now requests limit=200 (the endpoint max — anything beyond still falls back to the id honestly), and the id-before-strategy_id order is pinned with a comment for any future third list source. Review-of-record convention acknowledged; merge remains Dan's call.

dbrowneup:
Both review notes confirmed addressed in 9829278e (limit=200 on the /generated join with the endpoint-max rationale in a comment; id-field order pinned with the review reference). Review loop closed from my side — contract fit stands verified. Dan-gated for merge.

PR #1329, opened 2026-08-20
[backend] Resolve latest-backtest-per-strategy in SQL instead of loading the whole table

dbrowneup:
## Summary

`latest_backtests_by_strategy()` fetched **every** row for every requested strategy and reduced in Python, keeping one row per strategy and throwing the rest away. It loaded `N_refresh_cycles × N_strategies` rows in order to return `N_strategies`.

That would be merely wasteful if the table were bounded. It is not:

- `canonical_artifact_hash()` ([`backtest_mapper.py:119-127`](backend/archimedes/services/backtest_mapper.py#L119-L127)) hashes the **whole payload**, which includes `run_id` — a UTC timestamp minted per run ([`cli.py:160`](analytics-engine/src/archimedes_analytics_engine/cli.py#L160)) — and `timestamp_utc`.
- So `content_hash` is unique per run **by construction**, and `insert_backtest_if_missing` can never match an existing row.
- The scheduler runs a refresh on **every container start** (`_last_unresolved_missing` is a module global reset each process), so the table grows by ~30 rows per cycle, forever.
- Every row carries a **~489 KB `artifact_json`** plus a **~108 KB `equity_curve_json`**.

**Measured cost of the old read path: 0.58 MB of peak RSS per persisted row, perfectly linear** — 21 MB at 34 rows, 201 MB at 340, 395 MB at 680.

That staircase is what walked the 2026-08-19 backend from a flat 842 MB plateau to 2.6 GB against a **3,072 MB task budget shared with `auth` and `nginx`**, until the T+180s refresh spike crossed the limit and the OOM killer took the container — repeatedly, with `desiredCount=1`, so every kill was a full site outage.

## What changed

A `row_number()` window partitioned by `strategy_id` resolves the winners **in the database**; only those rows are hydrated.

The inner query deliberately selects **only `id`** — the window must not drag the 489 KB payload columns through the client buffer. Ordering semantics are unchanged (`created_at DESC`, then `id DESC`).

Window functions are available on Postgres (prod) and SQLite ≥ 3.25 (the hermetic test path; this machine has 3.53.1).

## Verification

```
pytest backend/tests/test_backtest_repository.py \
       backend/tests/test_backtest_scheduler.py \
       backend/tests/test_backtest_mapper.py -q      → 25 passed
ruff format --check + ruff check (changed files)     → clean
```

All five production callers exercised: `_generated_strategy_rigor`, `get_strategy_returns`, `needs_refresh`, `_remember_unresolved_missing`, `_load_backtests`.

### The guard, and the first version of it that was worthless

Per `CLAUDE.md` § "A guard must be shown to reject something", I mutation-checked by restoring the old `.all()` reader.

**My first attempt passed against the unfixed code.** It counted `session.identity_map` after the call — but the identity map holds **weak references**, so the 646 rows the old implementation loaded and discarded were garbage-collected before the assertion ran. It would have guarded nothing while looking like coverage. This is exactly rule 3 in that section, and I nearly shipped it.

The guard now counts hydration through the SQLAlchemy mapper `load` event, which fires once per row materialised from the DB regardless of what survives GC. Re-running the mutation:

```
AssertionError: hydrated 680 BacktestResultRecord rows to return 34;
                the reader is loading the whole table (680 rows on disk)
FAILED test_latest_backtests_hydrates_only_one_row_per_strategy
```

Restored → `9 passed`.

## Scope — what this does NOT fix

This bounds the **read**. It does **not** stop the table growing, and it does not reclaim the rows already accumulated in production. Two follow-ups, filed separately so they can be reviewed on their own merits:

1. **`canonical_artifact_hash` must exclude `run_id`/`timestamp_utc`** so `insert_backtest_if_missing` can actually dedupe. This changes hash semantics for existing rows and deserves its own review.
2. **The accumulated duplicate rows need a one-off cleanup**, and nobody has yet read the real row count in production — `SELECT strategy_id, count(*) FROM backtest_results GROUP BY 1 ORDER BY 2 DESC;` should be the first thing run.

Also unaddressed here, deliberately: whether a full 34-strategy backtrader re-run belongs inside the request-serving container at all (its own module docstring says re-triggering "burns the box's 2 vCPUs for ~15 min").

## Anti-goals honoured

- No change to write-path behaviour, to `insert_backtest_if_missing`, or to the hash.
- No change to ordering semantics or to the function's signature/return type.
- No new dependency; no raw SQL string (stays in the SQLAlchemy expression layer so SQLite and Postgres both work).

## Related

- Companion to #1328 (`/health` no longer loads the embedding model — 521 MB). Together these take the backend from a 2,956 MB peak toward roughly 800 MB inside the existing 3,072 MB budget, with **no memory bump and no cost increase**.
- Root cause traced to #1263, which fixed a real artifact-write `PermissionError` and thereby unmasked these two latent unbounded defects — before it, rows never persisted at all.

### DIFF
--- a/backend/archimedes/services/backtest_repository.py
+++ b/backend/archimedes/services/backtest_repository.py
@@ -11,6 +11,7 @@ import os
 from collections.abc import Iterable
 from datetime import UTC, datetime
 
+from sqlalchemy import func, select
 from sqlalchemy.orm import Session
 
 from archimedes.models.backtest import BacktestResult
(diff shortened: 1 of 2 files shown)


dbrowneup:
Full unit suite on this branch, using the exact command CI runs from the repo root (not a scoped `pytest backend/tests`, which collects a different set):

```
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest -m "not integration" -q
→ 3357 passed, 2 skipped, 2 deselected in 405.01s
```

All five production callers of `latest_backtests_by_strategy` are covered by that run: `_generated_strategy_rigor`, `get_strategy_returns`, `needs_refresh`, `_remember_unresolved_missing`, `_load_backtests`.

PR #1330, opened 2026-08-20
[infra] Make a verbatim .env.example bootstrap the local stack; refuse the placeholder in prod

dbrowneup:
## Summary

`cp .env.example .env` — the exact command the file's own header tells you to run — produced a `.env` that broke **every** `docker compose` command:

```
error while interpolating services.auth.environment.BETTER_AUTH_SECRET:
required variable BETTER_AUTH_SECRET is missing a value: Set BETTER_AUTH_SECRET in .env
```

`BETTER_AUTH_SECRET` shipped empty (`.env.example:361`) while [`docker-compose.yml:167`](docker-compose.yml#L167) requires it via `${BETTER_AUTH_SECRET:?…}`. It was the **only** compose-required variable shipped empty — `DATABASE_URL` and `POSTGRES_PASSWORD` both have working values.

Worse than a failed `up`: compose resolves interpolation **before it runs any action**, so this also broke `docker compose down`. You could not tear down a stack you had already started, and the error text points at a variable you may never have needed to think about.

The file header promises the opposite: *"Most defaults work out of the box for local development."*

## What changed

**1. `.env.example` ships a public placeholder** so a verbatim copy bootstraps the stack.

**2. That placeholder is refused in production.** It signs session cookies and is readable by anyone with the repo, so `createAuth()` throws when it sees exactly that value with `NODE_ENV=production`:

```js
if (production && secret === PLACEHOLDER_SECRET) {
  throw new Error('BETTER_AUTH_SECRET is still the public .env.example placeholder. …')
}
```

The literal lives in `auth/auth.js` as `PLACEHOLDER_SECRET`, and a test asserts `.env.example` still ships exactly that string — so the two cannot silently drift apart. The existing 32-character floor is untouched (the placeholder satisfies it).

**3. Corrected `REASONING_TRACE_REGISTRY_ADDRESS`** (both the bare and the `ARC_`-prefixed commented variant). It pointed at `0x42d8a23e…`, the **superseded v1 anchor-only registry**. An on-chain selector scan against Arc testnet:

| selector | `0x42d8a23e…` (was in .env.example) | `0x9d81bfbf…` (production) |
|---|---|---|
| `commit(address,bytes32,uint64,bytes32,bytes)` | **ABSENT** | PRESENT |
| `executeTrade(bytes32)` | **ABSENT** | PRESENT |
| `reveal(uint256,string,bytes)` | **ABSENT** | PRESENT |
| `getCommitment(uint256)` | **ABSENT** | PRESENT |
| `publishTrace(address,bytes32,bytes)` | PRESENT | PRESENT |

Production serves `0x9d81bfbf…` from `GET /api/config/contracts`. A local stack pointed at the old address silently **cannot do commit-before-trade** — it would fall back to bare anchoring while the UI claims ordering is contract-enforced.

## Verification

```
docker compose --env-file <verbatim copy of .env.example> config   → OK
same, with BETTER_AUTH_SECRET emptied                              → reproduces the original error verbatim
cd auth && npm ci && npm test                                      → 19 passed
```

**Mutation check** (`CLAUDE.md` § "A guard must be shown to reject something") — disabling the production guard:

```
✖ the public .env.example placeholder is refused in production
  AssertionError [ERR_ASSERTION]: Missing expected exception.
```

Restored → all pass.

Note: `auth/node_modules` was absent on a fresh clone, so `npm test` failed with `Cannot find package 'better-auth'` before `npm ci`. That is not introduced here, but it means `auth/`'s tests are easy to skip accidentally — see the follow-up note below.

## Anti-goals honoured

- **No real secret is committed.** The placeholder is deliberately non-secret and is rejected in production.
- `docker-compose.yml` and `docker-compose.production.yml` are **not** modified — the `${VAR:?…}` guards stay exactly as they are.
- The 32-char minimum, `revokeSessionsOnPasswordReset`, and `requireEmailVerification` are untouched.
- No change to any real contract address used by production.

## Follow-ups (not in this PR)

- ~~`auth/` has no CI job running `npm test`~~ — **retracted, this was wrong.** `.github/workflows/quality-gate.yml:158` runs `cd auth && npm ci --quiet && npm test` in the "Frontend + Better Auth" job, and it passed on this PR. The only real point was local: `auth/node_modules` is absent on a fresh clone, so `npm test` fails locally until you `npm ci`. CI is fine.
- The stale address also raises a live defect I filed separately: `supports_commit_reveal()` ([`trace_publisher.py:36-48`](backend/archimedes/chain/trace_publisher.py#L36-L48)) checks `hasattr()` against the **locally cached ABI**, not the chain. `contracts/abis/ReasoningTraceRegistry.json` contains `commit`/`reveal`, so that guard **always returns `True`** and its documented `publishTrace` fallback can never fire. `verifyTrace` is in that ABI but on neither deployed contract.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -295,7 +295,11 @@ PRICE_ORACLE_ADDRESS=0xe1c9f2b11be97097223a66a188fca541e07873a6
 SYNTHETIC_FACTORY_ADDRESS=
 AMM_ROUTER_ADDRESS=0xd5b829f9d364a8bbe1caf6c8b19cb05371b178f4
 VAULT_FACTORY_ADDRESS=0xca873414070844aeb98b0bf1051f81969c79cc32
-REASONING_TRACE_REGISTRY_ADDRESS=0x42d8a23edb897cbee203e9fa197eb05ab5106ca6
+# v1.5 registry — the one production runs (confirmed via GET /api/config/contracts).
+# The previous value here, 0x42d8a23e…, is the superseded v1 anchor-only deployment:
+# an on-chain selector scan shows commit/executeTrade/reveal/getCommitment ABSENT on
+# it, so a local stack pointed at it silently cannot do commit-before-trade.
+REASONING_TRACE_REGISTRY_ADDRESS=0x9d81bfbfadb683cf77acda480e94e64088012847
 ASSET_REGISTRY_ADDRESS=0x2d44550711137916df6175587d17886281a0fbc7
 STRATEGY_REGISTRY_ADDRESS=0x728C264d0681b71c4Cc1D26a4fb14Ec29D9a90e4
 
@@ -317,7 +321,7 @@ STRATEGY_REGISTRY_ADDRESS=0x728C264d0681b71c4Cc1D26a4fb14Ec29D9a90e4
 # ARC_AMM_ROUTER_ADDRESS=0xd5b829f9d364a8bbe1caf6c8b19cb05371b178f4
 # ARC_SYNTHETIC_FACTORY_ADDRESS=
 # ARC_VAULT_FACTORY_ADDRESS=0xca873414070844aeb98b0bf1051f81969c79cc32
-# ARC_REASONING_TRACE_REGISTRY_ADDRESS=0x42d8a23edb897cbee203e9fa197eb05ab5106ca6
+# ARC_REASONING_TRACE_REGISTRY_ADDRESS=0x9d81bfbfadb683cf77acda480e94e64088012847
 # ARC_ASSET_REGISTRY_ADDRESS=0x2d44550711137916df6175587d17886281a0fbc7
 # ARC_STRATEGY_REGISTRY_ADDRESS=0x728C264d0681b71c4Cc1D26a4fb14Ec29D9a90e4
 #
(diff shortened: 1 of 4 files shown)


dbrowneup:
**Correction to my own PR body.** I claimed "`auth/` has no CI job running `npm test` that I could find". That is **false**, and I should have checked before writing it rather than hedging with "that I could find".

`.github/workflows/quality-gate.yml:158` runs `cd auth && npm ci --quiet && npm test` in the *Frontend + Better Auth* job — and it passed on this very PR. I have struck the line from the body.

The only true residue is local, not CI: `auth/node_modules` is absent on a fresh clone, so `npm test` fails with `Cannot find package 'better-auth'` until you run `npm ci`. That is a developer-onboarding wrinkle, not a coverage gap.

Flagging it explicitly because an unbacked assertion in a PR description is the same defect class this PR is about — `CLAUDE.md` calls out prose claims the code does not support as "the same defect, just harder to grep for".

dbrowneup:
Fixed the `test_local_setup_contract.py:29` substring-replace bug: now that `.env.example` ships a non-empty `BETTER_AUTH_SECRET` placeholder (`insecure-local-dev-placeholder-change-me-before-any-deploy`), the old `template.replace("BETTER_AUTH_SECRET=", f"BETTER_AUTH_SECRET={TEST_AUTH_SECRET}", 1)` only matches the `BETTER_AUTH_SECRET=` prefix and **concatenates** `TEST_AUTH_SECRET` onto the placeholder instead of overriding it.

**Fix:** full-line substitution via `re.sub(r"^BETTER_AUTH_SECRET=.*$", f"BETTER_AUTH_SECRET={TEST_AUTH_SECRET}", template, count=1, flags=re.MULTILINE)`.

**Added assertion** (`test_composed_auth_secret_equals_the_test_override_exactly`) that the composed value reaching `services.auth.environment.BETTER_AUTH_SECRET` in `docker compose config` output equals `TEST_AUTH_SECRET` exactly.

**Mutation-proof** — reverted only the `re.sub` fix back to the old `str.replace` on this branch and re-ran the new assertion:

```
$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_local_setup_contract.py -q -k composed_auth_secret

FAILED backend/tests/test_local_setup_contract.py::TestLocalSetupContract::test_composed_auth_secret_equals_the_test_override_exactly
AssertionError: 'test-only-better-auth-secret-at-least-32-charactersinsecure-local-dev-placeholder-change-me-before-any-deploy' != 'test-only-better-auth-secret-at-least-32-characters'
1 failed, 4 deselected in 3.59s
```

Restored the `re.sub` fix — back to green:

```
$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest backend/tests/test_local_setup_contract.py -q
5 passed, 1 warning in 1.80s
```

Full unit suite (`pytest -m "not integration"` from repo root) and `ruff format --check` / `ruff check` on the touched file also clean — pushed to this branch.

PR #1387, opened 2026-08-20
[frontend] Leaderboard: fix false refresh promise, dead DSR sort, constant Forward column

### LINKED ISSUE #1365: [frontend] Leaderboard: a banner promising a refresh cycle that does not exist, a sort with no column, and a constant Forward column
## Summary

Three defects on `/app/leaderboard`, all verified against `main` @ `523a1bae`:

1. **The own-scope caveat banner promises a refresh that cannot happen.** It tells the user older figures "refresh on their next backtest cycle." The `own` cohort is generated-strategies-only, and no re-backtest path for a generated strategy exists anywhere in the backend. The banner talks the user out of the only action that would actually produce current numbers (generating again).
2. **A sort control with no visible column.** "Deflated Sharpe" is offered and honoured by the backend, but `deflated_sharpe_ratio` is never rendered — choosing it visibly reshuffles rows while every visible number stays put, which reads as a broken control. `out_of_sample_sharpe` carries 25% of the conviction weight and is likewise never printed.
3. **The Forward column is two constants.** `SB pending` / `P&L pending` are string literals, byte-identical on every row; `e.forward` is never read. The same caveat is already stated twice *above* the table, so the column costs width and conveys nothing — and it will keep printing "pending" the day a real status is emitted.

### 1 — the refresh promise

`ui/src/components/Leaderboard.jsx:202-204` (inside `{isOwn && (` at :189):

```jsx
<strong style={{ color: 'var(--warning, #b45309)' }}>One caveat on your figures.</strong>{' '}
Older strategies may still show numbers computed before the August engine corrections —
they refresh on their next backtest cycle.
```

The same false belief is written into the code comment at `:185`: `strategies that predate the corrections and refresh on their next / cycle`.

No such cycle exists for that cohort:

- `own` is generated-only — `backend/archimedes/api/leaderboard_routes.py:101-117` (`_own_cohort_responses` → `_owned_generated_strategy_responses`, `backend/archimedes/api/strategies_routes.py:1848`); curated rows are explicitly excluded.
- Those metrics are read verbatim off the persisted passport row — `backend/archimedes/api/strategies_routes.py:1751-1762` (`sharpe_ratio=record.sharpe_ratio`, … `out_of_sample_sharpe=record.out_of_sample_sharpe`).
- The only refresh loop is curated-only: `backend/archimedes/services/backtest_scheduler.py:78` stales off `default_provider().list_strategies()`, and its worker enumerates `strategy_dir.glob("*.py")` at `backend/archimedes/scripts/run_backtests.py:334` (matching `strategy_provider.py:493`). A generated strategy is a DB row with a DSL spec and **no `.py` file**, so it is never enumerated.
- The only writer of `real_*` metrics onto a passport is `_refresh_passport_real_metrics` / `_backtest_and_persist` in `backend/archimedes/agents/generation_pipeline.py:1394,1578` — both run once, inside the generation run that created the row. `update_rigor_gate_fields` (`backend/archimedes/services/backtest_repository.py:205-243`) writes `BacktestResultRecord`, not the passport, so it does not move these numbers either.

### 2 — sort with no column

`Leaderboard.jsx:18` offers it and `backend/archimedes/services/leaderboard.py:71` honours it:

```jsx
{ id: 'deflated_sharpe_ratio', label: 'Deflated Sharpe' },
```

Headers at `:283-290` are `# · Strategy · Conviction · Sharpe · CAGR · Max DD · Rigor · Forward`. The Rigor cell (`:322`) prints only `DSR conf=` and `· PBO`. `grep -n deflated ui/src/components/Leaderboard.jsx` returns exactly one hit — line 18, the sort option. `grep -n out_of_sample` returns **zero** hits, though it is 25% of the score (`services/leaderboard.py:39`) and appears only as an unlabelled segment of the 120px `ScoreBar` (`:52-68`) whose per-segment `title` is unreachable when the segment is `width: 0%`.

Self-demonstrating on current `main`:

```console
$ cd ui && node -e "
const s=require('fs').readFileSync('src/components/Leaderboard.jsx','utf8');
const b=s.match(/const SORT_OPTIONS = \[([\s\S]*?)\]/)[1];
for (const [,id] of b.matchAll(/id: '([a-z_]+)'/g))
  console.log(id, new RegExp('e\\\\.'+id+'\\\\b').test(s));"
conviction_score true
deflated_sharpe_ratio false   ← offered as a sort, never rendered
dsr_p_value true
sharpe_ratio true
cagr true
pbo_score true
```

### 3 — constant Forward column

`Leaderboard.jsx:324-327`:

```jsx
<span className="tag-muted" title="Per-strategy StockBench eval is pending">SB pending</span>{' '}
<span className="tag-muted" title="Live paper-P&L tracking is pending (testnet — paper)">P&L pending</span>
```

`grep -n forward ui/src/components/Leaderboard.jsx` returns **only the comment on line 11** — `forward.stockbench_status` / `forward.live_pnl_status` (`backend/archimedes/api/leaderboard_schemas.py:45-60`, entry field at `:97`) are dropped. The backend hardcodes them too (`services/leaderboard.py:157`, a bare `LeaderboardForwardAxis()`), so today the literals happen to be correct — and latently false the moment a real status is emitted. The same caveat already appears twice above the table: the `engine.disclaimer` banner at `:207-211` ("…render as 'pending' until that data flows") and the "Forward axis (pending)" card at `:221-228`.

## Acceptance criteria

All commands from the repo root, in the `archimedes` conda env (`export PATH="$(conda info --base)/envs/archimedes/bin:$PATH"`).

**1. The false promise is gone from both the banner and the comment:**

```console
$ grep -n 'refresh on their next' ui/src/components/Leaderboard.jsx | wc -l
       0
```

The replacement copy must state that the figures are fixed when the strategy is generated. Suggested, and pinned by the test below:

> **One caveat on your figures.** These numbers are **fixed at generation time** — a strategy generated before the August engine corrections keeps the numbers it was scored with. Generated strategies are not re-backtested; generate again to get a strategy scored by the current engine.

**2. Every sort option the UI offers renders a value from that same field:**

```console
$ cd ui && node -e "
const s=require('fs').readFileSync('src/components/Leaderboard.jsx','utf8');
const b=s.match(/const SORT_OPTIONS = \[([\s\S]*?)\]/)[1];
let bad=0;
for (const [,id] of b.matchAll(/id: '([a-z_]+)'/g))
  if (!new RegExp('e\\\\.'+id+'\\\\b').test(s)) { console.log('unrendered:', id); bad++; }
console.log('unrendered_count', bad);"
unrendered_count 0
```

`out_of_sample_sharpe` must also be rendered (`e.out_of_sample_sharpe`), labelled, even though it is not currently a sort option — it is 25% of the score.

**3. The constant column is gone:**

```console
$ grep -n 'SB pending\|P&L pending' ui/src/components/Leaderboard.jsx | wc -l
       0
```

**4. A CI-enforced regression test.** Extend the existing leaderboard block in `ui/test/app-visuals.test.js` (see Precedent) with:

```js
test("leaderboard renders every field it sorts by, and no constant forward column", () => {
	assert.doesNotMatch(leaderboard, /refresh on their next/);
	assert.match(leaderboard, /fixed at generation time/);
	const block = leaderboard.match(/const SORT_OPTIONS = \[([\s\S]*?)\]/)[1];
	for (const [, id] of block.matchAll(/id: '([a-z_]+)'/g)) {
		assert.match(leaderboard, new RegExp(`e\\.${id}\\b`), `sort option ${id} has no rendered value`);
	}
	assert.match(leaderboard, /e\.out_of_sample_sharpe/);
	assert.doesNotMatch(leaderboard, /SB pending/);
	assert.doesNotMatch(leaderboard, /P&L pending/);
});
```

```console
$ cd ui && npm test 2>&1 | tail -8
ℹ fail 0
```

(Baseline on `main` today is `ℹ tests 90 / ℹ fail 0`; the count rises by one.)

**5. Lint and build stay clean:**

```console
$ cd ui && npm run lint && npm run build
```
→ exit 0, no ESLint errors.

## Mutation check (required in the PR body)

Per CLAUDE.md § "A guard must be shown to reject something": after the fix, run

```bash
git stash push ui/src/components/Leaderboard.jsx && (cd ui && npm test); git stash pop
```

and paste the failing output. The new test must fail on the unfixed component — a source-text pin that passes both ways guards nothing.

## Anti-goals

- **Do NOT build a generated-strategy re-backtest loop in this PR.** The fix here is copy + rendering. If the loop is wanted, file it separately.
- **Do NOT touch the passport numeric-field data path** (`strategies_routes.py` `_to_strategy_response` / `_passport_to_strategy_response` rigor fields) — issue #1187 and open PR #1340 own that surface; a conflicting edit will collide.
- **Do NOT delete the #1344 assertions at `ui/test/app-visuals.test.js:149-161`** (`known to be incorrect`, `live execution currently interprets`, `awaiting live-trading sign-off`, the `{isOwn && (` gate). They are the standing guard against re-introducing a claim that is now false. The existing `before the August engine corrections` assertion at `:154` can stay — that clause is still true; only the refresh promise is not.
- **Do NOT widen the table without dropping the Forward column** — that column is where the width comes from. Preferred shape holds the count at 8: drop `Forward` (`<th>` at `:290`, `<td>` at `:324-327`), add a `DSR` column from `e.deflated_sharpe_ratio`, and fold OOS into the existing Rigor sub-line at `:322` next to `DSR conf` and `PBO`.
- **Do NOT keep the Forward column with hardcoded literals "for now."** Either bind it to `e.forward.stockbench_status` / `e.forward.live_pnl_status` or remove it — a cell that ignores the field it names is the defect.
- **Do NOT change `SORT_OPTIONS` by deleting "Deflated Sharpe"** to satisfy criterion 2. The value is real and worth showing; render it.

## Precedent

`ui/test/app-visuals.test.js:141-162` — the leaderboard block added by PR #1344: `readFileSync` the component, then `assert.match` / `assert.doesNotMatch` on load-bearing source substrings. Copy that shape exactly; the `leaderboard` binding at `:141-144` already exists. `ui/test/public-visuals.test.js` is the same pattern for the public shell. `npm test` runs unguarded in the blocking `frontend-and-auth` job (`.github/workflows/quality-gate.yml:157`), so these assertions are real CI enforcement, not documentation.

dbrowneup:
Closes #1365

## What was wrong

Three claim-integrity defects on `/app/leaderboard`, verified against `main` @ `523a1bae` (see issue for full evidence):

1. The own-scope caveat banner promised older figures "refresh on their next backtest cycle." No such cycle exists — `own` is generated-strategies-only, and no re-backtest path for a generated strategy exists anywhere in the backend.
2. "Deflated Sharpe" was offered as a sort option and honoured by the backend, but `deflated_sharpe_ratio` was never rendered — picking it silently reshuffled rows with nothing on screen to explain why. `out_of_sample_sharpe` (25% of the conviction weight) was likewise never printed.
3. The Forward column rendered two string-literal constants (`SB pending` / `P&L pending`), byte-identical on every row, never reading `e.forward`.

## What changed

`ui/src/components/Leaderboard.jsx`:

- Rewrote the own-scope banner copy (and the adjacent code comment making the same false claim) to state the true behavior: generated-strategy numbers are fixed at generation time; there is no refresh loop; generating again is the only way to get current numbers.
- Dropped the constant Forward column entirely (bind-or-remove per the issue's anti-goals — removed, since no real per-strategy status exists yet).
- Added a real `DSR` column bound to `e.deflated_sharpe_ratio`, holding the table at 8 columns.
- Folded `out_of_sample_sharpe` into the existing Rigor sub-line next to `DSR conf` and `PBO`.

`ui/test/app-visuals.test.js`: extended the existing leaderboard regression block (added by #1344/#1306) with a new CI-enforced test that pins the banner copy and asserts every `SORT_OPTIONS` id has a rendered `e.<id>` value in the component source, per the issue's Precedent section.

## Acceptance criteria — verified against this branch

```
$ grep -n 'refresh on their next' ui/src/components/Leaderboard.jsx | wc -l
       0

$ cd ui && node -e "... unrendered sort options check ..."
unrendered_count 0

$ grep -c "e.out_of_sample_sharpe" ui/src/components/Leaderboard.jsx
2

$ grep -n 'SB pending\|P&L pending' ui/src/components/Leaderboard.jsx | wc -l
       0

$ cd ui && npm test 2>&1 | tail -8
ℹ tests 128
ℹ suites 0
ℹ pass 128
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
```

(Baseline at this PR's base commit `bef41400` — before `main` moved on — was `tests 127 / pass 127`; count rises by exactly one over that base. `main` has since advanced well past `bef41400` — 155/155 as of `52a09bd5` — so treat "127" as the number at this PR's fork point, not as today's `main`.)

```
$ cd ui && npm run lint
> eslint .
(exit 0, no errors)

$ cd ui && npm run build
✓ 2359 modules transformed.
✓ built in 8.00s
(exit 0)
```

## Mutation check (guard shown to reject unfixed code)

Per CLAUDE.md § "A guard must be shown to reject something":

**Correction (2026-08-20):** the recipe originally published here used `git stash push
<path> && ... ; git stash pop`. `refs/stash` is a single ref in the shared `.git` common
dir, so **every worktree under `.claude/worktrees/` shares one stash stack.** On a clean
tree — the state anyone re-running this recipe post-merge is in — `git stash push <path>`
finds nothing to save, prints "No local changes to save", and exits 0 *without creating an
entry*; the `&&` then runs the test against the already-fixed file (proving nothing), and
the trailing `git stash pop` pops whatever unrelated entry happens to be on top of the
*shared* stack into this worktree. That's exactly what happened when this was re-run: it
popped an unrelated WIP (4 files, 316 insertions, from a different branch's session) into
this worktree. Recovered via `git stash store` back onto the shared stack; no data was
lost, but the recipe was unsafe and is replaced below with one that never touches
`refs/stash`, pinned to this PR's immutable pre-fix base commit so it stays reproducible
after later commits land on this branch:

```
$ git show bef41400e46cb12b1784fd2abe1d04c7bf7f7854:ui/src/components/Leaderboard.jsx > /tmp/leaderboard-unfixed.jsx
$ cp ui/src/components/Leaderboard.jsx /tmp/leaderboard-fixed.jsx
$ cp /tmp/leaderboard-unfixed.jsx ui/src/components/Leaderboard.jsx
$ (cd ui && npm test)
$ cp /tmp/leaderboard-fixed.jsx ui/src/components/Leaderboard.jsx
```

Result — with the component reverted to its unfixed pre-PR state (test file kept, i.e. the new guard runs against the old banner copy):

```
✖ leaderboard renders every field it sorts by, and no constant forward column (#1365) (2.897417ms)
  AssertionError [ERR_ASSERTION]: Expected "actual" to not match "expected"
  expected: /refresh on their next/
  operator: doesNotMatch

ℹ tests 128
ℹ pass 127
ℹ fail 1
```

The new test fails — and only that test fails — against the unfixed component: it catches the exact false claim named in the issue (`refresh on their next backtest cycle` is still present in the unfixed source). Restoring `/tmp/leaderboard-fixed.jsx` returns the fix; working tree confirmed clean afterward, and the full suite passes again (128/128). This result (the *builder's* original run, pre-commit with a dirty tree at the time) was genuine — only the published recipe for *reproducing* it was unsafe in a shared-stash worktree layout.

## Anti-goals honored

- No generated-strategy re-backtest loop added — copy + rendering only.
- `strategies_routes.py` (`_to_strategy_response` / `_passport_to_strategy_response`) untouched — confirmed via `git diff --name-only` (only `Leaderboard.jsx`, `app-visuals.test.js`, and `services/leaderboard.py` touched — the last per the Review follow-up section below).
- The #1344 regression assertions (`known to be incorrect`, `live execution currently interprets`, `awaiting live-trading sign-off`, the `{isOwn && (` gate) are all still present and passing.
- `SORT_OPTIONS` unchanged — `deflated_sharpe_ratio` was not deleted, it's now rendered.
- Table column count held at 8 (`#`, Strategy, Conviction, Sharpe, CAGR, Max DD, Rigor, DSR).
- Forward column is fully removed rather than kept with hardcoded literals.

## Review follow-up (2026-08-20)

Two more claim-integrity/rendering defects found by adversarial review, both fixed on this branch:

- **Dangling leading separator in the Rigor sub-line.** The gate that decides whether to render the sub-line was widened to `dsr_p_value != null || pbo_score != null || out_of_sample_sharpe != null`, but the ` · ` separators stayed hard-attached to the trailing fields. Rows with `dsr_p_value` NULL but `pbo_score`/`out_of_sample_sharpe` set (the July-era passports predating the dsr_p_value persist fix — see `generation_pipeline.py:1451-1453`) rendered a dangling leading `· PBO 0.42 · OOS 0.31`. `main` never showed this because the old gate was `dsr_p_value != null` only. Fixed by building the parts as an array and `.filter(Boolean).join(' · ')`.
- **Stale "pending" disclaimer.** `_DISCLAIMER` in `leaderboard.py` (rendered verbatim on the page) and the file-header comment in `Leaderboard.jsx` both said per-strategy StockBench/paper-P&L "render as 'pending' until that data flows" — true on `main` (literal `SB pending` / `P&L pending` cells) but no longer true after this PR removed the Forward column; nothing renders per-strategy "pending" now, only the unrelated "Forward axis (pending)" section header. Reworded both to say the forward axis "is not scored per strategy yet" instead of claiming a per-row render that no longer exists. `leaderboard.py` is outside the issue's anti-goals (only `strategies_routes.py` is listed), so this is in-scope.

Verified: `cd ui && npm test` → 128/128 pass (no test change needed; `e.out_of_sample_sharpe` and `e.pbo_score` still both appear in the source, so the existing #1365 guard still holds). `ruff format --check` clean on `leaderboard.py`.


dbrowneup:
Fix round applied (transferred from the strategy session's pipeline; its three findings all confirmed real and addressed):
- **major**: the #1365 guard now requires `fmt(`/`fmtPct(`-rendered sort fields — a bare `e.<id>` also matched null-checks inside the render gate. Mutation-proven: removing the OOS render while keeping its null-check fails the new guard (old guard passed that exact input); restored → 12/12.
- **minors**: both `leaderboard_schemas.py` docstrings reworded (API carries the explicit pending value; the UI omits the per-strategy axis — the 'renders an honest pending state' claim was stale), and the #1306 neighbor test retitled off the retired refresh residual.
Verified: ui suite 128/128, leaderboard backend suites 30/30, ruff clean. Branch pushed detached (the strategy session's worktree untouched). Still DRAFT pending re-verdict per house rules.

Issue #1184, opened 2026-07-28
[rigor] Five strategies have a zero-variance persisted return series, misreported as pending


dbrowneup:
## Summary

Five strategies have a persisted return series with realised annualised volatility of
exactly **0.00000** across 5,659 observations — a mathematically constant series (every
daily return identical, most plausibly all-zero). `compute_dsr()` in
`backend/archimedes/services/_rigor_helpers.py` explicitly guards on this: it checks
`np.ptp(arr) == 0.0` (peak-to-peak range) and returns `(None, None)` for a constant
series (confirmed at line 130 — the same guard also appears at line 481 in the
OOS-Sharpe helper). Because the rigor gate treats a `None` DSR as "not yet computed"
rather than as an error, these five strategies surface in the UI as **pending**,
indistinguishable from strategies that are legitimately still queued for evaluation. This
is a silent data-quality hole: a zero-variance 5,659-point return series is not "pending
more data," it is broken data (or a broken backtest), and it is currently invisible as
such.

## Scope (exact files)

- `backend/archimedes/services/_rigor_helpers.py` — `compute_dsr()` (~line 142, guard at
  line 130) and the OOS-Sharpe helper's equivalent guard (~line 481). Read-only reference
  for how `None` is currently produced; the fix should not weaken or remove this guard
  (a constant series genuinely has undefined Sharpe/DSR — the guard is correct behavior
  for the math).
- `backend/archimedes/services/rigor_evaluator.py` — wherever a `None` DSR/PBO result is
  translated into the gate's "pending" status for the API response; this is the surface
  that needs a new, distinct category.
- `backend/archimedes/models/daily_returns_store.py` (`strategy_daily_returns`,
  keyed by `stem`) and/or `backend/archimedes/models/backtest_store.py`
  (`backtest_results`, keyed by `strategy_id`) — wherever the zero-variance series was
  actually written; root-causing whether the backtest produced no trades (flat return by
  construction) versus the writer persisting zeros regardless of the true series requires
  reading the write path in whichever store holds the affected five strategies.

## Acceptance criteria

- [ ] The five affected strategies are identified by id/stem and listed explicitly in the
      PR (not just "five strategies" — name them, the way this issue names
      `capital_preservation_tbill` and `maillard_2010_risk_parity` in the companion
      universe-mismatch issue).
- [ ] Root cause identified for each: did the backtest genuinely produce zero trades (e.g.
      a strategy whose entry condition never fired over the backtest window, which would
[...]

Issue #1326, opened 2026-08-20
[backend][frontend] Persist per-generation cost and surface it: passport card, library column, quote-vs-measured


dbrowneup:
Follow-up to the #1314 cost meter (merged). Today the snapshot lives only on the Redis job record (JOB_TTL = 3600s) and surfaces only via `GET /api/generate/jobs/{id}/cost` and the jobs list — it evaporates an hour after the job ends and no UI shows it. Dan wants it durable and visible.

## Scope

**Persist:**
- [ ] At persist-winner time (and on rigor-failed/errored terminal paths where a strategy row exists), write the `cost_v1` snapshot to the DB, keyed to the strategy/job — new column(s) on the strategy store or a small `generation_costs` table (migration ships with its data rule)
- [ ] Store alongside it the quote that was in force at generation time (the literal `generation_payment.quote()` payload) so quote-vs-measured is an honest pairing of two recorded facts

**Surface:**
- [ ] Strategy passport: a generation-cost card — tokens in/out (with `usage_complete` honesty flag), wall seconds, dominant stage; quoted price shown next to the raw measurement
- [ ] Library table: a compact cost column (e.g. total tokens or wall time; design call in the PR)

**Honesty constraints (unchanged from #1314):**
- No $-conversion of token counts server-side — the flat quote is displayable because it is a recorded fact; converting measured tokens to dollars is #1217's remaining pricing work, not this issue
- A missing/incomplete measurement renders as unknown, never as zero or a fabricated figure

## Acceptance
- [ ] A strategy generated after this merge shows its cost card on the passport and survives Redis TTL expiry (verify: read after >1h or with the job record deleted)
- [ ] Suite green; migration chain guard passes; guard demos for the unknown-vs-zero rendering in the PR body

Anti-goals: don't touch quota/payment enforcement; don't backfill historical strategies (they have no measurements — render honestly as "not measured").

PR #1390, opened 2026-08-20
[frontend][rigor] Read rigor_gate_status tri-state so pending strategies stop rendering as failed

### LINKED ISSUE #1358: [frontend][rigor] The UI drops rigor_gate_status, so every "pending" strategy renders as a failed one
## Summary

The API already ships an honest tri-state. `StrategyResponse.rigor_gate_status` — `"pass" | "fail" | "pending"` (`backend/archimedes/api/schemas.py:221`), set from the live verdict at `backend/archimedes/api/strategies_routes.py:186` under the comment *"rigor_gate_status carries the honest tri-state"*. **No file under `ui/src/` reads it**: `grep -rn 'rigor_gate_status' ui/src/` → **0 hits** (verified on `c0e6400f`, 2026-08-20).

Every UI consumer branches on `passes_rigor_gate`, which is `False` for `pending` by design (`live_rigor_gate.py:88`). So a strategy on which **zero statistics were computed** is presented to the user as a strategy that **failed the rigor gate** — on the Library table, on the mobile cards, on the deployability chip, and on the passport.

This is not an edge case. `verdicts_for_strategies` deliberately degrades the **entire batch** to `pending` when the DB read raises (`backend/archimedes/services/live_rigor_gate.py:227-229`) or when the cohort PBO/correlation compute raises (`:254-256`), with the in-code rationale *"a pending badge, never a wrong PASS"*. A transient Aurora blip therefore paints a red X labelled **"Does not pass rigor gate"** on every row in the library. The backend's fail-closed design is converted into a false user-visible claim at the last hop — CLAUDE.md § *fail-soft is wrong for anything a claim depends on*.

## Evidence (all re-read on `c0e6400f`)

**1 — pending renders as failure, on three surfaces**

- `ui/src/components/Strategies.jsx:221-226` — two branches only:
  ```jsx
  {s.passes_rigor_gate === true && ( … aria-label="Passes rigor gate" … )}
  {s.passes_rigor_gate === false && ( … aria-label="Does not pass rigor gate" … )}
  ```
- `ui/src/components/StrategyPassport.jsx:189` `const passingRigor = s.passes_rigor_gate === true;` → `:609` renders the literal `"not Verified"`. The adjacent comment at `:196-199` even reasons about *"undefined = unknown yet (pending backtest)"*, but `passes_rigor_gate` is `bool = False` (non-optional, `schemas.py:215`) so it is never absent — **that branch is unreachable dead code**.
- `ui/src/components/Strategies.jsx:13-27` `DeployabilityChip`: `min_passing_level == null` → `not deployable`, titled **"Does not pass the rigor gate even at the most permissive level"**. `GET /api/selection-bias/gate/{id}` returns exactly `min_passing_level=None, blocked_by_floor=False, passes_all=False` for a never-scored strategy (`selection_bias_routes.py:601-616`, mirrored at `:343-361` for the cohort route), so the strongest-wording chip lands on the strategies with no verdict at all. `Strategies.jsx:685` stores only `{min_passing_level, blocked_by_floor}` — `gate_details` and `num_trials_scope` are on the wire and discarded (`grep -rn 'gate_details' ui/src/` → 0 hits).
- Backend counts collapse the same way: `selection_bias_routes.py:453-458`, `passing = sum(... r.passes_all)` then `failing=len(results) - passing`. Two buckets, so a row whose `gate_details` are every-key `MISSING (no backtest data)` is reported as **failing**.

**2 — the passport asserts a statistic from a field the API never sends**

`StrategyPassport.jsx:654-669` ternaries on `s.num_trials_in_selection`. `grep -n num_trials backend/archimedes/api/schemas.py` → **no match**; `GET /{strategy_id}` is `response_model=StrategyResponse` (`strategies_routes.py:1965`). The field is therefore always `undefined`, so the `else` branch always wins and the passport unconditionally states:

> *"This strategy is graded on its own Sharpe (num_trials = 1 — no multiple-testing correction applied)"*

The backend does persist a real value and does grade at `N > 1` for search-tracked pipelines (`selection_bias_routes.py:543-545`, `_SCOPE_GENERATED_SEARCH_POOL`; written at `generation_pipeline.py:1519`, `models/backtest_store.py:67`). `Architecture.jsx:496` makes the same claim but correctly scopes it to *"On the curated library"*; the passport does not scope it at all.

**3 — the passport undercounts the gate**

`StrategyPassport.jsx:673-674`: *"A strategy passes the rigor gate only when all three signals align."* The three named metrics are DSR, PBO, OOS. Look-ahead is omitted — and look-ahead is an always-on floor (`rigor_evaluator.py:560`) and therefore the criterion that produces the red **"blocked — correctness floor"** chip rendered 60 lines above at `:612-616`. `RigorExplainer.jsx:20` (*"four independent selection-bias controls"*) and `Strategies.jsx:782-783` (*"DSR + PBO + chronological OOS + look-ahead audit"*) are correct; the passport is the outlier. *Minor, same file:* the one sentence defining "correctness floor" (`StrategyPassport.jsx:303-305`) sits inside the `ROADMAP_SURFACES_ENABLED &&` block opened at `:289` and closed at `:362` — false by default (`featureFlags.js:27`) — so the red chip's own vocabulary is never defined next to it.

**4 — one number, two opposite meanings**

`leaderboard_schemas.py:82-89` documents the wire field: *"DSR confidence in [0,1] — HIGHER IS BETTER. Despite the legacy `p_value` name this is a confidence … not a classical p-value where lower is better."* `services/leaderboard.py:71` sorts it descending and `:105-106` reads it as a confidence. `Leaderboard.jsx:326` is the **only** UI surface that says so. Every other site calls it a p-value: `Strategies.jsx:242` (`DSR p=`), `StrategyPassport.jsx:628` (`p = 0.93`), `RigorStrictnessControl.jsx:127,147`, `RigorExplainer.jsx:32,49,83,263,323,324`, `FusionResult.jsx:112`, `PortfolioAdvisor.jsx:272,297,429`, `RejectedCandidates.jsx:235`. A quant reading the passport's bare `p = 0.93` reads catastrophic; it is the passing case.

## Acceptance criteria

`export PATH="$(conda info --base)/envs/archimedes/bin:$PATH"`, from the repo root.

```bash
# A1 — the tri-state is actually read.
grep -c rigor_gate_status ui/src/components/Strategies.jsx      # -> >= 1
grep -c rigor_gate_status ui/src/components/StrategyPassport.jsx # -> >= 1

# A2 — new ui/test/rigor-tristate.test.js (node --test source assertions,
#      same shape as ui/test/app-visuals.test.js) passes.
cd ui && npm test        # -> "# fail 0"
```
`ui/test/rigor-tristate.test.js` must assert, at minimum: (a) `Strategies.jsx` tests `rigor_gate_status === "pending"` **before** the `passes_rigor_gate === false` X branch; (b) `DeployabilityChip` has a pending branch that renders neither `tag-negative` nor the "even at the most permissive level" title; (c) `StrategyPassport.jsx` renders a third, non-`tag-positive` state for pending; (d) `StrategyPassport.jsx` no longer contains a two-branch ternary on `num_trials_in_selection` — a `null`/untrusted-scope branch exists that renders neither sentence.

**Mutation check (required, CLAUDE.md § "a guard must be shown to reject something"):** delete the pending branch from the `Strategies.jsx` rigor cell, re-run `cd ui && npm test` → non-zero exit, `# fail 1`, naming the pending assertion. Restore. Put the transcript in the PR body.

```bash
# A3 — three buckets on the wire, not two.
pytest backend/tests/test_selection_bias_routes.py -q   # -> "N passed, 0 failed"
```
`GET /api/selection-bias/gate` gains `pending`, and `passing + failing + pending == total`, with every all-`MISSING` row counted pending. **`backend/tests/test_selection_bias_routes.py:248` `test_gate_endpoint_counts_are_consistent` currently asserts `total == passing + failing` and must be updated in the same PR.** Mutation check: revert `failing=` to `len(results) - passing` → the new test fails.

```bash
# A4 — num_trials reaches the passport with its provenance.
python -c "from archimedes.api.schemas import StrategyResponse as S; k=S.model_fields; assert 'num_trials_in_selection' in k and 'num_trials_scope' in k, sorted(k)"   # -> exit 0, no output
pytest backend/tests/test_strategies_routes.py -q       # -> "N passed, 0 failed"

# A5 — copy.
grep -c "all three signals align" ui/src/components/StrategyPassport.jsx   # -> 0
grep -rn -- "p-value" ui/src/ | grep -cv "Not a classical p-value"          # -> 0
grep -rn -- "DSR p" ui/src/components/ | wc -l                              # -> 0

# A6 — whole gate green.
pytest -m "not integration" -q                          # -> "0 failed"
```

## Anti-goals

- **Do NOT make `passes_rigor_gate` nullable or `str | None`.** It is the fail-closed boolean by design (`schemas.py:215`, #821 contract). Add the pending branch on `rigor_gate_status`; leave the boolean's semantics alone.
- **Do NOT use `is_backtest_placeholder` as the pending signal.** It is `not has_real` (`strategies_routes.py:187-189`) — no `BacktestResultRecord` row — and misses the case that matters here: a strategy that *has* a backtest row but fewer than `_MIN_RETURNS_FOR_GATE = 10` usable returns (`live_rigor_gate.py:38,160`), and the whole-batch DB-failure degradation at `:227-229`.
- **Do NOT render pending with `tag-positive`, a check glyph, or any success affordance.** Neutral/muted only; the ask is a loud, visible *absence*.
- **Do NOT add jsdom, vitest, or @testing-library to `ui/`.** `ui/package.json:9` is `"test": "node --test"` and all seven `ui/test/*.test.js` are source-text assertions. A new top-level dep needs sign-off (CLAUDE.md § *When to ask*).
- **Do NOT rename the wire field `dsr_p_value`.** Documented legacy (`leaderboard_schemas.py:82-89`); renaming it is a separate migration. This is a UI-label change only — standardise on "DSR confidence".
- **Do NOT weaken any threshold or floor**, and do not move pending rows into `passing`.
- **Do NOT change `statusLabel`** (`Strategies.jsx:69-74`, `StrategyPassport.jsx:39-46`) — correct as-is.
- **Do NOT un-gate the ROADMAP deploy card** to surface the floor sentence; extract just the sentence at `StrategyPassport.jsx:303-305`.

## Precedent to copy

- **`ui/src/components/Leaderboard.jsx:40-48` — `rigorBadge()` is already the correct three-state shape**: it checks the no-data state *first* and returns a muted "No backtest" before ever reaching pass/fail. Copy this exactly, substituting `rigor_gate_status === "pending"` (strictly better than `is_backtest_placeholder`, see anti-goals).
- **`ui/src/components/Strategies.jsx:585-593`** — the same reasoning already written down in this file, applied to `status` but never to the rigor badge: *"Calling that 'Rejected' is misleading — surface it as 'pending_backtest' so users know what's actually happening."*
- **`backend/tests/test_strategies_routes.py:377,409,441,465,514`** — existing tri-state assertions (`PASS`/`FAIL`/`PENDING`) to model the new backend tests on.
- **`selection_bias_routes.py:126` `num_trials_scope`** — the repo's existing "carry an explicit provenance marker rather than a bare number" pattern; reuse it for the passport caption instead of inferring trust from the value.

## Coordination (all open right now — branch late, rebase before merge)

- **#1349** adds `board_level_fdr` to `RigorGateResponse` and leaves `failing=len(results) - passing` untouched (verified via `gh pr diff 1349`). The `pending` count lands in the same model — rebase on it.
- **#1340** (#1187) makes a pending strategy serve `dsr_p_value / pbo_score / deflated_sharpe_ratio / out_of_sample_sharpe = None`. Complementary, not overlapping: after it merges, a pending strategy shows blank metrics *next to* the red X. This issue is what removes the X.
- **#1348**, **#1331**, **#1334** all touch `ui/src/components/Strategies.jsx`; **#1334** also touches `StrategyPassport.jsx` and `backend/archimedes/api/schemas.py` (where A4's two new fields go).
- **#1184** is the inverse defect (broken zero-variance data misreported *as* pending). Complementary — fixing this one makes #1184's symptom more visible, not less.

## Suggested PR split (one logical change each)

1. Tri-state on the three UI surfaces + `pending` bucket on `GET /api/selection-bias/gate` (A1–A3).
2. `num_trials_in_selection` + `num_trials_scope` on `StrategyResponse`, passport caption made conditional (A4).
3. Copy sweep: "all three signals align" → four; "p-value" → "DSR confidence"; un-nest the correctness-floor sentence (A5).

dbrowneup:
## Summary

The API has served an honest four-state `rigor_gate_status` (`"pass"|"fail"|"pending"|"degenerate"`) since #1184, but nothing under `ui/src/` read it — every UI consumer branched on `passes_rigor_gate`, which is `False` for both `pending` and `degenerate` by fail-closed design. A strategy on which **zero statistics were computed** rendered as a strategy that **failed the rigor gate**: the Library table's X icon, the deployability chip, and the passport's "not Verified" badge.

Also fixes the two named companion defects: the passport's `num_trials_in_selection` ternary asserted a self-contained N=1 grading for strategies the gate never scored (the field didn't exist on `StrategyResponse` at all), and every non-Leaderboard rigor surface labeled the DSR confidence field "p-value"/"p" — the opposite of its documented "higher is better" meaning.

Closes #1358.

> **Merge ordering — read before merging.** This branch has `origin/feat/calm-precision-product-rebrand` (**PR #1469**) merged into it, so its diff against `main` currently contains the whole rebrand (`Landing.jsx`, `App.css`, `Security.jsx`, the font/OG assets, `DESIGN.md`, the `docs/plans/2026-08-2*` rebrand plans, `nginx/nginx.conf`, the `reports/repo-map-*` archives). That is **not** this PR's work and must not be reviewed as such. **Merge #1469 first**; once it lands on `main` the diff here collapses back to the rigor files listed under "What changed". Do not try to strip the rebrand out of this branch — it is a published merge commit.

## What changed

**Backend**
- `GET /api/selection-bias/gate` gains a `pending` bucket. `StrategyRigorResult.pending: bool` is set in both no-data branches (curated cohort loop + per-strategy generated path); `RigorGateResponse.pending: int` sits alongside `passing`/`failing`, with `passing + failing + pending == total` always — previously `total == passing + failing` and every no-data row counted as `failing`.
- `StrategyRigorResult.degenerate: bool`, set from `gate_result.is_degenerate` at **both** computed construction sites (cohort loop + `_generated_strategy_rigor`). This exists because a zero-variance persisted series leaves `dsr_p_value`/`oos_sharpe` at `None`, and `RigorGateResult.blocked_by_floor` treats a `None` on either as a floor failure — so a degenerate row arrives on the wire as `blocked_by_floor=True, min_passing_level=None`, byte-for-byte identical to a strategy that *was* fully graded and found broken. Verified against the real evaluator, not assumed:
  ```
  run_rigor_gate('x', [0.0] * 300, num_trials=1)
    -> is_degenerate=True, tri_state_status='degenerate',
       dsr_p_value=None, oos_sharpe=None,
       blocked_by_floor=True, min_passing_level=None
  ```
  `pending` cannot serve this case: there *is* data here, it is merely flat.
- `StrategyResponse` gains `num_trials_in_selection: int | None` + `num_trials_scope: str`. Curated strategies report the self-contained N=1 the live gate actually used, or `None`/`"unspecified"` when it never ran. Generated/fusion strategies delegate to the same `_num_trials_for_generated_row` provenance discriminator that `selection_bias_routes.py`'s own per-strategy gate uses, so this can never disagree with `GET /api/selection-bias/gate/{id}` for the same strategy.
- `schemas.py`'s `num_trials_in_selection` docstring says explicitly that the `None`/`"unspecified"` shape covers *both* "no persisted returns" *and* a batch/DB-read failure — a reader must not infer which one from this field alone.

**Frontend**
- `Strategies.jsx`: the rigor-cell icon is now a single exhaustive four-arm ternary. `unknown` → em-dash, `degenerate` → neutral alert-circle, `pending` → neutral clock, then the pass/fail icons. `DeployabilityChip` gains `pending` and `degenerate` branches, both checked **before** `blocked_by_floor` and the `min_passing_level == null` fallback — you cannot fail a floor nothing measured you against. Both are `tag-muted`, never `tag-negative`, never the "even at the most permissive level" claim, and the two never share a tooltip (pending says "no backtest data"; degenerate says "zero-variance persisted series" — reusing pending's sentence on a degenerate row would be a fresh false claim, since the data *is* there).
- `StrategyPassport.jsx`: the header rigor tag and the main "Verified" badge both gain neutral `pending` and `degenerate` states, tested ahead of `passingRigor` so a verdict nothing could compute can never reach a "Verified"/"not Verified" claim through a stale boolean. `blockedByFloor` is suppressed for a degenerate row at one place, so all five consumers of it agree; this loosens no permission (`deployable` additionally requires `minLevel != null`, which is `null` here either way) — only the *explanation* changes. The "blocked — correctness floor" chip gets a `title` defining the term, since the one sentence that defined it lived inside the `ROADMAP_SURFACES_ENABLED`-gated deploy card (off by default) — not un-gating the card, just surfacing the definition (per the issue's anti-goal). "A strategy passes the rigor gate only when all **three** signals align" → **four** (DSR, PBO, OOS, look-ahead — look-ahead is an always-on floor that was omitted).
- `StrategyPassport.jsx`'s `num_trials_in_selection` caption is a four-branch ternary, not three. `"generated_untracked_default"` (`schemas.py`) is distinct from "not graded yet": the strategy *was* graded, but its generation pipeline never proved it tracks its own selection-pool size, so `num_trials` was forced to `1` — reusing the true-self-contained N=1 sentence there claimed "only one candidate was ever considered" for a strategy where the honest claim is "we don't know how many candidates were considered." The "not graded yet" sentence no longer asserts "no persisted backtest returns" as a specific fact, since the same `unspecified` scope also covers the batch/DB-failure case.
- New shared module `ui/src/rigorGateStatus.js`: the four known states in one place, plus an `isUnknownRigorGateStatus` / `warnUnknownRigorGateStatus` pair. A fifth state added to the API in future renders an em-dash and a dev-only `console.warn` on **both** surfaces rather than silently falling through to a confident pass/fail. `null`/`undefined` is deliberately *not* "unknown" — coerced generated rows have never carried the field and already have a documented `passes_rigor_gate` tri-state fallback.
- DSR confidence labels standardized from "p-value"/"p"/"p-val" to "confidence"/"conf" across `Strategies.jsx`, `StrategyPassport.jsx`, `RigorStrictnessControl.jsx`, `RigorExplainer.jsx`, `RejectedCandidates.jsx`. UI-only — the wire field `dsr_p_value` is untouched (a separate migration).
- New `ui/test/rigor-tristate.test.js` (source-text assertions, `node --test`, same shape as `ui/test/app-visuals.test.js`) — 12 tests.

## Anti-goals honored

- `passes_rigor_gate` stays the fail-closed, non-nullable boolean — only `rigor_gate_status` and the gate route's new `pending`/`degenerate` fields carry the extra states.
- `is_backtest_placeholder` was **not** used as the pending signal (it's `not has_real`, which misses the "has a backtest row but < 10 usable returns" and whole-batch-DB-failure cases the issue calls out).
- Pending and degenerate never render `tag-positive`, a check glyph, or any success affordance — neutral/muted only.
- No new frontend dependency.
- `dsr_p_value` the wire field is unrenamed.
- `statusLabel` untouched.
- The `ROADMAP` deploy card stays gated; only the correctness-floor sentence's content was surfaced via a `title`, not the card itself.
- No deployability was loosened. The degenerate branch changes only which sentence is shown next to an already-disabled deploy button.

**Scope note (open, not auto-resolved):** issue #1358's own "Suggested PR split" names three logical changes; this PR carries a fourth (the correctness-floor `title` extraction) and a fifth (the `degenerate` state, which is the same defect class as `pending` and was found while fixing it — the fourth state was still rendering as a graded failure after the `pending` half landed). Whether to split is the merger's call; the diff is organized so the degenerate work is isolated in its own commits.

## Verification

All of the following were run on this branch **after** merging current `main` and resolving the conflicts (head `13ff4774`).

**Acceptance checks (issue's exact commands):**

```
grep -c rigor_gate_status ui/src/components/Strategies.jsx        # 6  (>= 1 ✓)
grep -c rigor_gate_status ui/src/components/StrategyPassport.jsx  # 8  (>= 1 ✓)
cd ui && npm test                                                  # tests 476, pass 476, fail 0 ✓
pytest backend/tests/test_selection_bias_routes.py -q              # 63 passed ✓
python -c "from archimedes.api.schemas import StrategyResponse as S; k=S.model_fields; \
  assert 'num_trials_in_selection' in k and 'num_trials_scope' in k"  # exit 0, no output ✓
pytest backend/tests/test_strategies_routes.py -q                  # 38 passed ✓
grep -c "all three signals align" ui/src/components/StrategyPassport.jsx    # 0 ✓
grep -rn -- "p-value" ui/src/ | grep -cv "Not a classical p-value"          # 0 ✓
grep -rn -- "DSR p" ui/src/components/ | wc -l                              # 0 ✓
pytest -m "not integration" -q                                     # 4127 passed, 2 skipped, 2 deselected, 0 failed ✓
```

**Mutation checks (CLAUDE.md "a guard must be shown to reject something"), each demonstrated and reverted:**

1. Deleted the `Strategies.jsx` rigor-cell `isPending` arm, re-ran `node --test ui/test/rigor-tristate.test.js`:
   ```
   ✖ Strategies.jsx checks rigor_gate_status pending before the does-not-pass X branch
   tests 12
   pass 10
   fail 2
   ```
   Restored — 12 passed.

2. Reverted `failing=len(results) - passing - pending` to `failing=len(results) - passing`, re-ran the selection-bias tests:
   ```
   2 failed, 61 passed
   (test_gate_endpoint_counts_are_consistent, test_gate_endpoint_pending_rows_excluded_from_failing)
   ```
   Restored — 63 passed.

3. Hardcoded the curated `num_trials_in_selection`/`num_trials_scope` wiring to always `(1, "curated_self_contained")` regardless of whether the live gate ran, re-ran `test_strategies_routes.py`:
   ```
   FAILED test_strategy_without_real_returns_is_pending
   AssertionError: <id> claimed a num_trials with no live gate run
   1 failed, 37 passed
   ```
   Restored — 38 passed.

4. Removed `degenerate=gate_result.is_degenerate` from `_generated_strategy_rigor` (the path the Passport actually calls for a generated strategy), re-ran `test_selection_bias_generated_gate.py`:
   ```
   1 failed, 13 passed
   ```
   Restored — 14 passed.

5. Reverted `StrategyPassport.jsx`'s `blocked_by_floor === true && !gateDegenerate` back to `blocked_by_floor === true`, re-ran the tri-state file:
   ```
   ✖ StrategyPassport suppresses the correctness-floor claim for a degenerate row
   tests 12
   pass 11
   fail 1
   ```
   Restored — 12 passed.

**New backend tests** (hermetic):
- `test_selection_bias_routes.py` — new: `test_gate_endpoint_pending_rows_excluded_from_failing`, `test_gate_endpoint_scored_row_is_not_pending`, `test_gate_endpoint_degenerate_row_is_distinguishable_from_a_floor_failure`, `test_gate_endpoint_ordinary_failure_is_not_marked_degenerate`. Updated: `test_gate_endpoint_counts_are_consistent` now asserts the 3-bucket invariant `total == passing + failing + pending` (it previously asserted `total == passing + failing`, which is exactly the bug).
- `test_strategies_routes.py` — num_trials-provenance assertions added to `test_strategy_without_real_returns_is_pending` (no live gate run ⇒ `None` / `"unspecified"`) and `test_single_strategy_endpoint_numeric_fields_equal_live_gate` (graded curated ⇒ `1` / `"curated_self_contained"`).
- `test_selection_bias_generated_gate.py` — new: `test_generated_strategy_zero_variance_series_reports_degenerate` (covers `_generated_strategy_rigor`, the path the Passport actually calls for a generated strategy: flat series ⇒ `degenerate=True`, `pending=False`, and `blocked_by_floor=True` anyway — which is the claim the UI must not repeat), `test_strategy_response_num_trials_matches_gate_for_search_tracked_engine`, `test_strategy_response_num_trials_forced_to_one_for_untracked_engine`, `test_strategy_response_num_trials_unspecified_without_backtest` — the last three proving `GET /api/strategies/{id}` agrees with `GET /api/selection-bias/gate/{id}` for the same generated strategy. The pre-existing happy-path test also now asserts `degenerate is False`, so hard-coding `True` fails rather than passes.

**Tallies (head `13ff4774`):**
- `backend/tests/test_selection_bias_routes.py` 63 · `test_strategies_routes.py` 38 · `test_selection_bias_generated_gate.py` 14 — 115 passed, 0 failed
- `pytest -m "not integration" -q` (full suite): 4127 passed, 2 skipped, 2 deselected, 0 failed
- `cd ui && npm test`: 476 passed, 0 failed (`node --test test/rigor-tristate.test.js` alone: 12 passed)
- `cd ui && npm run lint`: clean · `npm run build`: succeeds
- `ruff format --check .`: 622 files already formatted · `ruff check --select E9,F63,F7,F40 .`: all checks passed
- `alembic heads`: exactly one — `5cb798feef58` (no migration in this PR)

## Coordination

- **Base moved a long way.** `main` deleted `GET /api/strategies/advisor` and its helpers (`7c8fc27d`) after this PR was first opened. Two commits on this branch (`288dbfdc`, `9680ee88`) targeted `_build_rigor_summary` inside that endpoint and are now **dead against `main`** — they remain in the branch history but contribute **zero hunks** to the diff. `backend/tests/test_api_routes.py` is byte-identical to `main` in this diff. Any earlier body text describing an advisor `dsr_significant` / `dsr_significant_threshold` behaviour change no longer applies to anything shippable and has been removed.
- Conflicts from the `main` merge and how they were resolved: `strategies_routes.py` — took `main`'s advisor deletion wholesale, keeping only this PR's `num_trials` provenance (the import, the `_to_strategy_response` fields, `_num_trials_for_passport`, and its two call sites); `test_api_routes.py` — took `main` entirely; `docs/README.md` — union of both sides' new plan rows.
- `ui/src/components/PortfolioAdvisor.jsx` and `FusionResult.jsx` no longer exist on `main`; the label renames this PR originally made to them are gone from the diff with them.
- **Merge after PR #1469** (see the ordering note at the top).


onder-akkaya:
The core insight is right and still unfixed on main — `git grep rigor_gate_status origin/main -- ui/src` returns nothing, so the defect #1358 describes is live. This is worth landing. Two problems with the current shape.

**1. The branch is 316 commits / 10 days behind and no longer merges.** `git merge-tree origin/main` reports content conflicts in `strategies_routes.py`, `test_api_routes.py` and `Strategies.jsx`, plus modify/delete on `ui/src/components/PortfolioAdvisor.jsx` and `ui/src/components/FusionResult.jsx` — both deleted on main. Merging as-is resurrects two deleted files. The Round-2 DSR badge-floor fix also already landed on main (`strategies_routes.py:1084-1111`, `c3587702`), so that commit is dead weight now. The green check here describes a base that no longer exists, and re-running it just replays the frozen ref — needs a real rebase and a fresh read.

**2. It handles three of the four states, and `degenerate` is the one it misses.** `StrategyPassport.jsx:203` is `const rigorPending = s.rigor_gate_status === "pending";` — literal equality, nothing else. A degenerate row has `status="degenerate"` and `passes=False`, so it falls to the `s.passes_rigor_gate === false` arm and renders "rigor gate not passed" / "not Verified", with `Strategies.jsx:249` showing an X titled "Does not pass rigor gate".

It never renders as a pass, so it isn't the worst kind of claims violation. But `DeployabilityChip` is: degenerate isn't `pending` (the gate sets that only in the two <10-returns branches), so `min == null` and `Strategies.jsx:23-24` renders "not deployable" titled *"Does not pass the rigor gate even at the most permissive level"*. That's a positive claim about an evaluation that never validly happened on a series with no variance to grade. The backend already has honest copy for this (`portfolio_agent.py:183`, `chat_service.py:395`); the frontend would be the only surface collapsing it.

Timing note: I have a PR up against #1184 that makes the generated/fusion passport path emit `degenerate` for a zero-variance persisted series — today that path reports those rows as `pending` instead. So `degenerate` is about to start arriving on exactly the surfaces this PR touches, which turns the gap above from theoretical into visible. Worth coordinating the two.

**3. No exhaustiveness guard, and the two surfaces disagree on the fallthrough.** A fifth state with `passes_rigor_gate === false` renders "not passed" on both. A fifth state with null/undefined renders the pending clock in `Strategies.jsx` (`isPending = ... || s.passes_rigor_gate == null`) but renders literally nothing in the passport header — the trailing `&&` yields false. One is a plausible substitute, the other is a silent blank; neither is a loud absence. No `default:`, no em-dash, no dev-time warning.

**What's good:** `cd ui && npm run lint` is clean, `node --test` gives 161 passed. Field match confirmed end to end — `rigor_gate_status: str = "pending"` (`schemas.py:225`), emitted at `strategies_routes.py:199`, value set matches what the frontend reads. The backend test pair is the strongest thing here: `test_gate_endpoint_pending_rows_excluded_from_failing` dies against reverting `failing = len - passing - pending`, and `test_gate_endpoint_scored_row_is_not_pending` is a real inverse guard that dies against hoisting `pending=True` onto every row.

`ui/test/rigor-tristate.test.js` is weaker — it's source-text regex over JSX rather than rendering, so it dies against deleting the `isPending` ternary but would also die against a semantically identical refactor, and would pass code where `isPending` is computed and never reaches the DOM. Not vacuous, just brittle. Nothing covers `degenerate` on either side.

Suggested order: rebase first, since a third of the diff is obsolete or touches deleted files. While you're in there, fold `degenerate` into the same neutral treatment as `pending` (the honest copy already exists in `chat_service.py:395`), give `DeployabilityChip` a branch that doesn't claim "fails even at the most permissive level" for an unevaluable series, and add a `degenerate` case to the test file.

dbrowneup:
All three points addressed on `13ff4774` — the first two were sitting as unpushed local commits when you wrote this, completed and re-verified tonight. Point by point:

**1. Stale branch / dead weight — fixed this session.** Merged current main; `288dbfdc` (the Round-2 DSR fix that landed via #1489 as `c3587702`) and `9680ee88` (guarding the deleted `_build_rigor_summary`) now contribute **zero hunks** — `test_api_routes.py` is byte-identical to main, the `PortfolioAdvisor.jsx`/`FusionResult.jsx` hunks died with the files, and the advisor block took main's side wholesale (net `strategies_routes.py` diff is 47 lines, all `num_trials` provenance). The PR body discloses the two dead commits explicitly. Full suite on the merged result: 4127 passed / 0 failed; ui 476/476.

**2. `degenerate` — folded in as its own state, not pending's shadow (`ed2d1566`).** `DeployabilityChip` now has an explicit `deploy.degenerate` branch rendering a muted "unevaluable" tag titled *"zero-variance (broken data or a zero-trade backtest), not a real evaluation"* — the "fails even at the most permissive level" claim you flagged is gone, and the branch deliberately precedes `blocked_by_floor` so the floor can't outrank it. Passport and Library each give degenerate its own sentence (it *has* persisted returns — they're just flat — so it doesn't borrow pending's "no backtest yet" copy). Backend discriminator covered both directions in `a6d19fe1`. Your timing note was right on schedule: with #1519 merged, degenerate is now arriving on exactly these surfaces, and this branch renders it honestly.

**3. Exhaustiveness — one shared module, one loud absence.** New `ui/src/rigorGateStatus.js` holds the frozen four-state list consumed by both surfaces ("if the API grows a fifth state, both surfaces must start being honest about it on the same commit"), an em-dash `UNKNOWN_RIGOR_LABEL` per the #1326 convention with an explanatory title, and a dev-time alarm. The null/undefined generated-row fallback is documented as deliberate rather than treated as unknown, so the alarm can't drown. The fallthrough disagreement you found (clock on one surface, silent blank on the other) is gone — both render the em-dash.

Mutation evidence (each applied, shown to fail, reverted): deleting the `isPending` ternary arm → 2 UI tests fail; removing `degenerate=gate_result.is_degenerate` from `_generated_strategy_rigor` → 1 backend test fails; weakening `blocked_by_floor && !gateDegenerate` → 1 UI test fails. `rigor-tristate.test.js` now covers degenerate on both chip and passport — still source-text regex per the no-vitest constraint, brittle in the fail-safe direction.

Merge note: this branch carries the rebrand merge, so it lands **after #1469**; #1358 gets reopened rather than auto-closed (the frontend fix is not on main until this merges).

Issue #1346, opened 2026-08-20
[infra] Deploy starvation — a fast merge train cancels every build, so main silently diverges from prod for hours


dbrowneup:
## Summary

`deploy.yml` cancels superseded runs. Under a fast merge train that means **no deploy ever completes**, and `main` silently diverges from production for hours.

Observed live on 2026-08-20 while shipping the OOM fixes:

| run | outcome |
|---|---|
| #1200 | queued |
| #1201 | **cancelled** after 1m09s |
| **#1337** | **cancelled** after 2m15s ← the ~1 GB memory fix |
| #1327 | **cancelled** after 1m55s |
| #1335 | **cancelled** after 4m55s |

The last deploy that actually completed was #1328's at **07:36**. At **07:57** the live image was still `be68c2b9` (= #1328), despite five further merges. A build takes ~12–14 minutes; merges were landing every 2–5 minutes, so every build died before finishing.

`git merge-base --is-ancestor <fix-sha> <deployed-sha>` returned **false** the whole time: `main` contained the fix, production did not.

## Why this is its own defect, not #1309

#1309 describes the **deploy rollover window** (old target deregisters before the new task is healthy). That is real and still real. This is a third, distinct failure mode:

1. **Rollover window** — brief 502/503 *during* a deploy (#1309).
2. **OOM crash loop** — was killing the container every ~4 min independent of deploys (fixed by #1328/#1329/#1337).
3. **Deploy starvation** — *this*: under sustained merge pressure, deploys never complete at all.

Mode 3 is the most insidious because there is **no failure signal**. Every run shows `cancelled`, which reads as normal supersession. Nothing reports "production is now N commits and M hours behind `main`." On a build-on-deploy, main-only branch model (`docs/adr/build-on-deploy-main-only.md`), where merging to `main` *is* the deploy, that assumption is silently violated.

It also actively fights incident response: the faster you ship fixes for a production incident, the less likely any of them reach production.

## Acceptance criteria (machine-checkable)

1. A deploy that begins is allowed to finish, or is superseded only by a run that will itself complete. Concretely: change the workflow concurrency so runs **queue** rather than cancel —
   ```yaml
   concurrency:
     group: deploy-main
     cancel-in-progress: false
   ```
   — or add a debounce so a build starts only after N minutes of merge quiet.
   ```
   grep -n -A3 "concurrency:" .github/workflows/deploy.yml
   ```
   → `cancel-in-progress` is not `true` for the deploy job.
[...]

Issue #1309, opened 2026-08-20
[infra] Every deploy causes a ~2-minute 502/503 window — old target deregisters before the new task is healthy


dbrowneup:
## Summary

Observed live 2026-08-20 ~00:00 CT during a routine deploy: the ALB served 502 then 503 for ~2.4 minutes. ECS service events for `archimedes-backend` show the sequence:

```
23:59:04  started 1 tasks (new)
23:59:13  deregistered 1 targets (old) + taskSet began draining   ← 9s after new task START, long before it could pass health checks
00:01:37  registered 1 targets (new)
00:01:56  steady state
```

Between 23:59:13 and 00:01:37 the target group had no healthy target. With `desiredCount=1` and build-on-deploy (every merge to main rolls the service), **every merge carries this window** — busy merge nights multiply it.

`minimumHealthyPercent=100` / `maximumPercent=200` are set, so the old task should have stayed registered until the new one was healthy; the events show it did not.

## Scope

- [ ] Determine why the old target deregisters early despite minHealthy=100 (health-check grace period? TG health-check thresholds vs ECS's RUNNING signal? something in the deploy step forcing a stop?)
- [ ] Fix so the old target stays in service until the new target passes ALB health checks — or weigh `desiredCount=2` (cost vs. zero-blip deploys) as the blunt alternative
- [ ] Verify with a timed probe loop across a real deploy: 0 non-200s end to end

## Anti-goals

Don't reduce health-check strictness to "fix" this — a task that serves traffic before it's actually ready trades a visible blip for silent errors.

dbrowneup:
Cross-reference: three **distinct** failure modes were entangled in tonight's 502/503 observations. Recording them here so the fix for one isn't mistaken for the fix for another:

1. **Deploy-window gap** (this issue as filed): `desiredCount=1`; the old target deregisters ~9s after the new task starts, leaving ~2.4 min with no healthy target on every deploy, despite `minimumHealthyPercent=100`.
2. **OOM crash loop** (fixed in #1328 / #1329 / #1337): `/health` constructed a SentenceTransformer per liveness probe (+521 MB resident, ALB polls every 30s), an unbounded `.all()` pulled every `artifact_json` blob client-side, and a startup rigor backfill retained +1079 MB pinned by the suspended lifespan frame. The backend was OOM-killed roughly every 4 minutes **independent of deploys** — a control task receiving only `/health` died at T+213s.
[...]

PR #1395, opened 2026-08-20
[infra] Harden ALB-target health signal + commit a zero-downtime-deploy probe (#1309)

dbrowneup:
Refs #1309 — **which stays OPEN when this merges.** See "What this does not do" below.

## What this ships

1. **`infra/ecs.tf`** — gives the `nginx` container its own ECS Docker
   `healthCheck`. `nginx` is the actual ALB target
   (`load_balancer { container_name = "nginx" }`) but was the only essential
   container in the task without a health check — `backend` and `auth` both
   have one. Per AWS's documented rolling-deployment semantics, ECS computes a
   task's `healthStatus` from whichever essential containers declare a check; a
   container with none simply doesn't participate. That left the deployment-level
   health signal driven entirely by `backend`+`auth`, with no confirmation that
   `nginx` itself — started only after those two report `HEALTHY` via `dependsOn`,
   then still needing its own envsubst-render + `nginx -t` + listen — is actually
   up. It checks a new container-local `/nginx-health`
   (`nginx/nginx.conf`), **not** the proxied `/health`, so a 10s poller adds zero
   backend load.

2. **`infra/scripts/verify-zero-downtime-deploy.sh`** + the `runbooks/` Phase 5
   rewrite — turns the runbook's copy/paste bash into a committed, runnable
   probe: preflight, 1 req/s against the ALB directly for the whole rollout
   window, `force-new-deployment`, wait for steady state, fail loudly with every
   offending log line if any request was not a `200`.

3. **`backend/tests/test_local_setup_contract.py`** — a cross-file guard tying
   `ecs.tf`'s healthCheck to `nginx.conf`'s location and port, so an unrelated
   nginx edit can't silently turn every nginx container `UNHEALTHY`.

## Merging this does not land the fix — it is inert twice over

Stating this plainly so nobody reads a green merge as the window being closed
(raised by @onder-akkaya; he's right):

1. The `healthCheck` lives in `aws_ecs_task_definition.backend`, so it does not
   exist in AWS until someone runs a **manual `terraform apply`** — Dan-only, per
   the runbook's own header.
2. Only **after** that apply does CI carry it forward, and only on the **first
   deploy after it**: `.github/workflows/deploy.yml`'s `deploy-ecs` job clones the
   **currently registered** task definition (`aws ecs describe-task-definition`,
   `deploy.yml:633`) and rewrites only the three image fields. It never re-derives
   from `infra/ecs.tf`.

A probe run before that apply measures the OLD behaviour. Do not read such a run
— pass or fail — as evidence about this change.

## What this does not do (truth in labeling)

**This instruments the deploy window; it does not close it.** Every knob that
actually governs the gap #1309 measured is untouched here:

- `deregistration_delay` is never set anywhere in `infra/` → AWS default 300s.
  (And it governs how long a *draining* target holds in-flight connections, not
  when deregistration begins — it was never the lever.)
- `healthy_threshold = 2` × `interval = 30` (`infra/alb.tf:234-236`) unchanged, so
  a replacement target still needs ≥60s of passing checks before the ALB routes.
- `deployment_minimum_healthy_percent = 100` / `deployment_maximum_percent = 200`
  unchanged.
- `desiredCount` unchanged — the issue's own named "blunt alternative" is a
  recurring Fargate spend commitment, Dan's call.

**The healthCheck does not explain the observed incident**, and the `ecs.tf`
comment now says so. The timeline is `23:59:04 started 1 tasks (new)` → `23:59:13
deregistered 1 targets (old)` — 9 seconds. nginx only starts after
`backend`+`auth` are `HEALTHY` (backend's own check has `startPeriod = 30`), then
nginx's check has `startPeriod = 15` on top — earliest possible verdict ~T+45s,
with ALB target-group confirmation later still. Nothing here can have influenced
a T+9s deregistration. It ships as forward-looking deployment-health hygiene.
The two candidates that actually fit T+9s stay open and are unexamined from this
environment (no AWS credentials): (a) drift between the live service's applied
`deploymentConfiguration` and this file — it is **not** in the service's
`lifecycle.ignore_changes`, so it only takes effect on the next apply; (b) the
old task already being unhealthy/OOM-killed (mode 2, #1328/#1329/#1337) at
23:59:13 rather than drained by the deployment.

**Both health signals are liveness, not readiness — accepted, and out of scope
here.** `/nginx-health` is a literal `return 200 "ok\n"` the moment nginx listens.
`/health` (`main.py`, the chain-disconnected branch — grep
`HEALTH_CHAIN_DISCONNECTED`) returns **200 while degraded on purpose**: chain
disconnected logs `HEALTH_CHAIN_DISCONNECTED` and still answers 200 so an RPC blip
can't cascade the service down (#1039 N2). So a PASS from this script proves *the
ALB always had something answering 200*, not *the service was functional
throughout*. That's the criterion #1309 actually states, and it's what this
measures. Adding a fail-closed readiness endpoint is a real and separate change —
it would alter what the ALB target-group check gates on, and therefore when ECS
kills tasks, which is exactly the surface this PR is trying not to destabilise
while #1309 is live. Not smuggled in here.

**Accepted residual risk:** nothing guards that `wget` exists in
`nginxinc/nginx-unprivileged:1.31.2-alpine`. It does (busybox), but the blast
radius isn't small — with `deployment_circuit_breaker { rollback = true }`, a
broken `/nginx-health` fails every deploy. The new test guards the config-drift
half; the binary half is only provable by running the image, which is a CI-shape
change rather than a unit test.

## Review round 3 (2026-08-30) — the probe could never have passed

**1. HIGH — the probe form was TLS-impossible.** It sent
`-H "Host: archimedes-arc.com"` to `https://$ALB_DNS/health`. A `Host:` header is
an HTTP-layer thing set *after* the TLS handshake: it changes neither the SNI curl
sends nor the name curl verifies the certificate against. Both come from the URL's
host — the raw ALB DNS name — and the ALB listener certificate
(`aws_acm_certificate.main`, `infra/alb.tf`: `domain_name = var.domain_name`, no
`subject_alternative_names`) covers `archimedes-arc.com` and nothing else. **Every
sample would have been `000` and every run would have reported a total outage, on
a perfectly healthy service.** The one failure mode a zero-downtime verifier must
not have — and it means no previous run of this script proved anything.

Reproduced against a real host rather than argued from the docs:

```
$ curl -o /dev/null -s -w "http_code=%{http_code}\n" -H "Host: example.com" "https://172.66.147.243/"
http_code=000      # curl exit 35 — fails at TLS, request never sent
$ curl -o /dev/null -s -w "http_code=%{http_code}\n" --connect-to "example.com:443:172.66.147.243:443" "https://example.com/"
http_code=200      # same IP, same bypass-the-CDN intent, actually works
```

**Fix:** `--connect-to "$HOST_HEADER:443:$ALB_DNS:443"` with the URL as
`https://$HOST_HEADER/health`. SNI, certificate verification and `Host` all stay
`archimedes-arc.com` while the TCP connection still goes straight to the ALB —
which is the entire point of the probe (measure the target group, not the
CloudFront cache). Deliberately **not** `-k`: that would also "work" while masking
a genuinely broken listener certificate, which is itself an outage this probe
should catch.

**Plus a preflight, so this class of bug fails loudly next time.** The probe now
lives in one `probe_once()` function, and a preflight calls it once before
anything else. Unless that single sample is a `200` the run aborts — *before*
`update-service`, so we never trigger a deploy we have no ability to measure.
Because preflight and loop share the one function, a broken probe form cannot slip
past the preflight.

**2. The cross-file guard was a substring match over a 900-line file.** It asserted
`"http://127.0.0.1:8080/nginx-health " in ecs_tf` — a match *anywhere*. Shown to be
a real hole, not a theoretical one: delete the nginx container's entire
`healthCheck` block and leave the URL behind in a comment, and the old assertion
still passed while the thing it guards was gone. The guard now slices out the
`nginx` **container** (ecs.tf also defines `backend` and `auth`, both of which
already have healthChecks, so a file-wide match stays green with nginx's deleted),
finds a line-anchored `healthCheck = {` inside it, brace-matches that block, and
scopes every assertion to it. It also reads the port out of the healthCheck URL and
requires it to match both `listen <port>;` in `nginx/nginx.conf` and the
container's own `containerPort` — because the check is container-local, so if
`listen` moves the check polls a closed socket and, with the circuit breaker, every
deploy rolls back.

**3. The runbook credited the wrong thing.** Phase 5 said the #1309 violation was
"caught by the (manual, ad-hoc) version of the probe below". It was not, and per
finding 1 could not have been. It now credits what actually recorded it — the ECS
service event timeline, plus @onder-akkaya's 2026-08-30 `504` at 90.4s on `/health`
during a deploy — and explains why the old snippet was blind.

**4. Both of @onder-akkaya's closing notes were already fixed** in the earlier
review commit, and are re-verified here: the redundant `|| echo "000 0s"` (which
produced the mangled `000 0.000367s000 0s` double-verdict) is gone, and a failing
`aws ecs wait services-stable` no longer exits bare `255` — its code is captured,
the window is still scored, the log path is still printed, and a clean window under
a failed waiter reports `INCONCLUSIVE` (exit 1), never `PASS`.

## Earlier review rounds (kept for the record)

**Round 1** — the `nginx` healthCheck originally proxied `GET /health` straight
through to backend (Arc RPC + DB + Redis on every 10s poll, on the same endpoint
named as the OOM crash-loop cause), so "no-added-cost hardening" was false as
shipped; fixed with the container-local `/nginx-health`. The script could report
`PASS` without measuring anything (a bad `--interval` silently killed the probe
loop under `set -e` and the empty log read as "0 non-200s"); fixed with parse-time
`--interval` validation, a `probe_survived` liveness assertion, and a sample-count
sanity check. And it could report a false `FAIL` on pre-trigger baseline noise;
fixed by scoping the verdict to the rollout window and demoting baseline non-200s
to a warning.

**Round 2** — the window scoping compared raw `HH:MM:SS` strings, so a clean
rollout straddling UTC midnight (19:00 CDT — inside Dan's normal working window)
scored `0 probes` and FAILed; fixed by logging epoch seconds as field 1 and keying
all three `awk` comparisons off it. The sample-count check ran *before* `bad_lines`
was evaluated, so a real in-window outage — which depresses the sample count,
because a timed-out probe costs ~6s instead of ~1s, exactly #1309's own 144s-of-180s
shape — was reclassified as "not a reliable measurement" with the offending `000`
lines never printed; fixed by evaluating `bad_lines` first and demoting the
sample-count check to a warning on an otherwise-clean run.

## Guard demonstration

Probe script, against mocked `curl`/`aws` (round-3 scenarios; rounds 1–2 scenarios
re-run unchanged):

```
=== A. Preflight fails — curl returns 000, the exact pre-fix TLS symptom ===
::error::preflight probe returned '000 0.01s', expected 200 — refusing to trigger a deploy this run could not measure.
exit code: 1
  aws ecs update-service calls made: 0     <- never deployed something it couldn't measure

=== B. Preflight OK, clean rollout ===
Preflight OK (200 0.01s).
PASS — 0 non-200 responses across 5 probes during the rollout window.
exit code: 0
  probe argv actually issued:
  curl -o /dev/null -s -w %{http_code} %{time_total}s --max-time 5 \
    --connect-to archimedes-arc.com:443:fake-alb.us-east-1.elb.amazonaws.com:443 \
    https://archimedes-arc.com/health

=== C. `aws ecs wait services-stable` exits 255 (circuit-breaker rollback shape) ===
WARNING — 'aws ecs wait services-stable' exited 255 ...: the service did NOT reach steady state ...
Continuing to score the probe window anyway — the measurement below is the point.
INCONCLUSIVE — 0 non-200 responses across 6 probes, but 'aws ecs wait services-stable' exited 255, so the rollout was never confirmed complete.
Full log kept at /var/folders/.../rollout-watch.....log
exit code: 1
```

Cross-file guard test — six mutations, **all rejected**, with the unmutated
baseline green before and after each:

```
1. nginx healthCheck block deleted                           -> FAIL (rejected)
2. block deleted, URL left behind in a comment               -> FAIL (rejected)   <- OLD guard PASSED this
3. nginx.conf `listen 8080` -> `listen 9090`                 -> FAIL (rejected)
4. path typo /nginx-health -> /nginx-healthz                 -> FAIL (rejected)
5. `location = /nginx-health` renamed away                   -> FAIL (rejected)
6. `proxy_pass` added inside that location                   -> FAIL (rejected)
```

Mutation 2 is the one that matters: it is the exact regression the guard claims to
catch, and the previous substring assertion passed it.

## Verification tally

- `pytest -m "not integration" -q` (repo root, CI's exact command), after merging
  `origin/main`: **4120 passed, 2 skipped, 2 deselected** in 402.69s.
- `pytest backend/tests/test_local_setup_contract.py backend/tests/test_ecs_backend_secrets.py -q`: 28 passed.
- Guard mutation suite: 6/6 mutations rejected (above).
- `bash -n infra/scripts/verify-zero-downtime-deploy.sh`: clean. (`shellcheck` is
  not installed in this environment — not run.)
- Mocked probe-script scenarios: preflight-reject, clean-PASS, failed-waiter
  INCONCLUSIVE (above), plus rounds 1–2's eight scenarios unchanged.
- Real-network reproduction of the TLS defect and its fix (above).
- `ruff format --check .`: 622 files already formatted. `ruff check --select E9,F .`:
  all checks passed. `ruff check` on the changed test file: clean.
- `alembic heads`: exactly one — `5cb798feef58`. (No migrations in this PR.)
- Earlier rounds: `terraform fmt -check -diff` clean; `terraform validate` success;
  `nginx -t` against the rendered config in `nginxinc/nginx-unprivileged:1.31.2-alpine`
  successful, with a live check that `/nginx-health` returns 200 with **no backend
  process listening at all** while `/health` correctly 502s — proving the new check
  truly doesn't touch backend.
- **Not run:** anything requiring AWS credentials or a live ECS service — none exist
  in this environment. Nothing here is proven against the real service.

## Anti-goals honored

- Did not weaken `deployment_minimum_healthy_percent` / `deployment_maximum_percent`
  / the ALB health-check thresholds — the issue's own anti-goal is explicit.
- Did not bump `desiredCount` — the named cost decision, left to Dan.
- Did not touch the OOM-fix or CI-concurrency-fix surfaces (#1328/#1329/#1337, #1346).
- Did not use `curl -k` to make the probe "work" — that would have made a broken
  listener certificate invisible to a probe whose job is to catch outages.
- Did not quietly convert the ALB/container health checks to readiness — a real
  change with real blast radius, deliberately not smuggled into this PR.
- Did not let the probe script keep a guard that could not reject anything.

## What has to happen after this merges

1. **`terraform apply`** (Dan-only) so the task definition actually gains the check.
2. **One CI deploy after that apply**, so the registered task definition carries it.
3. **Run `infra/scripts/verify-zero-downtime-deploy.sh` across a real deploy.**
   #1309 closes on a measured, passing run — not on this merge. @onder-akkaya
   measured a `504` at 90.4s on 2026-08-30, so the window is live and current.


onder-akkaya:
Ran the probe script myself against mocked `curl`/`aws` rather than taking the PR body's word for it, and it does discriminate: clean rollout gives PASS/exit 0 across 13 probes; injecting 503s mid-window gives FAIL/exit 1 with the ten offending lines printed; a total outage with `curl` exiting 7 also gives FAIL/exit 1. `probe_survived` plus the `window_total -eq 0` guard close the "empty log reads as PASS" hole. That was my main worry with a committed probe and it's handled.

The unit test is discriminating too — `test_ecs_nginx_healthcheck_matches_a_real_nginx_conf_location` fails against deleting the `location = /nginx-health` block, against adding a `proxy_pass` into it, and against renaming the path on either side (the trailing-space anchor really does catch `/nginx-healthz`). 7 passed locally.

Two things I'd want said out loud before this merges:

**1. This observes the window, it doesn't close it.** Of the four levers, `deregistration_delay` is never set anywhere in `infra/` — so it's the AWS default of 300s; `healthy_threshold=2 / interval=30` (`infra/alb.tf:234-236`) is untouched; `deployment_minimum_healthy_percent = 100` / `maximum = 200` (`infra/ecs.tf:773-774`) is untouched. The only ECS-side change is the new nginx `healthCheck`. With `desired_count=1` and min-healthy 100 the window shouldn't happen at all, which points at drift or the old task dying rather than at anything this PR touches. I measured a 504 at 90.4s on `/health` during a deploy earlier today, so #1309 is live and current — please don't let it close on this merge.

**2. It's inert twice over, and the PR should say so.** The healthCheck only exists after a manual `terraform apply`, and `deploy.yml:633` clones the currently-registered task definition and swaps only the three images, deliberately never re-deriving from `ecs.tf`. So merging changes nothing; the check reaches prod on the first CI deploy *after* an apply. Worth stating plainly so nobody reads a green merge as the fix landing.

**Should-fix:** neither health signal measures readiness. `/health` (`main.py:559-575`) returns 200 while degraded on purpose — chain disconnected, oracle stale, `strategy_count=0` all warn and still 200. The new `/nginx-health` is a literal `return 200 "ok\n"` the moment nginx listens. So the ALB check, the container check and the probe's own success criterion are all liveness. A PASS from this script proves something answered, not that the new task was ready — which is the same distinction that makes the current window possible. Pointing the probe at a readiness endpoint would make it prove the thing it claims.

**Also worth a thought:** giving nginx a healthCheck makes it participate in task `healthStatus`, and with `deployment_circuit_breaker { rollback = true }` any `/nginx-health` breakage now kills every task. The new test guards the config-drift half; nothing guards `wget` existing in `nginxinc/nginx-unprivileged:1.31.2-alpine`. Low risk since alpine ships busybox wget, but it's unguarded and the blast radius isn't small.

Nit: `|| echo "000 0s"` in the probe loop is redundant — real curl still emits its `-w` output on failure, so the line comes out as `000 0.000000s000 0s`. `$3` is still `000` so the verdict is right; it's cosmetic. Separately, if `aws ecs wait services-stable` fails (circuit-breaker rollback) the script exits 255 with no verdict and doesn't report the log path.

dbrowneup:
@onder-akkaya — closing the loop on your review; the nit at the bottom turned out to be the tip of a worse bug, so this review earned its keep twice.

**Your two closing notes were already fixed** in the review commit that was sitting unpushed locally, and both were re-verified: the redundant `|| echo "000 0s"` is gone (it was producing a mangled double-verdict line), and a failing `aws ecs wait services-stable` no longer exits bare 255 — its code is captured, the window is still scored, the log path prints, and a clean window under a failed waiter reports `INCONCLUSIVE` (exit 1), never `PASS`.

**The worse bug:** the probe's one real-world invocation could never connect — `curl -H "Host: …" https://$ALB_DNS/health` cannot pass TLS verification (the cert is `archimedes-arc.com` only; a Host header changes neither SNI nor hostname checking), so every sample logged `000` and the script failed 100% of the time for reasons unrelated to any deploy. Fixed with `--connect-to "$HOST_HEADER:443:$ALB_DNS:443"` — still ALB-direct, still CloudFront-bypassing, valid cert. The runbook's attribution of the earlier incident to "the manual version of this probe" was corrected too (that form could only ever have produced TLS failures).

**(a) observes-not-closes:** agreed and now stated in the body's first line — merging this does not land the #1309 fix; the four untouched levers (deregistration_delay, ALB thresholds, min/max healthy percent) are listed, and **#1309 stays open** at merge. **(c) liveness-vs-readiness:** accepted as out of scope, said back explicitly rather than left implied. The cross-file guard now pins the healthCheck block and the 8080 pairing — your "URL left behind in a comment" case is one of the five demonstrated rejections.

Issue #1219, opened 2026-08-03
[ui] Breadcrumbs CRUMB_MAP references 10 pages that no longer exist — breadcrumbs silently render nothing


dbrowneup:
## Summary

`ui/src/components/Breadcrumbs.jsx`'s `CRUMB_MAP` is keyed on a **retired navigation scheme**. Of its 12 keys, only `explore` and `reasoning` exist in `App.jsx`'s `PAGE_TO_PATH`.

Dead keys: `strategies`, `trade`, `dashboard`, `mint`, `liquidity`, `vaults`, `create-vault`, `financial`, `vault-detail`, `risk`.

Because `Breadcrumbs` does `const info = CRUMB_MAP[page]; if (!info) return null`, the failure is **silent** — no error, no fallback, breadcrumbs simply do not render on the pages that actually exist. Notably `generate`, `leaderboard`, `library`, `corpus`, `architecture`, `marketplace` and `portfolio` have no entry at all, so the primary user path has no breadcrumbs.

This is the **fourth surface** carrying the retired scheme. The first three were fixed in #1192 (`sitemap.xml`, `index.html` JSON-LD, `/about` + `/imprint`).

## Why it survived

Same reason as the other three: **a dead route in this app fails silently rather than loudly.** `resolveRoute()` reads `pathname` and never `hash`, and `if (!info) return null` swallows the mismatch. Nothing 404s, nothing logs, so nothing surfaces.

## Acceptance criteria

- [ ] `CRUMB_MAP` keys are a subset of `PAGE_TO_PATH` keys. **Add a test asserting exactly that** — this is the part that prevents a fifth occurrence.
- [ ] Group labels reflect current navigation, not the retired `Markets` / `Portfolio` split.
- [ ] Pages on the primary path (`generate`, `leaderboard`, `library`, `corpus`, `architecture`) either have correct breadcrumbs or are deliberately excluded with a comment saying so.
- [ ] `npm run lint` clean in `ui/`.

## Verify

```bash
# every CRUMB_MAP key must exist in PAGE_TO_PATH
grep -n "CRUMB_MAP" -A20 ui/src/components/Breadcrumbs.jsx
grep -n "PAGE_TO_PATH" -A20 ui/src/App.jsx
```

## Anti-goals

- Do not add routes to `PAGE_TO_PATH` to make the map valid. The pages are gone on purpose; the map is what is stale.
- Do not touch `resolveRoute()`.

## Context

[...]

Issue #1318, opened 2026-08-20
[frontend] Accessibility residuals after the WCAG 2.2 AA fix pass (#1311)


dbrowneup:
The #1311 fix pass (PR to follow) implemented and independently verified ~31 findings. These verified gaps remain open, plus the never-assessed half of the standard:

- [ ] 1.4.3 — `--text-4` used as body text fails contrast in the portalled asset modals (base palette, 1.7–1.8:1) and on `.app-site` raised cards (3.9:1); ~175 call sites need triage to `--text-3`, not a token bump
- [ ] 1.4.3 — public `--text-4` is 3.94:1 on `--surface-3`
- [ ] 2.4.11 — `scroll-padding-top` clears the topbar but not the sticky `.explore-subnav`; a focused control can land beneath it on Explore
- [ ] 4.1.2 — Library `aria-controls` points at a row that only exists while expanded
- [ ] Formal conformance evaluation: screen-reader pass (VoiceOver/NVDA), 200%/400% reflow (1.4.4/1.4.10), text-spacing override (1.4.12), motion review, automated axe/Lighthouse run against the live app — none of this has been performed; no conformance claim may be made until it is

Context: the repo has no DOM-rendering test framework; the current guards are static source pins. A component-render a11y test setup would make several of these durable.

dbrowneup:
Status on the five residuals tracked here, after investigation:

**Fixed** (PR #1348):
- 1.4.3 — public `--text-4` was 3.94:1 on `--surface-3`. Confirmed via computed contrast; raised to `#8b9f97` (4.71:1 on `--surface-3`, 5.67:1 on `--surface-2`). No live call site currently pairs the two inside `.public-site` — the raise is defensive, matching how `--chart-label` is already handled.
- 4.1.2 — Library's `aria-controls` pointed at the expanded row's id unconditionally; the `<tr id={detailId}>` only exists while open, so every collapsed row referenced a nonexistent id. Now conditional on `open`, same pattern as `CustomSelect`.

**Investigated, not reproducible on current `main`**:
- 2.4.11 — `.explore-subnav` (the sticky bar this finding names) is dead CSS with zero call sites in `src/`. `git blame` traces it to a `Marketplace.jsx` page removed by a later UI rewrite; the pattern was never ported to today's `Explore.jsx`. There's nothing rendered that could obscure a focused control. Removed the orphaned rules in #1348.

**Still open** (not in #1348, needs more than a source-diff pass):
[...]

PR #1400, opened 2026-08-20
[frontend] Fix shell wayfinding: self-repeating breadcrumbs, exit-the-shell nav item, anonymous sign-out

### LINKED ISSUE #1370: [frontend] Shell wayfinding: self-repeating breadcrumbs, two "Home"s, a nav item that exits the shell, and an anonymous dead end
## Summary

Eight verified wayfinding/copy defects in the authenticated shell. None is individually
severe; together they make the shell read as unfinished on exactly the surfaces a first-time
visitor walks. Three are user-blocking enough to name up front:

- **Every breadcrumb on `/app/corpus` reads `Home / Discover / Corpus`, where `Home` and
  `Discover` are the same button** — both `setPage('explore')`.
- **Clicking `Architecture` in the sidebar destroys the sidebar.** It is listed in the
  DISCOVER nav group but routes to a *public* path, so the whole shell unmounts and the only
  way back drops the user on Explore, not where they were.
- **On the strategy passport — a page deliberately opened to anonymous visitors (#1194 rev d)
  — the page's own back button signs them out of it**, replacing the location with
  `/sign-in`.

All line numbers below verified against `origin/main` @ `523a1bae`.

## Evidence

**1. A breadcrumb trail visits the same page twice.**
`ui/src/components/Breadcrumbs.jsx:55` hard-codes the first crumb:

```jsx
{ label: 'Home', page: 'explore' },
```

and `:26-27` give both Discover children the same ancestor:

```jsx
corpus:       { group: 'Discover', groupPage: 'explore' },
architecture: { group: 'Discover', groupPage: 'explore' },
```

so `/app/corpus` renders `Home → explore / Discover → explore / Corpus`. `/app/explore`
renders `Home / Explore` where `Home` links to the page you are already on.
Related, same mechanism: `:30-32` label the mid-crumb `Strategy` but point it at `generate`,
whose own label is `Generate` (`Layout.jsx:90`). Same for `Position → Portfolio` and
`Market → Marketplace`. There is no section landing page, so the group crumb is a link to a
sibling wearing a section's name.

**2. `/app/account` renders no breadcrumb at all, undocumented.**
`account` is a real route (`ui/src/routes.js:15`, `'/app/account': 'account'`) and a real nav
item (`Layout.jsx:81`), but has no `CRUMB_MAP` entry (`Breadcrumbs.jsx:23-44`) and is not in
the file's own `Deliberately excluded:` list (`:17-20`), so `if (!info) return null` (`:48`)
silently drops the trail. This is the exact silent-failure mode #1219 filed the existing test
file to prevent; the test does not cover this direction.

**3. Two different "Home"s render at once on every interior page.**
`Layout.jsx:23` — `{ id: "landing", label: "Home", ... }` → `pageToPath('landing')` returns
`'/'` (`routes.js:125`), the public marketing page with no sidebar. `Breadcrumbs.jsx:55` —
`{ label: 'Home', page: 'explore' }` → `/app/explore`. Identical labels, ~40px apart, two
destinations, one of which leaves the app.

**4. A sidebar nav item exits the shell.**
`Layout.jsx:30` puts Architecture in DISCOVER beside Explore and Corpus, but
`pageToPath('architecture')` returns `'/architecture'` (`routes.js:126`), which is a
`PUBLIC_PATHS` key (`routes.js:5`), so `App.jsx:111-119` renders it inside `PublicLayout` —
no sidebar, no breadcrumbs, no proof rail. The only route back is `PublicLayout.jsx:26-28`
("Open app" → `/app`), which lands on Explore regardless of where the user came from.
Side effect: the `architecture` entry at `Breadcrumbs.jsx:27` is unreachable dead config —
`Breadcrumbs` is only mounted from `Layout.jsx:320`, which never renders on a public route.

**5. The nav-group comment describes a NAV that no longer exists.**
`Layout.jsx:12` — "The three labelled groups split the remaining surfaces" — while `NAV` has
five labelled groups (`:26`, `:34`, `:47`, `:66`, `:78`). `:16` describes POSITION as
"deployed vaults, on-chain audit, post-hoc review", but `portfolio` and `learnings` are in
`ROADMAP_PAGES` (`ui/src/featureFlags.js:34-42`) and `quant` defaults off in prod
(`backend/archimedes/feature_flags.py:36`), so in the shipped build POSITION is a group header
over the single link "Reasoning".

**6. The passport's back button ejects an anonymous visitor to `/sign-in`.**
`ui/src/components/StrategyPassport.jsx:226-231` (and the error state at `:175-180`):

```jsx
<button className="btn btn-outline btn-sm app-back-link" onClick={() => onNavigate("library")}>
  ← Back to Library
</button>
```

`library` is not in `ANON_APP_PAGES` (`routes.js:41` — `explore, leaderboard, corpus,
strategy`), so `App.jsx:65-68` fires `window.location.replace('/sign-in?next=/app/library')`.
The same component already handles `user == null` correctly 620 lines later
(`StrategyPassport.jsx:848-854`, sign-in link that preserves the passport in `?next=`), so the
back control is the odd one out. Compounding it: no anonymous-reachable page links *into* the
passport — the only `onNavigate('strategy', …)` call sites are `Strategies.jsx:648`,
`PaperTrading.jsx:215` and `RejectedCandidates.jsx:92`, all behind sign-in — so today the
anonymous passport is deep-link-only with a signed-out exit.

**7. Library copy.**
- `Strategies.jsx:927-932` renders `* Pre-backtest hypothesis — …`, but no `*` marker is
  rendered anywhere: `is_backtest_placeholder` occurs exactly twice in the file (`:615` sets
  it, `:927` gates the footnote), and no table cell prints an asterisk. The block also sits
  after every tab panel, so it renders under Generated too, not just Examples.
- `:782-784` — "only strategies that pass DSR + PBO + chronological OOS + look-ahead audit are
  surfaced here. That's the wedge." — contradicts `:761`, `const mainTableStrategies =
  [...passing, ...pending]`. Unscored strategies are surfaced in that table.
- One page, three names: nav "Library" (`Layout.jsx:37`), tab title `Library · Archimedes`
  (`App.jsx:77`), heading "Your Strategies" (`Strategies.jsx:703`) — over a view whose default
  tab includes an Examples set that is explicitly not the user's.

**8. Two false claims in shipped text.**
- `ui/src/components/AccountSettings.jsx:65` — `<p className="body mb-7">Better Auth user owns
  application data. Linked wallets prove on-chain control only.</p>` — user-facing copy naming
  the npm auth library as a noun, articles dropped.
- `ui/src/components/WalletGate.jsx:3` claims the gate covers "(Library, Portfolio,
  Learnings)". Only two call sites exist — `AuthenticatedApp.jsx:158` (Portfolio) and `:176`
  (Learnings); `case "library"` renders `<Strategies>` ungated at `:133-140`.

## Scope

`ui/src/components/Breadcrumbs.jsx`, `ui/src/components/Layout.jsx`, `ui/src/routes.js`,
`ui/src/components/StrategyPassport.jsx`, `ui/src/components/Strategies.jsx`,
`ui/src/components/AccountSettings.jsx`, `ui/src/components/WalletGate.jsx`,
`backend/tests/test_breadcrumbs.py`, `ui/test/routes.test.js`. Do exactly this, nothing more.

## Acceptance criteria

- [ ] **Three new invariants in `backend/tests/test_breadcrumbs.py`** (extend the existing
      file; reuse its `_extract_crumb_keys` / `_extract_page_to_path_keys` helpers rather than
      writing new parsers):
      1. `test_no_page_appears_twice_in_one_trail` — build each trail as
         `[<the Home crumb's page, parsed from the component source, not hard-coded>] +
         [groupPage if group] + [page]` and assert no page appears twice.
      2. `test_every_app_page_has_a_crumb_or_a_documented_exclusion` — every value in
         `APP_PATHS` is either a `CRUMB_MAP` key or named in the `Deliberately excluded:`
         comment block.
      3. `test_shell_nav_items_stay_inside_the_shell` — every `id:` in `Layout.jsx`'s `NAV`
         other than `landing` is a value in `routes.js` `APP_PATHS`.
- [ ] `pytest backend/tests/test_breadcrumbs.py -q` → `8 passed` (5 existing + 3 new).
- [ ] **Mutation check, all three.** **Do NOT use `git stash push <paths>` for this** — every
      worktree under `.claude/worktrees/` shares one `refs/stash` stack, so on a clean tree
      `git stash push <path>` finds nothing to save, exits 0 without creating an entry, and a
      later `git stash pop` pops whatever unrelated entry another worktree's session left on
      top of the *shared* stack into this one. (This is exactly what happened re-running the
      equivalent recipe on #1387/#1365 — see that PR's "Mutation check" section for the
      incident and the recovery.) Use the file-swap form instead, pinned to this issue's
      verified base commit so it stays reproducible after later commits land:
      ```
      $ for f in ui/src/components/Breadcrumbs.jsx ui/src/components/Layout.jsx ui/src/routes.js; do
          git show 523a1bae:$f > /tmp/$(basename "$f").unfixed
          cp "$f" /tmp/$(basename "$f").fixed
          cp /tmp/$(basename "$f").unfixed "$f"
        done
      $ pytest backend/tests/test_breadcrumbs.py -q
      $ for f in ui/src/components/Breadcrumbs.jsx ui/src/components/Layout.jsx ui/src/routes.js; do
          cp /tmp/$(basename "$f").fixed "$f"
        done
      ```
      Expect `3 failed, 5 passed`, with these exact messages (captured against `523a1bae`, so
      this is a real mutation, not a tautology):
      - `trails that visit the same page twice: {'explore': ['explore', 'explore'], 'corpus': ['explore', 'explore', 'corpus'], 'architecture': ['explore', 'explore', 'architecture']}`
      - `APP_PATHS pages with neither a crumb nor a documented exclusion: ['account']`
      - `sidebar nav items that route outside /app (shell disappears): ['architecture']`
      Paste this output in the PR body. Confirm `git status` is clean afterward (the swap-back
      restores the fixed files; nothing should be left modified).
- [ ] **Anonymous back-navigation.** The passport's back control resolves through a pure
      helper, and `ui/test/routes.test.js` gains a case asserting
      `isAnonymousAppPage(<page the helper returns for user === null>) === true` and that the
      signed-in branch still returns `'library'`.
      `cd ui && node --test test/routes.test.js` → `ℹ fail 0`, with the new case in the `✔`
      list (`ℹ pass` goes 14 → 15; #1331 is adding a case to the same file, so assert on
      `fail 0`, not on the total).
- [ ] `grep -c "Better Auth" ui/src/components/AccountSettings.jsx` → `0`
- [ ] `grep -c "That's the wedge" ui/src/components/Strategies.jsx` → `0`
- [ ] `grep -c "Library" ui/src/components/WalletGate.jsx` → `0`
- [ ] `grep -c "Your Strategies" ui/src/components/Strategies.jsx` → `0` — the page heading
      matches the one name used by the nav, the breadcrumb and the tab title. (The replacement
      wording is the implementer's call; this only forbids the odd name out.)
- [ ] `grep -n "is_backtest_placeholder" ui/src/components/Strategies.jsx` shows the footnote
      guard inside the `activeTab === 'examples'` panel (either the `*` marker is rendered on
      the affected rows, or the footnote text drops the orphan `*`).
- [ ] `cd ui && npm run lint` → exits 0 with no output, and `cd ui && npm test` → `ℹ fail 0`
      (`ℹ pass` 90 today, +1 for the new case).

## Verify

Clean clone, `archimedes` conda env, no docker, no `.env`, no network:

```bash
export PATH="$(conda info --base)/envs/archimedes/bin:$PATH"
pytest backend/tests/test_breadcrumbs.py -q          # 8 passed
cd ui && npm test                                    # ℹ fail 0
```

Then, signed out, deep-link `/app/strategy/<any id>` and click "← Back to Library": the URL
must stay under `/app` and must not become `/sign-in`.

## Anti-goals

- **Do NOT rename any nav group** without moving
  `backend/tests/test_breadcrumbs.py::test_crumb_map_group_labels_are_current_nav` in the same
  commit — it hard-codes `allowed = {None, "Discover", "Strategy", "Position", "Market",
  "Ops"}` (`:124`) and `CRUMB_MAP` mirrors those names.
- **Do NOT change the roadmap mid-crumb fallback** (`Breadcrumbs.jsx:56`,
  `!roadmapSurfaceHidden(info.groupPage)`) — verified correct; it is what keeps a crumb from
  linking into a 404.
- **Do NOT touch `PublicLayout.jsx:19`** (`research · rigor · custody`). That wording is
  #1169's call and its acceptance already covers `grep -rn "custod" … ui/src/`.
- **Do NOT rewrite code comments that correctly name the auth library** —
  `BacktestVisualizer.jsx:406`, `GenerationStatus.jsx:11`, `Insights.jsx:14`,
  `WelcomeProfileModal.jsx:73` are accurate and stay. Item 8 is about *rendered copy* only.
- **Do NOT touch `WalletGate.jsx:53`'s connect copy** — unreachable while portfolio/learnings
  are roadmap-hidden. Fix only the stale header comment at `:3`.
- **Do NOT add a `/app/architecture` route as a copy of the public one.** Either move
  Architecture out of the shell nav, or give it one canonical route — not two.
- **Do NOT touch `Strategies.jsx`'s Published tab** — open PR #1331 gates it behind
  `ROADMAP_SURFACES_ENABLED`; rebase on it rather than conflicting.
- **Do NOT weaken or delete any of the five existing breadcrumb tests** to make the new ones
  pass.

## Precedents to copy

- `backend/tests/test_breadcrumbs.py` — hermetic source-parsing test over `Breadcrumbs.jsx` +
  `routes.js`, already has the four extraction helpers. Extend it; do not start a new file.
- `ui/test/routes.test.js:152` (`roadmap surfaces are hidden by default…`) — the shape for a
  pure routing-helper assertion, and where #1331 is adding its sibling test.
- `ui/src/components/StrategyPassport.jsx:848-854` — the correct in-file pattern for a
  `user == null` branch that preserves the current page in `?next=`.
- `ui/src/components/Layout.jsx:240` (`.filter((group) => group.items.length > 0)`) — the
  existing precedent for "never render a group header over nothing", if item 5's POSITION
  group is addressed by moving Reasoning rather than by editing the comment.

dbrowneup:
## Summary

Eight verified wayfinding/copy defects in the authenticated shell (self-repeating
breadcrumbs, a missing breadcrumb, a sidebar item that unmounts the shell, two controls
both labelled "Home", an anonymous visitor's back button signing them out (and
mislabeled after the fix), and library-copy honesty bugs). All fixed in
`ui/src/components/Breadcrumbs.jsx`, `Layout.jsx`, `routes.js`, `StrategyPassport.jsx`,
`Strategies.jsx`, `AccountSettings.jsx`, `WalletGate.jsx`, `Insights.jsx`, plus tests in
`backend/tests/test_breadcrumbs.py` and `ui/test/routes.test.js`.

> **Diff scope warning.** The Files tab shows far more than the ten files above.
> `origin/feat/calm-precision-product-rebrand` ([#1469](https://github.com/a-apin/archimedes/pull/1469))
> is merged into this branch, so its files (`Landing.jsx`, `App.css`, `Security.jsx`,
> `PublicLayout.jsx`, font assets, three plan docs) ride along in the diff. They are
> **not** part of the #1370 review and are reviewed on #1469. See "Merge ordering" below —
> this PR must land *after* #1469.

Addresses #1370 — **not** `Closes`. Eight `CRUMB_MAP` sibling-aliasing mid-crumbs
(`library`/`paper`/`leaderboard` → Strategy/generate, `quant`/`reasoning`/`learnings` →
Position/portfolio, `publish`/`subscriptions` → Market/marketplace) share item 1's exact
shape and are deliberately **not** fixed here; they are tracked in
[#1405](https://github.com/a-apin/archimedes/issues/1405). #1370 stays open until #1405
lands. See "Adversarial-review fixes" item 5 for why they are out of scope for this PR.

**Update (adversarial review pass):** the original push of this branch claimed "all
fixed" while item 3 was genuinely untouched, and missed a mislabeled control the fix
itself introduced. Both are corrected below — see "Adversarial-review fixes" for the
full list and re-verification.

**Update (second adversarial review pass):** three more minor findings — a leftover false
rigor-gate claim four lines below where the first pass fixed the same claim
(`Strategies.jsx`), an overgeneralized code comment claiming item 1's pattern is avoided
everywhere when eight `CRUMB_MAP` entries still have it (`Breadcrumbs.jsx`), and this body's
no-rebase justification naming the wrong set of files as main's drift. All three corrected
in place below (marked "second adversarial-review pass" / "adversarial-review re-pass"),
plus a follow-up issue ([#1405](https://github.com/a-apin/archimedes/issues/1405)) for the
Breadcrumbs scope gap.

## What changed

**Breadcrumbs (item 1, 2, 4)** — `CRUMB_MAP`'s Discover-group pages pointed their mid-crumb
at the same page as the "Home" crumb, so `/app/corpus` read `Home / Discover / Corpus` with
two stops going nowhere new, and `/app/explore` listed `'explore'` twice in one trail.
`account` had no `CRUMB_MAP` entry and no documented exclusion, so its trail rendered
nothing (the #1219 failure mode). `Layout.jsx`'s Architecture nav item routed to the public
`/architecture` path — clicking it inside the app unmounted the whole authenticated shell.
Fix: drop the redundant Discover mid-crumb, add `account` under Ops (later flattened —
see below), remove Architecture from the shell nav (still reachable via `PublicLayout`'s
own nav and direct URL).

**Note on `/app` and `/app/explore`:** these now render **no breadcrumb `<nav>` element at
all** (previously duplicated "Home / Explore" with Home and the current-page crumb both
naming `explore`). This is intentional, not an oversight: `explore` *is* the Home crumb's
own target (`Breadcrumbs.jsx`'s "Deliberately excluded" comment), so giving it its own
`CRUMB_MAP` entry would trip `test_no_page_appears_twice_in_one_trail` by construction —
the same page can't be both the trail's anchor and the trail's leaf. Explore not showing a
breadcrumb is the same UX pattern as a site not showing breadcrumbs on its own homepage.

**Two different "Home"s (item 3)** — `Layout.jsx`'s sidebar anchor item and
`Breadcrumbs.jsx`'s Home crumb were both labelled "Home" ~40px apart with different
destinations: the sidebar item exits the app to the public marketing page (`/`), the
breadcrumb crumb returns to Explore (the anonymous-OK top of the app). Fix: renamed the
sidebar item to "Marketing site" (`Layout.jsx` NAV + `PAGE_LABELS.landing`) so the two
controls are no longer visually and semantically indistinguishable.

**Anonymous back-navigation (item 6)** — the strategy passport is deliberately reachable by
anonymous deep-link (#1194 rev d), but both "← Back to Library" buttons called
`onNavigate("library")` unconditionally. `library` is wallet-gated, so this tripped
`App.jsx`'s anonymous-page redirect and replaced the URL with `/sign-in` — signing out a
visitor who was never signed in. Fix: a pure helper `passportBackPage(user)` resolves to
`'explore'` (anonymous-OK) when signed out, `'library'` when signed in — **and** a sibling
`passportBackLabel(user)` keeps the button's own text ("← Back to Explore" /
"← Back to Library") in sync with where it actually navigates, so the control never reads
"Library" while landing on Explore.

**Library copy (item 7)** — heading read "Your Strategies" against nav/tab-title/breadcrumb
all saying "Library"; the rejected-only empty state claimed "only strategies that pass ...
are surfaced here" while the same block's `mainTableStrategies` includes pending candidates
too; the `is_backtest_placeholder` footnote rendered at the page level under every tab, not
just Examples. Fix: rename heading to "Library", rewrite the claim to match what's shown
(both the empty-state copy and a nearby code comment that still asserted the old, stricter
claim three lines above the `[...passing, ...pending]` line that disproved it — comment
and a second empty-state string both corrected in the adversarial-review pass), move the
footnote inside the Examples panel.

**Internal-library naming (item 8)** — `AccountSettings.jsx` named the npm auth library as a
user-facing noun; `WalletGate.jsx`'s header comment claimed it covers Library, which
actually renders ungated; `Insights.jsx`'s rendered `<p>` copy (not a comment — see
"Adversarial-review fixes") also named it. Fix: rewritten to describe what's actually true
in all three.

Item 5 (stale nav-group comment — "the three labelled groups" against an actual five) is
folded into the breadcrumb/nav-comment rewrite above as supporting context — no standalone
acceptance check targets it separately per the issue, and the comment now correctly
describes all five groups (Discover/Strategy/Position/Market/Ops).

## Adversarial-review fixes

A review pass on the pushed branch found six confirmed defects, fixed in a follow-up
commit with the same mutation-check discipline as the rest of this PR:

1. **Item 3 was genuinely untouched** despite the original "all fixed" / `Closes #1370`
   claim — fixed now (see "Two different Home's" above), so the claim is accurate.
2. **The passport back button's label didn't track its own routing fix** — it still read
   "← Back to Library" while landing anonymous visitors on Explore. Fixed with
   `passportBackLabel(user)`, both call sites.
3. **The passport regression guard in `routes.test.js` only imported `routes.js`** —
   reverting `StrategyPassport.jsx`'s actual wiring back to the pre-fix
   `onNavigate("library")` + hard-coded label left the suite fully green, so it proved
   nothing about the component itself. Added a source-text test over
   `StrategyPassport.jsx` (precedent: this file's own `strategiesSrc` test,
   `backend/tests/test_breadcrumbs.py`'s source-parsing pattern) asserting both back
   buttons call `onNavigate(passportBackPage(user))` and render
   `{passportBackLabel(user)}`, and that no bare `onNavigate("library")` or hard-coded
   Library label remains. **Mutation check:** reverted `StrategyPassport.jsx` to the
   exact prior regression, ran `node --test test/routes.test.js` → `16 passed, 1 failed`
   (only the new guard failed, with the exact assertion message below), restored, reran →
   `17 passed, 0 failed`.
   ```
   AssertionError [ERR_ASSERTION]: expected both back buttons to call
     onNavigate(passportBackPage(user)), found 0
   undefined !== 2
   ```
4. **`Insights.jsx:160` names the auth library in rendered `<p>` copy**, not a comment —
   the original Anti-goals bullet claimed all four "auth-library-naming" sites
   (`BacktestVisualizer.jsx`, `GenerationStatus.jsx`, `Insights.jsx`,
   `WelcomeProfileModal.jsx`) were comments and untouched, but `Insights.jsx` has both: an
   accurate comment at `:14` (correctly left untouched, matches the issue's own anti-goal)
   *and* rendered copy at `:160` the issue never called out. Rewrote `:160` to "canonical
   accounts" and corrected the Anti-goals bullet below.
5. **The new `account` breadcrumb entry reintroduced item 1's exact pattern** — a group
   crumb ("Ops") wearing a section name while actually linking to a sibling page
   (Insights), for a group with only two members and no real section landing page.
   Flattened `account` to `{ group: null, groupPage: null }`, same shape as the other
   anchor-only entries (`insights`, `portfolio`, `marketplace`, `generate`, `corpus`); the
   trail now reads "Home / Account". **Scope note (adversarial-review re-pass):** this
   only fixes the `account` instance. `library`/`paper`/`leaderboard` → Strategy/generate,
   `quant`/`reasoning`/`learnings` → Position/portfolio, and `publish`/`subscriptions` →
   Market/marketplace share the identical shape and are **not** touched by this PR — item
   1's own text names Strategy/Position/Market as further examples of the same mechanism,
   so those three are a deliberate deferral, not a fixed set, and are not implied fixed by
   this bullet or by the `Breadcrumbs.jsx` comment. No #1370 acceptance check requires
   them (those pages are genuinely distinct, so `test_no_page_appears_twice_in_one_trail`
   doesn't fire on them). **Correction (merge-readiness pass):** an earlier version of this
   bullet concluded "`Closes #1370` still holds on that basis" — it does not. Passing every
   automated acceptance check is not the same as closing the issue, and item 1's own text
   names the deferred siblings as instances of the defect. The link is now `Addresses #1370`
   (see Summary); #1370 stays open until
   [#1405](https://github.com/a-apin/archimedes/issues/1405) lands.
6. **`routes.js`'s `ANON_NAV_IDS` still listed `'architecture'`** after Architecture was
   dropped from the shell nav in this same PR — the identical unreachable-dead-config
   pattern the PR's own `Breadcrumbs.jsx` comment cites as the reason to delete the
   `CRUMB_MAP` `architecture` key. Dropped from the set.
7. **(Second adversarial-review pass)** `Strategies.jsx:838-839` still had the pre-fix
   claim four lines below where item 7 above ("Library copy") fixed the same claim in the
   same empty-state card: "They land in this table once the rigor gate clears" directly
   contradicted the corrected sentence right above it ("appear here once they've been
   backtested; the rigor verdict ... lands here after"). The table is built as
   `[...passing, ...pending]`, so a candidate lands before any verdict — the acceptance
   grep (`"That's the wedge"` → 0) doesn't cover this string, so it survived. Rewrote to
   "They land in this table once backtesting finishes, verdict or not."

Two more findings from the same pass were investigated and **not** changed:

- **Explore's breadcrumb disappearing entirely** was flagged as undisclosed. Traced to the
  `explore` page being the Home crumb's own target (documented in `Breadcrumbs.jsx`'s
  "Deliberately excluded" block) — adding a `CRUMB_MAP` entry for it would make
  `test_no_page_appears_twice_in_one_trail` fail by construction. Intentional; now stated
  explicitly above instead of left implicit.
- **The branch was behind `main`.** Resolved, not deferred: `origin/main` (`3dd63e5e`) is
  now merged in (repo convention is merge commits, not rebase — published history stays
  intact). Only one conflict, in `docs/README.md`: both sides appended rows to the
  "Plans and roadmaps" table (this branch's three rebrand plan docs vs main's
  `2026-08-30-intraday-paper-trading.md`). Resolved by keeping all four rows in date order
  — purely additive, no content dropped from either side.

  Two files are touched by both sides and auto-merged; both were checked semantically, not
  just for absence of conflict markers. `ui/src/components/StrategyPassport.jsx` — main
  added the owner-gated brief panel (#1547), this PR rewired the two back buttons; both
  survive (`grep` confirms `passportBackPage(user)`/`passportBackLabel(user)` at both call
  sites *and* the brief rendering). `ui/src/components/Strategies.jsx` — main added the
  paper-citation row and drawdown fix (#1541), this PR rewrote the heading and three
  empty-state strings; `git diff origin/main -- ui/src/components/Strategies.jsx` removes
  only this PR's own pre-fix copy and nothing of main's.

## Merge ordering (read before merging)

1. **Merge after [#1469](https://github.com/a-apin/archimedes/pull/1469)** (calm-precision
   product rebrand). This branch has `origin/feat/calm-precision-product-rebrand` merged
   into it, so its diff carries the rebrand's files (`Landing.jsx`, `App.css`,
   `Security.jsx`, the font assets, the three plan docs). Merging this PR before #1469
   would land the rebrand under the wrong PR's review.
2. **Merge before [#1437](https://github.com/a-apin/archimedes/pull/1437)** (Insights
   admin-gate + dashboard v2), and #1437 needs a fix afterwards. This PR's
   `test_shell_nav_items_stay_inside_the_shell` parses the `const NAV = [...]` block out of
   `Layout.jsx` (`_extract_layout_nav_ids`, hard `assert m, "NAV block not found in
   Layout.jsx"`). #1437 moves that block to a new `ui/src/navConfig.js` as `export const
   NAV` **and its copy still contains `{ id: "architecture", ... }`** — the exact
   exit-the-shell nav item this PR removes. So:
   - Landing #1437 *second, unchanged* fails CI loudly on `test_breadcrumbs.py` ("NAV block
     not found in Layout.jsx") rather than silently regressing — that is the safe order, but
     #1437 must then re-point the guard at `navConfig.js` **and** drop the `architecture`
     entry, or main gets the exit-the-shell bug back with a guard that no longer looks at
     the file the nav actually lives in.
   - Landing #1437 *first* is worse: this PR would have to re-do both fixes on top, and in
     the window between them main carries the bug.

## Guard transcripts

**Mutation check** (`backend/tests/test_breadcrumbs.py`, three new tests) — per repo
convention, reverted the fix and confirmed the new tests fail with the exact messages, then
restored:

```
$ git show HEAD:ui/src/components/Breadcrumbs.jsx > ui/src/components/Breadcrumbs.jsx  # (pre-fix)
$ git show HEAD:ui/src/components/Layout.jsx > ui/src/components/Layout.jsx            # (pre-fix)
$ git show HEAD:ui/src/routes.js > ui/src/routes.js                                    # (pre-fix)
$ pytest backend/tests/test_breadcrumbs.py -q
...
FAILED test_no_page_appears_twice_in_one_trail - AssertionError: trails that visit the
  same page twice: {'explore': ['explore', 'explore'], 'corpus': ['explore', 'explore',
  'corpus'], 'architecture': ['explore', 'explore', 'architecture']}
FAILED test_every_app_page_has_a_crumb_or_a_documented_exclusion - AssertionError:
  APP_PATHS pages with neither a crumb nor a documented exclusion: ['account']
FAILED test_shell_nav_items_stay_inside_the_shell - AssertionError: sidebar nav items
  that route outside /app (shell disappears): ['architecture']
3 failed, 5 passed
$ # restored fixed files
$ pytest backend/tests/test_breadcrumbs.py -q
8 passed
```

**Anonymous back-navigation guard** (`ui/test/routes.test.js`) — reverted `routes.js` +
`StrategyPassport.jsx`, confirmed the new test's import fails against unfixed code (the
helper doesn't exist yet), then restored:

```
$ git show HEAD:ui/src/routes.js > ui/src/routes.js                        # (pre-fix)
$ git show HEAD:ui/src/components/StrategyPassport.jsx > ui/src/components/StrategyPassport.jsx
$ node --test test/routes.test.js
SyntaxError: The requested module '../src/routes.js' does not provide an export named
  'passportBackPage'
tests 1, pass 0, fail 1
$ # restored fixed files
$ node --test test/routes.test.js
tests 17, pass 17, fail 0
```

**Passport-wiring source guard** (adversarial-review pass, `ui/test/routes.test.js`) —
reverted only `StrategyPassport.jsx`'s two call sites to the exact prior regression while
leaving `routes.js` fixed, confirmed only the new guard fails, then restored:

```
$ # onClick={() => onNavigate("library")} + hard-coded "← Back to Library" at both sites
$ node --test test/routes.test.js
✖ the strategy passport component actually wires the back-navigation helpers at both
  back-button call sites (#1370)
  AssertionError: expected both back buttons to call onNavigate(passportBackPage(user)),
  found 0
  undefined !== 2
tests 17, pass 16, fail 1
$ # restored fixed files
$ node --test test/routes.test.js
tests 17, pass 17, fail 0
```

## Verification tallies

Re-run in full **after merging `origin/main` (`3dd63e5e`)**. The counts below supersede the
earlier ones in this body (`17` / `129` / `3554`), which were measured against a much older
`main` and are no longer the numbers anyone should check against.

```
$ export PATH="$(conda info --base)/envs/archimedes/bin:$PATH"
$ pytest backend/tests/test_breadcrumbs.py -q
8 passed                                          # 5 existing + 3 new

$ cd ui && node --test test/routes.test.js
tests 23, pass 23, fail 0                         # grew with main's own route cases

$ npm test
tests 466, pass 466, fail 0

$ npm run lint
(exit 0, no output)

$ npm run build
✓ built in 1.80s

$ ruff check backend/tests/test_breadcrumbs.py
All checks passed!

$ ruff format --check backend/tests/test_breadcrumbs.py
1 file already formatted

$ pytest -m "not integration" -q          (repo root, full suite)
4122 passed, 2 skipped, 2 deselected in 384.85s

$ cd backend && python -m alembic heads
5cb798feef58 (head)                               # exactly one head, matches main's chain
```

## Adversarial pass on the hardened exclusion guard

Önder's review said `test_every_app_page_has_a_crumb_or_a_documented_exclusion` reads
stronger than it is: the exclusion set is parsed out of a `//` comment, so it guards against
a *silently* omitted crumb, not a missing one. The docstring now says that outright
("do not cite this test as evidence that breadcrumbs are complete"), and
`_extract_excluded_pages` enforces the bullet grammar it previously only documented.

A guard is only worth what its failing cases prove, so all five inputs that *should* fail
were written and run against the merged tree. Every one fails, with the message quoted:

| Mutation to `Breadcrumbs.jsx` | Result |
|---|---|
| `//   - account` (bullet, no ` — reason` half) | ✗ `malformed 'Deliberately excluded' bullet (no ' — reason' half)` |
| `//   - account — TODO` (placeholder reason) | ✗ `gives reason 'TODO' (4 chars, need >= 20)` |
| `//   - accountz — typo'd page name that is not a route at all` | ✗ `names pages that are not routes at all: ['accountz']` |
| `//   - corpus — …` (excluded *and* in `CRUMB_MAP`) | ✗ `pages listed as deliberately excluded but also present in CRUMB_MAP: ['corpus']` |
| delete the `account:` crumb entirely (the original #1370 item-2 regression) | ✗ `APP_PATHS pages with neither a crumb nor a documented exclusion: ['account']` |

All five pass the pre-hardening parser. Restored after each case; tree verified clean.

Acceptance-criteria greps, re-run against the merged tree. **One of them had to be
corrected — the check was wrong, not the code:**

```
$ grep -c "Better Auth" ui/src/components/AccountSettings.jsx        → 1   ← NOT 0 any more
```

This grep is stated as evidence elsewhere in this body as `→ 0`, and that is now
misleading. `main` (`496ab312`, the send-verification-email button) added an accurate
**code comment** at `AccountSettings.jsx:20` explaining that Better Auth's
`/send-verification-email` returns `{status:true}` on *queue*, not on delivery. #1370's
acceptance criterion is about the auth library appearing in **rendered copy**, and the
issue's own anti-goals explicitly protect accurate comments that name it. So the comment
stays and the grep is replaced with one that measures the actual criterion:

```
$ grep -n "Better Auth" ui/src/components/AccountSettings.jsx | grep -vc '^[0-9]*: *//'
0                                                                    # rendered copy only
```

Adversarially checked rather than asserted — injecting a rendered-copy regression
(`const REGRESSION = <p>Signed in with Better Auth</p>`) makes the corrected grep return
`1`; removing it returns `0` again. The naive `grep -c` returns `1` either way, so it could
no longer tell a regression from `main`'s comment.

The rest hold unchanged:

```
$ grep -c "That's the wedge" ui/src/components/Strategies.jsx        → 0
$ grep -c "Library" ui/src/components/WalletGate.jsx                 → 0
$ grep -c "Your Strategies" ui/src/components/Strategies.jsx         → 0
$ grep -c "rigor gate clears" ui/src/components/Strategies.jsx       → 0
$ grep -n "is_backtest_placeholder" ui/src/components/Strategies.jsx
664:     is_backtest_placeholder: true,
1010:          {examples.some(s => s.is_backtest_placeholder) && (
```

(Line numbers moved with the merge; `1010` is still inside the `activeTab === 'examples'`
block that opens at line `983`, which is the property item 7 actually asserts.)

## Manual verification note

The issue's "Verify" section also asks: signed out, deep-link `/app/strategy/<id>`, click
"← Back to Library" (now "← Back to Explore" for an anonymous visitor — see
"Adversarial-review fixes" item 2) — URL must stay under `/app`, never become `/sign-in`.
This session had no browser available to click through it live; verified by code trace
instead: `passportBackPage(null)` → `'explore'` → `pageToPath('explore')` → `/app/explore`,
and `'explore'` is in `ANON_APP_PAGES`, so `App.jsx`'s anonymous-page redirect effect
(`route.anonymousOk` short-circuit) never fires. Worth a real click-through before merge.

## Anti-goals respected

- No nav group renamed — `test_crumb_map_group_labels_are_current_nav` untouched and green.
- Roadmap mid-crumb fallback (`roadmapSurfaceHidden`) untouched.
- `PublicLayout.jsx` untouched **by the #1370 work**. ⚠️ **Correction (merge-readiness
  pass):** this bullet used to read "untouched (not in the diff)", and that is no longer
  true — `PublicLayout.jsx` now shows `+49/-10` against `main`. Those hunks come from the
  calm-precision rebrand merged into this branch (`ac7c0717` redesign-public-landing,
  `9f2e76a9` security-posture-page), i.e. from
  [#1469](https://github.com/a-apin/archimedes/pull/1469), not from this PR's own commits.
  Reviewing #1400 for #1370 should read `git diff` restricted to the ten #1370 files; the
  rebrand's share of the diff is reviewed on #1469 and is the reason for the merge-ordering
  constraint above.
- Auth-library-naming code **comments** that are accurate and untouched, per the issue's
  anti-goal: `BacktestVisualizer.jsx:437`, `GenerationStatus.jsx:11`, `Insights.jsx:16`,
  `WelcomeProfileModal.jsx:73`, and now also `AccountSettings.jsx:20` (added by `main`'s
  `496ab312` after this PR opened — see the acceptance-grep correction above). Line numbers
  refreshed against the merged tree; the earlier `:406`/`:14` were pre-merge positions.
  (Corrected from an earlier draft of this bullet, which mischaracterized `Insights.jsx`'s
  *rendered* copy at `:160` — a different location the issue never named — as one of these
  comments; that rendered copy is fixed, see "Adversarial-review fixes" item 4.)
- `WalletGate.jsx`'s connect copy untouched — only the header comment changed.
- No new `/app/architecture` route added — Architecture removed from the shell nav instead,
  per the issue's explicit either/or.
- `Strategies.jsx`'s Published tab (#1331) untouched.
- All five pre-existing `test_breadcrumbs.py` tests unchanged and still passing.


onder-akkaya:
**Heads up — this PR and 1437 collide semantically, and neither CI run can see it.** `git merge-tree` is clean in both directions, both PRs are green, and whichever lands second turns `main` red.

The collision:

- **#1400** adds `backend/tests/test_breadcrumbs.py`, which does `re.search(r"const\s+NAV\s*=\s*\[(.*?)\n\];", layout_text)` against `ui/src/components/Layout.jsx` and then `assert m, "NAV block not found in Layout.jsx"`.
- **#1437** deletes `const NAV = [` from `Layout.jsx` and recreates it in a new `ui/src/navConfig.js`.

So `test_shell_nav_items_stay_inside_the_shell` fails the moment both are on `main`.

There is a second, quieter half. `navConfig.js` reinstates `{ id: "architecture", label: "Architecture", icon: "i-lucide-network" }` — the exact nav item #1400 removes for #1370 item 4, because it routed to `PublicLayout` and unmounted the shell. If #1437 lands second it silently restores the bug #1400 fixed, and no test catches that either.

Both PRs are 88 commits behind `main` (opened 2026-08-20), so their green checks describe a base from nine days ago regardless. `CLAUDE.md` § *the green check may not mean what you think*, rules 1 and 2, cover both halves of this.

Suggested order: **land #1400 first, then rebase #1437 onto it.** That way #1437 has to reconcile `navConfig.js` against the removed Architecture item and against the NAV-location test, rather than reverting both by accident. Whichever way it goes, re-run `pytest -m "not integration"` from the repo root on the *merged* result before the second merge — not on either branch.

Separately, coupling a Python test to the declaration site of a `const` in one JSX file is brittle no matter which PR wins. Asserting on the exported nav list (or on the rendered nav) would survive a refactor that #1437 shows is already wanted.

onder-akkaya:
Verified on a tree merged with current `main`, `ui` after `npm ci`:

```
npm run lint                          exit 0, no output
node --test                           tests 411, pass 411, fail 0
pytest backend/tests/test_breadcrumbs.py -q    8 passed
```

The anonymous dead end is genuinely closed. Sign-out on a gated page goes through App.jsx's redirect to `/sign-in?next=`, and on `/app/strategy/<id>` — which is anonymous-OK — the back button now reads "← Back to Explore" and resolves to `explore`, which `isAnonymousAppPage` allows. I went looking for a route where an anonymous user lands with no forward affordance and didn't find one.

Mutations both fail, so the tests discriminate: restoring `explore: { group: null }` plus `corpus: { group: 'Discover', groupPage: 'explore' }` fails `test_no_page_appears_twice_in_one_trail` with the colliding trail printed; reverting both `StrategyPassport.jsx` call sites to `onNavigate("library")` gives `tests 23, pass 22, fail 1`. Worth noting the pure-helper test alone would have been vacuous — it never imports the component — and the source guard you added is what makes it real.

**`Closes #1370` overstates it.** Only `corpus` (plus the deleted `explore`/`architecture` keys) is fixed. `library`/`paper`/`leaderboard` → Strategy/generate, `quant`/`reasoning`/`learnings` → Position/portfolio, and `publish`/`subscriptions` → Market/marketplace all still carry a section-named mid-crumb linking to a sibling page. You document that at `Breadcrumbs.jsx:51-60` and #1405 tracks it, which is fine as scoping — but #1370 item 1's own text names those three, so auto-closing it puts "done" in the record for work that's still open. `Addresses #1370` and leave it.

**`_extract_excluded_pages` is weaker than it reads.** The documented-exclusion set is parsed out of a `//` comment block, so `test_every_app_page_has_a_crumb_or_a_documented_exclusion` is satisfiable by writing a comment — which is how `explore` and `architecture` clear it here. It's a guard against *silent* omission, not against a missing crumb. That's a legitimate thing to guard, but the docstring should say which one it is, or the next reader will over-trust it.

Small a11y item while the file is open: the terminal crumb is a non-interactive `<span className="breadcrumb-current">` with no `aria-current="page"`. `<nav aria-label="Breadcrumb">` and keyboard-reachable `<button>` crumbs are all preserved, and the missing attribute is pre-existing rather than introduced here — but #1318 tracks WCAG residuals and this one is a one-liner.

Also worth a sentence in the body: `/app` and `/app/explore` now render no breadcrumb `<nav>` at all, since `CRUMB_MAP['explore']` is gone and `if (!info) return null`. The sidebar `<nav>` landmark remains so it isn't a landmark-less page, but it does drop one from the app's most-visited anonymous route.

dbrowneup:
@onder-akkaya — fixed silently earlier, ack owed: the body reads 'Addresses #1370' now, not 'Closes' (your point about item 1's own text naming the three remaining sections was right; #1405 tracks them), and the exclusion-guard docstring now says which claim it makes (silent-omission guard, not missing-crumb guard). Nothing further requested from you.

PR #1401, opened 2026-08-20
[marketplace][infra] PAYMENTS_HALT kill switch + mainnet-gate startup assertions

### LINKED ISSUE #1240: [infra] Arc mainnet gate — Sept 16, payment rail only
Arc goes mainnet on 2026-09-16. This issue is the checklist that has to be green before anything of ours touches it, plus the list of things we are deliberately not shipping to mainnet in September.

## Scope

Mainnet is the **payment rail only**. The USDC path (x402 / Gateway / PaymentSplitter) goes to Arc mainnet. Vault, synth, and AMM execution stays on testnet behind a WIP banner.

This split is the whole point of the checklist. If it drifts back to "ship everything," most of the deferred list below becomes blocking again.

## Blocking item, already checked

- [x] **Does the pinned circlekit know an Arc mainnet chain?** Checked 2026-08-16. **No.**

  `circlekit/constants.py` at pinned SHA `09828f3999` has 23 entries in `CHAIN_CONFIGS`. Arc appears once, as `arcTestnet` (chain 5042002), in the testnet block. The 11 mainnet entries are ethereum, base, arbitrum, polygon, optimism, avalanche, sonic, unichain, worldChain, hyperEvm, sei. Upstream HEAD is the same commit (2026-03-24), so bumping the pin does not help.

  Related footgun in the same file: `CHAIN_ALIASES` maps `"mainnet"` to `"ethereum"`. Setting `GATEWAY_CHAIN=mainnet` resolves to a valid config and would settle real USDC on Ethereum rather than failing.

  Mitigation exists. `CHAIN_CONFIGS` is a plain module-level dict and `get_chain_config` reads it at call time, so an Arc mainnet `ChainConfig` can be registered at runtime with no fork. Eight of nine fields are available or derivable, and `MAINNET_GATEWAY_WALLET` / `MAINNET_GATEWAY_MINTER` are already constants in the SDK. The blocker is `gateway_domain`, Circle's internal domain ID (arcTestnet is 26), which we cannot invent.

- [ ] **Get Arc mainnet's Circle Gateway domain ID, chain ID, RPC URL, and confirmation the USDC sentinel address is unchanged.** Dan to Circle. Nothing else on this list matters if this doesn't land.

## Required before mainnet

- [ ] Arc mainnet chain ID and RPC published; `contracts/foundry.toml` gains a mainnet endpoint (today it has only `arc-testnet`). Dan
- [ ] `PAYMENTS_DRY_RUN` set explicitly in `infra/ecs.tf` — #1239
- [ ] `CIRCLE_API_KEY` and `CIRCLE_ENTITY_SECRET` in the ECS `secrets` block, not just the best-effort SSM boot fetch. A fetch failure otherwise means charge-state ambiguity. Dan
- [ ] `GATEWAY_CHAIN` set explicitly, never `marketplace/config.py`'s `arcTestnet` default. Dan
- [ ] `PaymentSplitter` deployed to mainnet and owned by a multisig (#1186 scoped to this one contract). Dan + Bogdan
- [ ] **Two-chain config.** Split `ARC_PAYMENTS_CHAIN_ID` / `_RPC_URL` from `ARC_EXECUTION_CHAIN_ID` / `_RPC_URL`. `GET /api/config/contracts` returns two blocks each with its own chain id. The three manifests and `ui/src/siwe.js:15` stop hardcoding `5042002`. This is the most under-estimated item here — the codebase assumes one chain everywhere
- [ ] **A written scope decision from Dan, in an ADR, not a Discord message.** It states: mainnet is payment-rail only; vault/synth/AMM stays testnet; `PAYMENTS_DRY_RUN=false` applies only to the caller-signed metered-API path, with the DCW browser subscribe flow staying in dry-run and staying blocked on #975. The justification is that a caller-signed payment means no platform wallet holds caller funds on the live surface, which is #975's actual acceptance criterion. Dan signs, Bogdan acknowledges the custody side
- [ ] **No dual-chain wallet UX at launch.** `ui/src/siwe.js:15` hardcodes chain 5042002, so a browser flow paying on mainnet while vaults sit on testnet needs chain switching. Programmatic callers don't have that problem, which is the argument for launching API/CLI-first and keeping the browser on testnet. Keep mainnet payments Gateway-mediated so no user-facing switch is needed
- [ ] **Mirror the circlekit commit somewhere we control**, or vendor it, plus a written risk acceptance. It is a git-pinned fork commit, not on PyPI, and it is the only path real money moves through. Pair with the fail-loud startup assertion below. Dan + Önder
- [ ] **Write the rollback procedure and rehearse it on testnet.** Flip `PAYMENTS_DRY_RUN=true` in Terraform, merge, deploy, plus an immediate manual lever (revert to a prior task definition via AWS CLI) for when a 10-minute deploy is too slow. Rehearse before it's needed
- [ ] **A documented yfinance licensing decision** (#1218). The obligation changes the moment we charge for a verdict derived from Yahoo data. Either accept and monitor with a named replacement candidate and its price, or swap the source. The swap itself isn't required before launch; the written decision is. Dan signs the cost commitment
- [ ] **A kill switch that isn't a redeploy.** `PAYMENTS_DRY_RUN` is read once into `MarketService`'s constructor. Add a per-charge `PAYMENTS_HALT` check so stopping real charges is an SSM change
- [ ] **Spend-cap / meter UI on `/subscriptions`.** There is none today. Charging real money with no visible cap is the fastest way to a furious user
- [ ] Fargate desired count pinned to 1 for the marketplace path (`market.publishers` is an in-process dict with a single-instance assumption)
- [ ] **Two startup assertions**, both turning silent degradation into a refusal to boot: `GATEWAY_CHAIN` must match what the RPC reports, and the process must not serve traffic when circlekit failed to import while `PAYMENTS_DRY_RUN=false`. `main.py` wraps that import in try/except today, so a broken install currently degrades to "marketplace absent"
- [ ] **One real cent proven.** See below

## Proof of life: one real cent, three pieces of evidence

A log line saying "charged" is not proof. `_charge_one` returns `(True, None)` in dry-run and logs the same shape. All three of these, recorded in `docs/runbooks/`:

1. A Circle facilitator settlement id from `middleware.settle`
2. An on-chain `PaymentSplitter` transfer tx hash on the mainnet explorer, non-zero USDC value
3. Publisher balance delta equal to subscriber balance delta, reconciled to the raw unit

Only then does the manifest's `paid` group flip to `status: "live"`.

## Deferred, not dropped

These are required only if vaults ever go to mainnet. Tagging them here so nobody has to guess whether they were forgotten.

- #1136 / #1129 fee-bps cap. Real money risk, hard-blocks mainnet vaults. Dan + Bogdan
- #1200 Vault NAV decimals for 6-dec synths, plus performance-fee share-mint dilution. Same reviewers
- `Ownable2Step` migration across all seven owned contracts, plus the ownership-transfer sweep. Not needed for PaymentSplitter alone, needed before any vault goes live
- #975 subscription fee custody. Stays deferred precisely because the browser subscribe flow stays in dry-run
- #1103 oracle price-snapshot writer and #1102 vault-address SSOT. Non-negotiable before mainnet vaults; a mainnet vault showing fabricated returns is indefensible
- #714 commit-reveal wiring. Required before "anchored on-chain" can be reclaimed in full
- #1065 / #1043 runner relocation and exactly-once lease lock
- `Deploy.s.sol` mainnet readiness: parameterised USDC address, no hardcoded testnet constants, dead Phase-4 branch removed
- The 281-synth and AMM mainnet redeploy

## Not gating on

The re-run's pass rate staying honest at 3/34 rather than being engineered upward. The corpus knowledge-graph build. Zero vaults with a transparent WIP banner, which beats one vault with fake returns.

dbrowneup:
Refs #1240

> **Merge ordering — this PR must land AFTER #1469 (the calm-precision
> rebrand).** Bringing the branch current required merging
> `origin/feat/calm-precision-product-rebrand` into it, so #1469's UI work is
> folded into this branch's history and shows up in this diff. That is
> deliberate and must not be stripped: rewriting it out would rewrite
> published history. Merge #1469 first and this diff collapses back to the
> payments/infra change it is. Reviewers: read the backend diff
> (`backend/`, `.env.example`, `infra/`) as this PR's own work; treat `ui/`
> as #1469's, already reviewed there.

## What this ships

#1240 is the Arc-mainnet launch-gate checklist. Most of it is genuinely
decision- or provisioning-gated on Dan (Circle's Gateway domain ID for Arc
mainnet, the written ADR scope decision, `CIRCLE_API_KEY`/`CIRCLE_ENTITY_SECRET`
in the ECS secrets block, the mainnet chain ID/RPC itself, multisig
`PaymentSplitter` deployment, the two-chain config split, yfinance licensing
(#1218), and the "one real cent" proof). This PR ships the decidable subset —
two checklist items that are pure engineering, independent of any pending
business decision, and explicitly spelled out in the issue text:

### 1. `PAYMENTS_HALT` — a kill switch that isn't a redeploy

> "PAYMENTS_DRY_RUN is read once into MarketService's constructor. Add a
> per-charge PAYMENTS_HALT check so stopping real charges is an SSM change."

`PAYMENTS_DRY_RUN` is the primary safety switch, but `MarketService` reads it
once at construction (`main.py` boot) and caches it — flipping it off
requires the full build → ECR → roll Fargate deploy pipeline. `PAYMENTS_HALT`
(`marketplace/config.py:payments_halted()`) is read fresh from `os.environ` on
**every** call instead. An operator sets it via the same `/archimedes/prod/*`
SSM prefix `secrets_service.load_ssm_secrets()` already sweeps into env at
boot, then cycles the ECS task (`aws ecs update-service
--force-new-deployment` — a same-image restart, not the deploy pipeline).

Wired into **six** fund-moving call sites — the ones the issue names, plus
three found across the adversarial review rounds (see "Post-review fixes"
below). Every gate lives in the callee, so a caller cannot forget it. The
authoritative inventory, including what is deliberately NOT gated and why,
is `marketplace/config.py:payments_halted()`'s docstring:

- `MarketService._charge_one` (the tick-charging rail, `payments_dry_run=False`
  path) — halted charges return `(True, None, True)`: `paid=True` so the
  subscriber isn't deferred (flipping the switch can't itself cascade into
  subscriber defer/halt logic; it only stops money moving), but the third
  element, `charge_suppressed`, tells every caller to persist `charged=False`
  — this is NOT the same shape as a real or dry-run settle, on purpose.
- `generation_payment.enforce_generation_payment` — the metered paywall.
  REFUSES with 503 rather than comping the paid product. `quote()` now reports
  `halted` alongside `dry_run` so callers see the true state. **This is the
  rail that most needs the switch, and the reason is easy to state wrong.**
  An earlier version of this section called generation "the one surface the
  scope decision allows `PAYMENTS_DRY_RUN=false` on at mainnet." That was true
  when this PR was opened and stopped being true with #1428 (2026-08-20),
  which split the switch. Prod today (`infra/ecs.tf` money-switch block) is
  `PAYMENTS_DRY_RUN="true"` (global stays dry), `GENERATION_PAYMENTS_DRY_RUN=
  "false"`, `GENERATION_PAYMENT_REQUIRED="true"`. So the conclusion survives
  but the named mechanism was wrong: generation is the only rail moving real
  money, and it gets there through the *generation-scoped* override, not the
  global. The consequence for the kill switch is the real point —
  `_payments_dry_run()`, which stops value moving everywhere else, is
  deliberately `False` here, so `PAYMENTS_HALT` is not a redundant second belt
  on this rail; it is the only no-redeploy way to stop a live charge.
- `SettlementSweeper.sweep_publisher` — the automatic per-tick sweep
  (Gateway withdraw → on-chain `gatewayMint` → `depositToPool`), added in
  review round 2.
- `SettlementSweeper.withdraw_publisher` — the Withdraw button's
  `PaymentSplitter.withdraw`, added in round 3.
- `SettlementSweeper.withdraw_subscriber` — the unsubscribe refund's DCW USDC
  transfer, added in round 3. Both withdraws are reachable directly from
  authenticated endpoints without passing through `sweep_publisher`, so they
  carry their own guard rather than inheriting one.
- `services.revenue_sweep.sweep_revenue` — the platform revenue sweep
  (Gateway balance → DCW tokens), added in round 4. It landed on `main` while
  this PR was open and is deliberately independent of `PAYMENTS_DRY_RUN`
  ("may we charge a user?" — not its question); it is **not** independent of
  `PAYMENTS_HALT` ("may a Circle call move USDC?" — `GatewayClient.withdraw`
  plainly does). Both entry points are covered, the hourly loop and the CLI.

### 2. Two startup assertions

> "Two startup assertions, both turning silent degradation into a refusal to
> boot: GATEWAY_CHAIN must match what the RPC reports, and the process must
> not serve traffic when circlekit failed to import while
> PAYMENTS_DRY_RUN=false."

Both fatal only under `PAYMENTS_DRY_RUN=false` (a dry-run/testnet boot logs a
warning and continues — this must not crash-loop testnet or CI):

- **`_assert_marketplace_live_or_dry`** — `main.py`'s `marketplace_routes`
  import is deliberately fail-soft (PR #958 prod incident: a broken circlekit
  install must never crash-loop the whole backend). Correct under
  `PAYMENTS_DRY_RUN=true`; stops being correct the moment an operator sets
  `PAYMENTS_DRY_RUN=false` — "the process boots fine, most routes 200,
  marketplace quietly 404s" is a silent trap for exactly that operator.
- **`_assert_gateway_chain_matches_rpc`** — circlekit's `CHAIN_ALIASES` maps
  `"mainnet"` → `"ethereum"` (the exact footgun #1240 documents), so a
  fat-fingered `GATEWAY_CHAIN` can silently resolve to a real, different chain
  than the one Arc RPC actually targets. Compares
  `get_chain_config(GATEWAY_CHAIN).chain_id` against a live
  `chain_client.get_chain_id()` RPC call (bounded by `AsyncHTTPProvider`'s
  existing request timeout, so it can't hang startup). An RPC connectivity
  failure is logged and swallowed — a different failure mode than "pointed at
  the wrong chain", which is the only thing this treats as fatal.

Both are pulled into standalone, importable functions specifically so they're
unit-testable without exercising the whole app/lifespan.

### Already satisfied, no code needed

- **`PAYMENTS_DRY_RUN`/`PAPER_TRADING` pinned explicitly in `infra/ecs.tf`** —
  #1239, merged.

### Correction (post-review): "Fargate desired count pinned to 1" is NOT satisfied

An earlier version of this section (and the matching #1240 comment) claimed
the ECS service's `lifecycle.ignore_changes = [task_definition, desired_count]`
(`infra/ecs.tf`) "pins" `desired_count` to 1 so CI/autoscaling "can't drift it
back up." That has the Terraform backwards: `ignore_changes` means a later
`terraform apply` will never *revert* an out-of-band change to that field —
it does not prevent one, and it cannot un-drift a scale-up someone (or
something) already made. `ecs_service_desired_count` (`infra/variables.tf`)
only sets the value at first apply/bootstrap; a bare `aws ecs update-service
--desired-count 2`, or the CPU-target autoscaling policy already wired in
this same file (`aws_appautoscaling_target.backend`, which can legitimately
drive `desired_count` up to `ecs_service_max_count = 4`), sticks permanently
and nothing in Terraform corrects it back down. Nothing today pins Fargate
desired count to 1.

What actually bounds the risk this checklist item is worried about — the
in-process `market.publishers` dict's single-instance assumption causing
double-ticking/double-charging across replicas — lives at the application
layer, not Terraform, and is already merged:
`MarketService._run_loop` takes a per-strategy Redis leader lock
(`state.try_acquire_leader` / `renew_leader` / `release_leader`, a
fencing-token compare-and-swap) before any replica is allowed to `tick()` a
given strategy, and `_charge_one`'s `SettlementIntent` unique index on
`(strategy_id, tick_id, sub_id, step)` is a second, independent guard against
a double-settle even if a lock race ever slipped through. Those two
mechanisms are real; "desired_count pinned to 1" is not. This PR does not
close that checklist item — leaving it open on #1240 rather than marking it
satisfied. A real fix (dropping `desired_count` from `ignore_changes`, or a
deploy-time assertion on the live service's `desiredCount`) is infra/CI
wiring and needs Dan's alignment before landing, per this repo's "ask before
touching deployment configs" rule.

## Guard demonstrations (revert → run → confirm failure → restore)

**Kill switch** (commit `751807f`): reverted `config.py`/`service.py`/
`generation_payment.py` to pre-fix, ran the new tests:
- `test_config.py` — `ImportError` at collection (`payments_halted` didn't exist)
- `test_payments_halt.py` — 2 failed (halt didn't short-circuit the charge)
- `test_generation_payment.py::TestPaymentsHalt` — 2 failed (halt/quote key
  absent); the adversarial companion (`test_halt_false_reaches_the_facilitator`)
  passed both before and after, proving the halted test isn't a tautology

All 21 pass after restoring the fix.

**Startup assertions** (commit `363f37e`): temporarily removed both new
blocks from `main.py` via direct edit (not `git stash` — see note below), ran
the new tests:
- `test_startup_assertions.py` — 10 failed, `AttributeError` (neither function existed)
- `test_marketplace_kill_switch_boot_gate.py` — the subprocess harness that
  forces `circlekit` unimportable via `sys.modules["circlekit"] = None`
  (reproducing the PR #958 failure mode without uninstalling the package):
  the `PAYMENTS_DRY_RUN=false` case printed `BOOTED` instead of refusing
- `test_lifespan_no_rigor_backfill.py` / `test_main_ssm_prod_gate.py` —
  unaffected throughout (2+2 passed both before and after), confirming the
  new assertions don't touch unrelated startup behavior

All 12 pass after restoring (byte-identical diff against a pre-edit backup).

## Test tallies

- New/touched test files: **37 passed**, 0 failed (`test_config.py`,
  `test_payments_halt.py`, `test_generation_payment.py`,
  `test_startup_assertions.py`, `test_marketplace_kill_switch_boot_gate.py`,
  plus the two pre-existing `main.py`-import tests re-verified unaffected)
- Full unit suite, `pytest -m "not integration"` from repo root: **3573
  passed, 2 skipped (pre-existing, unrelated), 2 deselected, 0 failed**
- `ruff format --check` / `ruff check --select E9,F63,F7,F40`: clean on all
  touched files
- Post-review-round + rebased onto current `main` (was 34 commits behind):
  full unit suite from repo root, **3588 passed, 2 skipped (pre-existing,
  unrelated), 2 deselected, 0 failed** — the exact `pytest -m "not
  integration"` command CI runs
- Post-review-round-2 (4 new regression tests for the round-2 findings
  below): full unit suite from repo root, **3592 passed, 2 skipped
  (pre-existing, unrelated), 2 deselected, 0 failed**, 332s — the exact
  `pytest -m "not integration"` command CI runs. `ruff format --check` /
  `ruff check --select E9,F63,F7,F40`: clean on all touched files.

## Post-review fixes (adversarial review round)

An adversarial review of this PR surfaced 2 blockers, 3 majors, and 3 minors.
All are fixed on this branch. One (the desired-count item below) is a
retraction of a wrong claim in this body rather than a code change — see the
correction inline above; not repeated here:

- **[blocker] The tick ledger could claim `charged=True` for a halted
  charge.** `_charge_one` now returns a third element, `charge_suppressed`,
  distinguishing "PAYMENTS_HALT suppressed this charge" from a real (or
  dry-run) settle. Both charging call sites (`_charge_step`,
  `_rebalance_phase`) now persist `charged=False` /
  `halt_source=HaltSource.PAYMENTS_HALT` for a suppressed charge — the
  subscriber is still NOT deferred (unchanged design intent: the switch
  stops money moving, not service) but the ledger no longer lies about it.
  New end-to-end regression test drives the real `tick()` pipeline (not a
  mocked `_charge_one`) and asserts no persisted record ever claims
  `charged=True` while halted (`test_per_step_charging.py::
  test_payments_halt_never_persists_charged_true`).
- **[blocker] "Fargate desired count pinned to 1" was not true.** Corrected
  above — `lifecycle.ignore_changes` prevents Terraform from *reverting*
  drift, it does not prevent it. Retracted, not code-fixed (infra changes
  need Dan's alignment per this repo's rules); the real mitigation
  (per-strategy Redis leader lock + `SettlementIntent` idempotency) is
  documented in its place.
- **[major] The lifespan startup wrapper's `except RuntimeError: raise`
  could abort boot on a plain connectivity error, even under
  PAYMENTS_DRY_RUN=true.** `_assert_gateway_chain_matches_rpc` now raises a
  dedicated `GatewayChainMismatch(RuntimeError)` on a confirmed mismatch
  only; the call site re-raises that type specifically, letting any other
  RuntimeError (e.g. `get_chain_id()` failing on a closed session) fall
  through to the non-fatal warning branch.
- **[major] `payments_dry_run` could go unbound before assertion 2 ran,
  silently disabling it.** It's now read once before step 3's try block
  instead of inside it, so a NameError there is no longer possible. New file
  `test_gateway_chain_lifespan_wiring.py` drives `main.lifespan` itself
  (not just the isolated function) end to end: aborts on a real mismatch,
  survives a connectivity RuntimeError under both dry-run and live, and
  proves the guard still fires even when step 3 is forced to raise early
  (the exact regression scenario). All 4 confirmed to fail against the
  pre-fix code before being confirmed to pass against the fix.
- **[major] PAYMENTS_HALT on the metered generation API served the paid
  product for free, unverified.** `enforce_generation_payment` now returns
  503 (`reason: "payments_halted"`) while halted instead of accepting any
  Payment-Signature value without decode/verify/payer-binding/settle —
  refusing service is the honest reading of a kill switch on a pay-per-call
  surface with no pre-existing subscriber relationship (unlike the
  marketplace tick rail, where "stays paid" is the correct call). Tests
  updated to assert the 503, plus a new adversarial case confirming even a
  malformed header is rejected (not silently accepted) while halted.
- **[minor] PAYMENTS_HALT was undocumented in `.env.example`.** Added,
  mirroring the PAYMENTS_DRY_RUN block, including the operator procedure
  (SSM param + `aws ecs update-service --force-new-deployment`).
- **[minor] A test docstring overstated per-call freshness as the
  operational win.** Corrected: nothing mutates `os.environ` mid-process,
  so per-call freshness and a cached read are equally stale within one
  running task. The real property is that PAYMENTS_HALT lives in SSM/env
  rather than the task definition, so flipping it needs a same-image
  restart, not a new task-definition revision.
- **[minor] Branch was 34 commits behind `origin/main`.** Rebased; full
  unit suite re-run on the rebased result (see Test tallies).

## Post-review fixes, round 2 (second adversarial review)

A second adversarial review of the round-1 fixes above surfaced 1 more
blocker, 2 more majors, and 2 more minors. All are fixed on this branch
(code) or corrected (prose):

- **[blocker] `_halt_publisher` still hardcoded `charged=True` on every
  publisher-halting tick.** Round 1's `charge_suppressed` fix only reached
  the two charging call sites (`_charge_step`, `_rebalance_phase`); it never
  reached `_halt_publisher`, which every step's exception path and every
  `result.halted` path (v_check fail, no-drift, a step raising — the most
  common tick outcomes) calls with `active` — the very survivors
  `_charge_step` just returned, some of whom may have had *this step's*
  charge suppressed by PAYMENTS_HALT. `_charge_step` now returns
  `(survivors, charge_suppressed_ids)`; `tick()` threads
  `charge_suppressed_ids` into `_halt_publisher`, which persists
  `charged=False` for any subscriber in that set instead of unconditional
  `charged=True`. New regression test
  (`test_per_step_charging.py::test_payments_halt_publisher_halt_never_persists_charged_true`)
  drives a real NO_DRIFT_DEDUP halt (not the always-succeeds path the round-1
  test drives) while PAYMENTS_HALT=true and asserts no PUBLISHER-halt record
  claims `charged=True`. Confirmed it fails against the pre-fix code (revert
  → run → confirm failure → restore) before confirming it passes.
- **[major] `_rebalance_phase` let PAYMENTS_HALT swallow a real execution
  failure.** The `if charge_suppressed: ... elif not mirrored: ...` order
  meant a genuine vault-mirror failure while halted was recorded with
  `halt_source=PAYMENTS_HALT` and a fabricated-sounding "rebalance still
  applied" reason — the real `RuntimeError` from `_apply_to_subscriber` was
  dropped from the ledger entirely, even though the subscriber is still
  deferred and a liability is still recorded. The two conditions are now
  composed instead of shadowed: `not mirrored` always takes
  `halt_source=EXECUTION` with the real error text (appending a PAYMENTS_HALT
  note when both are true); `charged=not charge_suppressed` is unchanged.
  New test `test_mirror_failure_while_payments_halted` asserts the real
  error text survives on the ledger. Confirmed to fail pre-fix, pass
  post-fix.
- **[major] The automatic per-tick settlement sweep ignored PAYMENTS_HALT
  entirely.** `SettlementSweeper.sweep_publisher` — Gateway withdraw →
  on-chain `gatewayMint` → `depositToPool`, running every tick with no human
  approval step — was gated only on the cached `payments_dry_run`, so real
  USDC kept moving with the kill switch on. `.env.example`'s claim that
  PAYMENTS_HALT "stops real money moving" was true for the charge/paywall
  rails but not this one. Fixed by adding the same `payments_halted()` check
  `SettlementSweeper` already uses for `payments_dry_run`, inside
  `sweep_publisher` itself (not only at the `tick()` call site) — same
  defense-in-depth reasoning the class docstring already documents for the
  dry-run guard, after the M1' manual-withdraw-endpoint bypass. `.env.example`
  updated from "both real-money call sites" to the three that now exist.
  New tests `test_payments_halt_sweep_publisher_is_noop` +
  adversarial companion `test_payments_halt_false_sweep_publisher_still_runs`
  (PAYMENTS_HALT=false must NOT block the sweep). Confirmed the noop test
  fails pre-fix, passes post-fix.
- **[minor] The "What this ships" section's `_charge_one` bullet still
  described the pre-round-1-fix return shape.** It said halted charges
  return `(True, None)`, "the same shape as dry-run" — the opposite of round
  1's fix, whose whole point is the 3-tuple with `charge_suppressed=True` so
  callers persist `charged=False`. Corrected in place above.
- **[minor] The #1240 issue comment crediting this PR with the leader-lock
  and `SettlementIntent`-uniqueness double-charge mitigations was corrected**
  — both predate this PR on `main` and are untouched by this diff; the
  comment said "already merged in #1401," which is not accurate. Edited on
  the issue.

## A worktree-hygiene note for whoever reviews this

Mid-session, a `git stash pop` in this worktree picked up an unrelated
in-flight change from a **different** worktree/session (branch
`dbrowneup/1370-...`, a `StrategyPassport.jsx`/`routes.js` fix) — `git
stash`'s ref is shared across all worktrees of one repo, not scoped to the
worktree it was pushed from, so two sessions stashing concurrently can cross-
contaminate. That foreign diff was identified and discarded from this branch
before committing (never staged, never pushed); at the time this note was
written, this branch contained only the two commits described above (it has
since gained the post-review-round commits listed in "Post-review fixes").
Recovering afterward was done by re-applying
the known-good edits directly rather than touching `git stash` again. Flagging
in case the `dbrowneup/1321-backlog` session (chain-status-pill work, still
sitting in this repo's shared stash as of this PR) needs it recovered — that
stash entry was left untouched, not mine to manage.

## Remaining decision-gated scope (not this PR)

Everything else on #1240's checklist stays open, blocked on Dan (or Dan +
Bogdan / Önder): the Circle Gateway domain ID for Arc mainnet, the mainnet
chain ID/RPC itself, `CIRCLE_API_KEY`/`CIRCLE_ENTITY_SECRET` in the ECS
secrets block, `GATEWAY_CHAIN` set explicitly at deploy time, mainnet
`PaymentSplitter` + multisig, the written ADR scope decision, the two-chain
config split (`ARC_PAYMENTS_CHAIN_ID`/`_RPC_URL` vs
`ARC_EXECUTION_CHAIN_ID`/`_RPC_URL`, `GET /api/config/contracts` two blocks,
`ui/src/siwe.js`'s hardcoded `5042002`), mirroring/vendoring the circlekit
fork commit, the rollback-procedure rehearsal, the yfinance licensing
decision (#1218), the spend-cap/meter UI on `/subscriptions`, and the "one
real cent" proof. This PR does not close #1240.

## Post-review fixes, round 3 (third adversarial review)

A third adversarial review of the round-2 fixes above surfaced 2 majors and 1
minor. All are fixed (code) or corrected (public issue comments) on this
branch:

- **[major] Round 2's `PAYMENTS_HALT` guard on `sweep_publisher` was only
  half the fix — `withdraw_publisher` and `withdraw_subscriber` (both
  `SettlementSweeper` methods, both reachable directly from authenticated
  endpoints — the Withdraw button and unsubscribe, respectively — with no
  dependency on `sweep_publisher`) still moved real on-chain value
  (`PaymentSplitter.withdraw` / an ERC-20 `transfer` from the subscriber's
  DCW) with only `payments_dry_run` gating them.** `PAYMENTS_HALT=false` was
  never the failure — the gap was PAYMENTS_HALT=true, `PAYMENTS_DRY_RUN=false`
  (live) hitting either withdraw path directly. Both methods now carry the
  same `payments_halted()` check `sweep_publisher` already had, placed
  immediately after the `payments_dry_run` check and before any chain touch —
  same "gate in the callee, not just one caller" reasoning the class
  docstring already states. `.env.example` updated from "three real-money
  call sites" to the five that now exist, naming both new ones and why they
  needed their own guard (not inherited from the sweep). Four new tests: a
  halt-is-noop case + an adversarial "PAYMENTS_HALT=false must still run"
  companion for each method. The `withdraw_subscriber` halt test initially
  asserted only on `_get_executor` (mirroring the `withdraw_publisher`
  pattern) and passed even against the unfixed code, because
  `withdraw_subscriber`'s try/except already swallows the mocked
  `_usdc_balance_of` failure before `_get_executor` is ever reached — a
  same-shape trap to the one CLAUDE.md's testing-conventions section warns
  about. Rewritten to assert `_usdc_balance_of.assert_not_called()` (the
  first mock the live path touches), which does distinguish the two. See the
  guard demonstration below for the full revert → run → confirm-failure →
  restore transcript, including the corrected assertion's before/after.

- **[major] `PR #1401`'s "Edited on the issue" correction to #1240
  (round 2's last bullet) replaced the target comment's entire body with the
  literal, never-expanded shell arg `@/tmp/issue1240_comment_new.md` —
  publishing a local temp path on the public repo with no readable
  correction — and the original comment carrying the retracted claim
  ("Fargate desired count pinned to 1 ... `lifecycle.ignore_changes` keeps
  CI/autoscaling from drifting it") was never edited at all, left standing
  on a checklist someone reads before a mainnet cutover.** Fixed directly on
  the issue (not a code change): `gh api -X PATCH
  repos/a-apin/archimedes/issues/comments/5358929218` now carries the actual
  correction text (verified: `gh api .../comments/5358929218 --jq .body`
  returns the real 3-paragraph correction, not the literal `@`-prefixed
  path). `gh api -X PATCH .../comments/5357916114` strikes the wrong
  desired-count line (`~~...~~`) with an inline retraction pointing at the
  correction comment, and adds a small edit-note that the "wired into both
  real-money call sites" line in that same comment is now stale (five sites,
  per this round) rather than silently leaving it inaccurate. Both edits
  verified live on the issue after the PATCH — see the transcript below.

- **[minor] `_halt_publisher`'s `charge_suppressed_ids` parameter defaulted
  to `None`, so a future caller that forgot to pass it would silently get
  `suppressed = charge_suppressed_ids or set()` → empty set → every survivor
  recorded `charged=True` — the exact tick-ledger lie round 2 fixed, made
  available again by default. Unlike `settlement.py`, where the same round
  argued the guard must live in the callee specifically so a caller can't
  forget it, this one relied on convention alone.** Dropped the `= None`
  default (`charge_suppressed_ids: set[str]`, required); both real call
  sites (`tick()`'s exception path and its `result.halted` path) already
  pass it positionally, so this is a no-op for them. A new direct unit test,
  `test_halt_publisher_requires_charge_suppressed_ids`, calls
  `_halt_publisher` without the argument and asserts `TypeError` — confirmed
  to fail (`DID NOT RAISE TypeError`) against the pre-fix default-`None`
  signature before confirming it passes post-fix.

### Guard demonstration, round 3 (revert → run → confirm failure → restore)

**`SettlementSweeper` PAYMENTS_HALT guard** (`withdraw_publisher` +
`withdraw_subscriber`): `git diff backend/archimedes/marketplace/settlement.py
> /tmp/finding1_fix.patch`, then `git checkout --
backend/archimedes/marketplace/settlement.py` to revert to pre-fix (no
`git stash` — plain diff/checkout/apply per this repo's rule), ran the 4 new
tests against the reverted file:

```
$ pytest backend/tests/marketplace/test_settlement.py -q
...
FAILED test_payments_halt_withdraw_publisher_is_noop
  - AssertionError: Expected 'mock' to not have been called. Called 1 times.
FAILED test_payments_halt_withdraw_subscriber_is_noop
  - AssertionError: Expected 'mock' to not have been called. Called 1 times.
2 failed, 24 passed, 1 warning in 4.18s
```

The 2 failures are exactly the halt-detection tests; the 2 adversarial
companions (`..._false_..._still_runs`) are in the 24 that still pass —
proving they're not tautologically skipping regardless of the flag. First
draft of `test_payments_halt_withdraw_subscriber_is_noop` asserted only
`_get_executor.assert_not_called()` and passed 26/26 even against this same
reverted file (a false green — the mocked `_usdc_balance_of` failure was
already being swallowed by `withdraw_subscriber`'s own try/except before
`_get_executor` was ever reached, so the assertion held regardless of the
guard). Rewritten to assert `_usdc_balance_of.assert_not_called()` instead —
re-ran against the reverted file and got the correct failure shown above.
Restored the fix (`git apply /tmp/finding1_fix.patch`); full file:
`26 passed, 1 warning in 3.71s`.

**`_halt_publisher` required parameter**: `git diff
backend/archimedes/marketplace/service.py > /tmp/finding3_fix.patch`, then
`git apply -R /tmp/finding3_fix.patch` to revert just that hunk (signature
back to `charge_suppressed_ids: set[str] | None = None`):

```
$ pytest backend/tests/marketplace/test_per_step_charging.py -q -k requires_charge_suppressed_ids
FAILED test_halt_publisher_requires_charge_suppressed_ids
  - Failed: DID NOT RAISE <class 'TypeError'>
1 failed, 10 deselected, 1 warning in 3.32s
```

Restored (`git apply /tmp/finding3_fix.patch`); full marketplace suite:
`173 passed, 1 warning in 24.72s`.

**Issue comment corrections**: verified live via
`gh api repos/a-apin/archimedes/issues/comments/5358929218 --jq .body`
(now the real correction text, `updated_at` bumped) and
`gh api repos/a-apin/archimedes/issues/comments/5357916114 --jq .body`
(desired-count line struck, retraction note present, `updated_at` bumped).
No code path to mutation-test here — this is a public-comment content fix,
verified by re-fetching the comment body post-PATCH rather than a revert
cycle.

### Test tallies, round 3

- `backend/tests/marketplace/test_settlement.py`: 26 passed (22 pre-existing
  + 4 new: `test_payments_halt_withdraw_publisher_is_noop`,
  `test_payments_halt_false_withdraw_publisher_still_runs`,
  `test_payments_halt_withdraw_subscriber_is_noop`,
  `test_payments_halt_false_withdraw_subscriber_still_runs`)
- `backend/tests/marketplace/test_per_step_charging.py` +
  `backend/tests/marketplace/`: 173 passed (172 pre-existing + 1 new:
  `test_halt_publisher_requires_charge_suppressed_ids`)
- `ruff format --check` / `ruff check --select E9,F63,F7,F40` on all touched
  Python files: clean

## Post-review fixes, round 4 (dossier review + merge to current `main`)

- **Merged current `origin/main`** (was 27 commits behind; base is now
  `3dd63e5e`). One conflict, `docs/README.md` — an additive doc-index collision
  where this branch's three #1469 plan rows and main's new
  `2026-08-30-intraday-paper-trading.md` row landed on the same line. Resolved
  by keeping both sides in date order; no row dropped. `alembic heads` prints
  exactly one head (`5cb798feef58`), so the merge introduced no migration fork.

- **[major] Extended `PAYMENTS_HALT` to `services/revenue_sweep.py`.** It
  landed on `main` while this PR sat open, calling `GatewayClient.withdraw`
  from an unattended hourly loop and from a CLI — the same transfer class
  `sweep_publisher` already refuses, and the one fund-moving path the switch
  did not cover. Gated inside `sweep_revenue` before `_client()`, so both
  entry points are covered by construction. `check_revenue` is untouched:
  reading a balance moves nothing and an operator mid-incident needs the
  number. Adversarial demonstration (gate deleted, tests re-run): 3 failed,
  2 passed — the 2 survivors are the un-halted companion and `check_revenue`,
  neither of which depends on the gate.

- **[major] A structural guard so a *seventh* call site cannot escape.** The
  gap above already happened once: per-site tests cannot catch a site that
  does not exist yet. `tests/marketplace/test_payments_halt.py` now asserts
  that every module under `archimedes/` importing `circlekit` — the only way
  USDC leaves this backend — also references `payments_halted`, or sits in
  `_READ_ONLY_CIRCLEKIT` with a written reason (one entry today: `main.py`,
  which imports the pure `get_chain_config` name→chain_id lookup). Modelled on
  main's `TestNoFifthCallSite`. Ships with a guard-on-the-guard (the scan must
  actually see `settlement.py`, `payments.py`, `revenue_sweep.py`) and a check
  that the allowlist cannot rot into a dumping ground. Dropping the
  `payments_halted` import from `revenue_sweep.py` makes it name exactly that
  module. A backstop inside `payments.charge` logs at ERROR and answers
  `False` if it is ever reached while halted — reaching it means a caller went
  around `_charge_one`'s gate, i.e. a bug.

- **[major] The credit ledger boundary is now proved in tests, not asserted in
  a comment.** #1441/#1498 and this PR were written in parallel and meet for
  the first time in the merged tree. Should the switch stop
  claim/settle/consume/void/restore too? **No** — those are DB rows recording
  an entitlement a Circle settle *already* created, and halting them would be
  actively harmful (`consume`/`restore` are what return an unspent credit to a
  payer whose generation died). The load-bearing new test: a halted paid start
  503s **and leaves no PENDING row**, because a pending claim reads as
  `in_flight` forever and would lock the caller's Idempotency-Key out
  permanently — `_paywall_with_credit`'s `except BaseException: void()` catches
  the 503 the same way it catches a 402, so the halt window costs the payer
  nothing. Plus: the same key succeeds once the halt lifts; a payer holding an
  unspent credit still gets their generation while halted (paywall replaced by
  a sentinel that raises if reached).

- **[minor] Reconciled the step-3y boot check with #1495's `gateway_chain()`
  accessor.** Merging main brought in the single `GATEWAY_CHAIN` accessor plus
  a structural test forbidding `DEFAULT_GATEWAY_CHAIN` references outside
  `config.py`; this PR's boot assertion predated it and read the environment by
  hand. Now reads through `config.gateway_chain()` — not just to pass the test:
  a boot check that resolves the chain differently from the money paths it
  protects is verifying the wrong name, which is precisely the #1495 failure.

- **[claim honesty] Corrected the stale "one surface allowed
  `PAYMENTS_DRY_RUN=false`" claim** in `generation_payment._payments_halted()`'s
  docstring and in §1 of this body — see the §1 bullet above for the full
  correction. The `$2.00` price and main's `quote()["chain"] = gateway_chain()`
  are untouched; the stale `$0.15` was **not** resurrected.

- **[claim honesty] Recorded a KNOWN LIMITATION on both startup assertions
  (§2) — documented, not fixed.** Both are fatal only under the **global**
  `PAYMENTS_DRY_RUN`, which `infra/ecs.tf` pins to `"true"` in prod. The rail
  actually settling real money is the generation paywall, live via the
  generation-scoped override — a switch neither assertion reads. **So in prod
  as configured today both assertions are inert: neither can refuse a boot
  even though real value is moving.** It bites hardest on
  `_assert_gateway_chain_matches_rpc`: the live paywall quotes
  `chain: gateway_chain()` to real payers, so a fat-fingered `GATEWAY_CHAIN` is
  exactly what that function exists to catch, and today it would only be
  logged as a warning. Widening the trigger to "any rail is live" changes boot
  behavior and wants its own PR; recorded in both docstrings so neither guard
  is mistaken for protection it does not currently give. **This PR ships the
  assertions knowing they are currently inert in prod** — that is a real
  limitation of what §2 delivers, not a detail.

- **[minor] Dropped two pieces of dead defensiveness.** (a)
  `suppressed = charge_suppressed_ids or set()` in `_halt_publisher` — round 3
  made the parameter required precisely so a forgetful caller gets a
  `TypeError` instead of a silent empty set, which leaves the `or set()`
  unreachable for its stated purpose and re-implements the fallback the
  required parameter exists to remove. (b) `assert result is None` in
  `test_payments_halt_sweep_publisher_is_noop` — `sweep_publisher` is `-> None`
  on every path, so it held whether or not the guard fired: the "passes against
  the unfixed code" shape CLAUDE.md warns about. Removed, with a comment naming
  the real discriminator. Re-confirmed adversarially after the edit (guard
  replaced with `if False:`): the test still fails
  (`Expected 'mock' to not have been called. Called 1 times.`), so the vacuous
  assertion was carrying no weight.

### Test tallies, round 4

- Full unit suite from repo root, the exact command CI runs
  (`pytest -m "not integration" -q`): **4171 passed, 2 skipped (pre-existing,
  unrelated — `test_chain_settings_addresses.py`), 2 deselected, 0 failed**,
  455s.
- `cd ui && npm ci && npm test`: **464 passed, 0 failed.**
- Targeted: `test_per_step_charging.py` + `test_settlement.py` +
  `test_payments_halt.py` + `test_revenue_sweep.py` → 62 passed;
  `test_generation_payment.py` + `test_generation_credits.py` +
  `test_startup_assertions.py` + `test_gateway_chain_lifespan_wiring.py` +
  `test_gateway_chain.py` → 73 passed.
- `cd backend && alembic heads` → exactly one head, `5cb798feef58`.
- `ruff format --check` clean on all touched files; `ruff check` clean
  (the merge left `test_settlement.py`'s import block unsorted — I001 — which
  this round fixed; `main` itself is clean).


dbrowneup:
Transferred to the contracts session. Parked deliberately: this branch touches `generation_payment.py` + `.env.example`, which the sequenced payment-flip stack (#1427→#1414→#1426→#1428) also modifies — fixing/rebasing before that stack lands guarantees a second conflict round. Plan: after #1428 merges, rebase this to 0-behind, apply the outstanding review findings, mutation-prove the kill-switch guard (input that SHOULD trip it, shown tripping), then flip from draft. Note: the strategy session's cleanup-verdicts file lists 1401 in its `fixed` set while its handoff cites 1-major+4-minors outstanding — reconciling against the PR's actual review threads is step one of the fix round.

PR #1408, opened 2026-08-20
[backend] Generation admission control: run-cap semaphore + bounded wait queue

dbrowneup:
## What

POST /api/generate/start currently fire-and-forgets an unbounded number of concurrent pipelines into the single shared Fargate task. One pipeline averages ~65% of the task's vCPU for ~48s, so N parallel generations starve auth, SSE, and the ALB health check — worst case the task is killed and every in-flight job dies.

- At most `GENERATION_MAX_CONCURRENT` (default 1) pipelines run at once (asyncio.Semaphore, per-event-loop singleton).
- Up to `GENERATION_MAX_QUEUE` (default 10) more wait; a waiting job stays `queued` and its SSE stream receives a `job_queued` event (position + cap) — the existing 15s heartbeat keeps the connection alive while it waits.
- Beyond that, /start refuses **429 `generation_queue_full` BEFORE the payment gate** — a caller is never charged for a slot that doesn't exist. Quota still runs first (cheapest check keeps its place).
- Cancel of a waiting job releases its queue slot.

## Guard demonstrations (per CLAUDE.md rule 4)

Reverted the routes file to origin/main and re-ran the new suite: **7/7 fail** (overlap observed, 202 instead of 429) — the tests reject the unfixed code. Restored: 7/7 pass; neighboring generate suites (payment gate, model gating, heartbeat, cancel scoping) 28/28 pass.

## Notes

- `.env.example` gains the two vars (commented) — env-contract change, flagging for review.
- Prod default stays 1 deliberately: raising the cap without raising task CPU just moves the thrash. The scale path is cap+CPU together (see the deployment cost analysis in the private docs repo).


dbrowneup:
Rebased onto current main (post-train). One semantic merge: main added a per-job heartbeat inside `_run_with_cleanup` — resolved so the heartbeat starts at entry (a QUEUED job waiting on the admission gate is alive and must not read as stalled) and cancels in the outer finally, gate logic unchanged inside the try. Concurrency + heartbeat + payment + cancel + job-status suites: 36/36 on the merged result.

Issue #1402, opened 2026-08-20
[infra] archimedes-runner EC2 wedges repeatedly — impaired instance checks + dead SSM agent (3 incidents)


dbrowneup:
The oracle/agent runner EC2 (`archimedes-runner`, t-class, launched 2026-07-28) has now wedged three times: a 13-day SSM `ConnectionLost` stretch (discovered 2026-08-19), and twice within 24h on 2026-08-20 (SSM dead 05:55–~10:15 CT and again by ~13:30 CT, the second with `InstanceStatus: impaired` while `SystemStatus: ok`). Each time a reboot recovered it; a reboot is not a fix.

**Signature:** instance-level status check fails, system check passes, SSM agent stops pinging — consistent with OS-level resource exhaustion (memory pressure from the docker pair + kb work is the prime suspect; the backend image the containers run was measured at 654 MB RSS on this box) rather than hardware.

**Impact when wedged:** oracle updater and agent runner stop (currently contained — zero vaults, oracle already deadlocked per #1341, Explore reads yfinance directly per #1371) and the DB tunnel jump host (`db-tunnel.sh`) is unusable.

**Investigate/fix candidates:**
1. Read `/var/log/messages` + `dmesg` after the next wedge (or from this boot) for OOM-killer traces; add swap or right-size the instance if confirmed.
2. CloudWatch alarm on `StatusCheckFailed_Instance` with an EC2 auto-recover/reboot action — turns multi-hour outages into ~2-minute blips without human paging.
3. Memory limits on the two docker containers so a leak degrades one container instead of the host.
4. Consider whether the runner belongs on ECS (scheduled/service) like the kb task — removes the pet-box class entirely (relates to #1044's architecture reeval).

**Acceptance:** root cause identified from logs (not guessed); an auto-recovery mechanism in place; runbook note in `docs/runbooks/` for the wedge signature; 7 days without a manual reboot.

dbrowneup:
Fix in PR #1413: 1GiB swap + per-container memory caps (oracle 512m / agent 900m), applied idempotently by deploy-runners' SSM host-prep (live box ignores user-data) and mirrored in runner-user-data.sh for future instances. A breaching container now OOM-kills and systemd-restarts instead of wedging the host. Right-size verdict: t3.micro downsize rejected — halving RAM worsens this exact failure mode; revisit after ≥2 quiet weeks. A StatusCheckFailed_Instance alarm for this box is being added in the dashboard consolidation branch.

[...]

PR #1413, opened 2026-08-20
[infra] Runner box #1402: swap + per-container memory caps (idempotent host-prep)

dbrowneup:
Fixes the 3x wedge class in #1402 (OS memory exhaustion → status check impaired + SSM dead → console reboot).

**Mechanism** — the live instance carries `lifecycle.ignore_changes = [user_data]`, so the fix ships two ways:
1. `deploy-runners.yml` SSM host-prep (runs before every image deploy, idempotent): create 1GiB swapfile (swappiness 10) if absent; insert `--memory=512m` (oracle) / `--memory=900m` (agent) into the unit files only if uncapped; **FATAL exit 5 if either unit ends uncapped** (a silent skip would leave us one leak from wedge #4).
2. `runner-user-data.sh` mirrored for any future instance, with a keep-in-sync note.

**Why these numbers**: observed steady RSS 654MB (agent side) on a 2GB box; caps sum to 1412m leaving ~600m for OS+docker+SSM. A container breaching its cap is OOM-killed and restarted by systemd (`Restart=always`) — a visible, self-healing container restart instead of an invisible host wedge. AGENT_DRY_RUN=true in prod, and the runner lease + next-tick re-evaluation make a mid-tick restart recoverable.

**Guard demonstrations**: host-prep logic run twice on a copy of the real unit text → exactly 1 cap inserted (idempotent); run against a pattern-mismatched unit → FATAL exit 5. YAML parses.

**Right-sizing**: t3.micro downsize is rejected, not deferred — halving RAM makes this failure mode worse for $7/mo. Revisit only after ≥2 quiet weeks.

Takes effect on the next `deploy-runners` dispatch (RUNNER_DEPLOY_ENABLED is set).


dbrowneup:
Amended after the strategy session's #1402 diagnostics (log-silent agent container spinning 100% CPU for 4.5h — a second wedge signature memory caps alone don't address): host-prep now also inserts `--cpus=0.50` (oracle) / `--cpus=1.00` (agent). Whatever spins, the host keeps ≥0.5 of the 2 vCPU for the kernel + SSM agent, so the box stays reachable and diagnosable. Same idempotent-sed + loud-FATAL pattern; adversarial demo re-run (1 cap after 2 runs, mismatch still FATALs). Rebased onto current main.

Issue #1325, opened 2026-08-20
[backend] wallet_provisioner sends a payload the Circle wallets API rejects (never exercised under PAYMENTS_DRY_RUN)


dbrowneup:
Found live while provisioning the platform wallet: `marketplace/wallet_provisioner.py::_create_circle_wallet` posts `{"blockchain": ..., no walletSetId}` to `POST /v1/w3s/developer/wallets`, and the API rejects it with 400: `'blockchains' field may not be empty` + `'walletSetId' field may not be empty`. The current API wants `blockchains` (array) and a `walletSetId`.

This has never fired in prod because `PAYMENTS_DRY_RUN=true` keeps subscriber/publisher provisioning dormant — but it means marketplace wallet provisioning will 400 on the first real run.

- [ ] Fix the payload (`blockchains: [CIRCLE_BLOCKCHAIN]`, `walletSetId` resolved via `GET /v1/w3s/walletSets` or a `WALLET_SET_ID` env)
- [ ] Response parsing: the 201 body nests `data.wallets[0]`, not `data.wallet`
- [ ] Test that pins the request shape against a mocked transport, per house guard rules (the pre-fix payload must be demonstrated failing)

Not MVP-blocking (marketplace is roadmap-hidden per #1266), but must land before `PAYMENTS_DRY_RUN=false` on the marketplace rail.

Issue #834, opened 2026-06-30
[infra] Feature-flag flip-list + dead-flag audit — the go-live checklist (don't lose track)


dbrowneup:
The canonical inventory of feature flags that gate in-progress work. **Flipping these is the FINAL go-live step.** Keep this issue updated as flags are added/flipped/removed.

_Audited + rewritten 2026-07-07. Two go-live blockers remain (#588 contract redeploy, #975 payments custody) — this issue stays OPEN until both close._

## ✅ RESOLVED since the original list (verified in code 2026-07-07)

| Flag | Resolution |
|---|---|
| **`PRICE_SOURCE`** | Flipped to `cascade` in prod box `.env` 2026-07-04 (Pyth live). Gating PR #826 MERGED; cross-check lineage #835→#898 MERGED. Code default stays `yfinance` for local/CI safety. ⚠️ **Fargate parity:** now also set in `ecs.tf` (was missing — cutover would've reverted the backend to yfinance). |
| **`ARCHIMEDES_DEBATE_ENABLED`** | **RETIRED.** #880 (Phase-3 cutover) made the debate society the *unconditional* sole generation pipeline and deleted the flag — no `os.getenv()` reads it anymore. Setting it in `.env` is a **no-op**. The architect/agent/standalone-fusion branches are gone from `generation_pipeline._pick_pipeline` (returns `debate` unconditionally). |
| **`REQUIRE_SIWE_FOR_GENERATION`** | **DELETED** (#1300, 2026-08-19): the flag, `gate_generation`, and `_generation_auth_required` are gone — generation/chat auth is unconditional (`require_current_user` / `require_linked_wallet`). Setting it anywhere is a no-op. (History: secure-by-default via #851.) |

## 🚩 FLIP-WHEN-READY (still gated)

| Flag | Default | Gates | Flip when |
|---|---|---|---|
| **`AGENT_DRY_RUN`** | `true` (SSM `/archimedes/prod/AGENT_DRY_RUN`, SecureString; re-confirmed 2026-08-20) | The agent runner's live-signal switch — while `true`, no live trades fire from strategy signals. | **Dan's explicit live-money go-ahead**, after the F2/F3 interpreter-parity fixes (PR #1320) are merged + deployed so live behavior matches the backtests being trusted. Was missing from this list — added 2026-08-20. |
[...]

PR #1414, opened 2026-08-20
[infra] Payment-flip wiring: recipient TF_VAR (DCW default) + $2.00 generation price

dbrowneup:
Unblocks the #834 paywall flip: the recipient was hardcoded `""` in ecs.tf, so setting it required a code PR at flip time. Now it follows the exact PLATFORM_ADMIN_WALLETS pattern — `TF_VAR_generation_payment_recipient=0x... terraform apply` once the platform DCW address exists (Circle DCW provisioning is the open half; #1325 fixed the payload). Empty default preserves today's behavior bit-for-bit: paywall off, and flag-on-without-recipient stays a deliberate 503.

`terraform fmt -check` clean. No task-definition value changes until the TF_VAR is supplied.

⚠️ Same drift caveat as the admin allowlist: any future `terraform apply` without the TF_VAR empties it again — documented in the runbook (docs repo).


dbrowneup:
Updated: the variable now DEFAULTS to the platform generation-revenue DCW address (`0xffa7…56c1`, located by Dan in the Circle Console) — a public receiving address, committed like the ARC_*_ADDRESS contract addresses. This removes the silently-empties-on-apply drift risk entirely; the TF_VAR override remains for rotation. Plain `terraform apply` after merge seeds the recipient with no TF_VAR needed.

PR #1425, opened 2026-08-20
[frontend] Drop-in company/social links footer (github/discord/x/company)

dbrowneup:
## Summary

Drop-in company/social links footer for both UI shells, ahead of the owner
supplying the real Discord invite and X handle (aprin.ai — APRIN Labs — is
also not live yet; domain is being registered).

- `ui/src/companyLinks.js`: single config `{ github, discord, x, company }`.
  `github` is real today (this repo). `discord`/`x`/`company` are `''`.
  **Rule (claims-must-be-true):** an empty string means the link renders
  **nowhere** — no placeholder/dead links ship.
- `ui/src/socialLinks.js`: `activeSocialLinks(links)` — the one place that
  enforces the empty-string filter, plus the icon/label metadata per link.
- `ui/src/components/CompanyLinksFooter.jsx`: shared component both shells
  mount; renders one icon-only `<a target="_blank" rel="noopener noreferrer">`
  per non-empty link, or nothing at all.
- **Public shell** (`PublicLayout.jsx`): the footer that used to live only in
  `Landing.jsx` (so Architecture had none, and Landing would have gotten a
  duplicate) is moved up to `PublicLayout.jsx`, which wraps Landing,
  Architecture, and not-found — one footer, consistently, everywhere in the
  public shell, brand+tagline on the left, links on the right.
- **App shell** (`Layout.jsx`): new layout-bottom `<footer className="app-footer">`
  under `<main>`, right-aligned, mirroring the public footer's role.
- Icons reuse the existing UnoCSS idiom (`i-simple-icons-github/-discord/-x`,
  `i-lucide-globe` for the company link) and the existing `.topbar-icon-btn`
  hover/focus treatment (new `.footer-icon-link` class). New icon classes
  added to `uno.config.js`'s safelist (they come from a JS object, not a
  static string UnoCSS's scanner can see).

## Owner drop-in steps

1. Fill in `ui/src/companyLinks.js`:
   - `discord: "<invite URL>"`
   - `x: "<profile URL>"`
   - `company: "https://aprin.ai"` — **only once the site actually resolves**
2. Update the three pinned-empty assertions in
   `ui/test/company-links.test.js` (`companyLinks.discord === ''` etc.) —
   they fail on purpose the moment a value is filled in, so this step can't
   be skipped silently.
3. Flip the PR out of draft.

No other code changes needed — icons and rendering are already wired and
will appear automatically.

## Tests

`ui/test/company-links.test.js` (node --test, source-assertion style,
matches the rest of `ui/test/`):

- Pins today's `companyLinks.js` state (github real, discord/x/company `''`).
- Unit-tests `activeSocialLinks()`'s filter against synthetic input
  (independent of `companyLinks.js`'s current values, so it stays a
  permanent regression guard).
- Confirms `activeSocialLinks(companyLinks)` (the real config) yields exactly
  one active link today (github).
- Source-asserts `CompanyLinksFooter.jsx` uses the shared filter (not an
  inline re-check), has the `links.length === 0` → `null` guard, and every
  anchor carries `target="_blank" rel="noopener noreferrer"` +
  `aria-label`.
- Source-asserts both shells wire `CompanyLinksFooter` into their existing
  footer patterns, and that `Landing.jsx`'s old per-page footer is gone
  (no duplicate).
- Asserts every icon class in `SOCIAL_LINK_META` is present in
  `uno.config.js`'s safelist.

### Mutation-check transcript (guards-need-an-adversarial-pass)

Per the repo's guard discipline: before pushing, I disabled the filter in
`socialLinks.js` (commented out `.filter((key) => Boolean(links[key]))`, so
`activeSocialLinks` maps every key unconditionally — the exact "renders
unconditionally" bug the tests exist to catch) and re-ran the suite:

```
$ node --test test/company-links.test.js
✔ companyLinks: github is real, discord/x/company are pinned empty today
✖ activeSocialLinks: empty-string entries are dropped entirely, non-empty entries pass through with their icon/label
✖ activeSocialLinks: all-empty config renders nothing — mutation-check target
✖ activeSocialLinks: today's real companyLinks.js produces exactly one active link — github
✔ CompanyLinksFooter: uses the shared filter, not an inline re-check, and renders nothing when it returns empty
✔ CompanyLinksFooter: every anchor is a real external link with an accessible name
✔ PublicLayout: mounts the shared footer once for every public page (Landing, Architecture, not-found)
✔ Layout: mounts the shared footer at the bottom of every /app page
✔ App.css: both footers style the shared link row with the existing icon-button idiom, not a new one
✔ uno.config.js: every brand icon used in SOCIAL_LINK_META is safelisted

ℹ tests 10
ℹ pass 7
ℹ fail 3
```

3 of 10 tests fail exactly as expected — the "today's real config" test
reported `['github', 'discord', 'x', 'company']` instead of `['github']`,
and the synthetic-input test reported `url: ''` anchors would have been
returned. Reverted the filter immediately after; full suite is green again
(`193 passed, 0 failed`).

## Verification

```
$ node --test        # 193 passed, 0 failed
$ npm run lint        # clean
$ npm run build       # succeeds; pre-existing >500kB chunk warning unrelated
                       # to this change (verified against a baseline build of
                       # unmodified main: the same-sized shared chunk already
                       # exists there, named after health.js instead — the
                       # public shell's Architecture.jsx and the app shell's
                       # Layout.jsx already share that module across the
                       # eager/lazy boundary; CompanyLinksFooter.jsx joining
                       # that same shared-chunk group added ~0.8kB)
```

## Deviations from spec

- Consolidated the public-shell footer into `PublicLayout.jsx` rather than
  only touching `Landing.jsx`, and deleted the old per-page footer in
  `Landing.jsx`. The task named "the public shell — Landing/Architecture
  layout" as the target file, and leaving the old footer in place would have
  produced either a duplicate footer on Landing or no footer at all on
  Architecture. No test pinned the old two-`<span>` footer structure, so
  this was a safe, minimal-risk consolidation.
- Company link's icon is `i-lucide-globe` (generic site icon) rather than a
  brand mark, since APRIN Labs has no distinct brand icon shipped in this
  repo's icon collections.

Claude-Session: https://claude.ai/code/session_01HCr9L8iz189eQ6v1QTPz2M

### DIFF
--- a/ui/src/App.css
+++ b/ui/src/App.css
@@ -4520,7 +4520,48 @@ button:where(.tag) {
 .public-footer .public-shell {
 	display: flex;
 	justify-content: space-between;
+	align-items: center;
 	gap: 20px;
+}
+
+.public-footer__brand {
+	display: flex;
+	flex-wrap: wrap;
+	align-items: baseline;
+	gap: 16px;
+}
+
+/* Shared footer icon-link row — public shell (public-footer, above) and app
+   shell (app-footer, below) both mount CompanyLinksFooter.jsx into one of
+   these, undecorated so it inherits whichever shell's --text-3/--accent/
+   --glass-border/--accent-muted tokens are in scope. Mirrors the existing
+   .topbar-icon-btn idiom (Layout.jsx) rather than inventing a new one. */
+.footer-links {
+	display: flex;
+	align-items: center;
+	gap: 10px;
+}
+
+.footer-icon-link {
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
+	width: 28px;
+	height: 28px;
+	border-radius: 6px;
+	border: 1px solid var(--glass-border);
+	color: var(--text-3);
+	transition:
+		color 0.15s ease,
+		border-color 0.15s ease,
+		background 0.15s ease;
+}
+
+.footer-icon-link:hover,
+.footer-icon-link:focus-visible {
+	color: var(--accent);
+	border-color: var(--accent);
+	background: var(--accent-muted);
 }
 
 .public-site .page-content {
@@ -4853,6 +4894,17 @@ button:where(.tag) {
 .app-site .main-area {
 	min-height: 100vh;
 	background: var(--app-canvas);
+}
+
+/* Layout-bottom footer (Layout.jsx) — mirrors .public-footer's role in the
+   public shell: the one place CompanyLinksFooter.jsx mounts for every /app
+   page. Right-aligned, unlike the public footer's brand+links split, since
+   the sidebar already carries the Archimedes wordmark on every app page. */
+.app-site .app-footer {
+	padding: var(--space-4) var(--space-6);
+	border-top: 1px solid var(--glass-border);
+	display: flex;
+	justify-content: flex-end;
 }
 
 .app-site .sidebar {
--- a/ui/src/companyLinks.js
+++ b/ui/src/companyLinks.js
@@ -1,0 +1,29 @@
+// Company and social links rendered in both shells' footers — the public
+// site (PublicLayout.jsx, wraps Landing + Architecture) and the
+// authenticated app (Layout.jsx). Both consume this through
+// socialLinks.js's activeSocialLinks(), which is the one place that reads
+// these values; nothing else in the UI should import this object directly.
+//
+// THE RULE (claims-must-be-true, see CLAUDE.md's "hard constraint, above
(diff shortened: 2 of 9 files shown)


dbrowneup:
Closing per tonight's triage: the public-shell half is replaced by the #1469 rebrand (direct file collision on all four meaningful files), and without real link values the claims-must-true rule leaves only the GitHub link rendering. Intent preserved in a follow-up issue; the branch work remains referenceable here.

Issue #1367, opened 2026-08-20
[auth][privacy] Account management is read-only — no email/password change, no session revocation, no deletion, and no privacy or terms page


dbrowneup:
## Summary

`/app/account` is a read-only `<dl>` plus wallet link/unlink and a Sign out button. A signed-in user cannot change their password or email, cannot see or revoke their other sessions, cannot export their data, and cannot delete their account — and there is **no privacy policy and no terms page anywhere in the product**, while a verification mail is sent to a real address on every signup. #1323 / PR #1332 close the *lockout* half (reset + resend verification); this is the residual account-management and legal-surface half.

Deletion in particular is **not a flag flip**: the schema would orphan the user's rows rather than remove them, including the Fernet-encrypted email.

### 1. The server never enables the capabilities

`auth/auth.js:104` is the entire `user` config:

```js
user: { modelName: 'auth_users' },
```

No `changeEmail`, no `deleteUser`. Both are opt-in in the installed Better Auth `1.6.25` (`auth/package.json:13`) and the endpoints refuse without them:

```
auth/node_modules/better-auth/dist/api/routes/update-user.mjs:288
  if (!ctx.context.options.user?.deleteUser?.enabled) {
auth/node_modules/better-auth/dist/api/routes/update-user.mjs:434
  if (!ctx.context.options.user?.changeEmail?.enabled) {
```

The rest are already live server-side and simply have no client:

```
dist/api/routes/update-user.mjs:75   createAuthEndpoint("/change-password", ...)
dist/api/routes/session.mjs:375      createAuthEndpoint("/list-sessions", ...)
                                     // + /revoke-session, /revoke-sessions, /revoke-other-sessions
```

### 2. The client exports none of them

`ui/src/auth-client.js` (40 lines) exports exactly six calls: `getSession`, `getProviders`, `signInEmail`, `signUpEmail`, `signInSocial`, `signOut`. Open PR #1332 adds three more (`requestPasswordReset`, `resetPassword`, `resendVerificationEmail`). **Even after #1332 merges there is still no `changePassword`, `changeEmail`, `listSessions`, `revokeSession`, or account deletion.**

### 3. `/app/account` has no editor

`ui/src/components/AccountSettings.jsx:69-73` is the whole Profile section:

```jsx
<dl className="grid grid-cols-[auto_1fr] gap-x-4 gap-y-2 text-sm">
  <dt className="caption">Name</dt><dd>{user?.name || '—'}</dd>
[...]

Issue #1323, opened 2026-08-20
[auth] No password reset exists — a forgotten password permanently destroys the account that owns the user's strategies


dbrowneup:
## Summary

There is **no password reset anywhere in the product** — no "Forgot password?" link, no reset endpoint wired, no recovery path of any kind. Since #1194 made the account the owner of a user's work, a forgotten password now means **permanent, unrecoverable loss of every strategy that account owns.**

## Evidence

**Server side — the handler is never configured.** `auth/auth.js:74-81`:

```js
emailAndPassword: {
  enabled: true,
  autoSignIn: false,
  minPasswordLength: 12,
  maxPasswordLength: 128,
  revokeSessionsOnPasswordReset: true,
  requireEmailVerification: emailVerificationEnforced(env),
},
```

Better Auth's `forgetPassword` flow requires `emailAndPassword.sendResetPassword`. It is absent:

```
grep -n "sendResetPassword" auth/auth.js auth/server.js
```
→ **no match.**

Note `revokeSessionsOnPasswordReset: true` is set — the config was written *expecting* reset to work. Only the delivery callback is missing.

**Client side — no entry point.** `ui/src/components/AuthPage.jsx` renders Email, Password, Sign in, and "New to Archimedes? Create account". There is no reset affordance:
```
grep -rni "forgot\|reset password" ui/src/
```
→ **no match.**

**The mailer already exists.** `auth/mailer.js` provides a working two-mode sender (`ses` in deployed envs, `console` for local) and is already wired for `sendVerificationEmail` at `auth/auth.js:85-94`. This is a small amount of work, not a new subsystem.

## Why this is P1, not polish

- `minPasswordLength: 12` — long passwords are exactly the ones users fail to recall.
- The sign-in page itself tells the user the stakes: *"Account owns your strategies and settings."*
- Generated strategies are the product's compounding artifact. Losing the account loses the library.
- Related lockout, same root: `requireEmailVerification` is enforced via `emailVerificationEnforced(env)`, and there is **no resend-verification UI**. A user whose verification mail is lost is equally stuck.

## Acceptance criteria (machine-checkable)

1. `sendResetPassword` is configured against the existing mailer:
   ```
   grep -n "sendResetPassword" auth/auth.js     # matches
   cd auth && npm test                          # 0 failing
   ```
[...]

PR #1429, opened 2026-08-20
[backend] D3: account-deletion cascade-vs-anonymize policy (#1367)

dbrowneup:
> ## ⚠️ Merge order: this PR lands AFTER #1438
>
> Migration `85ca5310b7a1` is serialized behind #1438's last revision (`9c2e7b5a1f4d`) and
> **drops a constraint #1438 creates** (`fk_paper_deployments_owner_user_id`). It cannot run
> on a chain that lacks #1438. Until #1438 merges, this branch carries #1438's commits as a
> merge so the chain resolves and CI is honest; once #1438 is on `main` those commits become
> ancestors of `main` and drop out of this PR's diff on their own — nothing to do here.

## Scope

Issue #1367 bundles five separately-PR-able deliverables (D1–D5). This PR ships **D3 only**
— the cascade-vs-anonymize migration + its orphan/PII test — for the reasons below. It does
**not** auto-close the tracking issue (see 1367); see the tracking comment on the issue for what remains.

Why D3 alone, not the whole issue in one PR:
- **D1** (privacy/terms pages) needs real legal copy, which the issue's own anti-goals
  reserve for Dan ("Do NOT generate the privacy/terms copy with an LLM and merge it
  unreviewed. The text is Dan's call.") — genuinely decision-blocked, not buildable here.
- **D2** (password/email change, session list+revoke) is a new auth surface across both the
  Node Better Auth service and the UI. Per `CLAUDE.md` § *Security ships with the product,
  not after*, security/PII-adjacent surfaces "get more careful review than the diff size
  suggests" — bundling it into a migration PR works against that, not for it.
- **D4** (`deleteUser: { enabled: true }`) is explicitly gated by the issue's own anti-goal:
  *"Do NOT set `deleteUser: { enabled: true }` before D3's migration and orphan test are
  green."* Shipping D4 in the same PR as D3 would mean enabling account deletion in the same
  breath as proving it's safe — the sequencing point is to let D3 land and be verified on
  its own first.
- **D5** (dead Profile menu affordance / orphan `POST /api/user/profile`) turned out to be
  entangled with D2/D3 in a way that needs a real design call, not a mechanical fix — see
  the issue comment for the specific fork.

## What this PR does

`b7e3f1a2c9d4` gave `strategy_store`, `strategy_passports`, `strategy_proposals`,
`user_profiles`, and `vault_metadata` a nullable `owner_user_id` FK'd to `auth_users`, all
`ON DELETE SET NULL`, uniformly — without asking whether "detach and keep the row" is the
right answer for every one of them. A sixth ownership column,
`paper_deployments.owner_user_id`, had **no FK at all** when this issue was filed; #1438 has
since given it one — also `SET NULL`, for consistency with the five rather than as a
considered per-table call.

Migration `85ca5310b7a1` makes an explicit, per-table call for each of those six, recorded in
the migration's own docstring, and enforces it at the database level:

| Table | Decision | Why |
|---|---|---|
| `user_profiles` | **CASCADE** (changed from SET NULL) | Holds the Fernet-encrypted email — the one row that exists purely to carry PII. SET NULL would detach ownership but leave the encrypted address in the DB forever. |
| `paper_deployments` | **CASCADE** (changed from SET NULL) | A private per-user paper-trading ledger — no other account reads or depends on it. Its child table already cascades off `paper_deployments.id`. |
| `strategy_store`, `strategy_passports`, `strategy_proposals` | SET NULL (unchanged) | Can be public marketplace/leaderboard/audit artifacts other accounts reference by id — deleting the row out from under those references would break other users' data. |
| `vault_metadata` | SET NULL (unchanged) | Describes an on-chain vault anchored by `owner_wallet` independent of the Better Auth account; not solely the deleting account's to remove. |

### Scope of "all six" — stated so it can't be read as more than it is

**"All six" means the six `owner_user_id` columns in that table, and nothing else.** It is
NOT a claim that every user-referencing column in the schema has been ruled on. Two more
have landed on `main` since this issue was filed and are **deliberately out of scope**, named
here and in the migration docstring rather than left to be discovered later:

- **`payment_receipts.user_id`** (`1752121b8d7c`, #1456)
- **`generation_credits.user_id`** (`a3f19c7d2e84`, #1441)

Neither carries a foreign key to `auth_users.id` at all today, so neither has an `ON DELETE`
action for this revision to change. Giving them one is an additive FK retrofit in #1438's
shape, not a policy edit in this one — and both raise a question this revision has no answer
for: a receipt is a financial record, and "delete it with the account" vs "keep it" are both
defensible. That is a call for whoever owns the money path, not a default to be picked
silently inside a deletion-policy migration. Tracked as follow-up work.

### Reconciliation with #1438

#1438 creates `fk_paper_deployments_owner_user_id` (SET NULL) **and**
`ix_paper_deployments_owner_user_id`. This revision therefore:

- **ALTERs** that constraint — `DROP CONSTRAINT` by #1438's name, then re-add with
  `ON DELETE CASCADE` — instead of creating one. Creating it would abort the Postgres deploy
  with *constraint … already exists*.
- **does not touch the index** in either direction. It is #1438's to own; `downgrade()` does
  not drop it.
- `downgrade()` restores `paper_deployments` to `SET NULL` rather than removing the
  constraint, so downgrading to this point leaves exactly the schema `9c2e7b5a1f4d` built.

`op.batch_alter_table` is used throughout for SQLite compatibility with this repo's Alembic
test harness (`test_alembic_migrations.py` runs real `alembic upgrade head` against SQLite
tmp files).

## Prod pre-flight — run this before the deploy

The `paper_deployments` re-add is a plain `ADD CONSTRAINT` (no `NOT VALID`), so **Postgres
validates it against every existing row as part of the statement and aborts the whole
migration on the first orphan.** #1438 avoided that for the *creating* step by adding the FK
`NOT VALID` and validating separately behind a live orphan gate; this revision's re-add gets
no such gate, because a re-add that skipped validation would silently downgrade a constraint
`9c2e7b5a1f4d` may have just validated.

The control is therefore an operator check, not a code path. Run against prod Aurora before
deploying, and expect **0** from every row:

```sql
-- (1) Orphans that would ABORT this migration's paper_deployments re-add.
--     Must be 0. Same query 9c2e7b5a1f4d's gate uses for this constraint.
SELECT COUNT(*) AS would_abort_the_deploy
FROM paper_deployments pd
LEFT JOIN auth_users au ON pd.owner_user_id = au.id
WHERE pd.owner_user_id IS NOT NULL AND au.id IS NULL;

-- (2) Same, for the user_profiles re-add.
SELECT COUNT(*) AS would_abort_the_deploy
FROM user_profiles up
LEFT JOIN auth_users au ON up.owner_user_id = au.id
WHERE up.owner_user_id IS NOT NULL AND au.id IS NULL;
```

Non-zero on either means: do **not** deploy this revision yet. Clean the orphans (or decide
they are expected and re-shape the re-add as `NOT VALID`) first — a hard abort mid-migration
is exactly the failure mode #1438's whole `NOT VALID` design exists to prevent, and this
revision must not reintroduce it by accident.

Two further counts, informational rather than blocking — they say how much data the new
policy will actually destroy the first time an account is deleted, which is worth knowing
*before* D4 flips `deleteUser` on:

```sql
-- Rows the two CASCADEs will destroy along with their account.
SELECT
  (SELECT COUNT(*) FROM user_profiles     WHERE owner_user_id IS NOT NULL) AS profiles_cascaded,
  (SELECT COUNT(*) FROM paper_deployments WHERE owner_user_id IS NOT NULL) AS deployments_cascaded;

-- Rows the BEFORE DELETE trigger reaches that the FK alone cannot:
-- unclaimed, but on a wallet some account has proven control of.
SELECT
  (SELECT COUNT(*) FROM user_profiles up
     WHERE up.owner_user_id IS NULL
       AND up.wallet_address IN (SELECT address FROM linked_wallets)) AS profiles_via_trigger,
  (SELECT COUNT(*) FROM paper_deployments pd
     WHERE pd.owner_user_id IS NULL
       AND pd.owner_wallet IN (SELECT address FROM linked_wallets)) AS deployments_via_trigger;
```

## Test

`backend/tests/test_account_deletion_cascade.py` seeds one `auth_users` row owning one row in
each of the six tables, issues a bare `DELETE FROM auth_users` (mirroring the real trigger —
either the Node Better Auth process or the Python backend can issue this; the DB's own FK
actions must do the work, not application code only one service would run), then asserts:

- (a) zero rows in any of the six tables still carry the deleted user's `owner_user_id`
- (b) the `user_profiles` row itself — and its encrypted email — is gone from the DB, not
  merely detached; same for `paper_deployments`
- the SET NULL tables' content survives with `owner_user_id` nulled, not deleted
- the second-linked-wallet gap the FK cannot reach is closed by the `BEFORE DELETE` trigger,
  and a stranger's unclaimed row on a never-linked wallet is **not** swept up

**Against the migrated schema, not the ORM's.** Every test runs a real `alembic upgrade head`
into a throwaway SQLite **file** (once per module, copied per test) — not
`Base.metadata.create_all()`, and not in-memory. On Postgres the migration alone shapes the
schema, so a `create_all()`-built database would be a guard over `models/user_profile.py`'s
`ondelete=` kwarg rather than over the migration, and it would not carry the trigger at all.

**SQLite does not enforce FK `ON DELETE` actions by default** — `test_identity_schema.py`'s
own module docstring documents this and is why every other FK test in this suite checks
presence via introspection rather than enforcement. This test's whole point is to prove the
actions actually fire, so it turns `PRAGMA foreign_keys=ON` on for its own engine. That
satisfies the issue's anti-goal — stay on sqlite, no Postgres/Redis — by enabling real
enforcement on sqlite rather than widening to an integration test.

### One guard deliberately reads Postgres DDL instead of sqlite

`test_migration_alters_the_paper_deployments_fk_rather_than_creating_a_second_one` renders
`alembic upgrade 9c2e7b5a1f4d:85ca5310b7a1 --sql` against the postgresql dialect (no live
connection — the URL only selects the dialect) and asserts the `DROP CONSTRAINT` precedes the
`ON DELETE CASCADE` re-add, and that `ix_paper_deployments_owner_user_id` is never named.

It has to, because **the failure is invisible on sqlite.** `batch_alter_table` there is a
reflect-copy-drop table rebuild that silently absorbs a `create_foreign_key` for a name that
already exists, leaving one correct-looking FK behind. Measured, not assumed: with the
`DROP CONSTRAINT` removed, all four sqlite tests stay **green**. Postgres is the only place
this policy runs and the only place the collision is fatal, so that is where the guard reads.

`test_migrated_schema_is_what_is_under_test` now also covers all six columns rather than just
the two CASCADE ones, and **counts** the FK rows on each `owner_user_id` before reading the
action — a `{from_column: on_delete}` dict collapses two constraints on one column into a
single entry and cannot see a duplicate at all.

```bash
ENV_BIN=/path/to/conda/envs/archimedes/bin    # `conda env list` prints the prefix
"$ENV_BIN/pytest" backend/tests/test_account_deletion_cascade.py -q
# 5 passed
```

### Mutation checks (CLAUDE.md § "A guard must be shown to reject something")

Each applied to the migration, run, and reverted:

| Mutation | Result |
|---|---|
| Re-create the FK **and** re-create the index (the pre-#1438-reconciliation shape) | **1 failed, 4 errors** — `alembic upgrade head` itself dies on the duplicate index, before any assertion runs |
| Remove only the `DROP CONSTRAINT`, leaving the plain re-add (the sqlite-invisible one) | **1 failed, 4 passed** — *only* the new Postgres-DDL guard catches it: `this revision does not DROP fk_paper_deployments_owner_user_id before re-adding it` |
| `user_profiles` CASCADE → SET NULL in the migration (ORM model untouched) | **4 failed, 1 passed** — `migrated user_profiles.owner_user_id is ON DELETE SET NULL, expected CASCADE`, plus both behavioural tests |
| Trigger not created in `upgrade()` | **2 failed, 3 passed** — *the second linked wallet's unclaimed profile survived account deletion; its Fernet-encrypted email is still in the DB*, plus the schema assertion naming the missing trigger |
| Trigger broadened, `linked_wallets` scoping dropped | **1 failed, 4 passed** — *an unrelated wallet's profile was destroyed by another account's deletion*. This test exists precisely so a purge that would sweep up every wallet-only visitor's row cannot pass. |

All reverted; suite green again (5 passed, confirmed above).

### Regression + gate runs (after merging #1438 and `main` @ `3dd63e5e`)

```bash
"$ENV_BIN/pytest" backend/tests/test_alembic_migrations.py -q
# 23 passed

cd backend && "$ENV_BIN/python" -m alembic heads
# 85ca5310b7a1 (head)   — exactly one head, unbroken chain

"$ENV_BIN/python" .github/scripts/migration_chain_guard.py --base-ref origin/main
# migration_chain_guard: OK — extends origin/main's head (5cb798feef58) cleanly

# Full CI unit-suite gate, from repo root:
"$ENV_BIN/pytest" -m "not integration" -q
# 4134 passed, 2 skipped, 2 deselected, 0 failed
```

### Lint

```bash
"$ENV_BIN/ruff" format --check backend/archimedes/models/paper_store.py \
  backend/migrations/versions/85ca5310b7a1_account_deletion_cascade_policy.py \
  backend/tests/test_account_deletion_cascade.py
# 3 files already formatted

"$ENV_BIN/ruff" check <same files>
# All checks passed!
```

## Anti-goal compliance

- ✅ Did **not** set `deleteUser: { enabled: true }` anywhere — grep confirms:
  `git grep -n "deleteUser" auth/auth.js` → no match on this branch.
- ✅ Did **not** touch #1323 / PR #1332's surface (`sendResetPassword`, `/reset-password`,
  resend-verification) — no changes under `auth/` or `ui/src/auth-client.js` in this diff.
- ✅ Did **not** widen the orphan test into an integration test needing Postgres or Redis —
  sqlite only (a throwaway file rather than in-memory, so a real `alembic upgrade head` can
  build it; still no service dependency).
- ✅ Did **not** flip `EMAIL_VERIFICATION_ENFORCED` — no changes under `infra/`.

## Review

Per `CLAUDE.md` § *"Security ships with the product, not after"*, this touches account
deletion semantics (auth + PII-adjacent), so it warrants the same careful review as the
issue itself calls for even though the diff is small. No contracts, no funds, no UI or
auth-service changes in this PR.

Two things a reviewer should specifically weigh in on:

1. **The pre-flight orphan check above is an operator step, not an automated gate.** If you
   would rather this revision could never abort a deploy, the alternative is to re-add both
   constraints `NOT VALID` (matching #1438) and accept that they stay unvalidated until
   someone validates them. Not done unilaterally — it trades a deploy risk for a weaker
   constraint, and that is a call, not a cleanup.
2. `payment_receipts.user_id` / `generation_credits.user_id` being out of scope (above).

Closes nothing by itself — progresses #1367 (D3 of 5). Follow-up status posted as a comment
on the issue.


dbrowneup:
Triage (transferred to contracts session): the auto-close blocker is FIXED — the body's own "**not** close #1367" was matched by GitHub's closing-keyword parser (verified closingIssuesReferences now empty). Two majors from review remain REQUIRED before this leaves draft:
1. `test_account_deletion_cascade.py` guards the ORM models, not migration `85ca5310b7a1` — on Postgres the migration alone shapes the schema, so the guard passes even if the migration's CASCADEs are gutted. The test must assert against the migrated schema (run the migration into the test DB and inspect FK rules), not model metadata.
2. The migration's CASCADE reaches at most ONE `user_profiles` row per account (`owner_user_id` UNIQUE + the link path refusing second claims) — a user with a second linked wallet whose profile row predates linking is left orphaned on deletion. Needs either a broadened cascade path or an explicit anonymize step for unclaimed rows, with a test that constructs exactly that two-wallet fixture.
Recommendation: keep as DRAFT and fix (the D3 scope is sound; the defects are in the guard + edge-case, not the design). Not worth closing/respeccing.

PR #1430, opened 2026-08-20
[backend] Fix _weights_present: huggingface nests the cache under hub/ (regression from #1328)

dbrowneup:
## Summary

**This fixes a regression I introduced in #1328.** Production has been reporting `paper_rag: "degraded"` and `corpus_embedded: false` ever since it deployed — on a container where the model loads perfectly.

Right now, live:

```
$ curl -s https://archimedes-arc.com/health | jq '{paper_rag, corpus_embedded}'
{ "paper_rag": "degraded", "corpus_embedded": false }

$ curl -s https://archimedes-arc.com/health/paper-rag
{"paper_rag":"live","reason":"model=all-MiniLM-L6-v2"}
```

The probe endpoint proves the model loads. The cheap check says it's broken.

## Cause

`_weights_present()` scanned **one level** with `root.iterdir()`. huggingface_hub **nests** its cache. From the actual production image:

```
$ docker run --rm --entrypoint sh <prod-image> -c 'ls -1 $HF_HOME'
hub
xet

$ find $HF_HOME -name "*.safetensors"
/app/model_cache/hub/models--sentence-transformers--all-MiniLM-L6-v2/snapshots/1110a243…/model.safetensors
```

The top level of `/app/model_cache` holds only `hub` and `xet`. Neither contains the model slug, so the loop matched nothing and returned `False`.

**This is worse than the fail-soft it replaced.** `degraded` means specifically "a load was attempted and failed". Nothing was attempted. It asserts a failure that did not happen — on `corpus_embedded`, which is a public claim surface. I introduced exactly the class of defect I've spent tonight filing against other code.

## Fix

Walk the tree instead of one level; match the model directory by full slug **or** bare name (`all-MiniLM-L6-v2`, so the `models--sentence-transformers--` prefix is not load-bearing); require an actual `.safetensors`/`.bin` **inside** it, so an empty directory is not mistaken for a present model. Memoised into `_weights_present_cache` — the cache is a static build artifact and the ALB polls `/health` every 30 seconds.

## Why it shipped, and what changed in the tests

The #1328 tests mocked the function that was broken:

```python
with patch.object(rag_module, "_weights_present", return_value=True):
```

That exercises the caller and never the callee — `CLAUDE.md` § "Mock at boundaries, not internals". The mock asserted the very thing that was wrong.

Replaced with a real temp directory laid out exactly as the image bakes it (`hub/models--sentence-transformers--<model>/snapshots/<sha>/model.safetensors`, plus a sibling `xet/`), covering: the nested layout, the `pytorch_model.bin` variant, a model directory that exists but holds no weights, `HF_HOME` unset, and a memoisation assertion.

## Verification

**Inside the real production image, with the real baked cache:**

```
weights_present : True
/health  status : ready
/health  reason : model=all-MiniLM-L6-v2 present, not loaded (no retrieval yet)
RSS delta       : 0.0 MB          torch imported: False
probe=True      : live
```

So the regression is fixed *and* #1328's actual benefit — never loading 521 MB of torch on the ALB probe path — is preserved.

```
pytest backend/tests/test_paper_rag.py -q     → 37 passed
ruff format --check + ruff check              → clean
```

**Mutation check:** restoring the `iterdir()` scan fails **3** tests, including `test_finds_weights_in_the_real_nested_hf_layout` and `test_ready_when_weights_present_but_unloaded` (`AssertionError: assert 'degraded' == 'ready'`).

### A second bug the tests could not have caught

While fixing this I removed `_embedding_model` and `_embedding_load_attempted` (module globals that sat inside the replaced region) and **the suite still passed 37/37** — because `_reset_model_cache()` in the test file *assigns* those globals, creating them at runtime and masking their absence. Only running the module standalone in the prod image surfaced `NameError: name '_embedding_model' is not defined`. Both are restored, and a module-level name diff against `main` now shows only the two intended additions (`_scan_for_weights`, `_weights_present_cache`).

Worth noting for reviewers generally: a test helper that assigns a module global will hide that global's deletion.

## Anti-goals honoured

- Does **not** relax `/health` back into loading the model — the RSS-delta and `torch not in sys.modules` guards from #1328 are intact and still mutation-checked.
- Does **not** widen `corpus_embedded` to accept `ready`. Weights on disk is still not proof the model loads; `corpus_embedded` remains strictly `status == "live"` and flips on first real retrieval.
- No change to `_get_embedding_model()`, the `torch.set_num_threads(1)` CPU guardrail, or the TF-IDF fallback.

### DIFF
--- a/backend/archimedes/services/paper_rag.py
+++ b/backend/archimedes/services/paper_rag.py
@@ -121,45 +121,66 @@ def paper_rag_health(probe: bool = False) -> PaperRAGHealth:
     return PaperRAGHealth(status="degraded", reason="embedding model unavailable, TF-IDF fallback")
 
 
-def _weights_present() -> bool:
-    """Cheap on-disk check that the baked model cache holds our model.
-
-    The Docker image bakes the weights into ``HF_HOME`` at build time
-    (backend/Dockerfile:70) and runs with ``HF_HUB_OFFLINE=1``, so "the
-    directory for this model exists and is non-empty" is the strongest claim
-    available without importing torch. It is deliberately NOT reported as
-    ``live`` — presence on disk is not proof the model loads.
-    """
-    hf_home = os.getenv("HF_HOME")
-    if not hf_home:
-        return False
-    slug = _model_name().replace("/", "--")
-    root = Path(hf_home)
-    try:
-        for candidate in root.iterdir():
-            if not candidate.is_dir():
-                continue
-            name = candidate.name
-            if slug in name or name in (slug, f"models--sentence-transformers--{slug}"):
-                return any(candidate.rglob("*.safetensors")) or any(candidate.rglob("*.bin"))
-    except OSError:
-        return False
-    return False
-
-
 def _semantic_enabled() -> bool:
     """Check the feature flag. Default ON for production."""
     val = os.getenv("FUSION_SEMANTIC_RETRIEVAL", "true").strip().lower()
     return val in _TRUTHY
 
 
-# ── Embedding engine ─────────────────────────────────────────────
-
 # Lazy-loaded embedding model (sentence-transformers).
 _embedding_model: Any = None
 # Sentinel: True once a load attempt has been made and failed, so repeated
 # health checks don't retry indefinitely (the result is cached after first try).
 _embedding_load_attempted: bool = False
+
+_weights_present_cache: bool | None = None
+
+
+def _weights_present() -> bool:
+    """Cheap on-disk check that the baked model cache holds our model.
+
+    The Docker image bakes the weights into ``HF_HOME`` at build time
+    (backend/Dockerfile:70-74) and runs with ``HF_HUB_OFFLINE=1``.
+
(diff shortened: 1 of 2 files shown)


dbrowneup:
Adversarial review (two independent lenses + verification of every raw finding): **pass** — no functional defects confirmed. Two minor findings, both addressed:

1. `TestWeightsPresent` mutated the module-global `_weights_present_cache` with no `teardown_method`, unlike the file's three sibling classes that touch module globals — the memoised `True` could leak into later tests in the same pytest process. Fixed in 5470ca94 (teardown mirrors the sibling pattern; 37/37 still pass).
2. PR body said the module-level name diff vs `main` shows "three intended additions"; the AST-level diff shows two (`_scan_for_weights`, `_weights_present_cache` — the other relocated globals pre-exist on `main`). Body corrected.

The PR's core claims were independently re-verified: guard rejects a weightless model dir, memoisation walks the tree once, `corpus_embedded` stays gated on `paper_rag == "live"`.

Issue #1420, opened 2026-08-20
[auth][ux] OAuth account_not_linked redirect lands on the bare landing page and renders nothing


dbrowneup:
With `disableImplicitLinking: true` (correct security posture), an existing email/password user clicking Continue-with-Google/GitHub is redirected to `/?error=account_not_linked` — the marketing landing page, which has no sign-in form and rendered no message at all: a silent failure. Surfaced during the #1289 enablement review (that issue is the external app-registration checklist and correctly states no UI change is needed for the buttons themselves — this is a distinct UX defect on the error path).

Fix (built in the PR that closes this): route `?error=...` on `/` to the sign-in surface and render an honest, mapped message; copy must not promise an account-linking flow (none exists — follow-up candidate).

Acceptance: `resolveRoute('/', '?error=account_not_linked')` redirects to `/sign-in?error=...`; the mapped message renders as a `role=alert` banner; unknown error codes get a generic honest line; all mutation-proven in ui tests.

PR #1433, opened 2026-08-20
[auth] Account linking: explicit link/unlink UI + session-freshness parity on /link-social

dbrowneup:
## Summary

Builds the account-linking flow noted as a follow-up candidate in #1420 / #1419: the
owner has one email/password account and wants Google + GitHub to reach the **same**
account, so any of the three signs in as one identity.

`#1420` itself is already closed (it fixed the silent `?error=account_not_linked`
redirect) and is not reopened by this PR — see **Refs** below, not Closes. What ships
here is the linking flow that #1419's PR description explicitly flagged as unbuilt:
> "A linking flow ... is a follow-up candidate, not covered here."

**What ships today is the explicit link/unlink flow below.** An earlier revision of
this PR also shipped a trusted-provider *implicit* auto-link path; round-2 review found
it bought nothing while the explicit flow was the only thing driving real user value,
and reverted it before merge (see **Round 2** below) — it is not part of what this PR
ships, only background for why the `accountLinking` config looks the way it does.

## Explicit link/unlink — the path that ships, works today regardless of verification state

`auth/package.json` pins `better-auth@1.6.25`. Every claim below was checked directly
against `auth/node_modules/better-auth/dist/{oauth2/link-account.mjs,api/routes/
{account,callback}.mjs}` — quoted inline in the long comment on `accountLinking` in
`auth/auth.js` — and then proven behaviorally in `auth/test/auth.test.js` by driving the
real endpoints (not by asserting against my own reading of the source).

`ui/src/components/AccountSettings.jsx` → **Connected accounts**: lists linked providers
(`GET /list-accounts`), **Link Google** / **Link GitHub** (`POST /link-social`, redirect
flow — state/PKCE/CSRF entirely library-managed, nothing hand-rolled), **Unlink**
(`POST /unlink-account`).

`linkSocialAccount`'s `if (link)` branch in `callback.mjs` does **not** check the base
account's `emailVerified` at all — proof of ownership is the live session `/link-social`
was called with. It **does** still enforce, unconditionally today:
- the provider's own `emailVerified` claim — this holds **only because `trustedProviders`
  is absent** from the shipped config; see **Round 3** below for the correction to an
  earlier, wrong claim that this check was independent of `trustedProviders`, and
- **`allowDifferentEmails` (kept `false`, unchanged)** — the OAuth account's email must
  equal the signed-in account's email, or the callback redirects with
  `?error=email_doesn't_match`.

**Deviation from my own initial assumption, caught by writing the test, not by reading
the source alone:** I expected the explicit flow to need no email match ("session proves
ownership"). It does still gate on `allowDifferentEmails` — proven with a real round trip
in `auth/test/auth.test.js` ("explicit link is refused when the provider's email differs
..."). Documented instead of asserted away; not weakened.

**Unlink guard**, both server- and client-enforced:
- Server: `/unlink-account` throws `FAILED_TO_UNLINK_LAST_ACCOUNT` (400) when
  `accounts.length === 1 && !allowUnlinkingAll` — `allowUnlinkingAll` stays `false`.
- Client: `canUnlink()` in the new `ui/src/account-linking.js` disables the Unlink
  control in the same state, so it's never even clickable — belt and suspenders per the
  task brief, not a substitute for either half alone.

**#1419's message updated**: `ui/src/auth-errors.js`'s `account_not_linked` copy now
reads *"...or sign in with your password and link the provider under Account Settings →
Connected accounts so it signs you in directly next time."* — replacing the old copy
that (correctly, at the time) said no linking flow existed. The old test asserting that
absence is updated to assert the opposite, since that absence is precisely what this PR
resolves.

## Security invariants (built to survive an adversarial read)

**Round-4 review finding (major): the first bullet below was false** — its own
counterexample is documented in this PR's own Round 3 section, below. Rewritten to state
exactly what is enforced, naming the actual mitigation, instead of a blanket guarantee the
code does not back.

- **Attaching a provider identity always requires a session that was live at initiation —
  but the attach step itself performs no session check of its own.** `/link-social`'s
  initiation is gated on a live session (401 with none) plus the same freshness check
  `/unlink-account` uses (Round 2 below); the linking user's id is bound into the signed
  OAuth `state` at that point. The step that actually writes the new `auth_accounts` row —
  `callback.mjs`'s `if (link)` branch, reached only after the OAuth round trip completes —
  trusts that signed `state` alone and never re-checks the session, so a session that goes
  stale (or signs out) between initiation and callback can still complete an attach that was
  validly started while fresh. What actually bounds that window: the OAuth `state`'s
  600-second TTL (`oauth2/state.mjs`), its double-submit `better-auth.state` cookie (defeats
  a bare state/URL replay without the matching cookie), and `allowDifferentEmails` staying
  `false` (the provider's email must match the linking account's). This is a documented,
  bounded gap — see Round 3 below for the full analysis — not the "no path" guarantee this
  bullet used to claim.
- **Unlinking the last credential is impossible** — proven server-side (`auth.test.js`)
  and guarded client-side (`account-settings.test.js`), each independently mutation-tested.
- **CSRF/state stays library-managed for the tests that exercise it.** `auth.test.js`'s
  email-mismatch, unverified-provider-email, and successful-link tests (plus the freshness
  tests) drive the real `/link-social` → `/callback/:id` handshake (the `state` param + its
  `better-auth.state` double-submit cookie) through Better Auth's own code; nothing here
  reimplements it. **Round-4 review finding (minor): this used to say "every test" — it
  doesn't; the config-shape, mailer, and other non-OAuth tests in the same file exercise
  none of this and were never meant to.**

## Mutation-proofs (done by hand before this commit; not committed as parametrized tests
— matches this repo's existing pattern)

| Flip | Test that must fail | Result |
|---|---|---|
| `allowUnlinkingAll: false → true` in `auth/auth.js` | "unlinking the account's only remaining credential is refused" | fails: `400 !== 200`. Reverted. |
| `canUnlink()` hard-coded to `return true` in `ui/src/account-linking.js` | "canUnlink: false at exactly one account..." + "...not just 'always true'..." | both fail. Reverted. |

(A third row from an earlier revision — adding `requireLocalEmailVerified: false` — is
dropped here: it described the round-1 implicit-auto-link config this PR no longer
ships, and only the config-shape test actually fails against today's tree. See Round 3.)

## Docs

`docs/account-authentication.md`: new "Account linking (#1420 follow-up)" section with
the exact verified semantics above and pointers to every implementing file.
`docs/README.md` row updated to mention it.

## Deviations from the brief

- **"No email-match needed" for the explicit flow was wrong** (see above) — documented
  honestly instead, `allowDifferentEmails` unchanged.
- No new dependency, no `docker-compose*`/CI/env-contract file touched, no contract or
  on-chain code touched.

## Refs

Refs #1420 (context — already closed, not reopened by this PR), refs #1419 (introduced
the message this PR updates).

## Round 2 (review fixes)

A second review pass surfaced one blocker, one major, and several minor findings, all
applied on top of the original PR (still draft):

**Blocker — session-freshness asymmetry.** Better Auth gates `/unlink-account` behind its
own `freshSessionMiddleware` (a session must be younger than `freshAge`, default 24h) but
left `/link-social` ungated. With `session.expiresIn` at 7 days, that meant a >24h-old
session could still permanently **attach** a new sign-in credential while the legitimate
owner's own stale session could not **remove** one — a one-way door not even closed by a
password reset (`revokeSessionsOnPasswordReset` revokes sessions, it doesn't touch
already-linked accounts). Fix: `session.freshAge` is now pinned explicitly in `auth.js`
(24h, tracked by a test instead of inherited silently), and a `hooks.before` mirrors the
library's own freshness check onto `/link-social`. Proven against the **real HTTP handler**
(`auth.handler`, what `toNodeHandler`/production actually serve) with a session aged past
`freshAge`: both endpoints now 403 `SESSION_NOT_FRESH` identically; a fresh session and a
no-session request are both unaffected. Mutation-proven by hand: reverting the hook's path
check regresses `/link-social` back to 200 while `/unlink-account` stays 403 — confirming
the test actually exercises the new code. UI: `AccountSettings.jsx` now recognizes
`err.code === 'SESSION_NOT_FRESH'` (added to `auth-client.js`'s thrown `Error` alongside
`.message`), shows an honest "sign in again" message instead of the library's raw string,
disables both Link and Unlink once discovered, and offers a "Sign in again" action.
(Round 3 finding: this covers link-social's *initiation* only — see below.)

**Major — implicit auto-link reverted.** The original PR's Half 1
(`disableImplicitLinking: false` + `trustedProviders: ['google', 'github']`) was found to
buy nothing today — the explicit link/unlink flow (Half 2, the part that actually ships)
never reads `disableImplicitLinking` — while quietly discarding the provider's own
`emailVerified` signal for the day `EMAIL_VERIFICATION_ENFORCED` flips on, at which point
an anonymous "Continue with Google/GitHub" click could silently attach to an existing
account. Reverted `accountLinking` to `disableImplicitLinking: true` with no
`trustedProviders` — identical to `main` before this PR. All tests updated to match: the
"verified account" case that used to succeed now asserts it stays refused; mutation-proven
by flipping the flag back and confirming that assertion fails. Implicit auto-link stays a
possible future change, made and reviewed on its own once email verification is genuinely
enforced. (Round 3 finding: the reasoning above about `trustedProviders` only affecting
this implicit path was itself wrong — see below.)

**Minor fixes:**
- Linking/unlinking a sign-in credential now emails the account owner (`databaseHooks.
  account.{create,delete}.after` in `auth.js`) — an out-of-band signal for exactly the
  account-takeover pattern the blocker above was about. Skips the redundant mail for the
  `credential` row created at signup (already covered by the verification mail). Must
  never throw — unlike the existing `sendResetPassword`/`sendVerificationEmail` patterns,
  Better Auth itself `await`s `create.after`/`delete.after` as part of the write, so an
  uncaught throw would 500 the request being reported on. Mutation-proven: removing the
  try/catch makes a throwing mailer crash the link request instead of degrading silently.
  (Round 3 finding: this guard didn't cover OAuth signup — see below.)
- The `?linked=` success notice on Account Settings rendered straight off the URL with no
  validation, so any `?linked=<anything>` showed a fabricated "Linked anything." toast. It
  now only fires for a recognized provider (`isKnownConnectedProvider`, new exported helper)
  that is actually present in the freshly reloaded account list. (Round 3 finding: still
  spoofable — see below.)
- A rejected `listAccounts()` left the Connected-accounts section reading "Loading…"
  forever, because the empty-array state doubled as the loading sentinel. Tracked with an
  explicit `connectedLoaded` flag instead.
- `account-linking.js`'s header comment pointed at a test file
  (`ui/test/account-linking.test.js`) that doesn't exist; repointed to
  `ui/test/account-settings.test.js`, where those helpers are actually tested.
- `docs/account-authentication.md` updated to match all of the above; `ui/src/auth-errors.js`
  and its test's stale "requireLocalEmailVerified is why this refuses" narrative corrected
  to the unconditional `disableImplicitLinking: true` refusal now in effect (this also makes
  `routes.js`/`routes.test.js`'s existing `disableImplicitLinking` comments accurate again,
  with no edit needed — they were describing `main`'s posture, which this round restores).

**Tests:** `cd auth && npm test` → 39/39 (was 33; 6 net new/rewritten). `cd ui && node
--test` → 283/283 (was 275; 8 new). `cd ui && npm run lint` clean; `cd ui && npm run build`
succeeds. Every new/changed guard was mutation-proven by hand before this commit (flip the
guard, confirm the relevant test fails, revert) — session-freshness parity, the
`accountLinking` revert, `isKnownConnectedProvider`, and the notification try/catch.
**Round-4 correction:** `isKnownConnectedProvider` and this test for it were removed as dead
code in Round 4 (finding #6) once `isLinkableProvider` took over its one real call site —
the count above is this round's own historical snapshot and stays as it was, but is no
longer the current ui total; see Round 4's "Tests:" line at the end of this PR body for the
current, accurate counts.

## Round 3 (review fixes)

A third review pass found one blocker and several major/minor issues in round 2's own
fixes and writeup — including in round 2's security reasoning itself, not just the code.
All applied on top of round 2, still draft:

**Blocker — the round-2 notification fix didn't cover OAuth signup, so brand-new users
got a fabricated account-takeover alert.** `notifyAccountChange`'s guard only special-cased
`account.providerId === 'credential'` — the row email/password signup creates. But
`databaseHooks.account.create.after` also fires for the account row OAuth *signup*
creates: a plain "Continue with Google/GitHub" for an email that owns no account at all
takes `link-account.mjs`'s `createOAuthUser` path (`dbUser` undefined → `isRegister =
true`), bypassing the `accountLinking` gate entirely and writing a `'google'`/`'github'`
row — which the old check let straight through. Reproduced against the real handler: a
first-time Google sign-in got "A sign-in method was added... remove it and reset your
password" as its very first email — impossible advice, since that account has no other
sign-in method (`canUnlink()` refuses to even render an Unlink control for it) and no
password to reset. Fixed by testing account count at hook time instead of providerId:
`databaseHooks.account.create.after` runs only after its triggering write commits (see
the comment in `notifyAccountChange`), so `internalAdapter.findAccounts(account.userId)`
already includes the just-created row — `<= 1` total means registration (either shape),
`> 1` means a genuine link onto an existing account. New test drives a real OAuth signup
round trip for a brand-new email and asserts zero mail sent. Mutation-proven: reverting to
the old `providerId === 'credential'` check makes that test fail with exactly the
fabricated alert present.

**Major — the PR's central "verified against the INSTALLED library" claim about
`trustedProviders` was itself false, in the direction that risks a future silent
weakening.** Round 1 and round 2 both stated `trustedProviders` is read in exactly one
place (the implicit-link path) and does nothing for the explicit flow. `grep -rn
trustedProviders auth/node_modules/better-auth/dist/` shows three call sites, not one:
`oauth2/link-account.mjs:21` (implicit), and — on the **explicit** flow this PR actually
ships — `api/routes/callback.mjs:98` and `api/routes/account.mjs:176`. At both explicit
call sites `trustedProviders` is the same switch: it turns off the requirement that the
*provider's* own `emailVerified` claim hold. Today that requirement holds on the explicit
path only because `trustedProviders` is absent; a future engineer who reads round 2's
comment, believes it only affects implicit linking, and reintroduces
`trustedProviders: ['google', 'github']` to arm implicit auto-link would silently drop
that check from the explicit link callback too. Corrected the comments in `auth.js` and
`docs/account-authentication.md`, and added a source-pinned test on `callback.mjs`'s guard
clause alongside the existing `link-account.mjs` pin, so a future `trustedProviders`
reintroduction has to confront the explicit-path consequence.

**Minor — the `?linked=` success toast (round 2's fix) was still spoofable.**
`isKnownConnectedProvider` — round 2's gate — is true for `'credential'`, which is never
linkable from this UI but is present in `connectedAccounts` for every password user, so
`?linked=credential` passed both of round 2's checks and asserted a link that never
happened; a stale/shared `?linked=google` could likewise re-assert an old link. Added
`isLinkableProvider` (google/github only, not credential) plus a one-shot
`PENDING_LINK_KEY` marker that `link()` sets in `sessionStorage` immediately before its
own redirect and the notice effect consumes exactly once — so the toast can only fire for
a link this tab itself just initiated, never an arbitrary or replayed URL.

**Minor — the freshness-parity docs overstated symmetry.** The round-2 `hooks.before`
gates only `/link-social`'s *initiation*; the step that actually attaches the credential
(`callback.mjs`'s `if (link)` branch, reached at `/callback/:id` after the OAuth round
trip) performs no session check at all — it acts on the signed `link` payload in OAuth
`state`. That window is bounded by the 10-minute state TTL plus the double-submit state
cookie and the `allowDifferentEmails` email match, not by a second freshness check —
which is why this is a documented bound, not an unclosed hole. Reworded the `hooks.before`
comment in `auth.js` and the docs to say so precisely.

**Minor — a failed provider-discovery fetch silently hid the Link controls.**
`getProviders().catch(() => {})` left `connectedProviders` at `{google: false, github:
false}` on any failure, indistinguishable from "no OAuth providers configured" — so a
real fetch failure hid the Link buttons with no explanation, while the section's own copy
kept promising them. Now surfaced via a `connectedProvidersError` alert, the same pattern
round 2 used for `connectedError`.

**Minor — the docs still read as if implicit auto-link were live.** `docs/README.md`'s
index row said "account linking, implicit + explicit"; `docs/account-authentication.md`
said the two paths "are both live" one paragraph before saying path 1 "stays off." Both
reworded to say implicit stays off.

**Minor — this PR's own description was stale after round 2.** The title advertised
"trusted-provider auto-link", which round 2 reverted before merge; the body's first
section described the reverted config as something that "genuinely ships" and carried
test counts superseded by round 2's own numbers 40 lines later; a mutation-proof table row
covered a config this PR no longer ships. Retitled the PR and rewrote the body above
top-down to describe only what ships today, folding the round-2 corrections in rather
than leaving them to be found 40 lines later.

**Tests:** `cd auth && npm test` → 41/41 (was 39; +2: the `callback.mjs` source pin, and
the brand-new-OAuth-signup notification test). `cd ui && node --test` → 287/287 (was 283;
+4: `isLinkableProvider` unit + mutation tests, the pending-link-marker wiring test, the
provider-fetch-error wiring test). `cd ui && npm run lint` clean; `cd ui && npm run build`
succeeds (pre-existing >500kB chunk-size warning, unrelated). Every new/changed guard
mutation-proven by hand before this commit (flip back to the pre-fix behavior, confirm the
new/updated test fails, revert; no residue left in the working tree): the account-count
notification check, `isLinkableProvider`'s exclusion of `credential`, the pending-link
marker's recency check, and the provider-discovery error surfacing.

## Round 4 (review fixes)

A fourth, adversarial review pass confirmed 11 findings across round 3's code and its own
writeup — including, again, in the writeup itself, not just the code. All applied on top of
round 3, still draft.

**Major — stale-session UX was dishonest about the freshness contract.** `/link-social` and
`/unlink-account` are fresh-session gated (24h, `session.freshAge`) while the app session
lives 7 days (`session.expiresIn`) — so for most of a session's life, both Link and Unlink
rendered enabled and any click 403'd (Unlink even popped its destructive `window.confirm`
first). Round 2 already made the UI attempt the action and catch the 403 — what was wrong
was what happened *next*: the "Sign in again" affordance called `logout()`, which lands on
`/`, a page with no sign-in form and no path back to Account Settings. Fixed with a
dedicated `reauthenticate()` (ends the session the same way `logout()` does — required,
because `AuthPage`'s own `user && window.location.replace(next)` guard bounces an
already-signed-in visitor straight past the sign-in form before it can render — then routes
to `/sign-in?next=%2Fapp%2Faccount`, the same `?next=` convention every other anonymous
bounce in this app uses: `routes.js`'s `postAuthPath`/`safeNextPath`, `Layout.jsx`,
`Leaderboard.jsx`, `StrategyPassport.jsx`). The message itself now reads "For security, you
need to sign in again before changing sign-in methods." (`ui/src/auth-errors.js`). The
server-side freshness gate in `auth/auth.js` is untouched — neither removed nor lengthened —
and the client still never precomputes session age; only the catch-handling itself (attempt
→ react to the server's actual response) moved, into a new `connectedActionErrorState`
(`ui/src/account-linking.js`) shared by both `link()` and `unlinkConnected()`, so it is
directly unit-tested instead of duplicated inline. New tests: 4 direct unit tests on
`connectedActionErrorState` (a mocked `{code: 'SESSION_NOT_FRESH'}` error must produce
*only* `{stale: true, message: null}` — the honest re-auth state, no fabricated notice, no
raw library string standing in for it) plus wiring tests pinning the
`reauthenticate()`/`?next=` redirect, the shared-handler dedup, and the "never sets a notice
in this branch" property. Mutation-proven by hand: reverting `connectedActionErrorState` to
always return `{stale: false, message: err?.message}` fails its direct unit tests;
reverting the "Sign in again" button back to `onClick={logout}` fails the wiring test. Both
reverted.

**Major — the PR's headline security invariant was false.** "No path attaches a provider
identity without a live session" — this PR's own Round 3 section documents the
counterexample (`callback.mjs`'s attach step performs no session check of its own; bounded
instead by the OAuth `state` TTL + double-submit cookie + `allowDifferentEmails`). Rewrote
the Security invariants section (top of this PR body) to state exactly what is enforced and
name the real mitigation, and narrowed the sibling "every test drives the real handshake"
claim to the tests that actually do — the config-shape, mailer, and other unit-level tests
in `auth.test.js` never touch `/link-social`/`/callback/:id` and were never meant to. No
code changed for this finding — it was a prose defect, fixed as prose, per this repo's
guard-adversarial-pass rule that a false claim in a PR description is the same defect as an
unenforced guard, just harder to grep for.

**Minor — the provider-emailVerified guard on the explicit-link path was proven only by
regex over vendored source.** The round-3 source-pinned test confirms `callback.mjs:98`'s
guard clause exists; nothing drove it end to end. Added a behavioral test in
`auth/test/auth.test.js`: same email as the signed-in account (isolates this guard from
`allowDifferentEmails`, already covered by its own test) but the provider's own claim is
`emailVerified: false` — asserts the callback redirects with `error=unable_to_link_account`
and nothing is attached to the account. Mutation-proven: temporarily arming
`trustedProviders: ['google']` (which round 3 already established also disables this exact
guard on the explicit path, not just the implicit one) makes the new test fail — the account
list gains `'google'` and the redirect carries no error. Reverted.

**Minor — the account-change notification had a silent failure mode.** Of
`notifyAccountChange`'s two failure paths, only "mailer.send() throws" logged anything
(`account ${action} notification email send failed:`); "the recipient can't be resolved at
all" (`internalAdapter.findUserById` returns nothing) hit a bare `return` with zero log
line — a second, silent way for this account-takeover signal to simply not go out, with no
operator-visible trace. Both modes now share one greppable marker,
`ACCOUNT_CHANGE_NOTIFY_FAILED`; neither throws — the notification stays fail-open for the
user-facing link/unlink flow (a mail outage must not block linking, the actual security
controls are the freshness gate and the server-side link/unlink guards, both untouched) — it
just never again fails *invisibly*. `notifyAccountChange` is now exported from `auth.js` for
direct unit testing of the previously-silent branch, which is impractical to reach through
the real HTTP handler on demand (`internalAdapter` always resolves the just-committed row in
practice). Mutation-proven: reverting either log call (back to a bare `return`, or to a
marker-less message) makes its respective new/extended test fail. Both reverted.

**Minor — the account-change notification was undocumented.**
`docs/account-authentication.md` gained a subsection: trigger (the account-count
distinguisher that tells a genuine link apart from either signup shape), content shape
(subject/body), and the failure semantics from the finding above.

**Minor — `isKnownConnectedProvider` was dead code.** `grep -rn isKnownConnectedProvider
ui/src ui/test` before removal matched only its own definition/comments and one test file —
zero real consumers; round 3 had already moved the one real call site (the `?linked=` toast
gate) to the narrower `isLinkableProvider`. Removed from `ui/src/account-linking.js`; its
dedicated test in `ui/test/account-settings.test.js` is removed, and the prototype-safety
assertions it carried (`''`, `'__proto__'`, `'toString'`) are migrated onto
`isLinkableProvider`'s own test so that coverage is relocated, not lost. This is also what
makes Round 2's "Tests:" ui count (283, and by extension round 3's 287) one test higher than
current — corrected inline at both locations, with the accurate current count below.

**Minor — Connected-accounts copy promised Google/GitHub linking unconditionally.** Even
when the provider-discovery fetch (the same `GET /api/auth/providers` call gating the Link
buttons themselves) says neither is configured on this deployment. New
`connectableProvidersIntro` (`ui/src/account-linking.js`) derives the sentence from that
same result — both, one, or neither, matching exactly what the buttons below it will show.
Mutation-proven: hard-coding it to always return the "both" sentence fails the "promises
nothing when neither is configured" and "names only the one configured" tests. Reverted.

**Minor — the provider-discovery error alert pointed the wrong direction.** It said the
Link buttons "may be missing below" while rendering physically after (below) the button
block in the DOM — backwards, since the (possibly absent) buttons sit above this text, not
below it. Changed "below" to "above" in `AccountSettings.jsx`; mutation-proven by reverting
the word and confirming the new wiring test fails.

**Minor — one line-pin into the installed better-auth source was wrong.** `auth.js`'s
`hooks.before` comment cited `dispatch.mjs:139` for the `matcher: () => true` line the fix
depends on; the real line is 143 — `:139` is `function getHooks(authContext) {` itself, four
lines above the matcher. Checked every other numbered pin in the same file against
`auth/node_modules/better-auth/dist/` by hand (`account.mjs:117`, `account.mjs:229`,
`callback.mjs:98`, `link-account.mjs:21`, `account.mjs:176`, `session.mjs:359-371`,
`create-context.mjs:148` — all correct) and corrected only the one that was wrong.

**Minor — the PR body's mutation-proof table overstated coverage by one test after the
round-2 revert.** Re-ran (not just re-read) all three mutations in the Summary's
"Mutation-proofs" table against today's tree: `allowUnlinkingAll: false → true` fails the
named test (plus two collateral failures — the config-shape test and the two-linked-accounts
test — which the table never claimed exhaustively, so that's not the defect); `canUnlink()`
hard-coded to `return true` fails exactly the two named tests; the dropped third row's own
claim ("only the config-shape test actually fails" for a `requireLocalEmailVerified: false`
addition) is exactly right, re-verified by hand rather than trusted. The actual one-test
overstatement was Round 2's "Tests:" recap line's ui count, which included a test for
`isKnownConnectedProvider` — removed as dead code in this round (finding above) — corrected
inline at that location.

**Minor — remaining implicit-auto-link-as-delivered text.** Audited `docs/README.md` and
`docs/account-authentication.md` end to end (`grep -rn "implicit\|auto-link\|
trustedProviders" docs/*.md`, plus a repo-wide `grep -rln` for the same terms across all
`*.md` files to rule out a third location): both already correctly describe implicit
auto-link as staying off and explicit link/unlink as what ships — round 3's fix for this
already covers it. No further doc change needed beyond this re-verification.

**Tests:** `cd auth && npm test` → 43/43 (was 41; +2: the unverified-provider-email
behavioral test, and the previously-silent `notifyAccountChange` failure-mode test). `cd ui
&& node --test` → 299/299 (was 287; net +12: −1 for the removed `isKnownConnectedProvider`
test, +13 across `connectedActionErrorState`'s 4 direct unit tests,
`connectableProvidersIntro`'s 4 direct unit tests, and 5 `AccountSettings.jsx` wiring tests
for the shared-handler dedup, the `reauthenticate()`/`?next=` redirect, the
notice-never-fabricated property, the intro-copy gating, and the "above" wording). `cd ui &&
npm run lint` clean; `cd ui && npm run build` succeeds (same pre-existing >500kB chunk-size
warning, unrelated to this round). No backend Python file touched this round, so `pytest -m
"not integration"` was not re-run for this PR's own changes (unrelated to this branch's
diff). Every new/changed guard in this round mutation-proven by hand before this commit
(flip back to the pre-fix behavior or a plausible-looking wrong fix, confirm the new/updated
test fails, revert; no residue left in the working tree): `connectedActionErrorState`,
`connectableProvidersIntro`, the `reauthenticate()`/`?next=` wiring, the "above" wording, the
unverified-provider-email guard, and both `ACCOUNT_CHANGE_NOTIFY_FAILED` log branches.

---


dbrowneup:
## Round 4 — disposition table

All 11 confirmed findings from this pass fixed. Full detail in the PR body's new
**Round 4** section (top-down: Security invariants section rewritten, Round-2 recap
corrected, Round 4 section appended). Head SHA: `9fb14617d579bfbd42fd76a4292c1a3e93545bf8`.

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | Stale-session UX: Link/Unlink render enabled for most of a session's life and 403; "Sign in again" led nowhere useful | Major | Fixed. `connectedActionErrorState` (`ui/src/account-linking.js`) shared by `link()`/`unlinkConnected()`; `reauthenticate()` ends the session then routes to `/sign-in?next=%2Fapp%2Faccount`; message now "For security, you need to sign in again before changing sign-in methods." Server gate untouched. 4 direct unit tests + 2 wiring tests, mutation-proven. |
| 2 | PR body's "no path attaches a provider identity without a live session" was false; "every test drives the handshake" overclaimed | Major | Fixed (prose only). Security invariants section rewritten to name the actual mitigation (state TTL + double-submit cookie + `allowDifferentEmails`); handshake claim scoped to the tests that exercise it. |
| 3 | Provider-emailVerified guard on explicit-link path proven only by source regex | Minor | Fixed. New behavioral test in `auth/test/auth.test.js` (unverified provider email → `unable_to_link_account`, nothing attached). Mutation-proven. |
| 4 | ATO notification: one failure mode logged nothing | Minor | Fixed. Both failure modes now log `ACCOUNT_CHANGE_NOTIFY_FAILED`; still fail-open. `notifyAccountChange` exported for direct testing. 1 new test + 1 extended test, mutation-proven. |
| 5 | Account-change notification undocumented | Minor | Fixed. New subsection in `docs/account-authentication.md` (trigger, content shape, failure semantics). |
| 6 | `isKnownConnectedProvider` dead export | Minor | Fixed. Removed (zero consumers verified via grep); prototype-safety assertions migrated onto `isLinkableProvider`'s test. |
| 7 | Connected-accounts copy promises Google/GitHub unconditionally | Minor | Fixed. New `connectableProvidersIntro`, derived from the same provider-discovery result gating the Link buttons. 4 unit tests + 1 wiring test, mutation-proven. |
| 8 | Provider-discovery error alert says "below," renders below the controls | Minor | Fixed. Changed to "above." Mutation-proven. |
| 9 | Wrong better-auth source line-pin | Minor | Fixed. `dispatch.mjs:139` → `:143` (the actual `matcher: () => true` line); every other numbered pin in the file checked against installed source and confirmed correct. |
| 10 | PR body mutation-proof table overstated coverage by one test | Minor | Fixed. Re-ran (not just re-read) all three mutations to verify the table's claims hold; corrected the actual drift, which was Round 2's "Tests:" ui count including the now-removed `isKnownConnectedProvider` test — annotated inline at that location. |
| 11 | Remaining implicit-auto-link-as-delivered text | Minor | Verified, no change needed. Repo-wide grep across `docs/*.md` confirms Round 3 already corrected both `docs/README.md` and `docs/account-authentication.md`. |

**Tallies:** 11/11 findings fixed (2 major, 9 minor). `cd auth && npm test` → 43/43 (was
41). `cd ui && node --test` → 299/299 (was 287). `cd ui && npm run lint` clean; `cd ui &&
npm run build` succeeds (pre-existing >500kB chunk-size warning, unrelated). No backend
Python file touched this round, so `pytest -m "not integration"` was not re-run. Every
new/changed guard mutation-proven by hand before commit (flip, confirm the test fails,
revert; no residue left in the working tree).

**Deviations:** none from the brief. Stays draft.

Issue #1217, opened 2026-08-03
[cost] Instrument one end-to-end generation — CPU-seconds, wall time, Bedrock tokens, rows written


dbrowneup:
## Summary

**We do not know what a generation costs.** The only measured figure is Bedrock inference — ~$0.0012/generation — and that is the *language model term only*. It excludes the backtest compute, which is almost certainly the dominant cost: walk-forward out-of-sample plus CSCV combinatorial splits, run across every candidate the debate society produces.

The rigor gate is the expensive part of this product, not the LLM. Nobody has measured it.

**Until this number exists, "near-zero marginal cost" is an assumption wearing the clothes of a finding**, and every pricing decision downstream of it is unsupported.

## Why this blocks

The MVP is *"a trader connects a wallet, funds it with testnet USDC, **pays for a generation**, and gets a verdict they can trust."* You cannot set that price without knowing the floor. This is the single measurement the business model rests on.

## What to measure

One complete generation, end to end, recording:

| Metric | Source |
|---|---|
| **ECS CPU-seconds** | task-level CloudWatch `CpuUtilized` × duration, or `docker stats` equivalent locally |
| **Wall-clock time** | request start → verdict persisted |
| **Bedrock input/output tokens** | `response.usage`, per call, summed across the debate society |
| **Aurora rows written** | `strategy_store`, `backtest_results`, `strategy_daily_returns`, `strategy_proposals` |
| **Peak memory** | to size the task correctly |

Break the CPU-seconds down by phase — corpus retrieval / debate / backtest / rigor gate — because the whole point is finding out which phase dominates.

## Acceptance criteria

- [ ] A reproducible measurement script or documented procedure lands in the repo.
- [ ] One full generation is measured and the per-phase breakdown recorded.
- [ ] The result is expressed as **$/generation at current ECS pricing**, with the LLM and compute terms stated separately.
- [ ] The measurement is repeated for a **multi-candidate** run (debate society N>1) so the scaling in N is visible, not assumed.
- [ ] Result recorded in the private docs repo's business-model doc, replacing the current assumption.

## Verify

```bash
grep -rn "BEDROCK_MODEL\|response.usage" backend/archimedes/services/llm_backend.py | head
aws ecs describe-tasks --cluster <cluster> --tasks <task> --query 'tasks[0].{cpu:cpu,memory:memory}'
```

## Anti-goals

- Do not estimate. An estimate stacked on top of an unmeasured base is what this issue exists to replace.
[...]

PR #1437, opened 2026-08-20
Insights admin-gate + dashboard v2

dbrowneup:
## Merge order (read before merging)

1. **After PR #1469 (calm-precision rebrand).** This branch already merged
   `origin/feat/calm-precision-product-rebrand` (commit `237abc7a`) to reconcile the
   insights gate's guards against the rebrand's reformatting of `App.jsx`. That merge
   cannot be stripped without redoing the reconcile, so #1469 must land first — otherwise
   merging this PR drags the rebrand onto `main` under this PR's number.
2. **After PR #1400 (shell wayfinding / #1370).** NAV collision: #1400 deletes the
   `architecture` item from the NAV array while NAV still lives inline in `Layout.jsx`;
   this PR moves NAV out to `ui/src/navConfig.js`. The two edits do not textually
   conflict, so a clean merge would silently resurrect the item #1400 deleted. Pre-empted
   here — `navConfig.js` ships **without** the Architecture entry (commit `f8ca45c2`), so
   either order lands on the same shell nav. #1400 also renames `landing`'s label to
   "Marketing site" in `Layout.jsx`'s `PAGE_LABELS`, which is a normal textual conflict
   git will surface; take #1400's side there.
3. **Ops pre-merge check:** `PLATFORM_ADMIN_WALLETS` must be non-empty in the **running**
   ECS task definition. Per `infra/README.md`, a bare `terraform apply` resets it to `""`
   — and with it empty, `require_platform_admin` 403s *every* wallet, so the owner sees an
   unexplained NotFound on `/app/insights`. That is the gate working as designed, not a
   bug; verify the value before concluding the feature is broken.

## Owner directive — supersedes issue #1028 D8

Per the 2026-08-20 owner directive, **this PR supersedes issue #1028 D8 ("public Insights page")**. `/app/insights` is no longer public app surface — it is **ADMIN-ONLY** (`PLATFORM_ADMIN_WALLETS` holders), and it stays the owner's traction dashboard. The supersession is recorded in `ui/src/components/Insights.jsx`'s header docstring and in `backend/archimedes/api/metrics_private_routes.py`'s module docstring, in addition to here.

## Part 1 — the gate (server truth, not UI theater)

- **New `GET /api/metrics/private/whoami`** on the existing `metrics_private_router` (inherits `require_platform_admin`, same as every other route on that router). Returns `{admin: true, wallet}` on success. There is no `admin: false` 200 — non-admin is always a `401` (no session) or `403` (verified, non-admin wallet).
- **Frontend probes on entry, not on trust.** `App.jsx` calls the shared `fetchAdminProbe()` (`ui/src/adminProbe.js`) as soon as routing lands on `insights`, *before* `AuthenticatedApp`/`Layout` ever mount for that page. A denied or anonymous probe renders **the exact same `NotFound` component** a truly unknown route gets (extracted from App.jsx's own not-found branch into `ui/src/components/NotFound.jsx` so both call sites can never drift apart) — no "you need admin access" message, which would itself advertise the page exists. The pre-existing anonymous→`/sign-in?next=...` redirect is explicitly bypassed for `insights` for the same reason (bouncing to sign-in would confirm the page is real).
- **The Ops nav item renders only after a successful probe.** `Layout.jsx` runs the same shared, TTL-cached probe and filters the `insights` nav entry out of the "Ops" group until it resolves `true`; it defaults **closed** (`isInsightsAdmin` starts `false`), so it never flashes visible-then-hidden. Its *effect* is keyed on `[userId, walletAddr]` (round-2 fix, §4 below) so a wallet swap re-probes — but the **cache itself is not keyed on anything**: `adminProbeCache.js` is a single global `cachedPromise`/`cachedAt` slot. (An earlier revision of this line claimed the cache was "keyed on account id". It never was, and `AuthContext.signOut`'s own comment says so out loud — "the shared probe cache is keyed on nothing session-specific, so it has to be cleared explicitly here". Correctness under identity change comes from the explicit resets in the next bullet, not from a key.)
- **Cache correctness under identity change.** The probe result is shared (one ~30s TTL cache, `ui/src/adminProbeCache.js`) across both mount points so a single `/app/insights` load doesn't double-probe, and is explicitly reset on sign-out (`AuthContext.signOut`) and on wallet change (`AuthenticatedApp`'s existing `wallet-changed` handler) so a stale admin/non-admin determination can never survive into a different session/wallet.
- **Public aggregate endpoints are untouched.** `/api/metrics`, `/api/metrics/funnel`, `/api/metrics/visitors` stay public — only the page and the two new per-entity endpoints (`/whoami`, `/engagement`) are gated. Confirmed by re-running `test_no_public_metrics_route_carries_a_wallet_response_model` and `test_public_metrics_stays_public_and_pii_free` — the **AC2 walk test** (walks `metrics_router`'s routes checking for wallet-bearing response models) — both green, unmodified.

## Part 2 — dashboard v2

**One new admin-gated endpoint**, `GET /api/metrics/private/engagement` (`backend/archimedes/services/engagement_metrics.py`), current-schema only, every query verified against the real ORM models (and covered by DB-seeded tests, not just read):

| Metric | Source | Status |
|---|---|---|
| Accounts total / new 7d / new 30d | `auth_users.created_at` | **Shipped** |
| Linked wallets total | `linked_wallets` row count | **Shipped** |
| Strategies generated total + 7d daily trend (zero-filled) | `strategy_store.created_at` | **Shipped** |
| Generation-costs measured count + input/output/total token sums | `generation_costs.measurement_json["llm"]` | **Shipped** — corrupt/unreadable rows are skipped from the sum, never counted as measured, and flagged via `calls_missing_usage` rather than silently contributing zero (round 4 fix); token totals cover measured, strategy-producing jobs only, not an all-time platform figure — see round 4 §4 |
| Paper deployments active / stopped | `paper_deployments.status` | **Shipped** |
| Repeat-generator proxy (accounts w/ >1 distinct generation day) | `strategy_store.owner_user_id` → `auth_users.id` (real FK, #1028 D1 retrofit — **verified live**, not assumed) | **Shipped**, scoped to non-NULL `owner_user_id`; pre-account (wallet-only, SIWE-era) generations are excluded from both the numerator and denominator rather than silently rounded into either |
| Money paid | none — `PAYMENTS_DRY_RUN` gates every settlement path | **Dry-run marker only** (claims-must-be-true): `{dry_run: true, settled_volume_usd: null, note}` — the real field name settlement wiring will populate, not a placeholder `$0` |

Country breakdown reuses the **existing** `Intl.DisplayNames` pattern in `Insights.jsx` unchanged (`ZZ` → "Unknown / not provided") — nothing to build there, it already does the right thing.

**Map decision: deferred to Phase 2.** Evaluated a world map beside the country bar chart under the zero-new-deps constraint. A faithful world map needs real polygon/TopoJSON geometry data — that is not achievable as a "small hand-bundled inline SVG," and pulling in a mapping library or geometry dataset would be a new dependency this task explicitly ruled out. Kept the existing bar-chart country table as-is; no map shipped this round.

### Phase 2 (deferred — schema-relations work in flight separately)

| Deferred item | Missing relation / reason |
|---|---|
| World map visualization | Not a schema gap — needs real geometry data (TopoJSON/GeoJSON), which fails the zero-new-deps constraint as a "small" asset. Not attempted this round; bar chart stays canonical. |
| Token totals → USD cost | `generation_costs` intentionally carries **measurement only** (`assert_measurement_only` enforces this at write time); there is no pricing-rate table to convert tokens → dollars. Roadmap pricing work (#1217), out of this PR's scope. |
| Live Bedrock/infra spend, cost-per-user, cost-per-generation | Pre-existing `GET /api/metrics/private/cost` — still draft `null` placeholders pending AWS Cost Explorer + Bedrock token-metering wiring. Unrelated to this PR; explicitly still not rendered on the Insights page (see that endpoint's own docstring). |
| Settled payment volume (live) | No durable settlement-ledger table exists — `PAYMENTS_DRY_RUN` gates every path in `services/generation_payment.py` / `marketplace/settlement.py`, so there is nothing for a query to sum. The real slot (`settled_volume_usd`) is wired and will read from that record once it exists. |
| Repeat-usage for pre-account (wallet-only) generators | `strategy_store.owner_wallet` exists, but there is no reconciled identity join that would let a wallet-only history and a later-linked account count as the same person without risking double-counting — that reconciliation is the identity-substrate work tracked separately, not this PR's join. |

## Respect-for-users (public code)

No new per-user drill-down. `/engagement` returns aggregates only — nothing beyond what the existing admin roster endpoints (`/wallets`, `/wallets/connections`) already expose. Naming is neutral (engagement/adoption), not surveillance-flavored.

## Guard verification (mutation-proven, not just written)

Per the repo's "a guard must be shown to reject something" rule — each guard below was actually broken, the test run to confirm it fails, then reverted before pushing:

**Backend — admin-membership check (`require_platform_admin`)**
```diff
-    if wallet not in _platform_admin_wallets():
-        raise HTTPException(status_code=403, detail="Admin access required.")
-    return wallet
+    return wallet  # MUTATED
```
Result: `test_private_cost_403_for_non_admin_wallet`, `test_whoami_403_for_non_admin_wallet`, `test_engagement_403_for_non_admin_wallet` all failed (`assert 200 == 403`). Reverted; all three pass again.

**Backend — full gate removal (router-level `dependencies=` + per-route `Depends`)**
Removing both layers together (the per-route `Depends(require_platform_admin)` param *and* the router-level dependency) produced 7 of 9 `/whoami` + `/engagement` tests failing (`assert 200 == 401`, `assert 200 == 403`, wrong wallet in body). Reverted; clean.

**Frontend — `adminProbe.js`'s network-failure vs. authoritative-denial branch**
```diff
-		throw err;
+		return { admin: false, wallet: null }; // MUTATED
```
Result: the dedicated guard test (`adminProbe.js: a genuine network/parse failure ... is rethrown so adminProbeCache does not cache it`) failed as expected. Reverted; 17/17 admin-probe tests pass.

## Tests

- **Backend**: full six-case pattern for **both** new endpoints (401 no session / 401 bad cookie / 403 non-admin / 200 admin / case-insensitive admin match / shape guard) in `test_metrics_private_routes.py`, plus 11 DB-seeded correctness tests in `test_engagement_metrics.py` (real tmp-sqlite rows against the actual ORM models — accounts windows, wallet counts, strategy trend bucketing, token summing with a corrupt-row case, paper-deployment status split, repeat-generator distinct-day logic, dry-run payments marker). Full suite (`pytest -m "not integration"`, exact CI command, from repo root) confirmed **3601 passed, 2 skipped, 0 failed** immediately before the six-case round-out commit; that commit adds 3 purely-additive test functions (no production code touched) to a file re-verified green in isolation afterward (`test_metrics_private_routes.py` + `test_metrics_routes.py` + `test_engagement_metrics.py` + `test_api_docs_drift.py` together: **48 passed, 0 failed**).
- **UI**: `admin-probe.test.js` (TTL cache behavior, classification logic, wiring checks for App.jsx/Layout.jsx/Insights.jsx/NotFound.jsx) + a regression fix to the pre-existing not-found-title a11y test (its exact-match regex needed updating for the new `deniedInsights` branch — same property, new source shape). Full `node --test` suite: **274 passed, 0 failed** at the time of that round — **494 passed, 0 failed** on the current branch after merging `main` (2026-08-30), plus `npm run check:sitemap` → OK. Backend: `pytest -m "not integration" -q` → **4161 passed, 2 skipped**.
- `npm run lint` (eslint) — clean. `npm run build` (vite) — clean (pre-existing chunk-size warnings only, unrelated to this change).
- `ruff check` / `ruff format --check` on every changed backend file — clean.
- `make docs-check` — 0 broken links, 0 orphaned docs, `docs/api/admin-private.md` updated with `whoami`/`engagement` sections and its stale "no admin UI today" closing note corrected.
- Fixed a docs-drift failure the new routes introduced (`test_every_live_route_is_documented_or_allowlisted`) by documenting both new endpoints in `docs/api/admin-private.md` rather than allowlisting them undocumented.

## Deviations from the literal spec

- The "Ops nav item" in the task description refers to the existing "Insights" entry inside Layout.jsx's pre-existing `"Ops"` NAV group (Insights + Account) — gated the `insights` item specifically; `Account` is untouched and still visible to any signed-in user.
- "Cheap daily-bucket time series from existing timestamps" was scoped to the strategies-generated 7-day trend (the one explicitly requested with "+ 7d trend" in the spec); accounts got 7d/30d point-in-time counts as specified, no separate trend series, to avoid scope creep beyond what was asked.
- Daily bucketing is done in Python over a bounded query window rather than a SQL `GROUP BY date(...)`, so the same code path is portable across the sqlite-in-test / Postgres-in-prod split without leaning on a dialect-specific date-truncation function neither test suite exercises today.

## Round 2 — adversarial review findings fixed (2026-08-20)

An adversarial review against this branch found 11 issues (2 blockers, 6 majors, 3 minors — some of the "major" count folded multiple sub-claims in the same file). All 11 are fixed as of the current tip, each mutation-verified per the repo's "a guard must be shown to reject something" rule: reverted, confirmed the new/updated test fails, restored before pushing.

1. **Blocker — `engagement_metrics.get_strategy_metrics()` counted platform-seeded rows as user-generated strategies.** `main.py` seeds `is_example=True` curated rows into `strategy_store` on every startup; the total/trend queries had no filter at all, so "Strategies generated (total)" and the 7d trend both counted rows no user ever generated. Fixed by adding `.filter(StrategyRecord.is_example.is_(False))` to both queries — the same exclusion `strategies_routes.py`'s library listing already applies to this table. Added `test_strategy_metrics_excludes_platform_seeded_example_rows`; mutation-verified by removing the filter and confirming the test drops from `total == 1` to `total == 3`.
2. **Blocker — a DB failure rendered as a measured-looking `0`, not an honest absence.** Every getter's `except` branch returned `{"total": 0, ...}` on a query failure, indistinguishable on the page from a real zero (`Insights.jsx`'s `Stat` component only renders `None` as `"—"`; a number, including `0`, renders as a number) — directly contradicting the honesty rule this same PR applies to the money-paid tile (`settled_volume_usd` stays `None`, never `0`). Fixed: every `except` branch across all six getters now returns `None` per numeric/list field plus `unavailable: True`. Added `test_db_failure_degrades_to_none_not_zero` (monkeypatches `get_session` to raise, asserts the full `None`-shaped payload from every getter); mutation-verified by reverting `get_account_metrics`'s branch to the old `0`-shape and confirming the assertion fails.
3. **Major — nginx had no `/app/insights` carve-out, so the anonymous-bypass logic in `App.jsx` was dead code at the deployed edge.** The PR body claimed the anonymous→`/sign-in` redirect "is explicitly bypassed for insights," but `nginx/nginx.conf` (untouched by the original PR) had no carve-out for `/app/insights`, so it fell through to the auth_request-gated `^~ /app` block and 302'd every anonymous GET to `/sign-in?next=/app/insights` before a single line of client JS ran. Fixed by adding a dedicated `location ^~ /app/insights` block that serves the SPA shell unauthenticated (same shape as the existing Explore/Leaderboard/Corpus/strategy carve-outs, but for a different reason — documented inline so it isn't mistaken for an anonymous-browse page) — the real gate stays server-side in `whoami`'s 401/403, this only stops nginx from pre-empting it. Added `test_nginx_serves_insights_shell_without_auth_request_gate` (`backend/tests/test_local_setup_contract.py`) asserting the carve-out exists, is declared before the catch-all `^~ /app` block, and does NOT itself `auth_request`-gate; mutation-verified by removing the block and confirming the assertion fails (`ValueError: substring not found`). Also validated `nginx -t` against the real envsubst-template render path (`docker run` with the actual `NGINX_*` vars from `docker-compose.yml`) — syntax OK both before and after.
4. **Major — a wallet swap did not re-trigger either admin probe, so a stale admin determination survived it.** `AuthenticatedApp`'s `wallet-changed` handler correctly calls `_resetAdminProbeCache()`, but nothing consumed that reset: `Layout.jsx`'s probe effect was keyed on `[userId]` only (unchanged by a wallet swap on the same account), and `App.jsx`'s page-gate effect was keyed on `[route.page]` only (unchanged while sitting on `/app/insights`). Since `require_platform_admin` checks `PLATFORM_ADMIN_WALLETS` membership against the connected wallet (`X-Wallet-Address` header) — not just the account — an account with more than one linked wallet, only one of them admin, could switch to the non-admin wallet and keep seeing the Ops nav item and the live dashboard indefinitely. Fixed: `Layout.jsx`'s effect now depends on `[userId, walletAddr]` (already a prop); `App.jsx` has no equivalent state, so it now listens for `wallet-changed` directly and bumps a counter (`walletChangeSeq`) that the page-gate effect depends on alongside `route.page`. Added two regression tests to `admin-probe.test.js` asserting both dependency arrays; mutation-verified by reverting each to its old array in turn and confirming the corresponding test fails.
5. **Major — `measured_count` counted rows whose JSON never decoded.** The count came from a separate `COUNT(*)` taken before the decode loop, so a corrupt `measurement_json` row was simultaneously "skipped from the sum" and "counted as measured" — the docstring's own claim ("skipped, not counted") was false, and a reader computing `total_tokens / measured_count` as an average would silently understate it whenever any row was corrupt. Fixed by incrementing `measured_count` inside the same loop that decodes and sums, only for rows that actually contribute; added a separate honest `rows_total` field for "how many rows exist at all." Updated `test_generation_cost_metrics_sums_tokens_and_skips_corrupt_rows`'s assertion from `measured_count == 3` to `measured_count == 2` (+ `rows_total == 3`); mutation-verified by reverting to the old `COUNT(*)`-up-front shape and confirming the test fails (`assert 3 == 2`).
6. **Major — the docstring claimed no durable settlement record exists; it does.** `settlement_intents` (`models/marketplace.py`) is a real table with a `status` column that reaches `"settled"`/`"failed"` via `marketplace/service.py`'s `_finalize_settlement_intent` — the record exists. The actual, narrower gap: its `amount_usdc` column is declared and never written by any code path (`grep -rn amount_usdc backend/archimedes/` returns only the declaration). Fixed the module docstring and `get_payments_snapshot`'s docstring + its `PAYMENTS_DRY_RUN=false` note to state the true, narrower fact instead.
7. **Major — the repeat-generator proxy's `note` overclaimed exactness.** `upsert_strategy`'s content-hash dedup can attach an account to a row it did not create on the day it was created (a content-identical regeneration returns the existing row with no new `created_at`; an ownerless legacy row backfilled with an owner keeps its original pre-account `created_at`), so `strategy_store.created_at` spans do not exactly equal "days a user generated." Reworded both the function's docstring and the response's `note` field to state the proxy's actual definition rather than the informal framing the field name suggests. No query-logic change — the underlying dedup behavior is shared, wide-blast-radius code this PR doesn't otherwise touch.
8. **Minor — `adminProbe.js`/`adminProbeCache.js`/docs named `Insights.jsx` as a probe mount point; it is `App.jsx`.** `Insights.jsx` never imports the probe — by the time it mounts, `App.jsx` has already resolved `admin === true`. Corrected the comments in both files and the closing note in `docs/api/admin-private.md`.
9. **Minor — accounts' and strategies' "(7d)" tiles used different window definitions.** `get_account_metrics`'s `new_7d` was a plain rolling 168-hour window; `get_strategy_metrics`'s `new_7d` was (and stays) calendar-day-anchored (midnight UTC, `_TREND_DAYS` days ago). `Insights.jsx` renders "New accounts (7d)" and "Generated (7d)" side by side as if they meant the same span. Unified accounts' `new_7d` to the same calendar-day window (`new_30d` is untouched — it has no adjacent "(30d)" metric to disagree with). Added `test_account_new_7d_uses_same_calendar_day_window_as_strategies`, pinning a boundary timestamp that the old rolling window would include and the new calendar window excludes; mutation-verified by reverting to `now - timedelta(days=7)` and confirming the assertion fails (`assert 1 == 0`).
10. **Minor — a naive/aware timezone mismatch between `strategy_store.created_at` and `auth_users.createdAt`.** The former is a bare, timezone-NAIVE column always written via `datetime.now(UTC)` (naive-UTC by convention); the latter is genuinely `DateTime(timezone=True)`. Binding a tz-aware Python value straight against the naive column pushes the aware/naive cast onto the DB session's `TimeZone` setting on Postgres, which could shift both the trend window boundary and day-bucketing on a non-UTC session. Added `_to_naive_utc()` and applied it to the window bind in `get_strategy_metrics` and to every timestamp `_daily_buckets` buckets. Added `test_to_naive_utc_normalizes_aware_and_naive_datetimes`; mutation-verified by making the helper a no-op and confirming the assertion fails.
11. **Minor — the "does not advertise the page exists" claims overstated what the mechanism closes.** `Layout.jsx` probes for every signed-in user on every page (a non-admin's devtools Network tab shows a 403 on `/api/metrics/private/whoami` on first render of any app page), and `/app/insights` plus the `Insights` component ship in the same statically-imported main chunk as every other app page rather than behind a lazy import — so the page's existence is discoverable by an attentive visitor even though the denial screen and the sign-in-bounce vector are genuinely closed. Rather than restructure the probe-timing/lazy-loading (real behavior change, worse trade-off: gating the probe to the insights route alone would break the Ops-nav-discoverability the mechanism exists to provide), added a scoping note to `adminProbe.js` (with short pointers from `NotFound.jsx` and `metrics_private_routes.py`) stating precisely which two vectors are closed and which two are not.

**Backend**: `pytest -m "not integration"` (exact CI command, from repo root): **3609 passed, 2 skipped, 0 failed** (was 3604 immediately before this round's 5 new regression tests: 4 in `test_engagement_metrics.py`, 1 in `test_local_setup_contract.py`). `ruff check` / `ruff format --check` on every changed file: clean. `make docs-check`: 0 broken links, 0 orphaned docs.
**Frontend**: `npm test`: **276 passed, 0 failed** (was 274; +2 new regression tests in `admin-probe.test.js`). `npm run lint`: clean. `npm run build`: clean (pre-existing chunk-size warnings only, unrelated).
**nginx**: config validated with `nginx -t` via the real envsubst-template render path (the actual `NGINX_*` vars from `docker-compose.yml`), both before and after the fix.

## Round 3 — adversarial review findings fixed (2026-08-20)

A second adversarial review found 9 more issues (3 majors, 6 minors — none blockers). All 9 are
fixed as of the current tip, each mutation-verified per the repo's "a guard must be shown to
reject something" rule: reverted (or the historical bug reproduced), confirmed the new/updated
test fails, restored before pushing.

1. **Major — every gate assertion in `admin-probe.test.js` was a regex over source text, none of
   which executed the line that actually consumes a probe result.** `Layout.jsx:229`'s
   `setIsInsightsAdmin(admin)` and `App.jsx:118`'s `setInsightsAdmin(admin)` were unasserted — a
   mutation collapsing either to `setIsInsightsAdmin(true)` left all 17 admin-probe tests green
   while every signed-in non-admin got the Ops nav item and the live dashboard. Fixed by
   extracting the gate's actual decisions into three pure, zero-import functions
   (`ui/src/insightsGate.js`, same "stays unit-testable" discipline `adminProbeCache.js` already
   uses): `resolveInsightsAdminState(probeResult)` (the ONLY place either caller reads `.admin`
   off a probe result), `isInsightsPageBlocked(routePage, insightsAdmin)`, and
   `filterInsightsNavItem(items, isAdmin)`. `App.jsx`/`Layout.jsx` now route every probe-consuming
   decision through these; `admin-probe.test.js` gained 4 new tests calling them with real inputs
   instead of matching source text. Mutation-verified three ways: reverting either call site back
   to `setIsInsightsAdmin(true)` fails the new wiring-regex assertion (which now checks for
   `resolveInsightsAdminState(result)` by name, not just the shape of the `.then` callback);
   reverting `isInsightsPageBlocked` to its old `insightsAdmin === false` form fails its own new
   test (`false !== true`); reverting `filterInsightsNavItem`'s `isAdmin === true` to a bare
   truthy check fails its own test (`['insights','account'] !== ['account']`).
2. **Major — an unresolved probe was itself a disclosure vector the round-2 "does not advertise
   the page exists" claim didn't cover.** While `insightsAdmin` was `null` (probe in flight),
   `App.jsx` titled the tab "Insights · Archimedes" and rendered a chrome-free `Loading…` screen —
   neither of which a genuinely unknown route ever produces on first paint, contradicting the
   in-file claim that "a many-tabs user should not be able to tell 'unknown route' from 'gated
   route' apart." Fixed: `isInsightsPageBlocked` (see #1) treats `null` the same as `false` for
   BOTH the title computation and the render branch, so an unresolved probe now renders `NotFound`
   immediately — the loader state is gone entirely for this route. Mutation-verified by
   temporarily reverting `isInsightsPageBlocked` to the old `=== false` check and confirming its
   own regression test fails (see #1); `a11y.test.js`'s existing not-found-title test was updated
   to match the new computation and still passes.
3. **Minor — the linked-wallet half of the admin gate had zero test coverage on the two new
   routes.** `test_metrics_private_routes.py`'s `client` fixture authenticates through the
   autouse `_legacy_siwe_test_adapter` (`conftest.py`), which monkeypatches
   `wallet_routes.get_linked_wallet_address` to derive the wallet straight from the SIWE cookie —
   so session and linked wallet agree by construction in every existing test, and
   `require_linked_wallet`'s own 403 branch (`docs/api/admin-private.md`'s documented step 2: "a
   session with no linked wallet → 403") was never reached by `/whoami` or `/engagement`. Added
   `test_whoami_403_for_signed_in_account_with_no_linked_wallet` and its `/engagement` twin, each
   re-overriding the shim to return `None` for a valid session and asserting the specific
   `"A verified linked wallet is required"` detail (distinct from the admin-membership 403's
   `"Admin access required."`). Mutation-verified by removing `require_linked_wallet`'s
   `if wallet is None: raise 403` branch and confirming both new tests fail (the response falls
   through to the OTHER 403 message instead: `'Admin access required.' == 'A verified...'`
   assertion error).
4. **Minor — a wallet-header self-lockout had no diagnostics or documented recovery.**
   `get_linked_wallet_address` resolves against `X-Wallet-Address` when the frontend sends one,
   and `ui/src/api.js` always sends it whenever a browser wallet is connected — regardless of
   whether that wallet is linked to the signed-in account. So the admin's own browser, connected
   to an unlinked or non-admin-linked wallet, gets 403'd and — by this PR's own by-design denial
   UX — a bare not-found page with no explanation. Not a code defect (fail-closed is correct
   here); fixed by documenting the mechanism and the recovery path
   (`docs/api/admin-private.md`'s new "Self-lockout note"): disconnect/switch the browser wallet,
   or use `curl -b session.jar .../whoami` to distinguish the two 403 flavors from a real 404.
5. **Minor — `get_generation_cost_metrics`/`get_repeat_generation_metrics` ran unbounded
   full-table scans, materializing every row into a Python list.** Neither had a LIMIT or window,
   unlike `get_strategy_metrics`. A LIMIT would silently under-report an all-time total (worse —
   a claims-must-be-true violation), so instead both queries now stream via SQLAlchemy's
   `yield_per(500)` rather than `.all()`, bounding peak memory without changing what gets counted
   or any query semantics. No behavior change; covered by the existing DB-seeded tests continuing
   to pass against the streamed query.
6. **Major — `measured_count`/`total_tokens` silently discarded the source instrument's own
   incompleteness flags.** `services/cost_meter.py`'s `CostMeter.snapshot()` writes
   `calls_missing_usage`/`usage_complete` into each row's `llm` block specifically so a reader
   can't mistake a partial measurement for a complete one; `get_generation_cost_metrics` read
   only `input_tokens`/`output_tokens`, banking a `usage_complete: false` row into
   `measured_count` exactly like a fully-measured one. Fixed: `measured_count` now only counts
   rows whose usage accounting is complete; new aggregate `calls_missing_usage` (summed) and
   `usage_complete` (`== 0` check) fields are returned and rendered on `Insights.jsx` as a visible
   "⚠ Partial measurement" qualifier under the token tiles whenever `usage_complete` is `false` —
   the same treatment `settled_volume_usd` already gets, per claims-must-be-true. Added
   `test_generation_cost_metrics_partial_usage_row_is_not_banked_as_measured`; mutation-verified
   by reverting the `if usage_complete:` guard to an unconditional `measured_count += 1` and
   confirming the test fails (`assert 2 == 1`).
7. **Major — "Strategies generated" and "Repeat generators" both count distinct stored content,
   not generation events, with no qualifier on the headline tiles.** `upsert_strategy`'s
   content-hash dedup means two different users generating identical output, or one user
   regenerating, produces ONE `strategy_store` row — already disclosed for the repeat-generator
   proxy, but silently also true of `strategies.total`/`new_7d`/the trend, which carried no
   caveat. Fixed: `get_strategy_metrics` now returns an explicit `note` field with the same
   caveat, rendered directly under the tile row on `Insights.jsx` (not buried below the trend
   chart and the entire money-paid block, matching the relocation in #10). `test_db_failure_...`
   updated for the new `note: None` failure-shape field.
8. **Minor — `get_generation_cost_metrics` summed every row unconditionally, latent
   double-counting the moment K>1 returns.** `models/generation_cost.py`'s own docstring states a
   row is keyed to a `(job_id, strategy_id)` pair carrying the JOB's measurement, and "summing
   across rows would double-count" if a job ever persists more than one strategy. Fixed by
   de-duplicating on `job_id` before summing (currently a no-op at K=1, guards the future). Added
   `test_generation_cost_metrics_dedupes_by_job_id_for_k_greater_than_one` (two rows, same
   `job_id`, different `strategy_id`, same measurement); mutation-verified by removing the
   `seen_job_ids` guard and confirming the test fails (`total_input_tokens` reads 1000, not 500).
9. **Minor — round 2's `_to_naive_utc` tz-normalization fix missed one bucketing site.**
   `_daily_buckets` and `get_strategy_metrics`'s window bind were fixed in round 2; the identical
   hazard over the SAME `strategy_store.created_at` column in `get_repeat_generation_metrics`'s
   day-bucketing loop (`created_at.date()`, no normalization) was not. Fixed by routing that line
   through `_to_naive_utc` too. Added
   `test_repeat_generation_metrics_normalizes_tz_aware_timestamps_at_day_boundary` — since SQLite
   (this suite's backend) silently truncates a tz-aware bind at INSERT time rather than
   reproducing the Postgres session-TimeZone cast hazard, the query boundary is faked (a small
   stub session/query pair) to hand the loop two aware timestamps chosen so a raw `.date()` places
   both on the same day (not a repeat) while `_to_naive_utc(...).date()` correctly resolves them
   to two different UTC days (a repeat); mutation-verified by reverting the `_to_naive_utc(...)`
   wrap and confirming `repeat_users` collapses from 1 to 0.
10. **Minor — the "Repeat generators" tile was a bare count with no denominator, sitting next to
    the wrong implicit comparison.** The backend already ships `generating_users` as the intended
    denominator specifically so a bare number wouldn't imply full population coverage, but the
    only place both numbers appeared together was a caption below the trend chart and the entire
    money-paid block — four sections away from the tile itself, which a reader would instead pair
    against "Accounts (total)" two rows up. Fixed: the tile now renders as a ratio
    (`repeat_users / generating_users`, labelled "Repeat generators (proxy)"), and its
    explanatory `note` — plus the new strategies-dedup note from #7 — moved to sit directly
    beneath the tile row instead of below the payments block.

**Backend**: `pytest -m "not integration"` (exact CI command, from repo root):
**3614 passed, 2 skipped, 0 failed** (was 3609 immediately before this round's 5 new regression
tests: 3 in `test_engagement_metrics.py`, 2 in `test_metrics_private_routes.py`). `ruff check` /
`ruff format --check` on every changed file: clean. `make docs-check`: 0 broken links, 0
orphaned docs.
**Frontend**: `npm test`: **280 passed, 0 failed** (was 276; +4 new tests in
`admin-probe.test.js`, exercising the extracted `insightsGate.js` functions with real inputs).
`npm run lint`: clean. `npm run build`: clean (pre-existing chunk-size warnings only, unrelated).

## Round 4 — adversarial review findings fixed (2026-08-20)

A third adversarial review confirmed 11 findings (5 majors, 3 minors, plus a verification
check spanning 2 previously-claimed round-3 fixes — the task's own tally). All are addressed
as of the current tip, mutation-verified per the repo's "a guard must be shown to reject
something" rule: reverted (or the historical bug reproduced), confirmed the new/updated test
fails, restored before pushing. The verification check (item 8's second half) found both
prior fixes already correctly landed — no code change was needed there, only re-confirmation
against the current tip.

1. **Major — `ui/public/sitemap.xml` still listed `/insights`, undermining the PR's own
   "page does not exist for a non-admin" treatment by advertising the URL to every crawler
   and anyone who reads the sitemap.** Removed the entry; added an explanatory exclusion note
   to the sitemap's own header comment. `ui/scripts/check-sitemap.mjs` (a general "no gated
   `/app` page in the sitemap" checker) does not exist on this branch — it landed on `main`
   after this PR's branch point, and this fix round does not rebase 100+ commits of unrelated
   `main` history onto a PR still mid-review (see Deviations). Added `ui/test/sitemap.test.js`
   instead: a PR-scoped equivalent asserting no `<loc>` contains "insights", plus an
   anti-vacuity assertion (the sitemap still lists real public URLs, so the first assertion
   can't be vacuously true against an accidentally emptied file). Mutation-verified: re-adding
   the `/insights` `<loc>` fails the first test; emptying the sitemap's URLs fails the second.
2. **Major — the linked-wallet half of the admin gate had zero behavioral coverage.** Every
   200/403 case in `test_metrics_private_routes.py` ran behind the autouse
   `_legacy_siwe_test_adapter` (`conftest.py`), which monkeypatches
   `wallet_routes.get_linked_wallet_address` to derive the wallet straight from the SIWE
   cookie — so the REAL production resolver (a DB-backed `LinkedWallet` lookup keyed on
   `user_id` + `chain_id`, with an `X-Wallet-Address` header override and an `is_primary`
   fallback) never actually ran in any test on this router. Added four tests (two per
   endpoint) that restore the real `get_linked_wallet_address` and seed a real `AuthUser` +
   `LinkedWallet` row into a tmp-sqlite DB (`redirect_to_tmp_sqlite` precedent, same as
   `test_engagement_metrics.py`): a session identity distinct from both the admin and
   non-admin wallets, whose LINKED wallet (not its SIWE-cookie wallet) determines the
   admin/non-admin outcome. Mutation-verified: monkeypatching the resolver to always return
   `None` fails all four; monkeypatching it to always return the admin wallet (bypassing the
   DB lookup) fails the two non-admin-linked tests with `200 != 403`.
3. **Major — `calls_missing_usage_total` undercounted the most-incomplete rows as
   zero.** A row whose `measurement_json` failed to decode, or that decoded with no `llm`
   block at all, contributed NOTHING to `calls_missing_usage_total` (a bare `continue`) —
   so `usage_complete` could read `true` ("a complete accounting") while an entire row's
   usage was silently unaccounted for, the single MOST incomplete case reported as the
   least. Fixed: such a row now adds `+1` (the conservative "at least one call's usage is
   unaccounted for" floor), de-duplicated by `job_id` like every other branch (moved the
   `seen_job_ids` check to run before JSON decoding). Updated
   `test_generation_cost_metrics_sums_tokens_and_skips_corrupt_rows`'s assertions
   (`calls_missing_usage`/`usage_complete` were `0`/`True`, now `1`/`False`); added
   `test_generation_cost_metrics_corrupt_and_missing_llm_rows_count_as_missing_usage` (two
   distinct broken jobs, proving each contributes its own `+1`) and
   `test_generation_cost_metrics_dedupes_missing_usage_increment_by_job_id` (a K>1 job whose
   every row is corrupt still counts once). Updated the function's docstring and
   `docs/api/admin-private.md` to the true accumulator semantics. Mutation-verified: reverting
   either `+1` back to a bare `continue` fails the new tests; reverting the `seen_job_ids`
   check's new position fails the dedup test.
4. **Major — "Total LLM tokens" was framed as an all-time total.** `generation_costs` rows
   exist only for jobs that persisted ≥1 strategy row
   (`agents/generation_pipeline.py`'s `_persist_generation_cost`: "No strategy row ⇒ nothing
   to key a durable record to, so nothing is written") — a generation that consumed tokens
   but errored, was cancelled, or failed the rigor gate before producing a strategy leaves NO
   row at all, not a zero one. Relabelled the `Insights.jsx` tile "LLM tokens (measured
   jobs)"; added a `note` field to `get_generation_cost_metrics()`'s response stating the
   coverage gap explicitly, rendered directly under the tile row (not buried) and reflected in
   the failure-shape dict too. Added
   `test_generation_cost_metrics_sums_tokens_and_skips_corrupt_rows`'s neighbors above plus an
   `admin-probe.test.js` assertion that the old "Total LLM tokens" string is gone and the new
   label + `note` wiring are present. Updated `docs/api/admin-private.md`.
5. **Major — `docs/api/admin-private.md` was stale in four ways.** (a) Documented the
   pre-round-2-fix "zero shape" degraded response for `/engagement`, contradicting
   `services/engagement_metrics.py`'s own module docstring (the fail-soft shape has been
   `None`-per-field + `unavailable: true` since round 2). (b) Re-asserted the settled-volume
   "no durable record" claim round 2 already corrected in the code — the doc had never been
   updated to match: `settlement_intents` IS a real, durable table whose `status` column
   reaches `settled`/`failed`; the actual gap is narrower (`amount_usdc` is declared and never
   written). (c) Described `calls_missing_usage` as "summed across every row," which was never
   accurate once the dedup-by-`job_id` and corrupt-row `continue` guards are accounted for —
   now rewritten to the true two-part accumulator semantics from fix #3 above. (d) All three
   contradicted the module docstring's fail-soft paragraph. Rewrote the `/engagement` section
   end to end to the exact shipped contract (including the new `note` fields and `|null`
   response-shape annotations), each correction explicitly labelled "round 4 correction" so
   the history of what changed stays legible. `make docs-check`: 0 broken links, 0 orphaned
   docs; `test_api_docs_drift.py`: still green.
6. **Minor — the `insightsGate.js` nav-filter test only exercised a hand-built item
   array.** `admin-probe.test.js`'s `filterInsightsNavItem` test built its own
   `[{id:"insights"},{id:"account"}]` array rather than importing Layout's real "Ops" group,
   so a future rename/reorder of that group could silently stop being covered while the test
   kept passing. Extracted `NAV` out of `Layout.jsx` into a new plain-JS `ui/src/navConfig.js`
   (zero imports, matching `adminProbeCache.js`/`insightsGate.js`'s "stays testable under bare
   `node --test`" discipline); `Layout.jsx` now imports it. Added a test that runs
   `filterInsightsNavItem` against the REAL `NAV` array's "Ops" group, plus a loop asserting
   every OTHER group is untouched by the insights gate. Mutation-verified: removing the gate
   from `filterInsightsNavItem` fails both the old hand-built test and the new real-data test.
7. **Minor — "Real users (accounts)" could fabricate a `0` on a DB failure.**
   `services/user_stats.py`'s `get_distinct_user_count()` has always failed soft to `0` on a
   query error — correct for `/health`/`/private/cost`, which treat it as operational
   telemetry, but dishonest for `GET /api/metrics`'s `real_users`, which `Insights.jsx` renders
   as a measured fact ("Real users (accounts)") right next to the engagement section's
   "Accounts (total)," which has rendered an honest `None`→"—" since round 2. A DB outage
   during `/api/metrics` therefore looked exactly like a genuinely empty user table. Added
   `get_distinct_user_count_or_none()` (the honest variant — same TTL cache, `None` instead of
   `0` on failure); `metrics_routes.get_metrics()` now uses it, and `MetricsResponse.real_users`
   is now `int | None`. The legacy `get_distinct_user_count()` is unchanged for its own
   callers. No `Insights.jsx` change was needed — its `Stat` component already renders `None`
   as "—". Added direct unit coverage in `test_canonical_user_ownership.py` and an HTTP-layer
   test in `test_metrics_routes.py`; updated `docs/api/leaderboard-and-metrics.md`.
   Mutation-verified: reverting either call site back to the 0-fallback fails the new tests.
8. **Minor — `get_repeat_generation_metrics` lacked the `is_example` filter its sibling
   `get_strategy_metrics` applies.** A platform-seeded curated row (`main.py` seeds these on
   every startup) with a non-NULL `owner_user_id` could inflate that account's
   distinct-generation-day count on the "Repeat generators" tile even though the exact same
   row is excluded from "Strategies generated," leaving two adjacent tiles reading different
   populations. Added the identical `StrategyRecord.is_example.is_(False)` filter. Added
   `test_repeat_generation_metrics_excludes_platform_seeded_example_rows`; mutation-verified by
   removing the filter and confirming `generating_users`/`repeat_users` inflate from `0`/`0`
   to `2`/`1`.
   **Also verified (no fix needed):** the second calendar-bucketing site's `_to_naive_utc`
   normalization (`get_repeat_generation_metrics`'s day-bucketing loop) and the
   repeat-generators tile's proxy marker/denominator (rendered as `repeat_users /
   generating_users` with the `note` directly under the tile) were both already landed from
   round 3 — re-read against the current tip and confirmed correct, no changes made.

**Backend**: `pytest -m "not integration"` (exact CI command, from repo root):
**3623 passed, 2 skipped, 0 failed** (was 3614; +9 new tests: 3 in `test_engagement_metrics.py`,
4 in `test_metrics_private_routes.py`, 1 in `test_metrics_routes.py`, 1 in
`test_canonical_user_ownership.py`). `ruff check --select E9,F63,F7,F40,F82 .` (the blocking
`ruff-gate` subset) + `ruff format --check .`: clean, whole repo. `make docs-check`: 0 broken
links, 0 orphaned docs, 141/141 indexed.
**Frontend**: `node --test` (from `ui/`): **284 passed, 0 failed** (was 280; +4 new: 2 in
`ui/test/sitemap.test.js`, 2 in `ui/test/admin-probe.test.js`). `eslint` on every touched
file: clean.

## Deviations from the literal spec (round 4)

- ~~**`ui/scripts/check-sitemap.mjs` was not pulled in via a rebase.**~~ **RESOLVED
  2026-08-30.** `main` has since been merged in, so `check-sitemap.mjs` is present on this
  branch and passes (`npm run check:sitemap` → OK). The two files were deliberately kept
  separate rather than folded together, because they enforce opposite directions and
  neither implies the other: the script asserts *every route `routes.js` marks public is
  IN the sitemap*; `ui/test/sitemap.test.js` asserts *the admin-gated page is NOT in it,
  anywhere in the served bytes*. Both headers now say so.
- **`get_distinct_user_count()`'s legacy fail-soft-to-`0` shape was intentionally left
  unchanged** for `/health` and `/api/metrics/private/cost` — only the DISPLAYED public
  `/api/metrics` value (`real_users`, rendered on `Insights.jsx`) was given the honest-null
  treatment. Widening the null-safe variant to `/health` and `/private/cost` as well is a
  larger, separately-reviewable behavior change (both have their own callers/expectations)
  and was outside this finding's stated scope ("Insights.jsx").

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HCr9L8iz189eQ6v1QTPz2M


dbrowneup:
**Round 4 finding 1 (major) — DISPOSITION: FIXED.**

`ui/public/sitemap.xml` still listed `/insights`, defeating the page's own "does not exist for non-admin" treatment by advertising the URL. Removed the entry, added an exclusion note to the sitemap's own header comment, and added `ui/test/sitemap.test.js` (no `<loc>` may contain "insights", plus an anti-vacuity check that real URLs still exist).

`ui/scripts/check-sitemap.mjs` does not exist on this branch — it landed on `main` (114 commits ahead) after this PR branched. Rebasing that much unrelated history onto a PR mid-review was judged out of proportion to this fix round; shipped the narrower PR-scoped guard instead (see PR body Deviations).

Mutation-verified: re-adding the `/insights` `<loc>` fails the guard test; emptying the sitemap's URL list fails the anti-vacuity test. Both reverted before push.

dbrowneup:
**Round 4 finding 2 (major) — DISPOSITION: FIXED.**

Every 200/403 case in `test_metrics_private_routes.py` ran behind the autouse `_legacy_siwe_test_adapter` (`conftest.py`), which monkeypatches `wallet_routes.get_linked_wallet_address` to derive the wallet straight from the SIWE cookie — the real production resolver (DB-backed `LinkedWallet` lookup keyed on `user_id`+`chain_id`, `X-Wallet-Address` header override, `is_primary` fallback) had zero behavioral coverage.

Added 4 tests (2 for `/whoami`, 2 for `/engagement`) that restore the real `get_linked_wallet_address` and seed a real `AuthUser` + `LinkedWallet` row into a tmp-sqlite DB (`redirect_to_tmp_sqlite` precedent). The SIWE-cookie session identity is a THIRD wallet, distinct from both admin/non-admin wallets, so a pass can only be explained by the DB lookup actually running: admin-wallet-linked → 200; non-admin-wallet-linked → 403 "Admin access required." (not the linked-wallet-missing 403, proving the resolver found a real row).

Mutation-verified: monkeypatching the resolver to always return `None` fails all 4 new tests; monkeypatching it to always return the admin wallet (bypassing the DB lookup) fails the two non-admin-linked tests with `200 != 403`. Reverted before push.

dbrowneup:
**Round 4 finding 3 (major) — DISPOSITION: FIXED.**

Rows skipped for corrupt JSON or a missing `llm` block contributed 0 to `calls_missing_usage_total`, so `usage_complete` could read `true` ("a complete accounting") while an entire row's usage was silently unaccounted for — the most incomplete case, reported as the least.

Fixed in `get_generation_cost_metrics`: such a row now adds `+1` (conservative "at least one call's usage is unaccounted for" floor), de-duplicated by `job_id` (moved the `seen_job_ids` check ahead of JSON decoding so this applies uniformly). Updated the function's docstring and `docs/api/admin-private.md` to the true accumulator semantics: `calls_missing_usage` = each qualifying row's self-reported value + `+1` per unreadable row.

Updated `test_generation_cost_metrics_sums_tokens_and_skips_corrupt_rows`'s assertions (was `calls_missing_usage==0, usage_complete==True`, now `1`/`False`); added `test_generation_cost_metrics_corrupt_and_missing_llm_rows_count_as_missing_usage` (two distinct broken jobs, each contributing its own `+1`) and `test_generation_cost_metrics_dedupes_missing_usage_increment_by_job_id` (a K>1 job whose every row is corrupt still counts once).

Mutation-verified: reverting either `+1` back to a bare `continue` fails the new/updated tests; reverting the `seen_job_ids` check's new position fails the dedup test. Reverted before push.

dbrowneup:
**Round 4 finding 4 (major) — DISPOSITION: FIXED.**

"Total LLM tokens" was framed as an all-time total, but `generation_costs` rows only exist for jobs that persisted ≥1 strategy row (`agents/generation_pipeline.py`'s `_persist_generation_cost`: "No strategy row ⇒ nothing to key a durable record to, so nothing is written") — a generation that consumed tokens but errored, was cancelled, or failed the rigor gate first leaves NO row, not a zero one.

Relabelled the `Insights.jsx` tile "LLM tokens (measured jobs)"; added an explicit `note` field to `get_generation_cost_metrics()`'s response (and its failure shape) stating the coverage gap, rendered directly under the tile row in the UI rather than buried. Updated `docs/api/admin-private.md`.

Verified via `admin-probe.test.js`: asserts the old "Total LLM tokens" string is gone and the new label + `note` wiring are present.

dbrowneup:
**Round 4 finding 5 (major) — DISPOSITION: FIXED.**

`docs/api/admin-private.md` was stale in the four ways cited: (a) documented the pre-round-2-fix "zero shape" degraded response, contradicting `services/engagement_metrics.py`'s own fail-soft module docstring; (b) re-asserted the settled-volume "no durable record" claim round 2 had already corrected in code without the doc ever catching up — the narrower truth is that `settlement_intents` IS a real, durable table whose `status` column reaches `settled`/`failed`, but its `amount_usdc` column is declared and never written; (c) described `calls_missing_usage` as "summed across every row," never accurate once dedup-by-`job_id` and the corrupt-row guards are accounted for; (d) all three contradicted the module docstring's fail-soft paragraph.

Rewrote the `/engagement` section end to end to the exact shipped contract, matching finding 3's corrected accumulator semantics and finding 4's coverage-caveat addition, with `|null` response-shape annotations and each correction explicitly labelled "round 4 correction" so the edit history stays legible.

`make docs-check`: 0 broken links, 0 orphaned docs, 141/141 indexed. `test_api_docs_drift.py`: still green.

dbrowneup:
**Round 4 finding 6 (minor) — DISPOSITION: FIXED.**

`insightsGate.js`'s `filterInsightsNavItem` test only exercised a hand-built `[{id:"insights"},{id:"account"}]` array, not Layout's real "Ops" group — a future rename/reorder of that group could silently stop being covered while the test kept passing.

Extracted `NAV` out of `Layout.jsx` into a new plain-JS `ui/src/navConfig.js` (zero imports, matching `adminProbeCache.js`/`insightsGate.js`'s "stays unit-testable under bare `node --test`" discipline); `Layout.jsx` now imports it. Added a test that runs `filterInsightsNavItem` against the REAL `NAV` array's "Ops" group, plus a loop asserting every other group is untouched by the insights gate.

Mutation-verified: removing the gate from `filterInsightsNavItem` fails both the pre-existing hand-built test and the new real-data test.

dbrowneup:
**Round 4 finding 7 (minor) — DISPOSITION: FIXED.**

`Insights.jsx`'s "Real users (accounts)" fabricated a `0` on a DB failure, unlike the adjacent "Accounts (total)" tile's honest `None`→"—" (since round 2). Root cause was server-side: `services/user_stats.py`'s `get_distinct_user_count()` has always failed soft to `0` on a query error — correct for `/health`/`/private/cost` (operational telemetry), dishonest for `GET /api/metrics`'s `real_users`, which is *displayed* on Insights.jsx as a measured fact. A DB outage during `/api/metrics` therefore rendered as a plausible "0 real users," indistinguishable from a genuinely empty user table.

Added `get_distinct_user_count_or_none()` (same TTL cache, `None` instead of `0` on failure). `metrics_routes.get_metrics()` now uses it; `MetricsResponse.real_users` is now `int | None`. The legacy `get_distinct_user_count()` is unchanged for its own callers (`/health`, `/private/cost`) — narrower scope per the finding's own "Insights.jsx" framing (see PR body Deviations). No `Insights.jsx` change was needed: its `Stat` component already renders `None` as "—".

Added direct unit coverage in `test_canonical_user_ownership.py` and an HTTP-layer test in `test_metrics_routes.py`; updated `docs/api/leaderboard-and-metrics.md`. Mutation-verified: reverting either call site back to the 0-fallback fails the new tests.

dbrowneup:
**Round 4 finding 8 (minor) — DISPOSITION: FIXED + VERIFIED.**

`get_repeat_generation_metrics` lacked the `is_example` filter its sibling `get_strategy_metrics` applies — a platform-seeded curated row with a non-NULL `owner_user_id` could inflate that account's distinct-generation-day count on the "Repeat generators" tile even though the exact same row is excluded from "Strategies generated." Added the identical `StrategyRecord.is_example.is_(False)` filter. Added `test_repeat_generation_metrics_excludes_platform_seeded_example_rows`; mutation-verified by removing the filter and confirming `generating_users`/`repeat_users` inflate from `0`/`0` to `2`/`1`. Reverted before push.

**Verification sub-item:** re-checked both previously-claimed round-3 fixes against the current tip — both already correct, no changes made:
- The second calendar-bucketing site's `_to_naive_utc` normalization: `get_repeat_generation_metrics`'s day-bucketing loop already routes every `created_at` through `_to_naive_utc(created_at).date()`.
- The repeat-generators tile's proxy marker/denominator: `Insights.jsx` already renders it as `repeat_users / generating_users` labelled "Repeat generators (proxy)", with the explanatory `note` directly under the tile row.

dbrowneup:
Top-level review of the round-4 fixes: all 8 dispositions verified, with **one correction applied on top** (latest commit): the round-4 sitemap fix removed the `<url>` entry but left an XML comment in the *publicly served* sitemap.xml naming the admin-only page, its path, and its gate mechanism — and scoped the guard test to `<loc>` values to accommodate that comment, which inverted the priority (the served file is the disclosure surface, comments included; a human reading /sitemap.xml is exactly the audience a crawler-only check ignores). The rationale now lives in the test file (never served), the sitemap says nothing about the page, and the guard covers the whole file — mutation-checked: an injected `<!-- insights -->` comment fails the test, the clean file passes 3/3.

With that closed, this PR has my review pass. It stays draft pending the owner's look — it is his dashboard — and the branch will need its 100+-commit update at merge time (the watcher handles that; main's check-sitemap.mjs and api-surface gate will run against it then).

onder-akkaya:
**Heads up — this PR and 1400 collide semantically, and neither CI run can see it.** `git merge-tree` is clean in both directions, both PRs are green, and whichever lands second turns `main` red.

The collision:

- **#1400** adds `backend/tests/test_breadcrumbs.py`, which does `re.search(r"const\s+NAV\s*=\s*\[(.*?)\n\];", layout_text)` against `ui/src/components/Layout.jsx` and then `assert m, "NAV block not found in Layout.jsx"`.
- **#1437** deletes `const NAV = [` from `Layout.jsx` and recreates it in a new `ui/src/navConfig.js`.

So `test_shell_nav_items_stay_inside_the_shell` fails the moment both are on `main`.

There is a second, quieter half. `navConfig.js` reinstates `{ id: "architecture", label: "Architecture", icon: "i-lucide-network" }` — the exact nav item #1400 removes for #1370 item 4, because it routed to `PublicLayout` and unmounted the shell. If #1437 lands second it silently restores the bug #1400 fixed, and no test catches that either.

Both PRs are 88 commits behind `main` (opened 2026-08-20), so their green checks describe a base from nine days ago regardless. `CLAUDE.md` § *the green check may not mean what you think*, rules 1 and 2, cover both halves of this.

Suggested order: **land #1400 first, then rebase #1437 onto it.** That way #1437 has to reconcile `navConfig.js` against the removed Architecture item and against the NAV-location test, rather than reverting both by accident. Whichever way it goes, re-run `pytest -m "not integration"` from the repo root on the *merged* result before the second merge — not on either branch.

Separately, coupling a Python test to the declaration site of a `const` in one JSX file is brittle no matter which PR wins. Asserting on the exported nav list (or on the rendered nav) would survive a refactor that #1437 shows is already wanted.

onder-akkaya:
Traced the gate end to end and it holds up. `require_platform_admin` (`metrics_private_routes.py:54`) depends on `require_linked_wallet` (`wallet_routes.py:525`) → `get_current_user(request)` on the signed session → a DB `LinkedWallet` lookup keyed on `user_id` + `chain_id`. Nothing client-controllable gets in: the `X-Wallet-Address` header only selects among wallets already verified and linked to that session's account (`wallet_routes.py:509-522`, `filter_by(user_id=user.id)`), so it can't introduce an unlinked address.

Fail direction is deny in every branch I could find. `if wallet not in _platform_admin_wallets(): raise HTTPException(403)` — unset env gives an empty set and everyone gets 403, which is the right way round. Frontend matches: `fetchAdminProbe()` ends `.catch(() => ({ admin: false, wallet: null }))`, and `isInsightsPageBlocked` blocks on `insightsAdmin !== true`, so the in-flight `null` blocks too rather than flashing the page.

Router-level `dependencies=[Depends(require_platform_admin)]` at `:70` covers all four routes, and I checked for an ungated sibling returning the same data — there isn't one. `metrics_router`'s public `/api/metrics`, `/metrics/funnel`, `/metrics/visitors` are the pre-existing aggregates, not the new per-entity payloads.

The tests are real guards, not admin-can-enter tests. I ran the mutations on a main-merged worktree: deleting the 403 body gives 5 failed / 19 passed; `resolveInsightsAdminState → return true` fails 1; `isInsightsPageBlocked → return false` fails 1. `test_whoami_403_via_the_real_wallet_resolver_with_a_non_admin_linked_wallet` is the good one — it restores the production resolver against a seeded sqlite `LinkedWallet` and asserts the *admin*-403 detail rather than the missing-wallet one, so it proves a genuine non-admin refusal.

**One thing I'd want changed: `nginx/nginx.conf:291`.**

The new `location ^~ /app/insights` bypasses the `auth_request /_auth_session` gate that `^~ /app` applies. The effect:

```
anonymous GET /app/insights  -> 200 HTML
anonymous GET /app/library   -> 302 /sign-in
```

That difference is an existence oracle at the layer before any JS runs. Only the shell is served so no data leaks, but `adminProbe.js`'s "Scope of does-not-advertise" note doesn't list this vector, and right now the note claims more than the config delivers. Either restore the auth boundary for that path or amend the note to say the page's *existence* is observable pre-auth.

Same line, smaller: `^~ /app/insights` is a prefix match, so it also catches `/app/insightsfoo` and drops those off the auth boundary too. `location = /app/insights` (or `^~ /app/insights/`) is tighter.

Minor and pre-existing: the Insights page still renders the public `/api/metrics|funnel|visitors` data, so "admin-only page" isn't "admin-only data". You've documented that, just noting it survives this PR.

67 passed on the merged tree.

dbrowneup:
@onder-akkaya — your existence-oracle concern is closed: nginx now uses an exact-match location with auth_request for /app/insights, so the prefix-match oracle is gone. Ops gate verified live: the running task definition carries a non-empty PLATFORM_ADMIN_WALLETS.

PR #1438, opened 2026-08-20
[backend] Schema relations Phase 1: identity/ownership indices + gated FKs

dbrowneup:
## Summary

Phase 1 of the schema-relations design in `docs/database-relations.md`: additive-only
indices + gated foreign keys closing the identity/ownership gaps an audit surfaced, with
three corrections and one gap found while verifying the audit's own claims against
the running code before building anything. (Revised after an adversarial review round —
see "Review round 2" below for the full disposition.)

- **10 indices** on FK columns Postgres never auto-indexes, and on composite indices that
  led with the wrong column for the query shape that actually runs (a money table,
  `subscriber_liabilities`, had **zero** indices before this PR).
- **1 column widen** — `paper_deployments.strategy_id` VARCHAR(64) → VARCHAR(128),
  matching `strategy_store.id`'s width ahead of the FK below. The downgrade direction is
  documented as NOT unconditionally safe (see Deviations).
- **4 foreign keys**, each created `NOT VALID`:
  `linked_wallets.address → wallet_identities.wallet_address`,
  `paper_deployments.owner_user_id → auth_users.id` (`ON DELETE SET NULL`),
  `paper_deployments.strategy_id → strategy_store.id`,
  `paper_deployments.owner_wallet → wallet_identities.wallet_address`.
  Validation (`VALIDATE CONSTRAINT`) is a **separate follow-up revision**
  (`9c2e7b5a1f4d`), gated on a live orphan count either way — see "Why NOT VALID +
  a separate validate revision" below.
- **1 explicit skip, stated in the migration docstring**: no FK from
  `backtest_results.strategy_id` to `strategy_store.id` — see Deviations below (C3).

No writer changed. `docs/database-relations.md` carries the full audit summary, the
target ERD (Mermaid), and the Phase 2 proposal (not built — write-path changes only).
**This doc and this PR are both unmerged/draft** — nothing described here is live.

## Corrections to the original audit (verified before building anything)

- **C1** — "row counts <5k everywhere" was unsourced. The only in-repo **production**
  measurement is `backend/migrations/versions/f0ab58339d55_...py`'s docstring: 646 rows
  over 51 strategies. Every other table's count is unverified — and separately, a
  2026-07-05 data-architecture investigation found prod Aurora already contains
  unjoinable/orphaned populations on identity/ownership columns, so "unverified" here
  means "known to include nonzero cases elsewhere," not "probably zero." This is why
  every new FK is added `NOT VALID` (see below) rather than trusting a static
  pre-migration checklist.
- **C2** — Alembic is not "the" schema path; `backend/migrations/env.py` documents the
  two-path design (Alembic owns Postgres/prod, `create_all()` owns SQLite/tests). The
  existing parity gate (`test_alembic_migrations.py`) compared **columns only**, never
  constraints/indices — this PR extends the parity tests for both tables it touches
  (`linked_wallets`, `paper_deployments`) to also compare **index names/columns and FK
  targets** (`_index_signature` / `_fk_signature`), not just column names.
- **C3** — `backtest_results.strategy_id → strategy_store.id` is "plausible, not
  provable": both writers that establish id-space equality (`main.py`'s startup seed,
  `strategy_provider.py`'s unified-table sync) are explicitly best-effort / non-fatal on
  failure. **This PR does not add that FK.**
- **C4** (new, added during review — originally filed as gap "G2") — re-verified against
  the code: `paper_deployments` always carries `owner_user_id` from an authenticated
  session at INSERT time (`paper_routes.py`'s module docstring: "every deployment is
  owned by `owner_user_id`"; "the table shipped with this feature, so there are no legacy
  wallet-owned rows"). The "a paper deployment made before a wallet link is permanently
  unreachable" scenario the original audit described **cannot occur** — there is no
  pre-account `paper_deployments` row for `claim_legacy_wallet_data` to ever need to
  backfill. Dropped from the gap list; not part of the Phase 2 proposal.

**G1 is closed — on `main`, not by this PR.** The original audit's remaining gap said
generation payments were settled through Circle and then only logged, never persisted.
That was true when the audit ran and is **false now**: #1456 landed `payment_receipts`
on `main` (revision `1752121b8d7c`), backed by `PaymentReceiptRecord`
(`models/payment_receipt.py`) — `user_id`, `payer_wallet`, `amount_base_units`,
`price_usd`, `network`, `settlement_ref`, nullable `job_id`, `created_at`, indexed
`ix_payment_receipts_user_id` — written at the settle site
(`generate_routes.py::_persist_payment_receipt`) and read back by
`GET /api/payments/receipts`. `docs/database-relations.md` was corrected in this branch
rather than merged with the stale claim.

Two things about G1 that are deliberately **not** claimed closed, both verified:

- The receipt write is fail-safe by design (`_persist_payment_receipt` swallows every
  exception, since the payment has already cleared), so a row is best-effort in exactly
  correction C3's sense — "what we recorded", not provably "what we settled".
- **`payment_receipts.user_id` carries no FK to `auth_users.id`.** Same gap class this PR
  closes elsewhere, but the table did not exist when `fb8d0bae8112` was authored, so it is
  out of scope here. Phase 2's G1 line item is therefore no longer "build a table" — it is
  "**extend `payment_receipts`** with that FK", which is what § 4 of
  `docs/database-relations.md` now proposes.

## Why NOT VALID + a separate validate revision

A plain `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY` validates every existing row as
part of adding the constraint, and aborts the whole statement on the first violation. On
a repo where prod Aurora is known to carry orphaned identity/ownership rows (C1), that
turns an additive migration into a deploy-time hard failure. So every FK here is split
across two Alembic revisions:

1. **`fb8d0bae8112`** (this file) — indices, the column widen, and all four FKs added
   **`NOT VALID`** only. No live orphan query runs in this revision's `upgrade()` at all,
   which is also what makes it fully renderable under `alembic upgrade --sql` (see
   rendered SQL below). A bounded `SET lock_timeout = '5s'` runs first so a blocked
   `ACCESS EXCLUSIVE` acquire fails fast and the (single-transaction) migration rolls
   back cleanly, rather than queuing behind live traffic.
2. **`9c2e7b5a1f4d`** (new, added during review) — the only place any of the four
   constraints is passed to `VALIDATE CONSTRAINT`, gated on a live orphan-count query run
   against its own connection immediately before each attempt (same queries as before,
   kept byte-identical to `fb8d0bae8112`'s copy — drift-checked by
   `test_phase1_validate_orphan_sql_matches_source_revision`). An orphan count above zero
   is not a failure: it is the documented, expected outcome for
   `fk_paper_deployments_strategy_id` / `.owner_wallet`, and it leaves that one
   constraint `NOT VALID` (already enforcing all future writes) while the migration still
   completes and prints which constraint(s) it left unvalidated and why — no operator
   ritual required.

**Caveat stated honestly, not assumed away:** `migrations/env.py`'s
`run_migrations_online()` wraps a WHOLE `alembic upgrade` invocation in one transaction
(no `transaction_per_migration=True`), so if both revisions are applied via a single
`alembic upgrade head` call — the normal build-on-deploy case — they still share that
transaction. What the split still buys under that default: the *expected* failure mode
(a known orphan population) is turned into a log line by the live gate, never a raised
exception, so it can't trigger a shared-transaction rollback in the first place. Real
transaction isolation for a prod deploy with unmeasured orphan counts is available today
by running `alembic upgrade fb8d0bae8112` and confirming it before `alembic upgrade
head` — two invocations, two transactions, no code change required. Changing `env.py`'s
transaction model globally is a separate, bigger decision, flagged here for Dan rather
than applied unilaterally in this PR.

## Rendered offline SQL (`alembic upgrade --sql`, 5cb798feef58:9c2e7b5a1f4d)

Both new revisions render end-to-end with no live connection — the regression this PR
fixes (see Deviations #4):

```sql
BEGIN;

-- Running upgrade 5cb798feef58 -> fb8d0bae8112

SET lock_timeout = '5s';

CREATE INDEX ix_linked_wallets_address ON linked_wallets (address);
CREATE INDEX ix_paper_deployments_owner_user_id ON paper_deployments (owner_user_id);
CREATE INDEX ix_identity_events_wallet_time ON identity_events (wallet, occurred_at);
CREATE INDEX ix_subscriber_liabilities_sub ON subscriber_liabilities (sub_id);
CREATE INDEX ix_subscriber_liabilities_strategy_created ON subscriber_liabilities (strategy_id, created_at);
CREATE INDEX ix_settlement_intents_sub ON settlement_intents (sub_id);
CREATE INDEX ix_settlement_intents_status_created ON settlement_intents (status, created_at);
CREATE INDEX ix_marketplace_agents_subscriber_wallet ON marketplace_agents (subscriber_wallet);
CREATE INDEX ix_marketplace_agents_creator_wallet ON marketplace_agents (creator_wallet);
CREATE INDEX ix_generation_costs_recorded_at ON generation_costs (recorded_at);

ALTER TABLE paper_deployments ALTER COLUMN strategy_id TYPE VARCHAR(128);

ALTER TABLE linked_wallets ADD CONSTRAINT fk_linked_wallets_address_wallet_identity
  FOREIGN KEY(address) REFERENCES wallet_identities (wallet_address) NOT VALID;
ALTER TABLE paper_deployments ADD CONSTRAINT fk_paper_deployments_owner_user_id
  FOREIGN KEY(owner_user_id) REFERENCES auth_users (id) ON DELETE SET NULL NOT VALID;
ALTER TABLE paper_deployments ADD CONSTRAINT fk_paper_deployments_strategy_id
  FOREIGN KEY(strategy_id) REFERENCES strategy_store (id) NOT VALID;
ALTER TABLE paper_deployments ADD CONSTRAINT fk_paper_deployments_owner_wallet
  FOREIGN KEY(owner_wallet) REFERENCES wallet_identities (wallet_address) NOT VALID;

UPDATE alembic_version SET version_num='fb8d0bae8112' WHERE alembic_version.version_num = '5cb798feef58';

-- Running upgrade fb8d0bae8112 -> 9c2e7b5a1f4d

SET lock_timeout = '5s';

ALTER TABLE linked_wallets VALIDATE CONSTRAINT fk_linked_wallets_address_wallet_identity;
ALTER TABLE paper_deployments VALIDATE CONSTRAINT fk_paper_deployments_owner_user_id;
ALTER TABLE paper_deployments VALIDATE CONSTRAINT fk_paper_deployments_strategy_id;
ALTER TABLE paper_deployments VALIDATE CONSTRAINT fk_paper_deployments_owner_wallet;

UPDATE alembic_version SET version_num='9c2e7b5a1f4d' WHERE alembic_version.version_num = 'fb8d0bae8112';

COMMIT;
```

(The `VALIDATE CONSTRAINT` statements above render unconditionally for offline review;
the live orphan gate in `9c2e7b5a1f4d` is what actually decides whether each one runs
against real data — see `context.is_offline_mode()` in that file.)

## Test plan

- [x] `alembic upgrade head` from empty DB (SQLite) — clean, single head (`9c2e7b5a1f4d`).
- [x] Idempotent up/up, down/up/up — verified for the whole chain and for both new
      revisions specifically.
- [x] **Realistic-orphan fixture** (`_seed_phase1_fixture`): one wallet/strategy pair that
      satisfies every FK, one `paper_deployments` row whose `strategy_id` and
      `owner_wallet` point at nothing. `alembic upgrade head` succeeds, the orphan row is
      byte-for-byte untouched, every constraint still exists.
- [x] Raw orphan-count SQL (the exact strings both migrations run) verified against that
      fixture: 0/0/1/1 orphans — matches the fixture by construction.
- [x] **Drift guard**: `test_phase1_validate_orphan_sql_matches_source_revision` proves
      `fb8d0bae8112` and `9c2e7b5a1f4d`'s intentionally-duplicated orphan SQL cannot
      silently diverge.
- [x] **Offline rendering**: `alembic upgrade --sql` and `alembic downgrade --sql` both
      render end to end with no live connection (`test_phase1_offline_sql_renders_without_crashing`,
      `test_phase1_offline_sql_downgrade_renders_without_crashing`).
- [x] **Downgrade-narrow guard**: `_fail_if_narrow_would_truncate` raises a named error
      when the strategy_id narrow would truncate data, verified with a fake bind
      reporting a nonzero count (`test_phase1_downgrade_narrow_guard_rejects_rows_that_would_truncate`).
- [x] Model ↔ migration parity for both touched tables (`linked_wallets`,
      `paper_deployments`) — **column names, index names/columns, AND FK targets**, not
      column names alone (see C2 correction above).
- [x] `pytest -m "not integration"` from repo root, **after merging `main` @ `3dd63e5e`**:
      **4129 passed, 2 skipped, 2 deselected, 0 failed**.
- [x] `.github/scripts/migration_chain_guard.py --base-ref origin/main` → OK; `alembic
      heads` → exactly one head (`9c2e7b5a1f4d`).
- [x] `ruff check --select E9,F63,F7,F40,F82 .` (the actual blocking `ruff-gate` scope) +
      `ruff format --check` on every changed file: clean.
- [x] `make docs-check`: 1273 links / 154 files, 0 broken; 150/150 docs indexed, 0
      orphaned.

### Mutation-proof (CLAUDE.md rule 3/4 — a guard must be shown to reject something)

- **Parity test catches a removed FK**: temporarily removed
  `LinkedWallet.address`'s `ForeignKey(...)` in `models/account.py` →
  `test_phase1_linked_wallets_matches_a_fresh_create_all_schema` failed on the new
  `_fk_signature` assertion (column-name assertion still passed, confirming this was
  previously a blind spot) → reverted, test passes again.
- **Drift guard catches a changed orphan query**: temporarily replaced one orphan_sql
  string in `9c2e7b5a1f4d` with `SELECT 0` →
  `test_phase1_validate_orphan_sql_matches_source_revision` failed → reverted.
- **Dialect guard removed** (`if not is_postgres: return` disabled in `9c2e7b5a1f4d`, so
  SQLite falls through to the Postgres-only `VALIDATE CONSTRAINT` statement):
  `test_phase1_sqlite_never_reaches_the_postgres_only_validate_branch` and
  `test_phase1_migration_upgrade_head_survives_realistic_orphans` both failed → reverted.
- **Downgrade-narrow guard**: called with a fake bind reporting 3 offending rows → raises
  `RuntimeError` naming the count; called with 0 → no exception; called for SQLite/offline
  → never queries at all (proven via a bind that raises if queried).

- **Chain guard catches a stale `down_revision`** (added when re-serializing onto
  `main`'s current head): reverted `fb8d0bae8112.down_revision` to `a3f19c7d2e84` →
  `migration_chain_guard.py --base-ref origin/main` exits 1 naming the exact fix, and
  `alembic heads` prints **two** heads (`5cb798feef58`, `9c2e7b5a1f4d`) → restored,
  guard OK, one head.

Full `test_alembic_migrations.py` run after all reverts: **23 passed** (was 20 before
`main` was merged in; the merge added `main`'s three `brief_intent` migration tests —
see "Merged `main`" below).

## Deviations from the design as given

1. **Runtime FK gate instead of a manual pre-flight checklist** — same queries, executed
   by a migration against its actual connection instead of by a human against prod before
   deploying. Every FK is created `NOT VALID` and validated only on a zero orphan count.
2. **No FK added for `backtest_results.strategy_id`** — the design's own C3 correction
   says not to add it blind.
3. **NOT VALID and VALIDATE CONSTRAINT split into two revisions** (added during review,
   see "Why NOT VALID + a separate validate revision" above) — the original draft ran
   both in one migration/transaction; an unexpected validate failure could have rolled
   back the indices/widen/NOT-VALID-constraint creation along with it.
4. **Offline-mode rendering fixed** (added during review) — the original draft's live
   orphan-count query ran unconditionally and crashed under `alembic upgrade --sql`
   (`AttributeError: 'NoneType' object has no attribute 'scalar_one'`, since offline mode
   has no live connection to fetch a real result from). Fixed by removing the live query
   from `fb8d0bae8112` entirely (moved to the validate revision) and guarding
   `9c2e7b5a1f4d`'s orphan check with `context.is_offline_mode()`.
5. **Downgrade of the strategy_id widen is documented as conditionally safe, not
   unconditionally reversible** (added during review) — `_fail_if_narrow_would_truncate`
   checks for rows that would truncate-fail first and raises a named error instead of
   claiming blanket reversibility.
6. **Test file hardening**: the indices up/down test derives its "did every index land"
   check from a hardcoded literal set of the 10 required names, not the migration's own
   `_INDICES` list (so a regression that silently shrinks that list still fails the
   test). The two model↔migration parity tests were extended (added during review) to
   compare index names/columns and FK targets, not just column names.

## Merged `main` (@ `3dd63e5e`) — what changed and why

1. **Re-serialized onto `main`'s current chain head.** `fb8d0bae8112.down_revision` now
   points at **`5cb798feef58`** (was `f0ab58339d55` when authored, then `a3f19c7d2e84`).
   Four revisions have landed on `main` since this branch forked — `1752121b8d7c`
   (payment_receipts), `a3f19c7d2e84` (generation_credits), `5728d9ef1901`
   (debate_transcripts), `5cb798feef58` (strategy_store.brief_intent) — and leaving the
   old pointer forked the chain into two heads, which is the red CI this PR was carrying.
   None of the four touches a column, index or constraint named in `_INDICES` or
   `_GATED_FKS`: three add new tables only, the fourth adds one nullable column plus a
   backfill. `upgrade()` / `downgrade()` are byte-identical to before.
2. **One merge conflict, in `backend/tests/test_alembic_migrations.py`** — both sides
   appended a new test section at the same end-of-file point. Resolved by **keeping both**:
   this branch's Phase 1 section and `main`'s `brief_intent` backfill section. Verified by
   parsing the resolved file and asserting zero duplicate top-level definitions or
   module-level assignments; 23 tests pass, covering both sections.
3. **One claim correction in that same file.**
   `test_phase1_paper_deployments_matches_a_fresh_create_all_schema`'s docstring described
   `fk_paper_deployments_owner_user_id` as **"pre-existing"**. It is not.
   `b7e3f1a2c9d4` added `owner_user_id` FKs to five tables — `strategy_store`,
   `strategy_passports`, `strategy_proposals`, `user_profiles`, `vault_metadata` — and
   `paper_deployments` is **not** among them. **All three** of `paper_deployments`'
   foreign keys are added by this PR, and `_GATED_FKS` names all three.

## Merge ordering

**This PR must land before #1429** (`dbrowneup/1367-backlog`, account-deletion cascade
policy). #1429's migration `85ca5310b7a1` is serialized behind this PR's last revision
(`9c2e7b5a1f4d`) and it *alters* `fk_paper_deployments_owner_user_id` — the constraint
this PR creates — from `ON DELETE SET NULL` to `ON DELETE CASCADE`. Landing #1429 first
would leave it altering a constraint that does not exist yet.


dbrowneup:
## Adversarial review — per-finding disposition

Head SHA: `1aece684e2d5ec9c3092c11a5b2f543fd8b206bc`. This PR stays **draft** — Dan reviews
any migration touching prod before it merges.

| # | Finding | Disposition |
| --- | --- | --- |
| Major 1 | Parity tests compared column names only, couldn't detect constraint/index drift | **Fixed.** `_index_signature` / `_fk_signature` helpers added; both `test_phase1_*_matches_a_fresh_create_all_schema` tests now assert index names+columns and FK targets, not just column names. C2 in `docs/database-relations.md` and the PR body corrected to describe exactly this scope. Mutation-check: temporarily removed `LinkedWallet.address`'s FK — the extended test failed on the FK-signature assertion (column assertion alone still passed) — reverted. |
| Major 2 | No `lock_timeout`, migration runs one transaction | **Fixed.** `SET lock_timeout = '5s'` added at the top of both migrations (Postgres only, before any table-touching statement). The migration already runs as one transaction (`migrations/env.py`), so a timeout failure aborts cleanly with no partial schema change — safe to retry. |
| Major 3 | `docs/database-relations.md` G2 describes an unreachable-deployment scenario the code makes impossible | **Verified impossible, corrected.** `paper_routes.py`'s `deploy_paper()` always sets `owner_user_id=user.id` from an authenticated session; its module docstring states plainly there are "no legacy wallet-owned rows and no fallback tier" for this table. Folded into § 1.1 as correction **C4** (not a real gap) instead of leaving a false claim in the gap list; removed from the Phase 2 proposal. |
| Additional major (mine) | Prod Aurora has known orphan populations; a plain `ADD CONSTRAINT ... FOREIGN KEY` validates on add and would hard-fail the deploy | **Addressed — restructured.** All four FKs are added `NOT VALID` in `fb8d0bae8112` (no validation at all in that revision). `VALIDATE CONSTRAINT` moved to a **separate follow-up revision** (`9c2e7b5a1f4d`), itself gated on a live orphan count, so a known orphan is only ever *named* (printed) never a blocking failure. Splitting into two revisions means an unexpected validate-time failure can't roll back the indices/widen/NOT-VALID-constraint creation that already landed. Honest caveat documented in both the migration docstrings and the PR body: `env.py`'s current transaction model shares one transaction across a whole `alembic upgrade` invocation unless run as two separate invocations — flagged for Dan rather than changed unilaterally, since it's a global, bigger decision. |
| Minor 5 | PR says "three foreign keys," it's actually four | **Fixed in all four places**: migration docstring, PR body, `docs/database-relations.md` § 2.3 heading, and the ERD caption. |
| Minor 6 | `alembic upgrade --sql` crashes on this revision | **Fixed.** Root cause: `bind.execute(orphan_sql).scalar_one()` ran unconditionally and returns `None` from Alembic's offline mock connection. `fb8d0bae8112` no longer issues any live-bind query in `upgrade()` (moved to the validate revision); `9c2e7b5a1f4d` guards its orphan check with `context.is_offline_mode()` and renders the unconditional `VALIDATE CONSTRAINT` statements for review instead. Full rendered SQL (`alembic upgrade f0ab58339d55:9c2e7b5a1f4d --sql`) is in the PR body. Regression tests added: `test_phase1_offline_sql_renders_without_crashing` / `..._downgrade_renders_without_crashing`. |
| Minor 7 | Two index-justification comments in `identity.py` contradict the code | **Fixed.** The comment above `ix_identity_events_wallet_time` mis-described which existing index leads with which column (said "the first leads with event_type" — it actually leads with `wallet`). Corrected to accurately describe both existing indices' actual failure modes. |
| Minor 8 | `docs/database-relations.md` claims `status: current` and implies Phase 1 is live | **Fixed.** Front matter now reads `status: draft — pending merge (PR #1438, unmerged)`; the doc body states "Nothing in § 2 is live" and reframes § 2 as "what PR #1438 introduces," not what shipped. `docs/README.md`'s index row updated to match (`status: draft`). |
| Minor 9 | Downgrade narrows `strategy_id` back to VARCHAR(64) — hard-fails on longer values, doc claims "all reversible" | **Fixed — validate-then-narrow.** `downgrade()` now calls `_fail_if_narrow_would_truncate` first, which counts rows where `length(strategy_id) > 64` (Postgres + online only; no-op on SQLite/offline) and raises a named `RuntimeError` instead of a generic Postgres error if any exist. Documented as the one honestly-irreversible piece of this PR rather than claiming blanket reversibility. Adversarial test: fake bind reporting a nonzero count → raises; zero → no-op; SQLite/offline → never queries (bind that raises if queried proves this). |

### Verification run on this head

- `pytest -m "not integration"` from repo root: **3613 passed, 2 skipped, 2 deselected, 0
  failed**.
- `ruff check --select E9,F63,F7,F40,F82 .` (blocking `ruff-gate` scope): clean.
- `ruff format --check` on every touched file: clean.
- `make docs-check`: 1151 links / 146 files, 0 broken; 142/142 docs indexed, 0 orphaned.
- `alembic upgrade --sql` / `alembic downgrade --sql` for both new revisions: render
  end-to-end, no crash (full SQL in the PR body).
- `test_alembic_migrations.py`: 20/20 passed, including 3 new mutation-proofs (parity/FK,
  orphan-SQL drift guard, downgrade-narrow guard) alongside the two carried over from the
  original draft (wrong orphan SQL, dialect guard removed).

### One deviation flagged for Dan, not applied unilaterally

`migrations/env.py`'s `run_migrations_online()` wraps a whole `alembic upgrade`
invocation in one transaction (no `transaction_per_migration=True`). Splitting FK-adding
and FK-validating into two revisions only gives true transaction isolation between them
if they're applied as two separate `alembic upgrade` invocations; applied together in one
`alembic upgrade head` call (the normal build-on-deploy path) they still share a
transaction. The live orphan gate means a *known* orphan population can never trigger a
hard failure either way, but changing `env.py`'s transaction model globally to make the
isolation unconditional is a bigger, separate decision — noted in both migration
docstrings and the doc, not made here.

PR stays draft pending Dan's review of the migration.

PR #1439, opened 2026-08-20
[backend] Route generation-path market data through the cached provider seam

dbrowneup:
## Summary

PR #1282 built a market-data vendor seam (`archimedes.services.market_data_provider`) with a `MARKET_DATA_PROVIDER`-selectable provider and a Postgres read-through cache (`asset_daily_bars`), and wired three backend call sites through it (`strategy_signal_evaluator`, `oracle_updater`, `asset_market_service`).

Two GENERATION-path call sites were not among them and still fetched market data directly:

- `backend/archimedes/services/fusion_market_data.py:96-101` (`_fetch_one`) — called `archimedes_analytics_engine.data.fetch_ohlcv` directly.
- `backend/archimedes/services/portfolio_backtester.py:89-96` (`_fetch_price_panel`) — same direct import/call, once per symbol.

Both are vendor-swappable in the sense that `archimedes_analytics_engine.data.fetch_ohlcv` is itself a façade over `archimedes_analytics_engine.market_data.get_provider()` (analytics-engine's *own*, separate, DB-less seam) — but that seam has no cache, since analytics-engine has no Postgres dependency. So every generation-path backtest re-fetch (fusion/debate variant sweeps, portfolio-weights backtests) hit the vendor fresh, never the `asset_daily_bars` cache the other call sites benefit from. Part of #1218.

## Before / after

**Before** (`fusion_market_data.py:96-101`):
```python
def _fetch_one(yf_ticker: str, start: str, end: str) -> Any:
    _ensure_analytics_import()
    from archimedes_analytics_engine.data import fetch_ohlcv
    return fetch_ohlcv(yf_ticker, start, end)
```

**After**:
```python
def _fetch_one(yf_ticker: str, start: str, end: str) -> Any:
    from archimedes.services.market_data_provider import get_provider
    return get_provider().get_daily_ohlcv(yf_ticker, start, end)
```

**Before** (`portfolio_backtester.py:89-96`, inside `_fetch_price_panel`):
```python
    _ensure_analytics_import()
    from archimedes_analytics_engine.data import fetch_ohlcv
    ...
    for sym in symbols:
        try:
            df = fetch_ohlcv(sym, start, end)
```

**After**:
```python
    from archimedes.services.market_data_provider import get_provider
    provider = get_provider()
    ...
    for sym in symbols:
        try:
            df = provider.get_daily_ohlcv(sym, start, end)
```

## What changed in the seam itself

`market_data_provider.MarketDataProvider` gains a new abstract method, `get_daily_ohlcv(ticker, start, end) -> pd.DataFrame` (full OHLCV, not just close — the shape both call sites need: backtrader feeds need Open/High/Low/Close/Volume, the portfolio simulator needs Close + Volume for its Almgren-impact cost model). This sits alongside the existing `get_daily_close_batch` rather than replacing it, since the two serve different shapes.

- `YFinanceProvider.get_daily_ohlcv` delegates to `archimedes_analytics_engine.data.fetch_ohlcv` — the exact function both call sites called directly before this change — so the vendor-fetch logic is not duplicated and cold-cache output is unchanged.
- `CachingMarketDataProvider.get_daily_ohlcv` adds read-through caching against the *same* `asset_daily_bars` table `get_daily_close_batch` already uses, with its own coverage/freshness checks. A row primed by `get_daily_close_batch`'s writer (close-only, other columns NULL) is treated as a miss by `get_daily_ohlcv` and gets healed with the full bar on re-fetch, so the two methods never hand back a partial row as if it were complete.
- On a full cache miss, the vendor's frame is returned **unchanged** (no reshape/reindex) before any (best-effort) cache write — this is what guarantees cold-cache output is identical to the pre-seam direct-fetch path.

## Anti-goals honored

- No graded-number changes: with the provider mocked at its boundary, a cold cache returns the vendor's frame byte-identical (`pd.testing.assert_frame_equal`) — see `TestDailyOhlcvCache` and `TestCacheColdWarmThroughGenerationPath` in the test files below.
- Rigor gate, quota, and payment logic: untouched — this PR only touches the market-data fetch layer.
- No new dependency: reuses `archimedes_analytics_engine.data.fetch_ohlcv`, which both call sites already depended on before this change.

## Tests (hermetic — no live yfinance/network; provider mocked at its boundary)

- `backend/tests/services/test_market_data_provider.py` — added `TestDailyOhlcvCache` (cold-miss-primes, warm-hit-skips-vendor, stale-TTL-refetches, insufficient-back-coverage-refetches, close-only-row-is-not-a-valid-hit, vendor-error-propagates-uncaught) and `TestYFinanceProviderOhlcvBoundary` (delegates to `archimedes_analytics_engine.data.fetch_ohlcv` unchanged). Extended the existing `_FakeVendor` / `_FakeVendorPassthroughSpy` fakes with the new abstract method.
- `backend/tests/services/test_generation_market_data_seam.py` (new) — routing guards for both call sites (`get_provider().get_daily_ohlcv` is actually called), source-level guards that neither call site imports `archimedes_analytics_engine.data.fetch_ohlcv` directly any more, and end-to-end cold/warm cache tests exercised through the real call-site functions with a real `CachingMarketDataProvider` (fake inner vendor, isolated SQLite session).
- `backend/tests/services/test_portfolio_backtester.py` — updated `test_empty_dataframe_vector` to mock at the new boundary (`market_data_provider.get_provider`) instead of the old `sys.modules["archimedes_analytics_engine.data"]` boundary.

Tallies: the four seam-related test files together — 65 passed, 0 failed. Full suite: `pytest -m "not integration"` from repo root — 3621 passed, 2 skipped (pre-existing, unrelated), 2 deselected, 0 failed.

## Mutation-check demonstrations

Each guard was verified to actually reject the thing it claims to reject (revert → fails; restore → passes), via `git diff HEAD > patch; git checkout HEAD -- <file>; run test; git apply --include=<file> patch`:

1. **Fusion routing guard** — reverted `fusion_market_data.py` to its pre-change state: `TestFusionMarketDataRoutesThroughSeam`, the fusion `TestNoSecondDirectFetchPathRemains` case, and the fusion cold/warm cache test all failed (`get_daily_ohlcv` called 0 times instead of once; the reverted code took the old direct-fetch path and returned a differently-shaped frame). Restored — all 3 pass again.
2. **Portfolio-backtester routing guard** — same recipe on `portfolio_backtester.py`: the portfolio routing guard, the portfolio `TestNoSecondDirectFetchPathRemains` case, the portfolio cold/warm cache test, and `test_empty_dataframe_vector` (mocked at the new boundary) all failed against the reverted file. Restored — all pass again.
3. **Close-only-row guard** (`_read_cached_ohlcv`'s completeness check) — temporarily disabled the `any(... is None ...)` check: `test_close_only_cached_row_is_not_a_valid_ohlcv_hit` failed (a close-only row was wrongly served as a complete OHLCV hit). Restored — passes.
4. **Back-coverage guard** — temporarily disabled the start-reach-back check: `test_insufficient_back_coverage_refetches` failed (a short-window cache entry was wrongly served for a longer request). Restored — passes.
5. **Cold-cache pass-through (anti-goal) guard** — temporarily changed the miss path to `return fetched.reset_index(drop=True)` instead of `return fetched`: the cold-cache-equivalence tests in both `test_market_data_provider.py` and `test_generation_market_data_seam.py` failed on index-shape mismatch. Restored — passes.

After all five mutation checks, `git diff HEAD` was confirmed byte-identical to the pre-mutation-check patch (no drift left behind).

## Lint / format

`ruff format --check` and `ruff check` pass clean on all touched files.

### DIFF
--- a/backend/archimedes/services/fusion_market_data.py
+++ b/backend/archimedes/services/fusion_market_data.py
@@ -8,9 +8,11 @@ supplier:
 
 - resolves a spec's ``asset_universe`` to yfinance tickers via the
   Chainlink-only universe SSOT (``GLOBAL_ASSETS``, PR #842),
-- fetches real daily OHLCV through the analytics-engine's ``fetch_ohlcv``
-  (the single owner of the fetch+normalize contract — same reuse rationale as
-  ``portfolio_backtester``),
+- fetches real daily OHLCV through the market-data provider seam
+  (``archimedes.services.market_data_provider.get_provider().get_daily_ohlcv``,
+  #1218/#1282) — vendor-swappable via ``MARKET_DATA_PROVIDER`` and
+  cache-backed by ``asset_daily_bars``, same seam the request-path call sites
+  use; same reuse rationale as ``portfolio_backtester``,
 - strictly inner-joins the panel across assets (missing data is a lookahead /
   survivorship vector — same rule as ``portfolio_backtester._fetch_price_panel``),
 - hands back per-asset **feed factories**: a backtrader feed object is consumed
@@ -94,11 +96,16 @@ def _ensure_analytics_import() -> None:
 
 
 def _fetch_one(yf_ticker: str, start: str, end: str) -> Any:
-    """Fetch one symbol's normalized OHLCV DataFrame. Test seam — monkeypatch me."""
-    _ensure_analytics_import()
-    from archimedes_analytics_engine.data import fetch_ohlcv
+    """Fetch one symbol's normalized OHLCV DataFrame via the cached
+    market-data provider seam (#1218/#1282 —
+    ``archimedes.services.market_data_provider``), so this fetch honors
+    ``MARKET_DATA_PROVIDER`` and reads/writes the ``asset_daily_bars``
+    Postgres cache the same way the request-path call sites do — a
+    generation-path re-run of the same universe hits the cache, not the
+    vendor, after the first fetch. Test seam — monkeypatch me."""
+    from archimedes.services.market_data_provider import get_provider
 
-    return fetch_ohlcv(yf_ticker, start, end)
+    return get_provider().get_daily_ohlcv(yf_ticker, start, end)
 
 
 def resolve_universe(asset_universe: list[str]) -> dict[str, str]:
--- a/backend/archimedes/services/market_data_provider.py
+++ b/backend/archimedes/services/market_data_provider.py
@@ -445,6 +445,19 @@ def _read_cached_ohlcv(session, ticker: str, start_date: date, end_date: date, t
     if earliest > start_date + timedelta(days=_COVERAGE_TOLERANCE_DAYS):
         return None  # cache doesn't reach back far enough for this request
 
+    # Forward-coverage is load-bearing, not symmetry for its own sake: both
+    # call sites default `end` to "today" on every call, so a cache primed
+    # through day D is routinely re-read after the date rolls to D+1 while
+    # still inside the TTL. Without this check that read is a "hit" on a
+    # silently shorter frame — a backtest window truncation that moves every
+    # graded number with no signal. Known tradeoff: a symbol whose vendor
+    # data genuinely ends before `end_date` (delisted/halted) now misses and
+    # re-fetches every call; correctness over cache hits — recording a
+    # vendor-end marker at write time is the fix for that, out of scope here.
+    latest = rows[-1].trade_date
+    if latest < end_date - timedelta(days=_COVERAGE_TOLERANCE_DAYS):
+        return None  # cache doesn't reach forward to the requested end
+
     newest_fetch = max(r.fetched_at for r in rows)
     if newest_fetch.tzinfo is None:
         newest_fetch = newest_fetch.replace(tzinfo=UTC)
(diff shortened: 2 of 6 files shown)


dbrowneup:
Adversarial review (two lenses + verification): one **confirmed blocker**, now fixed on-branch.

**Finding:** `_read_cached_ohlcv`'s docstring promised the cache "holds through `end_date`" but the code only checked back-reach (`rows[0]` vs `start_date`) and TTL — never `rows[-1]` vs `end_date`. Both call sites default `end` to today on every call, so a cache primed through day D was routinely served as a hit after the date rolled to D+1 within the TTL — a silently shorter backtest window, i.e. changed graded numbers with no signal, violating this PR's own "no graded-number changes" anti-goal. None of the original six warm-cache tests advanced `end` forward against a warm cache.

**Fix (latest commit):** symmetric forward-reach check (`rows[-1].trade_date < end_date - tolerance → miss`) with a comment recording the deliberate tradeoff (a symbol whose vendor data genuinely ends early now re-fetches every call — correctness over hit rate), plus `test_insufficient_forward_coverage_refetches` mirroring the back-coverage test.

**Mutation check:** reverted only the provider fix → the new test FAILS with the exact silent-hit signature (`vendor calls == []`); restored (byte-identical to the patch) → 32/32 across both seam test files.

**Scope note:** the pre-existing `_read_cached_series` (close-batch path from #1282) has no explicit `end` parameter — its implicit end is "now" and its exposure is bounded by the TTL (≤12h ≈ 0–1 trading day), inside the 5-day tolerance this design already accepts on the start side. Left unchanged deliberately; a vendor-end marker at write time is the right future hardening for both paths.

Issue #1441, opened 2026-08-20
[payments] Generation charge settles before enqueue — a failed job keeps the user's payment (no refund/release path)


dbrowneup:
Found during the sprint-card overlap audit (cluster-5/cluster-6 verification), present in the SHIPPED x402 generation paywall (#1296/#1300), not in a draft.

## The defect

The generation payment settles **before** the job is enqueued (`backend/archimedes/services/generation_payment.py` — settle-then-enqueue ordering), and there is no refund, release, or reservation mechanism anywhere in the path. Consequences:

1. A job that fails after settlement — enqueue error, worker crash, LLM failure, container roll mid-run — keeps the charge with nothing delivered.
2. There is no `Idempotency-Key` (or equivalent) on the payment step, so a client retry after an ambiguous failure can double-charge.

Today `PAYMENTS_DRY_RUN=true` in prod, so no real money is at risk YET — but this becomes a live money-integrity defect at the exact moment the payment stack flips on (#1427→#1414→#1426→#1428 sequence). It must be fixed or explicitly accepted BEFORE that flip.

## Precedent in-repo

`backend/archimedes/services/service.py:834` has the sibling pattern for at-most-once semantics worth mirroring.

## Acceptance sketch

- A payment that settles but whose job never reaches a terminal `done` is either (a) automatically released/refunded, or (b) durably recorded as a credit redeemable by the next generation attempt — pick one, write it down.
- Payment step accepts an idempotency key; a retried request with the same key cannot settle twice (regression test + mutation check).
- The Terms page's payment description stays true in the same change (see PR #1432's coupling note).

## Anti-goals

- Do NOT weaken or bypass the paywall/quota; do NOT touch the flag values themselves. This rides the payment-stack lane and its owner's sequencing.

dbrowneup:
Dan's decision (2026-08-20 evening): **Option A** — the testnet flip proceeds tonight with this defect known and bounded (faucet money, $2/incident, re-drippable, settle-log-visible). **This issue is prioritized immediately after MVP core is verified working** — Dan: release-or-credit + idempotency is 'the kind of polish and professionalism I would expect in a high quality financial product.' Hard gate: **real-USDC mainnet (Sept 16) does not happen without this fixed** — recorded on the #834 flip-list. Ops interim: watch settle logs for post-payment job failures; manual redress = re-drip guidance.

dbrowneup:
[...]

Issue #1431, opened 2026-08-20
[frontend][claim-integrity] /architecture tells users to sign in with a wallet or passkey — prod supports neither, and the sign-in page says so


dbrowneup:
## Summary

`/architecture` tells visitors to **"sign in with any wallet or a passkey"** — twice. Neither method exists. The three that do work (email/password, Google, GitHub) are not mentioned anywhere on the page.

Production is authoritative on this:

```
$ curl -s https://archimedes-arc.com/api/auth/providers
{"emailPassword":true,"google":true,"github":true,"passkey":false}
```

No wallet provider, and `passkey: false`.

## The site contradicts itself

`ui/src/components/AuthPage.jsx:433` — the actual sign-in page — says the opposite in as many words:

> Circle wallet passkeys authorize Circle smart wallets. **They do not sign in to Archimedes.**

and `AuthPage.jsx:315`:

> Account owns your strategies and settings. **Wallets stay optional and link only after signature proof.**

So one live surface instructs the user to sign in with a passkey, and the surface it sends them to states that passkeys do not sign you in. `AuthPage.jsx` offers no wallet sign-in path at all — `grep -niE "siwe|connect wallet|sign-in with ethereum"` returns nothing.

## Evidence (verified on `main` @ `61952c4f`, and in the deployed bundle)

`ui/src/components/Architecture.jsx:236-239` — pipeline step 1:
```js
{
    title: "Your brief",
    sub: "plain English, optional asset classes, risk profile, and a model cost picker · sign in with any wallet or a passkey",
},
```

`ui/src/components/Architecture.jsx:1257-1259` — the closing CTA:
```js
{ROADMAP_SURFACES_ENABLED
    ? ROADMAP_COPY.ctaLede
    : "Describe a strategy in plain English, then sign in with any wallet or a passkey to see it through the rigor gate."}
```

**Note the second one is in the `else` branch** — it is the copy written *specifically for the currently shipped build*, not leftover roadmap text. Someone wrote it for today's product and got the sign-in methods wrong.

Both are live in production right now:
```
$ curl -s https://archimedes-arc.com/assets/index-C07Xo5rR.js | grep -o "wallet or a passkey" | wc -l
2
```

## Why it matters beyond a wording nit

[...]

Issue #1354, opened 2026-08-20
[frontend][claim-integrity] Every public surface still sells the vault + marketplace journey that #1266 removed from the build


dbrowneup:
## Summary

`#1266` hid the vault and marketplace **routes** behind `ROADMAP_SURFACES_ENABLED` (default off in every production build — `VITE_ROADMAP_SURFACES` is set nowhere outside `ui/.env.example:41`). It did not touch the **copy**. The shipped app therefore still advertises "deploy into a non-custodial vault", "five wallet signatures, all yours", and "publish → subscribe → 90/10 split" as the primary journey, while every door those claims point at resolves to `not-found`.

Live-verified on production today (2026-08-20), not inferred from source:

```
$ curl -s https://archimedes-arc.com/ | grep -o 'src="/assets/[^"]*\.js"'
src="/assets/index-CFrO5p7D.js"

$ grep -c "Deploy as vault"                assets/index-CFrO5p7D.js        → 1
$ grep -c "non-custodial vault on Arc"     assets/index-CFrO5p7D.js        → 1
$ grep -c "Pay creators, not the house"    assets/index-CFrO5p7D.js        → 1
$ grep -c "Withdraw assets from your vault" assets/index-CFrO5p7D.js       → 1
$ grep -c "Deployed Vault"                 assets/AuthenticatedApp-DGLNvB4K.js → 1
$ grep -c "Portfolio and Reasoning"        assets/AuthenticatedApp-DGLNvB4K.js → 1
$ grep -c "Core strategy journey"          assets/AuthenticatedApp-DGLNvB4K.js → 1
```

And the consequence, from the live funnel:

```
$ curl -s https://archimedes-arc.com/api/metrics/funnel
… {"stage":"landed","distinct_visitors":326} … {"stage":"vault_deployed","distinct_visitors":0}
```

326 visitors have been sold a journey whose fourth step is unreachable; zero have completed it. This is the `CLAUDE.md` §"claims must be true" rule failing on the app's front door.

**Relationship to existing work.** `#1266` shipped the gate; `#1324` / PR #1331 caught the Library "Published" tab (one dead door, already fixed — do not re-scope into it). `#1241`'s claims ledger carries the *wording* rows ("executing into your non-custodial vault on Arc" → soften); this issue is the executable slice for the **build-flag** half — gate the copy so it tracks the flag, rather than re-wording it once and having it drift again the next time the flag flips.

[...]

PR #1442, opened 2026-08-20
[docs] Sprint README: verified state refresh (2026-08-20)

dbrowneup:
The sprint README's State section (2026-08-16, "zero sprint work has landed") predates three days of merge waves — including the card author's own #1239/#1242 — and no longer describes reality. This adds a verified 2026-08-20 state section: all 118 actionable card items were checked one by one against current main (evidence: merged PR numbers, file:line reads, live-prod checks), scoring 35 DONE / 27 PARTIAL / 42 OPEN / 14 SUPERSEDED-OBSOLETE, with per-card dispositions and where the live remainder is executing.

Also records the owner's 2026-08-20 execution-style call on session rules 5/7 (written for a solo token-constrained session; superseded for sessions running the repo's parallel-agent pipeline per CLAUDE.md), while keeping rule 6's anti-goals in force.

Doc-only change. The original 2026-08-16 section is left intact below the new one as the historical record.

### DIFF
--- a/docs/sprint/README.md
+++ b/docs/sprint/README.md
@@ -21,6 +21,38 @@ Path corrections vs the source doc:
 
 ~20 working days remain to Sept 16 against a 24-day estimate. See the re-cut at the bottom.
 
+## State — 2026-08-20 (verified, main @ 0a22d034)
+
+The 2026-08-16 section below is superseded: far from "zero sprint work landed," a
+card-by-card verification of all 118 actionable items against current `main` found
+**35 DONE, 27 PARTIAL, 42 OPEN, 14 SUPERSEDED/OBSOLETE**. Highlights:
+
+| Card | Disposition |
+|---|---|
+| cluster-0-unblock | Code asks absorbed; residue is owner-only (creds, ADRs, PyPI, Bedrock) |
+| cluster-1-cost-ssot | #1242 shipped the core; every PARTIAL closes by landing #1379 |
+| cluster-2 / cluster-3 | Ride #1379 + two work packages in flight (fusion dates, sleeve label) |
+| cluster-4-strategies-route | #1340/#1374 landed both mains; two small items in a work package |
+| cluster-5-meter | **Retired** — #1194 + #1296/#1300 rebuilt this space (flat settle, no reservation); the missing refund/release path is now #1441 |
+| cluster-6-boot-paywall | A1/A2 sit in draft #1401; the `/api/v1` half needs an owner architecture call (`/api/rigor/verify` already occupies the space) |
+| cluster-7-ui-surface | Core retired by #1266's flag model; surviving tail dispatched as work packages |
+| cluster-8-returns-csv | Blocked on the same `/api/v1` architecture call |
+| a6-rerun | **Retired** — the re-run self-healed via #1263/#1306; read-path fix + copy item remain |
+
+Work packages dispatched 2026-08-20 evening (draft PRs, adversarially reviewed before
+merge): manifest honesty (E3/E5), UI dead-code + check scripts, api-surface status doc,
+quota on the bare generate endpoint, derived leaderboard selectivity copy. #1379's rebase
+follows the market-data seam PR (#1439) to avoid a same-file collision. Owner-gated items
+(payment flags, multisig, Circle mainnet data, `/api/v1` fork, QuantLab disposition) are
+with Dan.
+
+**Execution-style note (Dan, 2026-08-20):** rules 5 and 7 below ("no subagents/workflows",
+"max 2 merges/day") were written for a solo token-constrained session and are superseded
+for sessions running the repo's parallel-agent pipeline (see CLAUDE.md § parallel agent
+fan-out): builder agents + adversarial review + 0-behind merge watchers + post-burst union
+suites. The anti-goals in rule 6 remain in force except where explicitly shipped past
+(React.lazy and the Architecture.jsx rework landed via sanctioned #1194/#1123).
+
 ## Session rules — apply to every card
 
 1. **Anchor-trust.** Re-anchor with `grep -n "<symbol>" <file>` (~200 tok), then `Read` with


onder-akkaya:
Reviewed the state table against the tree at `baa173a`. The dispositions hold — I'm only flagging one card item that the cluster-5 retirement doesn't cover, because verifying it turned up something on a now-live payment path.

**cluster-5 is rightly retired on the metering design.** #1194 + #1296/#1300 did rebuild that space, and #1441 owns the refund/release gap. No argument.

**But B5 — the premium-model downgrade — isn't in either.** #1441 is refund/release; B5 is model selection. It's still in the tree at `generate_routes.py:335`:

```python
selected_model = req.model if is_allowed_model(req.model) else None
```

`is_allowed_model` (`llm_backend.py:71`) returns true only for `FREE_TIER_MODELS`, so an **entitled premium** id also falls through to the env default. To be fair to the code: this is *disclosed*, not silent — the comment at `:329-334` says exactly this and gives the reason (premium can't serve until Bedrock activation, roadmap T3.8), and `:337` logs the fallback. So it's a documented limitation, not a hidden one, and I'd have left it alone.

**What made me raise it is the flag state.** `a45e8e5` ("THE FLIP") set `GENERATION_PAYMENT_REQUIRED = "true"` (`ecs.tf:563`) with `GENERATION_PAYMENTS_DRY_RUN = "false"` (`:547`), so generation payments are enforced and settling for real. The other half of the pair, `PREMIUM_MODELS_ENABLED`, is **absent from `ecs.tf` entirely** — its prod value is whatever `model_gate.py:56` defaults to when unset, which is off.

So today it's safe. But it's *one env var* from not being: set `PREMIUM_MODELS_ENABLED=true` — no code change, no deploy gate between the two flags — and an entitled caller pays $2 and is served the env default model instead of the premium one they paid for. That's a charge-for-X-deliver-Y shape on a live rail.

Two suggestions, both small, take or leave:
1. **Pin `PREMIUM_MODELS_ENABLED` explicitly in `ecs.tf`, even to `false`.** Same reasoning as the `PAYMENTS_DRY_RUN` fix in #1239 — an unset money/entitlement switch whose prod value is an accident of a code default is a bug regardless of which way it points. (Related: `PAYMENTS_DRY_RUN` is pinned in `ecs.tf` but still unset in all three compose files and `infra/scripts/setup-ssm-secrets.sh`, so cluster-0's bullet is 1-of-4 done.)
2. **Make the pairing explicit in code** — refuse to honour an entitled premium request while it cannot actually be served, rather than downgrading it, so payment-enforced + premium-enabled can't quietly become charge-and-substitute.

Either would close B5 properly; without one, I'd keep the card item open rather than retire it with the rest of cluster-5.

Everything else in the table matched what I found. The rules 5/7 execution-style call is noted and I've adopted it in #1453 — that PR is scoped as the verification layer (what a `grep` finds, where a card's own text is wrong) to sit alongside this one as the dispatch state, and it says so explicitly. Happy for this to land first; I'll rebase onto it.

---
_Generated by [Claude Code](https://claude.ai/code)_

onder-akkaya (commented):
Verified the added section line by line against the tree. **This should land** — it's additive, it leaves the 2026-08-16 section intact as the historical record, and pinning the claim to `main @ 0a22d034` (real: `0a22d03`, "Merge pull request #1391", 2026-08-20) is exactly the right habit. Four findings, none blocking, one worth an answer.

### 1. Two rows were overtaken within hours of the snapshot

Both PRs the table treats as pending have merged, on the same day:

| Cited as | Actually |
|---|---|
| "every PARTIAL closes by landing **#1379**" | #1379 merged — `2c27a52`, 2026-08-20 |
| "#1379's rebase follows the market-data seam PR (**#1439**)" | #1439 merged — `b42fdc6`, 2026-08-20 |

That's the hazard of a snapshot rather than an error in it, but a reader today acts on the row. What actually closed when #1379 landed (I verified all three at `baa173a`): Engine C is on the cost floor (`fusion_evaluator.py:230-232`, `:393-395`, with `TestEngineCFloor` pinning both call sites), `cost_model_id` reaches every writer, and the fabricated `0.0` correlations now serve `None` — including `portfolio_backtester.py:447`'s fail-soft default. So cluster-1 reads better than the row claims.

What did **not** close: cluster-2/3's "two work packages in flight (fusion dates, sleeve label)". Both still absent — `grep -rn "dsl-fusion-sleeves\|n_independent_sleeves" backend ui/src` returns nothing, and the dead `backtest_start` condition is still at `fusion_evaluator.py:277` with `:430` still fabricating `date(2004, 1, 2)`.

Either add "as of `0a22d03`" to those two rows or refresh them — I'd take the one-word fix over a rewrite.

### 2. The `Architecture.jsx` attribution doesn't hold, and it's load-bearing

The execution-style note retires two rule-6 anti-goals as "explicitly shipped past (React.lazy and the Architecture.jsx rework landed via sanctioned #1194/#1123)."

`React.lazy` via #1194 checks out. **`Architecture.jsx` via #1123 does not.** #1123 is `a5a2934`, 2026-07-16. The churn that took that file 896 → 1329 lines is **ten commits on 2026-08-19/20** — `#1431`, `#1354`, `#1382`, plus the claim-integrity series — none of them #1123. Whitespace-insensitive that's +750/−324 on an 896-line file.

This matters because the attribution is what justifies retiring the anti-goal. The *work* looks right to me (it's mostly claim-integrity retractions, which is the good kind of churn), so I'd expect the sanction to be straightforward — I just can't trace it to the PR cited. Point it at the real ones and it's settled.

### 3. The docs index row didn't move

`docs/README.md:170` still reads `Last verified 2026-08-16` while this PR adds a 2026-08-20 verified section. `docs/CONVENTIONS.md` makes that column part of the contract, so as it stands the index understates the doc's freshness.

**Merge-order note:** my #1453 sets that same row to `2026-08-21`. If this lands first, #1453 carries the bump and there's no conflict. If #1453 lands first, this needs a trivial rebase. Either order is fine — just flagging so neither of us is surprised.

### 4. The cluster-5 row — still my one open question

Restating briefly since it's a row in this diff: cluster-5 is rightly retired on the metering design, but **B5's premium-model downgrade isn't covered by #1441** (refund/release) or by the rebuild. Detail in [my comment above](https://github.com/a-apin/archimedes/pull/1442#issuecomment-5393900257); the short version is that with `GENERATION_PAYMENT_REQUIRED = "true"` and `GENERATION_PAYMENTS_DRY_RUN = "false"`, flipping `PREMIUM_MODELS_ENABLED` charges $2 and delivers the env default model.

I've since opened #1482 (pins that flag to `false`, plus a comment in that block that contradicted its own value) and #1483 (nothing in CI parses a `.tf` file — assigned to you). If you'd rather the card row stay "Retired" and track B5 on #1482 instead, that works for me; I just don't think it should vanish silently between the two.

---

**Commenting rather than approving** only because of #4 — it's a row in this diff and I raised it before the review. Settle it either way (answer, or narrow the row to exclude B5) and I'll approve; #1 and #2 are corrections I'd happily see land as a follow-up rather than hold this up. #3 is a one-line rebase question.

---
_Generated by [Claude Code](https://claude.ai/code)_

dbrowneup:
Superseded — every unique item in this diff reached main via #1453 (merged 2026-08-28, c34815e) and its follow-ups, which carry the per-card dispositions in a re-verified, richer table. Önder's four asks from 08-24 are moot against the merged version. Closing per tonight's 14-PR triage (owner-approved dispositions).

Issue #1356, opened 2026-08-20
[frontend][backend] An API failure renders as an empty state or a measured zero — fail-soft sweep across six surfaces


dbrowneup:
## Summary

Six reader-facing surfaces convert a backend failure into a plausible, confident-looking answer instead of a visible absence.

- A 500 on `/api/strategies/generated` renders as **"No generated strategies yet"** and invites the user to generate more.
- A failed `/api/selection-bias/gate` silently deletes **every** deployability chip in the Library.
- A failed `/api/corpus/overview` sits on **"Loading overview..."** forever; a failed catalog fetch announces **"0 papers found"** in a `role="status"` live region.
- A total leaderboard outage reads **"No strategies match these filters yet."** on the flagship public page.
- An RPC failure inside `ConfigService` publishes `pools: {}` / `vaults: {}`, which the landing census renders as **"Arc census live · … 0 pools"** and `/architecture` renders as a measured `0` — on a page whose own header rule forbids exactly that.

#1309's recurring ~2-minute 502/503 window is the live trigger: every one of these is reachable in production today. This is CLAUDE.md § *"Fail-soft is correct for optional configuration and wrong for anything a claim depends on"* (`docs/architectural-principles.md:161-172`) — and one of the failures is currently **locked in by a passing test**.

All line numbers below re-verified against `main` at `c0e6400f`.

### Scope — owned elsewhere, do not fix twice

- `RiskAnalysis.jsx:493-516` (same class: `fetch` does not reject on 404, so the mock-data notice at `:537-544` never renders under `FEATURE_QUANT=false`). Mounted only from `QuantLab.jsx:260`, nav-hidden in prod → the Quant Lab prod-readiness issue owns it.
- `PaperTrading.jsx:235` day-0 `+0.00%` → Paper Trading claim-integrity issue.
- `Reasoning.jsx:322-337` anchored-only green tick → commit-reveal issue.
- Leaderboard *fixture-fallback numbers* → #1187 / PR #1340.

---

### A. Library — three of four feeds fail into an empty state, and the default tab has no loading guard

`ui/src/components/Strategies.jsx:665-692` — `Promise.allSettled` over four endpoints. Only `seedRes` has an `else`:

```js
665:      const [seedRes, genRes, gateRes, publishedRes] = await Promise.allSettled([
...
677:        setLoadError(seedRes.reason?.message || 'Failed to load examples')
679:      if (genRes.status === 'fulfilled') { ... }       // no else
[...]

PR #1443, opened 2026-08-20
[frontend] Leaderboard: derived selectivity headline (N evaluated, K clear the bar)

dbrowneup:
## Summary

Sprint card: [`docs/sprint/a6-rerun.md`](https://github.com/a-apin/archimedes/blob/main/docs/sprint/a6-rerun.md) — A6's rejection-rate item: "Reframe the headline from scoreboard to rejection rate ... a board that passes everything is marketing."

Adds a headline aggregate to `ui/src/components/Leaderboard.jsx` — "Of N candidates evaluated, only K clear the rigor bar." — presented with the page's existing accent-bordered callout styling (matches the `engine.disclaimer` box idiom already on the page).

**Both numbers are derived at render time from `data.entries`, the same array the results table below renders — never a hard-coded literal:**
- `evaluatedCount = data.entries.length` (every row currently fetched under the active scope/filters)
- `passingCount = data.entries.filter(e => e.passes_rigor_gate).length`

This directly follows CLAUDE.md's corollary: *"Don't quote a curated-library strategy pass count — anywhere ... Say 'unestablished', not a number."* The prior finding on this repo ("3 of 34" as a literal claim) was itself later shown to be wrong (equity-like series through a data-feed fallback) — this PR makes that class of error structurally impossible on this page by construction, not just by discipline.

## Design notes / deviations

- **Denominator includes any `is_backtest_placeholder` rows** (generated-but-not-yet-backtested strategies, own-scope only). I considered excluding them (they haven't really been "evaluated") but kept them in `evaluatedCount`, since including them can only make the reported pass rate look *more* conservative, never inflate it — the safer direction given the hard constraint. Called out here for review in case the team wants the stricter definition instead.
- Headline is gated on `!loading && !error && data && !data.degraded && evaluatedCount > 0` — identical guard shape to the existing results-table non-empty state — so it never renders while loading, on error, when the board is `degraded` (#1356's honest-degradation signal), or before any strategies exist (no "0 of 0" claim).
- No data-layer changes, no new endpoints, no feature flags — pure derivation from the existing `/api/leaderboard` payload the page already fetches.

## Guard test

`ui/test/leaderboard-selectivity.test.js` — mirrors the claim-integrity source-scan idiom in `roadmap-copy.test.js` / `signin-claims.test.js`:
- Raw source-text scan of `Leaderboard.jsx` for any literal `\b\d+\s+of\s+\d+\b` pattern (e.g. "3 of 34", "34 of 34") — must be absent.
- Anti-vacuity: the pattern is asserted to match its own canonical bad examples ("3 of 34", "34 of 34").
- Positive check that the derived-count code path exists (`data.entries`, `passes_rigor_gate`, `evaluatedCount`, `passingCount` all present in source) — so the guard can't pass merely because the feature was deleted.
- Guards the "no claim before data arrives" behavior via the `evaluatedCount > 0` gate.

**Mutation-check performed by hand before pushing** (per CLAUDE.md's guard discipline): temporarily inserted the literal string `"3 of 34"` into `Leaderboard.jsx`, ran `node --test test/leaderboard-selectivity.test.js` — the guard failed as expected (`AssertionError: Leaderboard.jsx contains a literal "<N> of <M>" pass-count...`), then reverted and reran to confirm green. Also caught and fixed a false-positive on myself: my own code comments described the "0 of 0" edge case using that literal digit shape, which the guard flagged — reworded to "zero-over-zero" to keep the guard meaningful rather than weakening it.

## Verify

```bash
export PATH="$(conda info --base)/envs/archimedes/bin:$PATH"
cd ui
node --test        # 332 passed, 0 failed
npm run lint       # clean
npm run build      # clean (pre-existing chunk-size warning only, unrelated)
```

## Test plan

- [x] `node --test` — 332/332 pass
- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] Guard test mutation-checked (fails on injected literal, passes after revert)
- [ ] Visual check on the live/local board once deployed (both curated and own-scope views, plus the "Rigor-gated only" filter toggle — note: with that filter active, `evaluatedCount === passingCount` by construction, which is correct/honest, not a bug)

### DIFF
--- a/ui/src/components/Leaderboard.jsx
+++ b/ui/src/components/Leaderboard.jsx
@@ -109,6 +109,19 @@ export default function Leaderboard() {
   const servedScope = data?.scope ?? (user ? 'own' : 'curated')
   const isOwn = servedScope === 'own'
 
+  // Selectivity headline (WP-7, docs/sprint/a6-rerun.md's rejection-rate item):
+  // "of N candidates evaluated, only K clear the rigor bar" — derived at
+  // render time from the SAME `data.entries` array the table below renders,
+  // never a hard-coded count (CLAUDE.md: "Don't quote a curated-library
+  // strategy pass count — anywhere"). evaluatedCount is every row currently
+  // fetched under the active scope/filters (including any is_backtest_placeholder
+  // rows without a real backtest yet — counting those in the denominator can
+  // only make the reported pass rate look MORE conservative, never inflate
+  // it). Both stay `null` until `data` arrives, so the loading/empty states
+  // below never render a zero-over-zero claim.
+  const evaluatedCount = data ? data.entries.length : null
+  const passingCount = data ? data.entries.filter((e) => e.passes_rigor_gate).length : null
+
   return (
     <div className="leaderboard-page" style={{ maxWidth: 1100 }}>
       <div style={{ marginBottom: 18 }}>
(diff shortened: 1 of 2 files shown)


dbrowneup:
Top-level review: good build — the derived-count discipline, the guard's anti-vacuity, and the self-caught "0 of 0" comment false-positive are all right. Two honesty corrections applied in the latest commit before this goes ready:

1. **Placeholder rows leave both counts.** A row with `is_backtest_placeholder` was never seen by the gate, so it cannot appear in a "graded" count — the direction of the resulting error doesn't matter; "claims must be true" is about literal truth, not conservative bias. Counts now derive from `entries.filter(e => !e.is_backtest_placeholder)`.
2. **Scope-true nouns.** Curated-scope copy now says "library strategies graded by the rigor gate" — leaderboard rows are graded library strategies, not generation-funnel "candidates" (89% of actual candidates die at generation time and never reach this page, so "candidates" would understate real selectivity while misnaming what's counted). Own-scope keeps "your strategies", now also phrased "graded by the rigor gate".

332/332 ui tests, eslint clean. The guard test needed no change (it forbids literals, not wording).

PR #1444, opened 2026-08-20
[frontend] Remove dead components, add orphan/sitemap check scripts

dbrowneup:
## Scope

Sprint card: [`docs/sprint/cluster-7-ui-surface.md`](https://github.com/a-apin/archimedes/blob/main/docs/sprint/cluster-7-ui-surface.md) — "Sprint subset (~0.4d) — claim-honesty only" table, specifically: the four named dead-file deletions, the `check-orphans.mjs`/`check-sitemap.mjs` pair from the card's "Tests" section, and the Generate-page copy line. Technical-only body per convention (business/roadmap context lives in the private docs repo).

## 1. Dead-file deletions — zero-importer proofs

For each file: grepped `ui/src` (recursive), `ui/index.html`, `ui/vite.config.js` for the basename, for any `import ... from '<path>'` / dynamic `import('<path>')` referencing it, and for any barrel/`index.js` re-export. All four came back with **zero real importers** (only prose mentions in unrelated comments, which are not code paths).

- **`ui/src/components/FusionResult.jsx`** — `grep -rn "FusionResult" ui/src ui/index.html ui/vite.config.js` hits only its own `export default function FusionResult(...)` and two prose comments in `roadmapCopyApp.js` / `StrategyPassport.jsx`. No `import` statement anywhere.
- **`ui/src/components/PortfolioAdvisor.jsx`** — same command for `PortfolioAdvisor` (word-boundary, distinct from the *live* `PortfolioAdvisorPanels.jsx`, which QuantLab.jsx does import). Hits are its own export plus prose comments describing the *unrelated* `PortfolioAdvisorPanels` component. No import statement.
- **`ui/src/data/promptLibrary.js`** — `grep -rn "promptLibrary" ui` → zero hits anywhere, including inside the file's own former importers.
- **`ui/src/assets/hero.png`** — `grep -rn "hero\.png\|assets/hero"` across `.js/.jsx/.html/.css/.json` → zero hits. (The only "hero" hits were the unrelated `.public-hero` CSS class and prose comments.)

Also checked: no dynamic `import()` in the codebase references any of these (the only dynamic import anywhere is `App.jsx`'s `lazy(() => import('./AuthenticatedApp'))`), and neither `components/` nor `data/` has a barrel `index.js`.

**Note on the card's 5th named file, `assets/logo_old.svg`:** it does not exist in the tree (`ls ui/src/assets/` → `flow-diagram.svg, hero.png, react.svg, vite.svg`). Not in this WP's assigned deletion list either — left alone, not fabricated.

**Cascading cleanup required by the FusionResult.jsx deletion:** `roadmapCopyApp.js` exported a `fusion` object (`deployAsVaultLabel`/`deployAsVaultTitle`) whose own comment states it existed *solely* for FusionResult.jsx's gated CTA — verified via `grep -rn "roadmapCopyApp\.fusion\|deployAsVaultLabel\|deployAsVaultTitle"` (no other consumer). Removed it and its file-header mention, and dropped `FusionResult.jsx` from `test/roadmap-copy.test.js`'s `SURFACE_FILES` list (the file it names no longer exists — that test's own anti-vacuity check, "every declared surface file exists," would otherwise fail loudly, which it did until fixed).

## 2. New check scripts — `ui/scripts/check-orphans.mjs` + `check-sitemap.mjs`

Per the card's explicit "do NOT stand up vitest" instruction, both are plain Node scripts using the same source-as-text idiom already used across `ui/test/*.test.js` (regex over file contents, no bundler). Wired as:

```json
"check:orphans": "node scripts/check-orphans.mjs",
"check:sitemap": "node scripts/check-sitemap.mjs",
"check:routes": "npm run check:orphans && npm run check:sitemap"
```

**`check-orphans.mjs`** — walks `src/`, builds an import graph (static + dynamic imports, comment-stripped so a prose mention can't fake a reference), and fails if any file under `src/components/` or `src/data/` has zero importers. Has an `ALLOWLIST` mechanism for genuine unreachable-by-import entry points.

Running it today surfaced a **second real orphan the card's list missed**: `src/components/WelcomeProfileModal.jsx` (217 lines, zero importers, verified the same way as the four above). It references a "wallet-menu edit-profile" entry point that doesn't exist yet, which reads like unfinished wiring rather than inert cruft, so **it was left in place** (out of this WP's authorized deletion list) and added to the script's `ALLOWLIST` with an explicit "known dead, deletion deferred, needs a human call" comment — not silently swept under a false "legitimate entry point" label. Flagging here for a delete-vs-wire-up decision.

**Mutation check** (per repo convention — a guard must be shown to reject something): added a temp orphan file `src/components/__TempOrphanForMutationTest.jsx`, ran the script → exit 1, listing it; deleted the temp file, ran again → exit 0. Not committed.

**`check-sitemap.mjs`** — parses `src/routes.js`'s `PUBLIC_PATHS`/`APP_PATHS`/`ANON_APP_PAGES` **as text** (no module import, so it never needs routes.js to export its private tables — zero route-file behavior change) and `public/sitemap.xml`'s `<loc>` entries, then fails if any route routes.js marks genuinely public is missing a sitemap entry.

Deliberately **one-directional** (routes.js-public ⊆ sitemap, not proven equal), and documented as such in the script's header comment. The reverse direction does not currently hold: **`/insights` is listed in `sitemap.xml` and `Insights.jsx`'s own header comment calls it "the public conversion + traction dashboard... Public-only,"** but neither `routes.js`'s `ANON_APP_PAGES` nor `nginx.conf`'s anonymous-browse carve-outs include it — so today an anonymous visitor hitting `/insights` is bounced to `/sign-in` by `App.jsx`'s redirect effect (the same class of bug the card cites for the `/marketplace` WalletGate mismatch). Fixing that needs an `ANON_APP_PAGES` + `nginx.conf` change — a routing/gating change, out of scope for this check-script-only PR per its anti-goals. Flagging for a follow-up issue rather than silently enforcing (which would fail this gate on unrelated pre-existing state) or silently ignoring (which would hide a real finding).

**Mutation check:** temporarily deleted the `/corpus` `<url>` block from `public/sitemap.xml`, ran the script → exit 1, naming `/corpus`; restored from a backup copy, ran again → exit 0, diffed against the backup to confirm byte-identical restoration. Not committed.

## 3. Generate.jsx copy — the 281/6/sleeves line

Added under the asset picker's existing caption:

> Steer the universe. Leave empty to use the full supported set — **{SUPPORTED_ASSETS.length} assets**. Engine C caps each spec at **{ENGINE_C_ASSET_CAP}** and grades them as independent sleeves.

Both numbers verified from live code rather than hard-coded blind:

- **281** — `{SUPPORTED_ASSETS.length}` reads the already-imported `SUPPORTED_ASSETS` export from `ui/src/data/assetUniverse.js` at render time, so it can never drift from what the picker actually offers. That file is itself autogenerated from the backend SSOT by `backend/archimedes/scripts/gen_ui_asset_universe.py` and CI-guarded against drift (`test_ui_asset_universe.py`); confirmed the current value is exactly 281 via `node -e "import('./src/data/assetUniverse.js').then(m => console.log(m.SUPPORTED_ASSETS.length))"` → `281`, matching the backend's independent 281-synth SSOT assertion in `backend/tests/test_synthetic_addresses.py`.
- **6** — `ENGINE_C_ASSET_CAP = 6`, a new local constant with a source-anchor comment pointing at `backend/archimedes/services/fusion_market_data.py:53 _MAX_ASSETS = 6` (confirmed via `grep -n "_MAX_ASSETS" backend/archimedes/services/fusion_market_data.py`). This mirrors the existing precedent in `ui/src/config.js` (`MIN_VAULT_FUNDS_RAW`, "Mirrors `MARKETPLACE_MIN_VAULT_FUNDS_RAW` on the backend") for a cross-language constant that can't be imported directly — it's a plain frontend/backend Python constant, so a literal JS import isn't possible; the comment is the drift guard, matching the exact phrasing (picker offers 281 assets, Engine C caps at 6 per spec, independent sleeves) that `docs/sprint/cluster-2-fusion-engine.md:59-61` specifies for this line.

## Verification

```
cd ui && node --test                        # 312 passed, 0 failed
npm run check:routes                        # both scripts OK
./node_modules/.bin/eslint <touched files>   # clean
npx vite build                               # clean; confirmed new copy text + confirmed
                                              # FusionResult/PortfolioAdvisor absent from
                                              # dist/assets/*.js; dist/ not committed
```

Backend suite untouched — `git diff --name-only` against `main` shows zero `backend/` paths.

## Anti-goals honored

No route changes (routes.js untouched — the sitemap script reads it as text, doesn't modify or even import it). No component rewrites (Generate.jsx's asset-picker JSX structure is unchanged; only the caption text and one new constant were added). Nothing behind `ROADMAP_SURFACES_ENABLED` touched behaviorally — the `roadmapCopyApp.js` edit *removes* a now-orphaned export, it doesn't change what renders under the flag.

## Deviations from the literal card list

1. **`assets/logo_old.svg`** (card's 5th named file) doesn't exist in the tree — not deleted (nothing to delete), not fabricated.
2. **`WelcomeProfileModal.jsx`** — new orphan found by the tooling this PR adds, not in the card's list, left in place + allowlisted with a "needs a human call" comment rather than deleted or silently ignored (see § 2 above).
3. **`check-sitemap.mjs` scope** — implemented as one-directional (routes.js-public ⊆ sitemap) per the explicit WP brief, not the stronger bidirectional "sitemap ≡ derived public set" the card's prose also gestures at — the bidirectional version would fail today on the pre-existing `/insights` gating gap (§ 2), and fixing that gap is a routing change outside this PR's anti-goals.


dbrowneup:
Top-level review: pass. Deletion proofs verified (zero-importer including dynamic-import/barrel checks), the cascading roadmapCopyApp/SURFACE_FILES cleanup is exactly right, and leaving `WelcomeProfileModal.jsx` on an explicit allowlist with a human-call note instead of silently deleting it is the correct instinct.

**Direction note on the `/insights` sitemap finding so nobody fixes it backwards:** the resolution is to REMOVE `/insights` from `sitemap.xml` (and its "Public-only" header claim), not to make the page anonymous-accessible — Insights is being converted to an admin-only surface in an in-flight branch, after which a sitemap listing would advertise a page visitors cannot reach. The in-flight admin-only PR should carry the sitemap removal; the one-directional scope of `check-sitemap.mjs` here is fine as-is and the stronger bidirectional check can land after that.

PR #1445, opened 2026-08-20
[backend] Route POST /api/strategies/generate through the shared generation quota

dbrowneup:
## Sprint reference

Work-package: WP-6 ("the unmetered generation endpoint" + cohort-divergence TODOs). Sprint card: `docs/sprint/cluster-4-strategies-route.md` § 4 ("The unmetered budget hole — 2 lines") and § 3 ("A7-cohort — note the divergence, fix in buffer").

## Premise verification (before writing any code)

Read both routes side by side:

- `backend/archimedes/api/strategies_routes.py` — `POST /api/strategies/generate` (`generate_strategy`, pre-fix at line 2208) was decorated only with `@limiter.limit("20/minute")`. It never imported or called `enforce_generation_quota`. This is the "unmetered budget hole": a second live, SIWE-gated, LLM-spending generation path with no daily-volume cap.
- `backend/archimedes/api/generate_routes.py` — `POST /api/generate/start` (`start_generation`, line 193) is rate-limited `5/minute` **and** calls `await enforce_generation_quota(request, user.id)` (line 211, gated `if not os.getenv("TESTING")`) before any other work — the 10/day/account + 20/day/IP caps in `backend/archimedes/services/generation_quota.py`.

Confirmed: the hole is real, exactly as the sprint card describes.

For the A7-cohort TODOs, read both filters:

- `strategies_routes.py::_live_rigor_results_for_strategies` (the `valid_returns` filter, ~line 347): `len(v) >= 10 and float(np.ptp(np.asarray(v, dtype=float))) > 0.0` — excludes zero-variance series.
- `live_rigor_gate.py::verdicts_for_strategies` (the `valid_returns` filter, ~line 268): `len(v) >= _MIN_RETURNS_FOR_GATE` — length only, no zero-variance exclusion.

Confirmed the divergence is still real (not yet unified) — added the paired `# TODO(A7)` notes rather than skipping them.

## What changed

1. **Quota wiring** (`strategies_routes.py`): `generate_strategy` now calls the identical `enforce_generation_quota(request, user.id)` the primary path calls, under the same `if not os.getenv("TESTING")` gate, run FIRST — before the `fusion_enabled()`/corpus checks — matching the primary path's "cheapest anti-abuse check before any other work" ordering. No new quota logic was written; this endpoint now goes through `services/generation_quota.py` exactly as `/api/generate/start` does. Because both calls key off `user.id` / client IP only (`archimedes:genquota:{scope}:{day}:{identity}` — never the endpoint), the daily counters are shared automatically, with no additional plumbing.
2. **Cohort-divergence TODOs**: paired `# TODO(A7)` comments at `strategies_routes.py` (~347) and `live_rigor_gate.py` (~268), each naming the divergence and pointing at the sprint doc, per the card's "buffer work" instruction — no filter behavior changed.

## Anti-goals honored

- No payment/paywall/x402 code touched — `git diff` confirms zero occurrences of "payment"/"x402" in the diff.
- No threshold changes.
- No behavior change to `/api/generate/start` — it calls the same function it always did; this PR only adds a second caller.

## Tests

New file `backend/tests/test_strategies_generate_quota.py` (5 tests, hermetic — `TestClient(app)`, signed legacy-SIWE cookies via `tests.auth_helpers.auth_cookies`, every boundary mocked: the Redis client behind `GenerationQuota._get_redis`, the job store, the fire-and-forget background task, the optional market-regime Redis read):

- `test_quota_enforced_on_strategies_generate_returns_429_before_fusion_work` — over-cap caller gets 429 with `reason=generation_daily_cap`, `scope=user`, `cap=1`; `fusion_enabled` spy proves fusion work never starts.
- `test_429_body_is_byte_identical_between_the_two_endpoints` — the same over-cap identity hitting both endpoints gets the exact same `detail` payload.
- `test_quota_counters_are_shared_across_both_endpoints` — cap=1, IP layer disabled; call `/api/generate/start` first (202, consumes the only slot), then `/api/strategies/generate` (429). Uses an in-memory `_FakeCountingRedis` that really persists per-key counts, so this is a genuine proof of one shared Redis key incremented twice, not two independent mocks each returning 1.
- `test_quota_counters_shared_in_the_other_call_order_too` — same proof, reversed call order (strategies-generate first).
- `test_quota_disabled_preserves_prior_strategies_generate_behavior` — both caps `<= 0` → a Redis-touch spy proves `GenerationQuota._get_redis` is never even called, and the endpoint reaches enqueue exactly as before (202 + job_id).

### MUTATION-CHECK

Per CLAUDE.md's "a guard must be shown to reject something": temporarily removed the `enforce_generation_quota` call from `generate_strategy` (kept the import to avoid an unused-import warning) and re-ran the new file.

Before restoring — 4 of 5 tests failed, exactly the ones that depend on the wiring:
```
FAILED test_quota_enforced_on_strategies_generate_returns_429_before_fusion_work - AssertionError: fusion_enabled must not run after a quota block
FAILED test_429_body_is_byte_identical_between_the_two_endpoints - assert 503 == 429
FAILED test_quota_counters_are_shared_across_both_endpoints - AssertionError: {"status":"queued","job_id":"job-strat-1"}
FAILED test_quota_counters_shared_in_the_other_call_order_too - AssertionError: {"job_id":"job-gen-1", ...}
4 failed, 1 passed
```
(The one still-passing test is the quota-disabled test, correctly unaffected since it asserts behavior when the caps are off regardless of wiring.)

Restored the wiring via `git diff` comparison against the pre-mutation patch (byte-identical — confirmed with `diff`), never via `git stash`. Re-ran: `5 passed`.

### Full suite

```
pytest -m "not integration" -q
# 3620 passed, 2 skipped, 2 deselected
```

```
ruff format --check backend/archimedes/api/strategies_routes.py backend/archimedes/services/live_rigor_gate.py backend/tests/test_strategies_generate_quota.py
# 3 files already formatted
ruff check backend/archimedes/api/strategies_routes.py backend/archimedes/services/live_rigor_gate.py backend/tests/test_strategies_generate_quota.py
# All checks passed!
```

## Deviations from the work package

None. Scope is exactly: quota wiring on `/api/strategies/generate` + the two `# TODO(A7)` comments. `git diff --stat` touches only `strategies_routes.py` (+19), `live_rigor_gate.py` (+3), and the new test file.


dbrowneup:
Top-level review: pass. Verified in the diff: the enforcement call is byte-for-byte the sibling route's idiom (same function, same key format, keyed on identity not endpoint — so counters are provably shared), ordered before any other work matching the primary path, and the TESTING gate mirrors `/api/generate/start`. The mutation demonstration (4/5 tests fail with the wiring removed) shows the tests exercise the enforcement despite the TESTING gate. The paired TODO(A7) markers land the cohort-divergence breadcrumb at both sites as the card asked.

PR #1446, opened 2026-08-20
[docs][backend] API surface status doc + completeness gate

dbrowneup:
## Summary

WP-5: adds a census doc + a mechanical completeness gate for the backend API surface.

- **`docs/api-surface-status.md`** — one row per `app.include_router(...)` call in
  `backend/archimedes/main.py` (30 rows, current registration block is lines 493-528):
  prefix, router module (`archimedes.api.<module>.<router_name>`), auth model
  (`public` / `session` / `private-admin` / `internal`, plus honest `mixed (...)`
  compounds where a router genuinely isn't uniform — e.g. `vaults_router`'s reads are
  public but `POST /create` needs a linked wallet), and status (`live` / `gated` /
  `degraded-capable`). A "Documented in" column links to the existing per-surface docs
  under `docs/api/` where one exists — 15/30 rows do; the other 15 are named as real
  documentation debt, not silently glossed over. Front matter follows
  `docs/CONVENTIONS.md`; indexed in `docs/README.md` in this same commit.
- **`backend/tests/test_api_surface_status.py`** — parses `main.py`'s source via `ast`
  (never imports `archimedes.main`, so no DB/Redis/SSM side effects) to get the live
  `include_router` set, parses the doc's table cells via regex for the same set, and
  asserts equality in both directions: a router registered but undocumented fails, and
  a stale doc row for a removed router also fails.

No router or route was added, removed, or changed — verified with `git diff` on
`backend/archimedes/main.py` (empty) before pushing.

Sprint card: [`docs/sprint/cluster-7-ui-surface.md`](https://github.com/a-apin/archimedes/blob/main/docs/sprint/cluster-7-ui-surface.md)
§ "Backend routers — leave every one registered, unchanged".

## Guard verification (mutation-tested per CLAUDE.md's "a guard must be shown to reject something")

Both directions were manually broken and confirmed to fail before this was pushed, then
reverted:

1. Deleted the `assets_router` row from the doc → `test_doc_and_main_agree_on_router_set`
   failed with `missing a row ... ['assets_router']`. Restored from a backup, diffed
   clean against the original.
2. Appended `app.include_router(totally_fake_router)` to `main.py`'s registration block
   → same test failed with `missing a row ... ['totally_fake_router']`. Restored from a
   backup, diffed clean against the original (`git status` shows no diff on `main.py`).

## Tallies

- `pytest -m "not integration"` from repo root: **3617 passed, 2 skipped, 2 deselected,
  0 failed** (282.64s).
- `make docs-check`: links 0/1168 broken across 147 files; index 143/143 docs indexed, 0
  orphaned (this doc included).
- `ruff check` + `ruff format --check` on the new test file: clean.
- Doc census: 30 rows (one per `include_router` call); 15/30 have a detailed doc under
  `docs/api/`.

## Deviations from the task brief

- The sprint card's own phrasing for this item asks for columns "prefix, UI consumer,
  status, plan" and cites `main.py:471-495`; the work-package brief that dispatched this
  PR specified different columns (prefix, router module, auth model, status) and a
  different line range. I followed the work-package brief (the more specific, more
  recently-given instruction) — the actual `include_router` block in current `main.py`
  is lines 493-528, not 471-495 (confirmed by reading the file, not trusting either
  cited range).
- Added a "Documented in" column beyond the four required columns, cross-referencing the
  existing `docs/api/*.md` per-surface reference (which already covers 15 of the 30
  routers in detail, including its own five-level auth vocabulary). This directory
  needed reading anyway to classify auth models accurately, and surfacing which routers
  still lack a detailed doc seemed like exactly the kind of finding this census doc
  exists to make visible rather than bury.
- Where a router's routes don't share one auth model (10 of 30 routers), the Auth model
  cell uses a `mixed (...)` compound plus a prose note spelling out the exact per-route
  split, rather than forcing a false single label — oversimplifying here would have
  produced a doc that reads as complete but misstates the real security boundary.

## Test plan

- [x] `pytest -m "not integration"` — 0 failed
- [x] `make docs-check` — 0 broken links, doc indexed
- [x] `ruff check` / `ruff format --check` on the new test — clean
- [x] Mutation check both directions (doc row removed; fake router added) — confirmed
      the guard fails, then reverted
- [x] `git diff` on `backend/archimedes/main.py` — empty (anti-goal: no route changes)


dbrowneup:
Top-level review: pass. AST-parse of main.py (no import side effects) + bidirectional doc equality is the right shape, both mutation demonstrations verified, and the `mixed (...)` labels are the honest choice over a forced single category. The extra "Documented in" column surfaces real doc debt (15/30 routers undocumented) — worth a follow-up issue after merge rather than scope creep here.

Heads-up recorded for in-flight branches: any PR that adds an `include_router` after this merges will fail the completeness gate until it adds its doc row — intended ratchet, failure message names the router.

PR #1447, opened 2026-08-20
[backend] Agent manifest: replace stale pending-T32 claims with true status

dbrowneup:
## What

`agent_manifest_routes.py` served a `_PENDING_T32` marker on the `deploy`,
`marketplace`, and `monitor` groups claiming those endpoints were "landing with
the T3.2 contract redeploy (issue #588) — endpoint exists but agent-driven use
isn't wired end-to-end yet." That is a false claim today:

- The T3.2 redeploy landed 289 contracts on Arc testnet on 2026-07-09.
- Issue #588's GitHub state has been terminal (non-open) since 2026-07-14,
  verified with `gh issue view 588 --json state,closedAt`.
- Reading the actual code paths (not just the manifest text) shows all three
  groups are genuinely wired to those redeployed contracts, with the same
  session + linked-wallet gating as every other group already marked `"live"`:
  - `POST /api/vaults/create` → `chain_executor.create_vault` — real
    `VaultFactory` call, then `setAgent` + `transferOwnership` for the
    non-custodial owner/agent split. No stub, no feature flag.
  - `GET /api/vaults/{address}/health` → `vault_monitor.get_vault_health` —
    real on-chain/Redis-backed read.
  - `POST /api/marketplace/publish` / `subscribe` → real DB rows, real
    on-chain vault creation, real Circle dev-controlled subscriber wallets,
    on-chain fee-cap enforcement.

Per this repo's "claims must be true" rule, a publicly served manifest cannot
keep asserting a blocked state that resolved over a month ago. This PR:

1. Removes `_PENDING_T32` and marks `deploy` / `marketplace` / `monitor` as
   `"live"`, matching every other functioning group.
2. Rewrites the module docstring's "Honesty note" and the stale in-line
   comment on the `paper` group, both of which repeated the same outdated
   claim.
3. Sources `auth.chain_id` from `ARC_CHAIN_ID` (module-level, same env var and
   default as `wallet_routes._SUPPORTED_CHAIN_ID` / `auth_siwe._EXPECTED_CHAIN_ID`)
   instead of a separately hardcoded `5042002` literal, so the manifest can no
   longer silently drift from the chain the rest of the backend actually
   enforces — the exact staleness class this PR exists to eliminate.
4. Adds a `chain_id` shape/value guard + tests (see "Guard demonstrations"
   below).
5. Updates `test_agent_manifest.py` / `test_agent_discovery.py` so the test
   suite itself no longer asserts the false "pending #588" state.

Sprint card: `docs/sprint/cluster-6-boot-paywall.md`, "Manifest honesty"
section (items referred to internally as E3/E5).

## Explicitly out of scope (held)

The same sprint-card section also asks to (a) add `"APIKey"` / `"x402"` to
`auth.schemes` plus a `pricing` block, and (b) replace the flat `chain_id`
with `chains: {payments: {...}, execution: {...}}`. Both are held out of this
PR:

- No settle proof exists yet for x402 on this surface — advertising a payment
  scheme the platform cannot yet honor would be the same class of dishonesty
  this PR is fixing in the other direction. `auth.scheme` (singular) is
  unchanged; nothing payment-related was added.
- The `chains{payments,execution}` split only makes sense once there is a
  distinct payment chain to disclose. Absent that, this PR keeps `chain_id` as
  a single scalar but makes it config-sourced and guards its shape/value
  instead — a narrower, compatible fix. **Deviation flagged for the sprint
  owner:** the full card asks for a test that `chain_id` appears "nowhere as a
  bare scalar," which presumes the `chains` split ships first; that test isn't
  implemented here because implementing it now would require either the held
  `chains` split or a test that fails against intentionally-unfinished work.
- Rebased onto current `main` before opening this PR (which had, among other
  things, a real x402 payment flow land for the generation-quote path) and
  re-verified: `infra/ecs.tf` still sets `PAYMENTS_DRY_RUN=true` for the
  DCW/marketplace-subscribe path specifically (the new x402 work is scoped to
  a different, caller-signed metered-API path). The "marketplace" group's
  `"live"` status in this PR is about the routes being wired to real
  contracts, not about payment settlement — the manifest makes no claim
  either way about billing, and this PR does not touch `PAYMENTS_DRY_RUN` or
  any payment flag.

**Also found, not fixed (separate surfaces, outside this PR's scope of
`agent_manifest_routes.py` + its tests):** `ui/public/.well-known/agent.json`
repeats the identical stale "landing with the T3.2 redeploy (#588)" text in
five places (three endpoint groups + two `skills` entries), and
`docs/agent-api.md` (line ~363) separately claims issue #588 "is still open,"
which is also no longer accurate. Both should be corrected in a follow-up
scoped to those files.

## Test plan

- `pytest backend/tests/test_agent_manifest.py backend/tests/test_agent_discovery.py -q`
  → 30 passed.
- `pytest -m "not integration" -q` from repo root → 3635 passed, 2 skipped
  (pre-existing, unrelated to this change), 2 deselected, 0 failed.
- Hermetic gate: `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m
  pytest backend/tests/test_agent_manifest.py backend/tests/test_agent_discovery.py -q`
  → 30 passed, 0 failed.
- `ruff format --check` + `ruff check` on all three touched files → clean.

### Guard demonstrations (revert → fail → restore → pass)

1. **Stale-status regression guard.** Reverted `agent_manifest_routes.py` to
   its pre-PR state while keeping the new/updated tests. Result: 3 failures,
   exactly as expected —
   `test_agent_manifest_deploy_and_marketplace_marked_live_post_t32` failed on
   `deploy status is "landing with the T3.2 contract redeploy (issue #588) —
   ..." not 'live'`; `test_agent_manifest_chain_id_is_a_well_shaped_scalar_sourced_from_config`
   and `test_chain_id_is_read_from_arc_chain_id_env_not_a_frozen_literal` both
   failed with `ImportError`/`AttributeError` on the now-absent `_CHAIN_ID`.
   Restored the fix (`git apply` from a saved patch) — back to 30 passed.
2. **chain_id shape/value guard.** `test_chain_id_guard_rejects_a_wrong_shaped_or_wrong_valued_input`
   directly proves `_assert_chain_id_is_a_well_shaped_scalar` rejects a
   stringly-typed value (`"5042002"`), both bools (`True`/`False` — `bool` is
   an `int` subclass in Python, so a bare `isinstance` check would wave them
   through), a float (`5042002.0`), and a wrong int value (`1`), while
   accepting the real shape/value.
3. **Config-sourcing guard.** `test_chain_id_is_read_from_arc_chain_id_env_not_a_frozen_literal`
   monkeypatches `ARC_CHAIN_ID` and reloads the module, proving `_CHAIN_ID`
   tracks the env var rather than being frozen at whatever value was
   hardcoded at some point in the past — the exact bug class `_PENDING_T32`
   itself was.

## Deviations from a literal reading of the sprint card

- Card's `auth.schemes` (plural) / `pricing` / `chains{payments,execution}`
  asks: held, per explicit scope for this work package (see "Explicitly out
  of scope" above).
- Card's "nowhere as a bare scalar" `chain_id` test: implemented a narrower,
  compatible guard instead (config-sourced scalar + shape/value rejection)
  since the full test depends on the held `chains` split.
- `.well-known/agent.json` and `docs/agent-api.md` carry the same stale claim
  and were deliberately left untouched (outside `agent_manifest_routes.py` +
  its tests) — flagged above for a follow-up.


dbrowneup:
Top-level review: pass. The premise-verification order was exactly right — reading `create_vault`/`vault_monitor`/`marketplace_routes` wiring before marking groups "live" is what makes the new statuses claims-backed rather than optimistic, and holding E1 plus narrowing the chain_id guard (with the deviation stated) were the correct scope calls. Mid-flight rebase over the x402 generation-quote landing with a re-check that `PAYMENTS_DRY_RUN` stays pinned was well handled. The out-of-scope stale text in `.well-known/agent.json` + `docs/agent-api.md` is now tracked as its own issue.

Issue #1368, opened 2026-08-20
[corpus] "Knowledge Graph" tab and unqualified "10,000 papers" chip survive #1315 — and the new guard scans neither file


dbrowneup:
## Summary

`/app/corpus` is anonymous-reachable (`ui/src/routes.js:41`). It offers a permanent tab labelled **"Knowledge Graph"** whose backing table is empty and has always been empty, and a header chip reading **"10,000 papers"** with no qualifier, while the very same API response says `processed_papers: 0`, `cluster_count: 0`, `pipeline_status: "skipped"`. A visitor who searches the KG tab gets `0 entities / 0 relations` and `No entities found.` — indistinguishable from "your query matched nothing", forever. The honest string that would say otherwise exists in the component but is dead code: it fires only on a 503 the endpoint never returns.

PR #1315 (merged 2026-08-20, refs #778) corrected the corpus **retrieval** claims and added a hermetic guard, `backend/tests/test_corpus_claim_integrity.py`. Its `PUBLIC_SURFACES` tuple does not list `CorpusExplorer.jsx` or `OnboardingTour.jsx`, and the merge touched neither file — so the two surfaces that still carry the over-claim are the two the guard cannot see. This is the #1 rule (claims must be true) with the guard-hole variant of CLAUDE.md § "A guard must be shown to reject something".

**Not a duplicate of #778.** #778 remains the parent and stays open for Part B (`#1090`/`#1091`/`#1092`/`#1093`); PR #1315's "Refs, not Closes" note explains why. This issue is the executable remainder of #778 item A on the two files #1315 skipped, plus a guard-coverage defect that postdates #778 by three months.

## Evidence

All verified against `origin/main` @ **523a1bae** and against live prod on **2026-08-20**.

**1 — The guard omits both offending files.** `backend/tests/test_corpus_claim_integrity.py:71-80`:

```python
PUBLIC_SURFACES: tuple[str, ...] = (
    "README.md",
    "docs/README.md",
    "docs/architecture.md",
    "docs/corpus-architecture.md",
    "docs/specs/architecture-page-design.md",
    "ui/src/components/Architecture.jsx",
    "ui/src/components/CorpusGraph.jsx",
    "ui/src/components/CorpusKG.jsx",
)
```

**And adding them changes nothing on its own.** Running the merged guard's own matcher against both files today:

```
$ PYTHONPATH=backend python -c "<load module>; print(m.find_overclaims(open(f).read()))"
[...]

PR #1450, opened 2026-08-20
[corpus] Fix Corpus search returning zero results for every term

dbrowneup:
## Summary

Typing anything into the Corpus catalog search box returns nothing — for any term, always. Reported by Dan; reproduced against production before the fix:

```
GET /api/papers/?search=momentum                       -> total=0
GET /api/papers/?search=momentum&processed_only=false  -> total=119   <- the real matches
GET /api/papers/?search=volatility                     -> total=0
GET /api/papers/?search=<empty>                        -> total=10000
```

## Why it fails, and why it looked fine

`processed_only` defaults to `True` and filters `cluster_id IS NOT NULL`. **The KB pipeline has never run in production** — `/health` reports `corpus_kg_built: false`, `corpus_kg_entities: 0` — so that predicate matches **zero rows for every query**.

Two `total == 0` fallbacks exist to rescue precisely this situation, but both are guarded by `and not category and not search`:

```python
if total == 0 and not category and not search:
    ...                       # re-query unfiltered
if total == 0 and not category and not search:
    ...                       # file-manifest fallback
```

That guard is deliberate and correct in intent — a user's search must not be silently widened into an unfiltered result set. But combined with a filter that matches nothing, it is what made search **dead rather than merely empty**:

- **No search term** → fallback fires → full 10,000-paper catalog renders → page looks healthy.
- **Any search term** → fallback skipped → `cluster_id IS NOT NULL` stands alone → zero results.

So the failure is invisible until the moment a user tries to use the feature. The same guard clause disabled the **category filter** for the same reason.

## The fix

Apply the `processed_only` filter only when at least one paper has actually been clustered:

```python
if processed_only and _any_paper_processed(session):
    query = query.filter(PaperRecord.cluster_id.isnot(None))
```

Where the pipeline **has** run, behaviour is unchanged and the filter still restricts to processed papers — covered by a test. Where it has not, the filter is skipped rather than zeroing the query. The honest reading of "processed only" when nothing is processed is not "show nothing".

`_any_paper_processed` is a `LIMIT 1` existence probe, not a count, since this runs on every catalog request. It is deliberately **not** memoised: the pipeline can land at any time, and a stale `False` would keep serving the unfiltered corpus after processing exists — the less honest of the two directions to be wrong in.

The two `total == 0` fallbacks are left exactly as they are; they still cover the genuinely-empty-DB case.

## Verification

```
pytest backend/tests/test_papers_routes.py -q     → 8 passed
ruff format --check + ruff check                  → clean
```

Five new tests, in `TestCatalogSearchOnUnprocessedCorpus`:

| test | pins |
|---|---|
| `test_search_returns_matches_when_nothing_is_kb_processed` | the regression itself |
| `test_search_matches_the_abstract_too` | abstract matching, not just title |
| `test_search_still_discriminates` | the fix does **not** turn search into "return everything" |
| `test_processed_only_still_filters_once_the_pipeline_has_run` | original intent preserved |
| `test_category_filter_also_works_on_an_unprocessed_corpus` | the same guard broke `category` |

**Mutation check** (`CLAUDE.md` § "A guard must be shown to reject something") — restoring the unconditional filter:

```
FAILED test_search_returns_matches_when_nothing_is_kb_processed
       - AssertionError: search over an unprocessed corpus returned nothing
FAILED test_search_matches_the_abstract_too              - assert 0 == 1
FAILED test_category_filter_also_works_on_an_unprocessed_corpus - assert 0 == 1
3 failed, 5 passed
```

Restored → `8 passed`. Note `test_search_still_discriminates` passes in **both** states by design — it is the guard against over-correcting, so it should not flip.

## Anti-goals honoured

- **Does NOT** widen the `total == 0` fallbacks to fire on a search. That would answer a user's query with unfiltered results, which is the thing the original guard exists to prevent.
- **Does NOT** change the default of `processed_only`, the `ilike` search predicate, or the file-manifest fallback.
- **Does NOT** touch the knowledge-graph surfaces (#1368) — this is the catalog search only.

## Related

- Search quality itself is untouched here: it remains `title ILIKE %term% OR abstract ILIKE %term%`. That is a reasonable floor and it does return sensible matches (119 for "momentum"), but ranking, stemming and author search are all absent — worth a separate issue rather than scope-creeping this fix.

### DIFF
--- a/backend/archimedes/api/papers_routes.py
+++ b/backend/archimedes/api/papers_routes.py
@@ -33,6 +33,19 @@ def _paper_row_to_dict(r) -> dict:
     }
 
 
+def _any_paper_processed(session) -> bool:
+    """True if the KB pipeline has clustered at least one paper.
+
+    Cheap existence probe (LIMIT 1), not a count — this runs on every catalog
+    request. Deliberately not memoised: the pipeline can land at any time and a
+    stale `False` would keep serving the unfiltered corpus after processing
+    exists, which is the less honest of the two failure directions.
+    """
+    from archimedes.models.corpus_store import PaperRecord
+
+    return session.query(PaperRecord.arxiv_id).filter(PaperRecord.cluster_id.isnot(None)).first() is not None
+
+
 @papers_router.get("/")
 async def list_papers(
     page: int = Query(1, ge=1),
@@ -63,7 +76,15 @@ async def list_papers(
     with get_session() as session:
         query = session.query(PaperRecord)
 
-        if processed_only:
+        if processed_only and _any_paper_processed(session):
+            # The filter is applied ONLY when the KB pipeline has actually
+            # clustered something. When nothing is processed — the live state,
+            # since the pipeline has never run (corpus_kg_built=false) — this
+            # predicate matches zero rows for EVERY query. The two `total == 0`
+            # fallbacks below rescue the unfiltered catalog, but both are
+            # guarded by `and not search`, so an empty catalog view looked fine
+            # while every search returned nothing. That combination is what made
+            # corpus search 100% dead rather than merely empty.
             query = query.filter(PaperRecord.cluster_id.isnot(None))
         if category:
             query = query.filter(PaperRecord.categories.contains(category))
--- a/backend/tests/test_papers_routes.py
+++ b/backend/tests/test_papers_routes.py
@@ -52,12 +52,12 @@ def _ctx_session(session):
     return _CtxSession(session)
 
 
-def _add_paper(session, arxiv_id, *, authors, cluster_id=None):
+def _add_paper(session, arxiv_id, *, authors, cluster_id=None, title=None, abstract="Abstract"):
     session.add(
         PaperRecord(
             arxiv_id=arxiv_id,
-            title=f"Paper {arxiv_id}",
-            abstract="Abstract",
+            title=title if title is not None else f"Paper {arxiv_id}",
+            abstract=abstract,
             authors=json.dumps(authors),
             primary_category="q-fin.PM",
             categories='["q-fin.PM"]',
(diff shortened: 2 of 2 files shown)


dbrowneup:
Reviewed directly (small diff, user-reported): **approve**. The probe's design reasoning is the right kind — unmemoised because a stale `False` would keep serving the unfiltered corpus after the pipeline lands, the less honest failure direction — and the test set covers both intents (search must hit on an unprocessed corpus; the filter must still apply where clustering exists; `test_search_still_discriminates` is the deliberate both-states guard against over-correcting into return-everything).

One non-blocking note for #1451's pile: with zero clustered rows the `LIMIT 1` probe scans until exhaustion on every catalog request (~10k rows — milliseconds today, and it short-circuits the moment the pipeline lands). If the corpus grows 10×, a partial index on `cluster_id IS NOT NULL` makes it constant.

dbrowneup:
Agreed on the probe-cost note, and here are the numbers so whoever picks it up under #1451 has them rather than re-deriving.

You're right that `LIMIT 1` buys nothing in the zero-match case — with no rows satisfying `cluster_id IS NOT NULL` the planner has to scan to exhaustion to prove absence. And `cluster_id` is not indexed (`backend/archimedes/models/corpus_store.py:35`, plain `mapped_column`, no `index=True`), so it's a seq scan.

The mitigating context is that this request was already scanning the same table before my change:

- `papers_routes.py:74` — `total = query.count()`, unfiltered in the zero-clustered case
- `papers_routes.py:88` — `fallback_query.count()`, a *second* full count on the same path

So the probe adds a third scan to a request that already did two, rather than introducing a new class of cost. Measured end-to-end against production just now:

```
GET /api/papers/?page_size=20   0.250 s
GET /api/papers/?page_size=20   0.139 s
GET /api/papers/?page_size=20   0.144 s
```

140–250 ms for the whole request at 10k rows. Your "fine at 10k, partial index if the corpus ever 10×s" is the right call, and a partial index on `cluster_id WHERE cluster_id IS NOT NULL` would collapse all three.

Worth noting for #1451 that the bigger win there isn't the probe — it's that the zero-clustered path does two full counts to answer one question. Whoever indexes this should look at collapsing those as well; the probe is the cheapest of the three.

PR #1454, opened 2026-08-21
[frontend][infra] Wallet-matrix payment unlock: connect flow, passkey signing, caps 100/200 !minor

dbrowneup:
Addresses #1298's field failures (Dan: MetaMask/Firefox, Circle Passkey/Safari, Coinbase/Chrome all unable to pay).

**Root cause 1 — the dead end**: with no in-session wallet connection, the pay panel said "use a browser wallet" with NO connect button (`getWalletClient()` only returns a client the in-app connect flow built). The no-wallet state now renders **Connect wallet →** dispatching the existing `open-wallet-modal` (EIP-6963 discovery — Brave and Phantom work through the same door).

**Root cause 2 — passkey guarded off**: the Circle smart account now signs the EIP-712 authorization via its viem `SmartAccount.signTypedData` (ERC-1271/6492 wire format, NOT re-wrapped — the #870/#871 lesson) and funds Gateway with approve+deposit as ONE batched userop through the existing `executeUserOp` bundler path. Facilitator acceptance of 1271 sigs is the remaining unknown — if it refuses, the exact facilitator error now surfaces verbatim in the UI (fix 3: `shortMessage`/`message` passthrough), so the next field test is diagnostic rather than mute.

**Also**: `GENERATION_DAILY_CAP_PER_USER` 10→100 and `_PER_IP` 20→200 (testing directive; per-IP must move too or the user cap is unreachable behind one IP).

Verification: 383/383 ui tests (3 new matrix guards + 1 repointed #1427-era pin whose behavior this supersedes); eslint clean; adversarial revert → 3/3 guards fail. **After merge**: UI rides the merge deploy; the caps need `apply.sh --apply` + the next deploy roll.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Hqgq6BzkRzuDrUL34xqZj7

PR #1461, opened 2026-08-21
[frontend] Wallet split-brain fix: announce wallet-changed on every connect/reconnect/disconnect !minor

dbrowneup:
## The bug (root cause of Dan's browser-payment failures in ALL three browsers)

Live-driven repro on prod Safari tonight (post-#1459 deploy): top bar shows the connected wallet chip (`Test_Wallet · 0x512b…9596`), but clicking **Approve & Generate** takes the *plain submit* path → 402 → the panel says **"Connect your wallet to pay"** — and clicking its Connect button can never fix it.

Why, verified in source:

- `Generate` copies the connected address into local state: `getAddress()` at mount + the `wallet-changed` event (`Generate.jsx:124,190-192`).
- The ONLY dispatcher of `wallet-changed` was `_onAccountsChanged` — the EOA runtime listener (`config.js`). `connectWallet()`, `connectCircleWallet()`, both `reconnectWallet()` branches, and `disconnectWallet()` all mutate `_address` silently.
- On page load, `reconnectWallet()` resolves AFTER Generate mounts → Generate's `walletAddr` is null forever. A manual connect via the modal calls `onConnect={setWalletAddr}` — AuthenticatedApp state only; Generate never hears about it either.

So the one-click pay path (#1459) never engages in a browser: not for Circle passkey (Safari), not for MetaMask (Firefox), not for Coinbase (Chrome). The agent/headless path never touches this component — exactly the observed split of "agent path proven end-to-end, every browser fails."

## The fix

One announcer helper in `config.js` dispatching `wallet-changed` with the module address; called at all five state transitions (EOA connect, passkey connect, both reconnect branches, disconnect). `_onAccountsChanged` reuses it; its revoke branch now relies on `disconnectWallet()`'s announce instead of double-dispatching.

Consumers already in place: `Generate.jsx` resyncs via its existing listener; `AuthenticatedApp`'s listener re-checks linkage on the same event (idempotent with its own reconnect `.then()`).

## Guard demonstrated to reject (per CLAUDE.md rule 3/4)

New pin test run against the PRE-fix `config.js` (stashed): **2/3 tests fail**. Against the fixed file: 3/3 pass. (The third is a consumer-side pin that holds both ways by design.) Full ui suite: **394/394**. ESLint clean.

## Field verification plan

After deploy: hard-reload the prod Generate tab in Safari → the pay panel should engage the one-click flow with the passkey wallet without any Connect prompt; then Dan retests MetaMask/Firefox and Coinbase/Chrome (errors render verbatim on-screen since #1454).

### DIFF
--- a/ui/src/config.js
+++ b/ui/src/config.js
@@ -41,17 +41,27 @@ let _lastSeenChainId = null
 // (Rabby/Brave/…) whose object is NOT window.ethereum (#921). Every handler
 // no-ops unless the event's provider is the one connectWallet selected
 // (_provider), so an announced-but-inactive wallet can't hijack the session.
+// Announce the module-level wallet state to the app. Components that copy
+// the connected address into their own state (Generate's payment panel:
+// getAddress() at mount + this event) can NOT see page-level setters like
+// AuthenticatedApp's onConnect — this event is their only resync. EVERY
+// transition of _address must fire it, or any panel mounted before the
+// transition keeps the stale address forever ("top bar says connected,
+// pay panel says connect wallet" split-brain).
+function announceWalletChanged() {
+  window.dispatchEvent(new CustomEvent('wallet-changed', { detail: { address: _address } }))
+}
+
 function _onAccountsChanged(provider, accounts) {
   if (provider !== _provider) return
   if (!accounts?.length) {
-    disconnectWallet()
-    window.dispatchEvent(new CustomEvent('wallet-changed', { detail: { address: null } }))
+    disconnectWallet() // announces the null address itself
     return
   }
   _address = accounts[0]
   if (_providerId) saveWalletMeta(_providerId, _address)
   _walletClient = createWalletClient({ account: _address, chain: arcTestnet, transport: custom(provider) })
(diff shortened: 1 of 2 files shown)


dbrowneup:
**Live prod verification (Safari, Circle passkey wallet, post-deploy c69ebd5b):**

Before this PR the panel dead-ended at "Connect your wallet to pay" with the wallet chip connected. After deploy, same tab, hard reload:

> Payment required. Generating costs \$2.000000 USDC on arcTestnet to 0xffa7…56c1 — **paid from your connected wallet (0x512b…9596)**.
> Payments run from your Circle Gateway balance… **Your balance: \$0.00.**
> [Pay \$2.00 & generate →] First payment? A one-time deposit of 20.00 USDC is included automatically

Clicking Pay ran fresh-402 → on-chain balance read → deposit user-op → WebAuthn, which surfaced verbatim as `Failed to request credential.` (ox `SignFailedError`) — expected under an automated (synthetic) click, since WebAuthn requires genuine user activation. Field test with a physical click is the remaining step; if a real click also fails, the suspect is transient-activation expiry across the two network hops between click and `credentials.get()`, and the fix is a second-gesture pattern for the passkey leg.

PR #1482, opened 2026-08-24
[infra] Pin PREMIUM_MODELS_ENABLED=false; fix a comment that contradicted its own value

onder-akkaya:
Two values in the backend task definition's env block, both following the money-switch pinning #1239 established. **Behaviourally a no-op today** — see Verification.

Raised from the #1442 review thread.

## 1. `PREMIUM_MODELS_ENABLED` was unset here

So prod's answer was whatever `model_gate.py:56` defaults to when the var is absent — an accident of a code default rather than a decision. That is the exact class of bug the `PAYMENTS_DRY_RUN` / `PAPER_TRADING` block a few lines above was pinned to close: *"an unset money switch is the bug regardless of which way it points."* This is the entitlement half of the same pair.

Pinned to `"false"`, which is the value prod already has in effect.

**Why it's worth pinning now rather than whenever.** `a45e8e5` ("THE FLIP", 2026-08-20) set `GENERATION_PAYMENT_REQUIRED = "true"` with `GENERATION_PAYMENTS_DRY_RUN = "false"`, so the generation rail settles for real at `GENERATION_PRICE_USD = "2.00"`. Meanwhile an entitled premium request still isn't servable — premium ids aren't in `FREE_TIER_MODELS`, so `generate_routes.py:335` falls back to the env default. The code is honest about this (`:329-334` explains it, `:337` logs it, roadmap T3.8/Bedrock).

Put those together and this flag is one env var from a charge-for-X-deliver-Y: flip it to `"true"` before the serving path is fixed and an entitled caller pays $2 and receives the env default model. No code change and no gate between the two flags would stand in the way. The rationale comment now sits at the flag, so the next person reaching for it sees the precondition rather than rediscovering it.

## 2. A comment that documented the opposite of its own value

```
# Generation payment gate (flip-list #834): flag stays "false" until
# Dan flips it deliberately …
{ name = "GENERATION_PAYMENT_REQUIRED", value = "true" },
```

Dan flipped it in `a45e8e5`; the comment was never updated. Rewritten to record the flip and its commit, keep the `GENERATION_PAYMENT_RECIPIENT` precondition (flag-on with no recipient is a deliberate 503), and keep the don't-drive-by-flip warning — now in both directions, since rolling it back matters as much as flipping it.

## Verification

There is no `terraform`/`tofu` in this environment, so I could not run `fmt -check`, `validate`, or `plan`. Validated textually instead, and stating that plainly rather than implying a plan was run:

- Braces `186/186` and brackets `48/48` balanced.
- Env entries `53 → 54` — exactly one added, no accidental deletion.
- No new duplicate env names. (`DATABASE_URL` appears twice both before and after — pre-existing, two containers in the task def.)
- Formatting matched to the surrounding single-line object entries; `git diff --check` clean.

**A `terraform plan` before apply would be the real check** — expected diff is one added environment entry on `aws_ecs_task_definition.backend` plus comment-only churn, i.e. a new task-definition revision with no behavioural delta.

## Cross-lane note

Infra is Dan's lane, so this wants his read even at two values — flagging rather than assuming, per the lead/coverage table. It touches a task definition on a live payment path, which is exactly where I'd rather over-communicate.

## Not fixed here

`PREMIUM_MODELS_ENABLED` is also unset in all three compose files. Same gap `PAYMENTS_DRY_RUN` still has after #1239 pinned only the `ecs.tf` site — worth one sweep across both flags and all four files rather than piecemeal, so I left it out of a two-value PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mq4GamnxERxpp2qdnhDMYD)_


onder-akkaya:
One thing about the green checks on this PR, since it's the shape `CLAUDE.md` § *the green check may not mean what you think* warns about:

**Nothing in CI validates Terraform.** `grep -rn "terraform\|tofu" .github/workflows/*.yml` returns exactly one hit, and it's a warning *string* inside `deploy.yml:760` about a grant needing `terraform apply` — not a step. `Infra — user-data size guard` (✅ on this PR) runs only `infra/scripts/check-user-data-size.sh`, a 16 KB limit check on EC2 user-data; it never parses a `.tf` file.

So the full green row here says the Python suite, ruff, import-guard and the frontend build are unaffected — which is true and is most of what a reviewer wants to know for a two-value env change. It says **nothing** about whether `infra/ecs.tf` still parses. I had no `terraform`/`tofu` in my environment either, so this change has been validated by textual checks only (braces/brackets balanced, env entries 53 → 54, no new duplicate names, formatting matched to neighbours).

**Please run `terraform plan` in `infra/` before applying.** Expected: one added `environment` entry on `aws_ecs_task_definition.backend`, plus comment-only churn — a new task-definition revision with no behavioural delta. If the plan shows anything else, that's the signal.

Separately, this looks like a genuine CI gap rather than a quirk of this PR — `infra/*.tf` changes currently merge with no syntax validation whatsoever. Happy to open an issue for a `terraform fmt -check` + `validate` job (path-filtered to `infra/**`, with the same non-required-check caveat as `contracts-test.yml` and `docs-gate.yml`) if you think it's worth having.

---
_Generated by [Claude Code](https://claude.ai/code)_

onder-akkaya:
Filed the CI gap as #1483 — path-filtered `fmt -check` + `init -backend=false` + `validate`, no credentials and no state access, with the not-a-required-check caveat that `contracts-test.yml` and `docs-gate.yml` already carry.

It's shaped as a proposal rather than a decision: the shape questions (does `fmt -check` block or just report, does the provider lockfile get pinned) are yours.

This PR still wants a `terraform plan` before apply regardless — #1483 would catch a syntax error, not a wrong value.

---
_Generated by [Claude Code](https://claude.ai/code)_

dbrowneup (commented):
Reviewed as the infra-lane read you asked for. The reasoning here is correct and I verified the whole chain independently — the charge-for-X-deliver-Y hazard is real, and pinning the flag is the right call. Three things to change before this lands, one of which the main-merge introduced after you wrote it.

## Verified correct (so you don't have to re-derive it)

- `model_gate.py:56` truthiness is `("1","true","yes","on")`, so "unset/anything-but-truthy = off" (line 566) is accurate.
- "the value prod already has in effect" is right: `PREMIUM_MODELS_ENABLED` appears nowhere in `infra/`, and nothing in `infra/scripts/*.sh` seeds it into SSM. Prod reads the code default today. The pin is a genuine no-op.
- The hazard trace holds end to end. Payment settles at `generate_routes.py:478` (`_paywall_with_credit`), entitlement runs after it at `:487`, an entitled premium id then fails `is_allowed_model` (premium ids deliberately absent from `FREE_TIER_MODELS`, `llm_backend.py:47-63`), so `selected_model = None` → env default. Charged $2, served something else. Correct, and worth pinning for exactly the reason you give.
- The `GENERATION_PAYMENT_REQUIRED` comment fix is right — `a45e8e5` did flip it and the old text did assert the opposite of its own value.
- Your new comment does not disturb `TestTheCommentsTellTheTruth` in `backend/tests/test_ecs_backend_secrets.py`; its anchored regions are lines 41-53 and 638-647, and the insertion sits at 579, outside both. Confirmed rather than assumed.

## 1. The new comment's line references went stale in the main-merge (blocking)

Lines 574-575 cite `generate_routes.py:335` for the fallback and `:329-334` for the disclosure. Those were **exactly right when you wrote them** — at the base of `b3fec15`, `:329-334` was the disclosure comment, `:335` was `selected_model = ...`, `:337` was the log. I checked.

They are wrong at the current head (`08cadec`, after the main merge). Those lines are now the `generation_credits.void(credit_id)` payment block; the free-tier selection moved to `:492-501`.

So the half of this PR that fixes a comment which contradicted its own code currently ships a comment that misdirects the reader by ~164 lines — the same failure mode, one merge later. That's the argument for citing symbols instead of lines, since symbols survive a rebase:

- `generate_routes.py:335` → the `selected_model = req.model if is_allowed_model(req.model) else None` line (free-tier selection, defense-in-depth)
- `:329-334` → the comment directly above that line
- `services/model_gate.py:56` (line 565) is still correct today, but same fragility — `model_gate._premium_globally_enabled` reads stably.

## 2. The pin doesn't actually close the hazard it documents (blocking)

`is_entitled_to_premium` (`model_gate.py:68-85`) is:

    wallet AND (_premium_globally_enabled() OR wallet in PREMIUM_MODELS_ALLOWLIST)

It's an **OR**. Adding one wallet to `PREMIUM_MODELS_ALLOWLIST` reaches the identical code path and produces the identical charge-$2-deliver-env-default outcome, with `PREMIUM_MODELS_ENABLED` still `"false"`. And `PREMIUM_MODELS_ALLOWLIST` appears nowhere in `infra/` or the SSM scripts either — it is unpinned in precisely the way this PR's own thesis condemns.

As written, lines 572-578 read as though this flag is *the* switch, so the next person trusts a door that has two locks and only one of them thrown. Suggest pinning both in the same block and saying so:

    { name = "PREMIUM_MODELS_ENABLED", value = "false" },
    { name = "PREMIUM_MODELS_ALLOWLIST", value = "" },

with the comment noting both halves of the entitlement OR must stay off.

## 3. Nothing enforces the pin — and the machinery already exists

`backend/tests/test_ecs_backend_secrets.py` already parses this exact block: `_container_block("backend")` → `_block(..., "environment")` → `_entries()`, exposed as the `backend_environment` fixture. `TestAntiGoals` already pins `REVENUE_SWEEP_ENABLED` *out* of that block for the same owner-gated reason. Your added entry matches `_ENTRY_RE` unchanged, so the guard is a few lines and needs no parser work:

    def test_premium_tier_stays_off(backend_environment):
        assert backend_environment.get("PREMIUM_MODELS_ENABLED") == "false"
        assert backend_environment.get("PREMIUM_MODELS_ALLOWLIST", "") == ""

Please run the adversarial pass before pushing — flip the value to `"true"` in `ecs.tf`, confirm the test fails, flip it back. A guard that has never rejected anything isn't known to be a guard.

This also sidesteps #1483 for this particular value: a Python test asserting on `ecs.tf` text needs no `terraform` binary, so the pin gets CI coverage today rather than after the fmt/validate job lands. Worth noting in #1483 that the text-assertion pattern already exists and covers *values*, which `validate` never would.

Right now line 578 says "do not flip this to unblock a demo" — and demo night is exactly when nobody re-reads a comment. That sentence wants to be a test.

## 4. Minor: `.env.example` belongs in "Not fixed here"

The PR body names the three compose files, but `.env.example` also lacks `PREMIUM_MODELS_ENABLED` entirely, while `PAYMENTS_DRY_RUN` gets a documented fail-safe block at `.env.example:215-218`. Worth adding to that sweep list so it isn't lost.

## Merge readiness

- **Draft**, so it can't land as-is.
- **174 commits behind `origin/main`** (`e152b74`) — GitHub still reports `MERGEABLE`/`CLEAN`, but that staleness is what invalidated the line refs in finding 1, so please merge main in again and re-check those citations after (or adopt symbol references and stop paying this tax).
- CI fully green on the head: backend units, ruff, import-guard, frontend, CodeQL all pass. The earlier quality-gate comment showing 41 ruff errors is stale.
- Your note that nothing in CI parses `.tf` is correct and matches what I found — `Infra — user-data size guard` runs only `check-user-data-size.sh`. Good catch, and #1483 is the right home for it.

Agreed on the cross-lane flag; this was worth the read. The analysis is sound — it's the citations and the missing second lock that need the pass.

danielscoffee (commented):
Stay draft. Not merge-ready. Do not APPROVE.

The pin itself is the right no-op vs today’s code default: `model_gate._premium_globally_enabled` is `os.getenv("PREMIUM_MODELS_ENABLED", "").strip().lower() in ("1", "true", "yes", "on")`, so unset = off. Pinning `PREMIUM_MODELS_ENABLED="false"` in `infra/ecs.tf` matches that default and the #1239 money-switch pinning convention. The `GENERATION_PAYMENT_REQUIRED` comment rewrite (flip-list #834, flipped in a45e8e5) is the right fix for a comment that contradicted its own value.

Do not treat this as landable while it is still a GitHub draft, and the 2026-08-31 review on `08cadec` is still open:

1. Comment line-refs `generate_routes.py:335` / `:329-334` went stale after the main merge (same failure mode this PR is fixing one comment up). Cite symbols (`_premium_globally_enabled`, free-tier `selected_model` / `is_allowed_model`), not lines.
2. Entitlement is an OR: wallet AND (`_premium_globally_enabled()` OR wallet in `PREMIUM_MODELS_ALLOWLIST`). Pinning only the global flag does not close the charge-$2-deliver-env-default hazard. Pin `PREMIUM_MODELS_ALLOWLIST=""` in the same block and say so.
3. Nothing enforces the pin. `backend/tests/test_ecs_backend_secrets.py` already parses this env block (`TestAntiGoals` / `backend_environment`). Add a guard that the two keys stay off, and show it reject a flip.
4. Re-merge main (this branch was already far behind when last reviewed) and re-check citations. No `terraform plan` has been run; CI does not parse `.tf`.

CI on the last head was green. That does not make a draft merge-ready.

Issue #1463, opened 2026-08-21
APIN - Infra - revenue_sweep cannot run in the web tier: Circle creds absent from ecs.tf backend task-def


dbrowneup:
## What was found (2026-08-21, one-off ECS run-task verification)

`python -m archimedes.services.revenue_sweep --check` run as a one-off Fargate task against BOTH task definitions fails:

- `archimedes-migrate`: `GENERATION_PAYMENT_RECIPIENT` / `REVENUE_WALLET_ID` unset (expected — migrate carries no app env).
- `archimedes-backend:118`: env vars resolve, then `circlekit/wallets.py` raises `ValueError: api_key is required. Pass it directly or set CIRCLE_API_KEY env var.`

The backend container definition has **no `secrets` block at all** — `CIRCLE_API_KEY` / `CIRCLE_ENTITY_SECRET` exist in SSM (`/archimedes/prod/CIRCLE_API_KEY`, `/archimedes/prod/CIRCLE_ENTITY_SECRET`) but are not injected. The paywall settle path doesn't need them (facilitator), which is why generation payments work regardless — but the sweep's Circle-wallet withdraw does.

Consequence: flipping `REVENUE_SWEEP_ENABLED=true` (#1457's scheduler) today would crash the sweep loop on first run. The revenue-collection demo is blocked on this too.

## Acceptance

1. `infra/ecs.tf` backend container gains `secrets` entries mapping `CIRCLE_API_KEY` and `CIRCLE_ENTITY_SECRET` from the SSM parameters above; `terraform apply` rolls the service.
2. One-off verification: `aws ecs run-task` (backend task-def, command override `python -m archimedes.services.revenue_sweep --check`) exits 0 and prints the Gateway balance line.
3. Anti-goal: do NOT put the key in plain `environment`; do NOT enable the scheduler in the same PR.

## Context

On-chain balance today (read-only, no creds): `cast call 0x0077777d7EBA4688BDeF3E311b846F25870A19B9 "availableBalance(address,address)(uint256)" 0x3600…0000 0xffa7…56c1` → `6000000` = **$6.00 USDC** — below the $10 `REVENUE_SWEEP_MIN_USDC` default, so no sweep is due yet anyway.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Hqgq6BzkRzuDrUL34xqZj7

PR #1487, opened 2026-08-27
[payments] Hold back Circle's fee when sweeping a Gateway balance (#1486)

### LINKED ISSUE #1486: [payments] Gateway sweep asks for the full balance; Circle adds its fee on top → every withdrawal 400s
## What happened

The first real platform revenue sweep failed in production, 2026-08-26:

```
POST https://gateway-api-testnet.circle.com/v1/transfer "HTTP/1.1 400 Bad Request"
ValueError: Withdrawal failed (status 400): {"success":false,
  "message":"Insufficient balance for depositor 0xffa7abba…56c1:
             available 36.000000, required 36.0035"}
```

No funds moved — the 400 comes back before any attestation is issued — but the
$36.00 of settled generation revenue could not be collected.

## Root cause

**Circle charges the Gateway withdrawal fee ON TOP of the burn amount**, so the
real constraint is `amount + fee <= available`. Both sweep paths asked for the
entire available balance, which can never satisfy it:

- `backend/archimedes/services/revenue_sweep.py` — `amount = balances.formatted_available`
- `backend/archimedes/marketplace/settlement.py` (Stage A) — the identical line

So this is not a revenue-only bug. **Marketplace settlement Stage A has the same
defect** and would fail the same way on its first real withdrawal.

A second, quieter problem sits behind it: neither call site passed `max_fee`, so
both inherited circlekit's `DEFAULT_WITHDRAW_MAX_FEE = 2_010_000`. That
authorises Circle to take **2.01 USDC — 5.6% of a $36 sweep** — without
complaint. The "~2.01 USDC withdrawal fee" comments in both modules had mistaken
that *cap* for the fee; the fee actually charged on this withdrawal was
**0.0035**.

## Why tests did not catch it

Both suites asserted against a bare `AsyncMock` for `withdraw`, which accepts any
argument and therefore certifies whatever the code does:

```python
client.withdraw.assert_awaited_once_with(amount="12.500000")   # revenue
instance.withdraw.assert_awaited_once_with(amount="15.00")     # settlement
```

This is repo rule 3 — *a test that passes against the unfixed code proves
nothing* — in its purest form: the mock could not fail, so it locked the bug in.

## Fix

PR #1489. Shared helper in `marketplace/config.py` computing
`available - reserve`, used by both call sites, with the reserve also passed as
`max_fee` so Circle can never charge more than was set aside. New
`tests/gateway_fake.py` models Circle's real `amount + fee <= available` rule, so
the old code provably fails.

## Measured context (Arc testnet, 2026-08-26)

- Fee on the 36.000000 withdrawal: **0.0035 USDC**
- Comparable Gateway ops: `submitBatch` 132,245 / 177,075 gas, `gatewayBurn`
  162,667, at 26–35 gwei
- Because USDC is the native gas token on Arc, the DCW's `gatewayMint` costs
  **~$0.011** — the revenue DCW needed funding first (it had nonce 0)

Runbook: `operations/runbooks/revenue-sweep.md` in the private docs repo.

dbrowneup:
Closes #1486.

## The bug, in production

The first real platform revenue sweep failed tonight:

```
POST https://gateway-api-testnet.circle.com/v1/transfer "HTTP/1.1 400 Bad Request"
ValueError: Withdrawal failed (status 400): {"success":false,
  "message":"Insufficient balance for depositor 0xffa7abba…56c1:
             available 36.000000, required 36.0035"}
```

**Circle charges its withdrawal fee on top of the burn amount.** The real
constraint is `amount + fee <= available`, and both sweep paths asked for the
entire available balance — which can never satisfy it.

It is not a revenue-only bug. `marketplace/settlement.py` Stage A carries the
same line and would have failed identically on its first real withdrawal.

Behind it sits a quieter problem: neither call site passed `max_fee`, so both
inherited circlekit's `DEFAULT_WITHDRAW_MAX_FEE = 2_010_000` — authorising
Circle to take **2.01 USDC, 5.6% of a $36 sweep**, without complaint. The
"~2.01 USDC withdrawal fee" comments in both modules had mistaken that *cap*
for the fee. The fee actually charged was **0.0035**. Comments corrected.

## Fix

`gateway_sweep_amount()` in `marketplace/config.py` — one helper, both call
sites, so the two paths cannot drift apart again:

```python
amount, fee_reserve_raw = gateway_sweep_amount(balances.available)
result = await client.withdraw(amount=amount, max_fee=fee_reserve_raw)
```

Passing the reserve as `max_fee` is the half that isn't obvious: it means an
out-of-band fee **fails loudly instead of being paid quietly out of revenue**.

Reserve defaults to **0.05 USDC** — 14x the observed fee, under 0.2% of a $36
sweep — overridable via `GATEWAY_WITHDRAW_FEE_RESERVE_USDC`. No input can drive
it to zero (garbage, empty, negative, and `0` all fall back), because a zero
reserve reproduces exactly this bug.

## Why CI was green through all of it

Both suites asserted `withdraw` against a bare `AsyncMock`, which accepts any
argument and therefore certifies whatever the code happens to do:

```python
client.withdraw.assert_awaited_once_with(amount="12.500000")   # revenue
instance.withdraw.assert_awaited_once_with(amount="15.00")     # settlement
```

That is repo rule 3 in its purest form — *a test that passes against the unfixed
code proves nothing*. The mock could not fail, so it locked the bug in and made
it look covered.

`tests/gateway_fake.py` replaces it with a fake that enforces Circle's actual
rule and reproduces its rejection messages, including the `maxFee` one.

## Guard demonstration (repo rule 4)

Fix stashed (call sites only), tests re-run against the old sources:

```
FAILED test_revenue_sweep.py::test_at_threshold_withdraws_available_less_fee_reserve
  ValueError: … "available 12.500000, required 12.5035"
FAILED test_revenue_sweep.py::test_min_override
  ValueError: … "available 4.000000, required 4.0035"
FAILED test_revenue_sweep.py::test_balance_under_the_reserve_is_not_swept
  ValueError: … "available 0.040000, required 0.0435"
FAILED test_settlement.py::test_stage_a_above_threshold_withdraws_less_fee_reserve
  AssertionError: assert '15.000000' == '14.950000'
FAILED test_revenue_sweep.py::test_reserve_bounds_what_circle_may_charge
  ValueError: … "available 36.000000, required 38"    ← the unbounded maxFee

5 failed, 39 passed
```

Restored: **44 passed**, and identically under `env -i HOME=$HOME PATH=$PATH
PYTHONPATH=backend` (hermetic gate).

`test_requesting_the_whole_balance_is_what_circle_rejects` pins the failure mode
itself, so if the fake ever goes slack the guard demonstration fails loudly
rather than silently becoming decorative.

## Measured on Arc testnet, 2026-08-26

| | |
|---|---|
| Fee on the 36.000000 withdrawal | **0.0035 USDC** |
| Comparable Gateway ops | `submitBatch` 132,245 / 177,075 gas · `gatewayBurn` 162,667 |
| Effective gas price | 26–35 gwei |
| Cost of the DCW's `gatewayMint` | **~$0.011** (USDC is the native gas token on Arc) |

## Not in this PR

`REVENUE_SWEEP_ENABLED` stays off, and #1463 (Circle creds absent from
`ecs.tf`) is untouched — the scheduler should not arm until a manual sweep has
succeeded once.

### DIFF
--- a/backend/archimedes/marketplace/config.py
+++ b/backend/archimedes/marketplace/config.py
@@ -4,7 +4,13 @@ Single source of truth for chain-name defaults so that payment charging,
 Gateway withdrawal, and settlement all read from the same constant.
 """
 
+from __future__ import annotations
+
+import logging
 import os
+from decimal import Decimal
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_GATEWAY_CHAIN = "arcTestnet"
 
@@ -28,3 +34,77 @@ def gateway_chain() -> str:
     chain is to ask for it.
     """
     return os.getenv("GATEWAY_CHAIN", DEFAULT_GATEWAY_CHAIN).strip()
+
+
+_USDC = Decimal(10) ** 6
+
+# Circle's Gateway withdrawal fee is charged ON TOP of the burn amount, not
+# deducted from it. Asking to withdraw the entire available balance therefore
+# always fails — observed in prod 2026-08-26 sweeping the revenue DCW:
+#
+#   POST /v1/transfer -> 400
+#   "Insufficient balance for depositor 0xffa7abba...56c1:
+#    available 36.000000, required 36.0035"
+#
+# So a full sweep must hold back at least the fee (0.0035 USDC on that
+# withdrawal). We reserve 0.05 — ~14x observed headroom, and still under two
+# thousandths of a $36 sweep — and pass the same number as the burn intent's
+# ``maxFee``. That second half matters: circlekit's default maxFee is
+# 2_010_000 (2.01 USDC), so an unbounded withdrawal authorises Circle to take
+# a 5.6% haircut off a $36 sweep without complaint. Bounding it means an
+# unexpectedly large fee fails loudly instead of being paid silently.
+DEFAULT_GATEWAY_FEE_RESERVE_USDC = "0.05"
+
+
+def gateway_fee_reserve_raw() -> int:
+    """Raw 6-decimal USDC held back from a Gateway withdrawal to cover Circle's
+    fee. Override via ``GATEWAY_WITHDRAW_FEE_RESERVE_USDC`` if Circle's fee
+    schedule moves; a bad value falls back to the default rather than risking
+    a zero reserve (which is the bug this exists to prevent)."""
+    raw = os.getenv("GATEWAY_WITHDRAW_FEE_RESERVE_USDC", "").strip()
+    if not raw:
+        raw = DEFAULT_GATEWAY_FEE_RESERVE_USDC
+    try:
+        value = Decimal(raw)
+    except Exception:
+        logger.warning(
+            "invalid GATEWAY_WITHDRAW_FEE_RESERVE_USDC=%r — using %s",
+            raw,
+            DEFAULT_GATEWAY_FEE_RESERVE_USDC,
+        )
+        value = Decimal(DEFAULT_GATEWAY_FEE_RESERVE_USDC)
+    if value <= 0:
+        logger.warning(
+            "GATEWAY_WITHDRAW_FEE_RESERVE_USDC=%r is not positive — using %s "
+            "(a zero reserve makes every full sweep fail)",
+            raw,
+            DEFAULT_GATEWAY_FEE_RESERVE_USDC,
+        )
+        value = Decimal(DEFAULT_GATEWAY_FEE_RESERVE_USDC)
+    return int(value * _USDC)
+
+
+def format_usdc(raw: int) -> str:
+    """Raw 6-decimal USDC → the fixed-point decimal string circlekit parses."""
+    return f"{Decimal(raw) / _USDC:.6f}"
+
+
+def gateway_sweep_amount(available_raw: int) -> tuple[str, int]:
+    """Plan a full sweep of ``available_raw``.
+
+    Returns ``(amount, fee_reserve_raw)`` where ``amount`` is the decimal
+    string to pass to ``GatewayClient.withdraw`` and ``fee_reserve_raw`` is
+    the ``max_fee`` to pass alongside it. ``amount + fee <= available`` holds
+    for any fee at or under the reserve, which is what Circle actually checks.
+
+    Raises ``ValueError`` when the balance cannot cover the reserve — callers
+    treat that as "nothing to sweep", never as an amount to clamp to zero.
+    """
+    reserve = gateway_fee_reserve_raw()
+    amount_raw = available_raw - reserve
+    if amount_raw <= 0:
+        raise ValueError(
+            f"available {format_usdc(available_raw)} USDC does not cover the "
+            f"{format_usdc(reserve)} withdrawal fee reserve"
+        )
+    return format_usdc(amount_raw), reserve
--- a/backend/archimedes/marketplace/settlement.py
+++ b/backend/archimedes/marketplace/settlement.py
@@ -1,7 +1,9 @@
 """Settlement sweep: Gateway balance → agent wallet → PaymentSplitter.depositToPool → withdraw.
 
 Three cadences:
-  Stage A — Gateway → wallet (threshold-based, ~2.01 USDC fee per withdrawal).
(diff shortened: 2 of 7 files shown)


onder-akkaya:
Reviewed the arithmetic and ran the mutations. The core fix is right, the integer/Decimal handling on the production path is exact, and `tests/gateway_fake.py` is the first test in either suite that could actually have caught #1486 — the `AsyncMock` it replaces certified the bug. Two things I'd want changed before this merges, plus a rebase.

**1. `config.py:62` — the zero-reserve guard checks the wrong value.**

The guard is `value <= 0` on the Decimal, but the reserve that gets used is `int(value * _USDC)`, and `int()` truncates. So a positive env value can still produce a zero reserve:

```
GATEWAY_WITHDRAW_FEE_RESERVE_USDC=0.0000005  -> reserve 0
GATEWAY_WITHDRAW_FEE_RESERVE_USDC=1e-9       -> reserve 0
```

That is #1486 reproduced, silently, with nothing logged. The docstring at `config.py:38-41` and the PR body both say no input can drive it to zero — right now the code doesn't enforce that. Check `int(value * _USDC) <= 0` instead of `value <= 0`.

Worth noting `test_bad_values_fall_back_never_to_zero` is blind to this: it covers `-1`, `0`, `0.0` and dies if you remove the `value <= 0` block, but there's no sub-microdollar case in it.

**2. `config.py:45-47` — the try only wraps `Decimal(raw)`.**

`"nan"` parses fine and then raises `InvalidOperation` on the `value <= 0` comparison; `"Infinity"` raises `OverflowError` on the `int()`. Both escape. `InvalidOperation` is an `ArithmeticError`, so it also slips past the `except ValueError` in `revenue_sweep.py:130` and out of `sweep_revenue`. Move the comparison and the conversion inside the try.

**3. Rebase before merging.** 71 commits behind and `mergeable: CONFLICTING` — main refactored `DEFAULT_GATEWAY_CHAIN` -> `gateway_chain()` (#1495) across the same import lines in all three files you touch. The green check on this PR describes a base that no longer exists, and re-running it replays the same frozen ref. Push the rebase and read the new result.

**On the flat-fee assumption** — not blocking, but worth an issue. 0.05 rests on one observation (0.0035 on a $36 sweep) and nothing here establishes the fee is flat rather than proportional; at a $1,000 sweep a proportional fee would blow through it. I checked and circlekit exposes no fee-discovery API, so hard-coding is the only option available. The saving grace is that `max_fee` makes the failure loud — Circle rejects rather than us silently overpaying — which is the right direction. But there's a log line and no alarm.

**Smaller stuff:**
- `test_gateway_fee_reserve.py:52` uses `int(float(amount) * 10**6)` — float on a token amount, in the test for a PR about exact money arithmetic. Exact for these values, won't be in general.
- `test_gateway_fee_reserve.py:57`: `assert available_raw + observed_fee_raw > available_raw` is a tautology.
- `GATEWAY_WITHDRAW_FEE_RESERVE_USDC` isn't in `.env.example`.
- Ruff informational: `I001` at `test_settlement.py:6`, `SIM117` at `test_revenue_sweep.py:118`.

Things I checked that are fine: `max_fee` is a real circlekit kwarg (`client.py:490`); dry-run is intact and `sweep_enabled()` at `revenue_sweep.py:161` tests the literal value `== "true"` rather than presence, which is the right shape given the `AGENT_DRY_RUN` history. Rounding direction on the amount is safe — integer subtraction, exact 6-dp formatting, dust left behind and recovered next sweep. Baseline 44 passed locally.

One gap in coverage: swapping `int()` for `round()` in the reserve parse fails no test, so the rounding direction is unpinned.

dbrowneup:
Rebased onto current main (`9dd1a708`, 82 commits). The conflict was with #1497's `gateway_chain()` centralisation in the same three files — resolved by keeping both: the chain now comes from `gateway_chain()` (Önder's #1495 fix) and the amount/max_fee from `gateway_sweep_amount()` (this PR). Touched suites: 44 passed post-rebase; format + lint clean. Note for reviewers: the fee-on-top bug is still live on main — the successful $35.95 sweep on 08-26 ran from this branch, not main.

Issue #1481, opened 2026-08-24
[rigor][claim-integrity] POST /api/rigor/verify returns passes:true on 4 bars — 1 of 4 legs decides the verdict, and the CLI exits 0 on it


onder-akkaya:
## Summary

`POST /api/rigor/verify` computes its top-level `passes` boolean over only the legs that turned out to be *evaluable*, and two of its four legs are hard-coded `not_evaluable` on every single request. On a short series only DSR is evaluable, so **one leg of four decides the verdict** — and `archimedes verify` exits `0` on that boolean, which is what a script or CI branches on.

The response body is honest: every leg carries its own status and reason, and the all-legs-unevaluable case is explicitly guarded. The **scalar and the exit code** are not.

This matters more than an internal helper would, because the endpoint is advertised to autonomous agents in all three discovery surfaces: `backend/archimedes/api/agent_manifest_routes.py:170`, `ui/public/llms.txt:45`, `ui/public/.well-known/agent.json:91`. An agent that reads `passes` and nothing else gets an unqualified "cleared the rigor gate."

## Mechanism (verified at `baa173a`)

```
backend/archimedes/api/rigor_verify_routes.py
:199  pbo_check        = RigorCheck(status="not_evaluable", reason=_PBO_NOT_EVALUABLE_REASON)        # always
:200  look_ahead_check = RigorCheck(status="not_evaluable", reason=_LOOK_AHEAD_NOT_EVALUABLE_REASON) # always
:203  evaluable = [s for s in statuses if s != "not_evaluable"]
:204  passes = bool(evaluable) and all(s == "pass" for s in evaluable)
```

Evaluability floors of the two legs that *can* run:

| Leg | Becomes evaluable at | Gate |
|---|---|---|
| DSR | **≥ 4 bars** with nonzero variance (`:137`) | `DSR_P_FLOOR = 0.50` (`rigor_profiles.py:37`) |
| OOS consistency | ≥ 10 bars total **and ≥ 21 OOS bars** (`:164`) — so ~70 bars at the 70/30 split | `OOS_ABS_FLOOR = 0.0` (`rigor_profiles.py:40`) |
| PBO | never | — |
| look-ahead | never | — |

So for roughly `4 ≤ n_bars < 70`, `evaluable` has exactly one element. A DSR p-value ≥ 0.50 on a 4-bar series returns:

```json
[...]

Issue #1305, opened 2026-08-20
CLI 0.1.0: login, meter, verify against the hosted API (+ usage & rigor-verify endpoints)


dbrowneup:
## Summary

`cli/` is a 0.0.1 name-reservation: the command tree and flags are settled but every subcommand exits `NOT_IMPLEMENTED`. Nothing tracks building it, and the CLI is an MVP deliverable. This issue is the 0.1.0 the stub's own README promises: `login`, `meter`, and `verify` working against the hosted API.

## Scope

**Backend (two small endpoints, account-session-gated via `require_current_user`):**
- `GET /api/account/usage` — the caller's generation count today vs the two daily caps (reuse `services/generation_quota.py` bucket reads; do NOT change enforcement), plus the current quote from `generation_payment.quote()` so the CLI's number and the API's number come from one place.
- `POST /api/rigor/verify` — body `{returns: [{date, daily_return}], trials?: int>=1 (default 1, self-attested)}`. Reuses the SAME gate functions the passport verdict uses (no reimplementation, no new thresholds). Rate-limit `5/minute`.

**Honesty contract for the verify verdict (load-bearing):** a single returns series can support DSR (deflated by the *declared* trials count, labeled self-attested) and the walk-forward OOS consistency check. PBO requires a trial matrix and the look-ahead audit requires strategy code — for a bare series they MUST be reported as `not_evaluable` with a reason, never silently passed or defaulted. Response shape: all four checks, each `{status: pass|fail|not_evaluable, reason?}` + the numbers. `passes` is true only if no evaluable check failed AND at least one check was evaluable.

**CLI (`cli/src/archimedes_cli/`):**
- `login` — the account model is Better Auth (#1194), NOT the wallet-signature flow the 0.0.1 docstring sketches: prompt email + password (or read `ARCHIMEDES_EMAIL`/`ARCHIMEDES_PASSWORD` for CI), `POST /api/auth/sign-in/email`, store the session cookie at `~/.config/archimedes/session.json` chmod 600. `--json` reports `{ok, email}`. Update the docstring.
- `meter` — GET `/api/account/usage` with the stored cookie; human table + `--json` passthrough; exit 2 (`AUTH`) when there is no valid session.
[...]

PR #1492, opened 2026-08-28
[docs] Retire four sprint-card findings that landed in the five days since #1453

onder-akkaya:
Follow-up to #1453 (merged 2026-08-28). Four of the findings that PR recorded were fixed within a day of it landing, so the doc on `main` currently asserts things that are no longer true. A doc whose job is cataloguing other people's stale claims doesn't get an exemption from the first rule.

Docs-only, one file.

## Closed since #1453 merged

Each re-verified at `b49609a` with a literal command, not taken from a commit message:

| Finding | Closed by | Verified |
|---|---|---|
| `/api/rigor/verify` passes on four bars | #1481 → **#1484** (`e8a0644`) | `passes = legs_evaluated == len(_RUNNABLE_LEGS) and all(...)` — a leg that *could not run* now makes the verdict `false` instead of invisible |
| cluster-2 A4 (fabricated `date(2004, 1, 2)`) | **#1485** (`2f1517e`) | `grep "date(2004, 1, 2)" fusion_evaluator.py` → gone |
| cluster-2 A8 (sleeve label) | **#1485** | label now reaches the row across 5 files |
| `metrics_source` | **#1491** (`b49609a`) | present in 4 files — and it went past what the card asked, also naming the display-metric chain's `stub_placeholder` link |

The `/api/rigor/verify` item is **struck through rather than deleted**, because cluster-8's row still refers to it and a dangling reference is worse than a retired one.

**cluster-8 is narrowed, not cleared.** Of the two prohibitions #1305's endpoint breached, the `passes` scalar is fixed; **A1 is not** — gate compute still runs on the event loop rather than `asyncio.to_thread`.

## Still standing at `b49609a`

`GATEWAY_CHAIN` unguarded in `main.py` · the `arcTestnet` default · B5's downgrade (see #1482) · the look-ahead OR · `siwe.js`'s non-existence.

## The decay note now carries a count

Seven findings have closed across two waves — three during the pre-merge rebases, four since. The note says so, and says the thing that's actually useful to the next reader: **every correction so far has been a finding closing, not a finding turning out to be wrong.** That's a better signal about how to read this doc than any of the fractions.

## Verification

- `make docs-check` → links 0 broken (1184 scanned), index 143/143.
- Every retirement re-verified at `b49609a` by command; the still-standing list re-checked at the same SHA.
- `git diff --check` clean. No code touched.

## Note on #1442

#1453 rewrote the section #1442 inserts into, so **#1442's insertion anchor ("Path corrections vs the source doc") no longer exists on `main`** and it will need a rebase. That's a conflict my merge created, and I flagged the merge-order risk in my review of it beforehand — happy to do the rebase for @dbrowneup if that's easier than redoing it, since I know what moved.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mq4GamnxERxpp2qdnhDMYD)_


onder-akkaya:
Rebased onto current `main` (`54bd648f`) and taken out of draft.

The green checks on this PR were generated against `b49609a`, eighteen commits back, so I re-ran every claim in the body against the new base rather than trusting them. All four retirements still hold:

| claim | re-verified at `54bd648f` |
|---|---|
| fabricated `date(2004, 1, 2)` gone | `grep -c` → 0 |
| `/api/rigor/verify` quorum | `rigor_verify_routes.py:246` — `legs_evaluated == len(_RUNNABLE_LEGS) and all(...)` |
| `metrics_source` | present in 4 files |
| sleeve label reaches the row | present |

Cluster-8's A1 is still open, as the body says: `grep -c to_thread rigor_verify_routes.py` → 0.

## One change beyond a rebase

Re-checking the still-standing list turned up something the original grep couldn't see. The cluster-6 row said `GATEWAY_CHAIN` "silently falls back to `arcTestnet` in three places" — that count is correct, but there's a fourth site with a different shape:

```
backend/archimedes/services/revenue_sweep.py:88
    return GatewayClient(chain=DEFAULT_GATEWAY_CHAIN, signer=signer, tx_executor=executor)
```

It passes the default as a bare kwarg and never reads the env var, so it doesn't match a `getenv("GATEWAY_CHAIN"` grep. With `GATEWAY_CHAIN=arcMainnet` set and `GatewayClient` stubbed to capture its argument:

```
generation quote chain        = arcMainnet
revenue_sweep client chain    = arcTestnet
```

After cutover the paywall would collect on mainnet while the sweep looked for a balance on testnet — and it would fail quietly, since an empty testnet balance reads as "nothing to sweep yet". Filed as #1495, labelled `mainnet-gate`. The row now names it.

I've kept that in this PR rather than splitting it, because the PR's whole argument is that a doc cataloguing other people's stale claims doesn't get an exemption from the first rule. Leaving a known-incomplete line in place while retiring four others would be the same failure in miniature.

`make docs-check`: 0 broken links of 1185 scanned, index 143/143.

PR #1493, opened 2026-08-29
[deps] Align environment.yml floors with backend/requirements.txt, test it, and fold in the five open Dependabot bumps

onder-akkaya:
> **Updated 2026-08-30.** This PR now also carries the five open backend
> Dependabot bumps. Rationale and verification in the first section; the
> original alignment write-up follows unchanged below it.

## Why the five Dependabot PRs are folded in here

The guard this PR adds is an ordinary unit test, so it runs inside
`pytest -m "not integration"` — the **blocking** half of `quality-gate.yml`.
That makes the merge order load-bearing. Landing the guard first turns
#1509, #1513, #1514, #1515 and #1516 red on a required check, because each
raises a floor in `backend/requirements.txt` and nothing mirrors it. Dependabot
cannot make that mirroring edit itself, so the fix would have to be pushed onto
a branch Dependabot force-pushes over on its next rebase.

I confirmed the guard really does reject that state rather than assuming it.
Applying all five bumps to `requirements.txt` alone, on top of this branch:

```
E   AssertionError: environment.yml has drifted from backend/requirements.txt.
E   Local dev would resolve a different version than Docker/CI for:
E       alembic:   environment.yml >=1.19.0   vs  requirements.txt >=1.19.1
E       botocore:  environment.yml >=1.43.60  vs  requirements.txt >=1.43.81
E       starlette: environment.yml >=1.4.1    vs  requirements.txt >=1.6.0
E       uvicorn:   environment.yml >=0.52.3   vs  requirements.txt >=0.52.4
E       yfinance:  environment.yml >=1.5.2    vs  requirements.txt >=1.7.0
→ 2 failed, 1 passed
```

So the five are raised here in **both** files, and the guard passes by
construction rather than by exemption. One merge, one image rebuild, one deploy,
instead of five rebase-and-deploy cycles against #1346's build starvation.

| package | from | to | Dependabot PR |
|---|---|---|---|
| starlette | 1.4.1 | 1.6.0 | #1514 |
| uvicorn[standard] | 0.52.3 | 0.52.4 | #1509 |
| alembic | 1.19.0 | 1.19.1 | #1516 |
| yfinance | 1.5.2 | 1.7.0 | #1515 |
| botocore | 1.43.60 | 1.43.81 | #1513 |

**starlette is the only one with real surface area.** 1.5.0 through 1.6.0 change
`GZipMiddleware` defaults, tighten `FileResponse` range handling, and add
`max_body_size`. The backend imports neither `GZipMiddleware` nor `FileResponse`,
and compression is done in nginx (`nginx.conf:92-97`), so none of it is reachable
from our code. Recorded in the comment above the pin.

**Consolidating is also the re-verification.** All five were green, but against a
base 25 commits behind `main` — a result about a tree that no longer exists. One
CI run here tests every bump against current `main`.

### What I checked and rejected

Making `environment.yml`'s pip block just `- -r backend/requirements.txt` would
delete this whole class of problem. It does not work: `requirements.txt:99` is
`--extra-index-url https://download.pytorch.org/whl/cpu`, so `-r` would force the
CPU-only torch wheel onto dev machines and lose MPS on Apple Silicon. That
constraint is the actual reason the two files are separate.

The permanent fix is to split a `requirements-base.txt` out and have both files
include it, leaving the index URL and torch in the Docker-facing file only.
That touches the Dockerfile COPY and CI, which is not a change I would make
seventeen days from the mainnet gate. Worth its own issue.

### Verification on this head

- `pytest -m "not integration"` from the repo root, CI's command → **3860 passed, 2 skipped, 0 failed**
- Hermetic gate on the guard → 3 passed
- `ruff format --check` + `ruff check` on the new test → clean

Mutations re-run against this head, since the file changed since the original set:

| mutation | result |
|---|---|
| revert one new env floor (starlette 1.6.0 → 1.4.1) | **2 failed**, 1 passed |
| set an env floor *above* prod (starlette 99.0.0) | **1 failed**, 2 passed |
| break `_FLOOR` so the parser matches nothing | **1 failed**, 2 passed |

The second and third matter together. An env floor above production is merely
strict and only trips the identical-floors assertion; one below production is the
harmful direction and trips both. The parser mutation is the guard on the guard —
without it, a regex that silently stopped matching would make the whole
alignment assertion trivially true.

---

`environment.yml` (local dev) and `backend/requirements.txt` (Docker + CI) had drifted on **18 of the 34 packages they share** — every single one with the local floor *below* the deployed one.

I found this while merging this week's Dependabot batch. Dependabot only ever edits `backend/requirements.txt`, and nothing mirrors the change across, so the gap reopens on its own every Monday. CLAUDE.md already says the two files must stay aligned; nothing enforced it.

## It was live, not theoretical

Installed versions in the `archimedes` conda env against the production floor:

| package | installed locally | production floor | |
|---|---|---|---|
| pandas | 3.0.3 | 3.0.5 | behind |
| cryptography | 49.0.0 | 50.0.0 | behind |
| starlette | 1.3.1 | 1.4.1 | behind |
| fastapi | 0.140.6 | 0.141.1 | behind |
| uvicorn | 0.51.0 | 0.52.3 | behind |
| alembic | 1.18.4 | 1.19.0 | behind |
| boto3 | 1.43.56 | 1.43.71 | behind |
| redis | 8.0.1 | 8.1.0 | behind |
| yfinance | 1.4.1 | 1.5.2 | behind |
| sqlalchemy | 2.0.51 | 2.0.52 | behind |
| numpy | 2.5.1 | 2.5.2 | behind |

Eleven of thirteen checked. The `pandas` one is the one that would have bitten: the old floor was `>=2.2`, so a fresh `conda env create` could legitimately resolve pandas 2.x while production runs 3.x — a major version apart, on the code path that grades strategies.

`cryptography` and `starlette` are the security-relevant pair. Both are pinned in `requirements.txt` specifically to close CVEs; local dev was resolving underneath both floors.

## What changed

18 floors raised to match. Comment columns preserved, comment text untouched except one:

- `starlette`'s comment said *"Floor raised to 1.3.1 to close CVE-2026-54282 / CVE-2026-54283"* while the floor is now 1.4.1. Extended rather than replaced, so the CVE history stays readable.

Nine of the 18 lines carried a comment asserting they were *"aligned with backend/requirements.txt"* — `alembic`, `redis`, `numpy`, `anthropic`, `cryptography`, `boto3`, `botocore`, `arxiv`, `pypdf`. Each of those claims was false when I found it. They're true now, which is the part of this that isn't just hygiene.

## The guard

`backend/tests/test_env_requirements_alignment.py`, three tests, stdlib only — no PyYAML, no `packaging`. Neither is a declared backend dependency, and a guard that only runs when an undeclared transitive import happens to be installed isn't a guard.

Raising the floors without this would just reset the clock to next Monday.

## Shown to reject

| mutation | result |
|---|---|
| revert `environment.yml` to the drifted state | **2 failed**, 1 passed — and the message names all 18 with both floors |
| break `_FLOOR` so the parser matches nothing | **1 failed** — `environment.yml pip block parsed to only 0 deps` |
| set an env floor *above* prod (`pandas>=99.0.0`) | identical-floors **FAILS**, below-deployed **PASSES** |

Mutation 2 is the guard on the guard: without that vacuity test, a parser that silently stopped matching would make the alignment assertion trivially true. Mutation 3 exists to show the third test isn't a restatement of the second — an env floor above production is merely strict, one below it is the harmful direction, and only the strict case separates them.

## Verification

- `pytest backend/tests/test_env_requirements_alignment.py` → 3 passed.
- Hermetic gate: `env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest ...` → 3 passed.
- `ruff format --check` + `ruff check` clean.
- Full unit suite on this branch's base (`54bd648`, after the nine Dependabot merges) → 3719 passed, 2 skipped.

## Flagging for review

`environment.yml` is on CLAUDE.md's ask-first list, so I'm not self-merging this. Two things worth a second opinion:

1. **`pandas>=2.2` → `>=3.0.5` is the one with teeth.** Anyone whose env predates the production bump will get a real upgrade on their next `conda env update`, and pandas 3 has breaking changes. My read is that this is correct anyway — production already runs pandas 3, so an env on 2.x is running code prod doesn't — but it's a real disruption and worth knowing about before it lands rather than after.
2. **The three env-only floors I did *not* touch**: `requests>=2.32,<2.34` and `urllib3>=2.7.0` already match, and `torch` matches. Dev-only tooling (`pytest`, `ruff`, `mypy`, `jupyter`, `fakeredis`, …) has no counterpart in `requirements.txt` and is correctly excluded from the comparison.

Related: I closed #1478 in the same pass. Dependabot kept rewriting the torch floor to `>=2.13.0+cpu`, which is invalid PEP 440 — a local version label can't sit on `>=`. A follow-up PR adds the `ignore` entry so it stops regenerating.


onder-akkaya:
Filed #1522 for the structural follow-up.

This PR polices the drift; #1522 removes it. `backend/requirements.txt:99`'s `--extra-index-url` is what stops `environment.yml` from just including the file, so the 32 shared floors get restated by hand and the same `grep -vE` that strips the torch block is copied twice into `quality-gate.yml` (:134 and :313). Splitting a `requirements-base.txt` out gives both consumers something they can include, and deletes the CI grep as a side effect.

Filed `post-mainnet` on purpose. It touches the Dockerfile COPY and the CI install path, which is not a change worth making two weeks out from the payment-rail cutover. The guard here is the part that matters before Sept 16.

onder-akkaya:
Tested this against a real upgraded env before merging, not just against CI.

I installed the new floors into my local conda env and re-ran the suite. **3860 passed, 2 skipped, 0 failed** — identical to the run before the upgrade. `pip check` reports no broken requirements. All backend modules still import (0 failures walking `archimedes.*`).

Machine: macOS arm64, Python 3.12, the `archimedes` conda env. 16 of the 18 packages were below the new floor on my box before this, so it was a real upgrade rather than a no-op.

## Two things to know before you update

**1. `anthropic` jumps a major version.** The floor is `>=0.120.2` with no upper bound, so pip resolves it to **1.2.0**. I checked every surface we actually use rather than assuming:

- `anthropic.Anthropic` and `anthropic.AnthropicBedrock` both still exist
- `messages.create` still takes `model`, `max_tokens`, `system`, `messages`, `tools` — the complete set of kwargs we pass at all four call sites (`llm_backend.py:138/192/261`, `portfolio_agent.py:363`)
- we never pass `temperature`, which is the one param that has left the signature

So our usage is compatible. Worth knowing anyway, because the unit suite mocks the LLM boundary and would not have caught it either way. 1.x also brings its own `httpx2`, which sits alongside `httpx` 0.28.1 rather than replacing it, so nothing else is disturbed.

**2. `conda env update -f environment.yml` does more than this PR.** `nodejs`, `terraform` and `awscli` are declared in the conda section but were never installed in my env — my node comes from Homebrew and I use OpenTofu rather than terraform. Running the full command installs all three from conda-forge and they will shadow your Homebrew equivalents once the env is active. That section is untouched by this PR. If you only want what changed here, update the pip block.

Also: if your env predates the production bump, `pandas>=2.2` becoming `>=3.0.5` is a major upgrade. Mine was already on 3.0.3 so I only moved a patch, and I have not exercised the 2.x to 3.x path. Production has been on pandas 3 for a while, so an env on 2.x was running something prod does not.

## What the PR is, in one line

It raises 18 floors in `environment.yml` to the values `backend/requirements.txt` already carried, folds in the five open Dependabot bumps so they land in both files at once, and adds a test so the gap cannot silently reopen. Docker and CI install from `requirements.txt` and are unaffected either way.

Issue #1462, opened 2026-08-21
[infra] archimedes-arc.com is spoofable: no SPF, no DMARC


dbrowneup:
## The problem

`archimedes-arc.com` publishes **three SES DKIM CNAMEs** and sends real
transactional mail (Better Auth verification + password-reset), but the zone
has **no SPF record and no DMARC record**. Verified 2026-08-21 against the live
zone `Z0326924NBP1KIQXCIGO`:

```
$ dig +short TXT archimedes-arc.com @8.8.8.8
"google-site-verification=..."      # <- the only TXT record
$ dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8
                                    # <- empty
```

Consequences today:

1. **Anyone can send mail as `@archimedes-arc.com`.** With no SPF and no DMARC,
   a receiving server has no published policy to check a forged envelope
   sender against. Our own DKIM signing does not prevent this — DKIM proves a
   message *we* signed is authentic; it says nothing about an unsigned forgery.
2. **We would never know.** With no DMARC record there is no `rua` address, so
   there are no aggregate reports and no signal that spoofing is happening.
3. **Deliverability of our real mail suffers.** Gmail and Yahoo require
   authenticated mail from bulk senders and increasingly penalise domains with
   no published policy — the sign-in and password-reset mail this app depends
   on is exactly the mail that must not land in spam.

This matters more now than it did last week: `/privacy` and `/terms` (PR #1432)
publish `privacy@archimedes-arc.com` as the contact address for privacy and
deletion requests. A spoofable domain on the address people are told to use for
data-protection requests is a genuinely bad combination — it is a ready-made
phishing pretext against our own users.

## Scope

Add to the `archimedes-arc.com` hosted zone:

1. **SPF** — a single TXT record at the apex authorising SES:
   `"v=spf1 include:amazonses.com -all"`.
   Note the apex TXT already holds the Google verification string; SPF must be
   **added as a separate TXT record**, and the existing string must not be
   disturbed. Deleting it un-verifies Search Console.
2. **DMARC** — TXT at `_dmarc.archimedes-arc.com`. Start at
   `p=none` with an `rua` so we can *observe* before enforcing:
   `"v=DMARC1; p=none; rua=mailto:<reports@...>; fo=1"`.
3. **A follow-up to move to `p=quarantine` then `p=reject`** once aggregate
   reports show all legitimate sending is aligned. Do not jump straight to
[...]

Issue #1483, opened 2026-08-24
[infra][ci] Nothing in CI parses a .tf file — add a path-filtered terraform fmt-check + validate job


onder-akkaya:
## Summary

`infra/` holds **18 `.tf` files** and took **28 commits in the last 30 days**, several of them on the live payment path (`#1414` payment-recipient TF_VAR, `a45e8e5` the generation-payment flip, `#1457` revenue sweep). **No workflow validates any of it.** A syntax error in `infra/ecs.tf` surfaces for the first time when Dan runs `terraform apply` — which is also the prod deploy path.

## Evidence

```bash
grep -rn "terraform\|tofu" .github/workflows/*.yml
```

Returns exactly one hit, and it is a warning *string* inside `deploy.yml:760` about a grant needing `terraform apply` — not a step.

The one infra-named job is a false friend:

```yaml
# .github/workflows/quality-gate.yml
name: Infra — user-data size guard
  - name: EC2 user-data stays under the 16 KB limit
    run: bash ./infra/scripts/check-user-data-size.sh
```

That is a byte-count check on EC2 user-data. It never parses a `.tf` file, but it is the row a reviewer sees go green on an `infra/**` PR — the exact "a signal was trusted for something it does not actually measure" shape `CLAUDE.md` § *the green check may not mean what you think* is about.

Concretely: #1482 changes `infra/ecs.tf` and carries a full green row. That row says the Python suite, ruff, import-guard and the frontend build are unaffected. It says nothing about whether `ecs.tf` parses.

## Why this is worth a job rather than a habit

`infra/ecs.tf` is where the money switches live — `PAYMENTS_DRY_RUN`, `GENERATION_PAYMENTS_DRY_RUN`, `GENERATION_PAYMENT_REQUIRED`, `GENERATION_PRICE_USD`, and the daily generation caps. Those are edited under time pressure (the flip landed the same day it was decided), by more than one contributor, and increasingly by agents. "Dan will catch it at apply time" is a single point of failure on the one file where a typo is most expensive.

## Proposed job

A new `.github/workflows/infra-gate.yml`, path-filtered to `infra/**`:

```yaml
- run: terraform fmt -check -recursive infra/
- run: terraform -chdir=infra init -backend=false
- run: terraform -chdir=infra validate
```

Three properties that make this safe and cheap:

- **`fmt -check` needs no init at all.**
[...]

Issue #1460, opened 2026-08-21
[infra] Codify privacy@ inbound mail (MX + SES receipt rules) in Terraform


dbrowneup:
## Context

`privacy@archimedes-arc.com` is the contact address published on `/privacy` and
`/terms` (PR #1432). It was stood up **by CLI on 2026-08-21** to unblock that
merge, with the explicit agreement that Terraform would follow. Until this
lands, the resources exist only in the live account and nothing in the repo
records that they exist — that is undocumented drift, and it is the whole
reason this issue exists.

## What exists in the account today (all `us-east-1`, account `037613907429`)

| Resource | Identifier |
| --- | --- |
| SNS topic | `arn:aws:sns:us-east-1:037613907429:archimedes-privacy-inbox` |
| Topic policy | `AllowSESPublish` — `ses.amazonaws.com` may `sns:Publish`, conditioned on `AWS:SourceAccount` |
| SNS subscription | `email` → the owner's inbox |
| SES receipt rule set | `archimedes-inbound` (**active**) |
| SES receipt rule | `privacy-inbox` — recipients `["privacy@archimedes-arc.com"]`, `SNSAction` (UTF-8), `ScanEnabled: true`, `TlsPolicy: Optional` |
| Route 53 MX | `archimedes-arc.com. 300 IN MX 10 inbound-smtp.us-east-1.amazonaws.com` (zone `Z0326924NBP1KIQXCIGO`) |

Also added by CLI the same day, and in scope here for the same reason:

| Route 53 TXT | `archimedes-arc.com. 300 IN TXT "google-site-verification=..."` — Google Search Console domain verification. Deleting it un-verifies the property. |

## Task

Bring all of the above under Terraform in `infra/`, importing the existing
resources rather than recreating them.

- `aws_sns_topic`, `aws_sns_topic_policy`, `aws_sns_topic_subscription`
- `aws_ses_receipt_rule_set` + `aws_ses_active_receipt_rule_set` + `aws_ses_receipt_rule`
- `aws_route53_record` for both the MX and the Google verification TXT

Note that `infra/alb.tf:108` exposes the zone as `data.aws_route53_zone.main`
— use that, do not add a second zone lookup.

## Acceptance criteria (machine-checkable)

1. `cd infra && terraform plan` reports **`No changes.`** against the live
   account. A plan that would destroy and recreate the rule set, the topic, or
   either DNS record is a **failure**, not a diff to accept — recreating the
[...]

PR #1503, opened 2026-08-29
[infra] Publish SPF and DMARC for archimedes-arc.com (#1462)

onder-akkaya:
Addresses #1462. **Not applied — this needs @dbrowneup's AWS account.** I can write and validate the config; I cannot `plan` or `apply` it, and the issue's acceptance criteria are all `dig` against live DNS, so they can only be met after apply. See "What I could not verify" below.

## Re-verified the premise before writing anything

```
$ dig +short TXT archimedes-arc.com @8.8.8.8
"google-site-verification=nHeZsrl8SxRsJeKWIQx0kaSQkHOlzPDdfRZU_ZCUqk8"

$ dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8
(empty)

$ dig +short MX archimedes-arc.com @8.8.8.8
10 inbound-smtp.us-east-1.amazonaws.com.
```

Still true on 2026-08-29. One TXT, no SPF, no DMARC.

## The apex TXT is the part to read carefully

Route 53 keeps **one record set per (name, type)**, so SPF cannot be added as a second `aws_route53_record` alongside the Google Search Console string. And that string is **not currently managed by Terraform** — nothing in `infra/*.tf` declares a TXT record at all.

So the new resource has to *adopt* the existing record set: `allow_overwrite = true`, carrying **both** strings. Getting this wrong is not subtle — a resource with only the SPF string plus `allow_overwrite` would replace the set and un-verify Search Console on the next apply, which is exactly the failure the issue warns about. The Google token is lifted into a variable so it is visible rather than buried in a list, with a comment saying why it lives there.

## `-all`, not `~all`

The issue specifies `-all` and I checked before agreeing. `auth/mailer.js` sends through `SESv2Client`; its only other mode is a console stub for local dev that logs instead of sending. Nothing else in the tree sends mail — no SendGrid, Mailgun, Postmark, nodemailer or SMTP config anywhere.

So there is no legitimate mail from another source to soft-fail. **If that changes, the include list must be extended in the same apply that enables the new sender**, or its mail hard-fails.

## DMARC stays at `p=none`

Deliberately. It publishes a policy and requests reports without changing how any receiver treats our mail, so it cannot silently drop our own sign-in and password-reset mail. `p=quarantine` → `p=reject` is evidence-led and belongs in a later change once the aggregate reports show every sending path is aligned. I've filed that as a follow-up rather than leaving it as a comment.

`p=none` isn't decorative in the meantime: Gmail's and Yahoo's bulk-sender rules require a DMARC record to *exist*, which is one of the three consequences the issue lists.

**One honest caveat on `rua`.** It defaults to `dmarc-reports@archimedes-arc.com`. The zone's MX already points at SES inbound, but the receipt rule that actually delivers to that address is **#1460's scope** — so until #1460 lands, the reports are requested and not collected. A reporter that cannot deliver drops them silently rather than bouncing at us, so this is a gap in observability, not a breakage. I'd rather say that plainly than let the record imply reports are being read.

## What I could not verify

- **No `plan`, no `apply`.** No credentials for account `037613907429`.
- **All three acceptance criteria are post-apply** (`dig` for SPF, `dig` for DMARC, and a test send showing `spf=pass`). None can be checked from here.
- What I did run: `tofu validate` → *Success! The configuration is valid.*, and `tofu fmt -check -recursive infra/` → clean, with the diff limited to one new file plus two variables.

Worth noting this is exactly the gap #1483 is about — nothing in CI parses a `.tf` file, so `validate` ran only because I ran it by hand.

## Out of scope, with reasoning

Item 4 (custom MAIL FROM domain in SES, so the envelope sender aligns for SPF as well as DKIM) is marked "consider" in the issue and I left it out. It needs an SES domain-identity resource that doesn't exist in Terraform yet, plus its own MX and SPF records on a subdomain — a bigger change that shouldn't ride along with the record that closes the immediate spoofing hole. Happy to do it next if you want it before mainnet.

## For the applier

```bash
cd infra && tofu plan -target=aws_route53_record.apex_txt -target=aws_route53_record.dmarc
```

The apex TXT plan should show a **replace carrying two strings** — if it shows one string, stop, because that apply would drop the Search Console verification.

After apply, the issue's own commands:

```bash
dig +short TXT archimedes-arc.com @8.8.8.8
dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8
```

### DIFF
--- a/infra/dns_email.tf
+++ b/infra/dns_email.tf
@@ -1,0 +1,71 @@
+# ── Email authentication for ${var.domain_name} (#1462) ──────────────────────
+#
+# The zone publishes three SES DKIM CNAMEs and the app sends real transactional
+# mail (Better Auth verification + password reset), but had no SPF and no DMARC.
+# Verified against live DNS on 2026-08-29:
+#
+#   dig +short TXT archimedes-arc.com @8.8.8.8
+#     "google-site-verification=nHeZsrl8SxRsJeKWIQx0kaSQkHOlzPDdfRZU_ZCUqk8"
+#   dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8
+#     (empty)
+#
+# DKIM alone does not close this. DKIM proves a message we signed is authentic;
+# it says nothing about an unsigned forgery, and with no published policy a
+# receiver has nothing to check a forged envelope sender against. This matters
+# more now that /privacy and /terms publish privacy@archimedes-arc.com as the
+# contact for data-protection requests — a spoofable domain on the address
+# users are told to trust is a ready-made phishing pretext.
+
+# ── SPF ──────────────────────────────────────────────────────────────────────
+#
+# CAREFUL: Route 53 holds ONE record set per (name, type), so the apex TXT
+# cannot be split across two `aws_route53_record` resources. The Google Search
+# Console verification string already lives there and is NOT currently managed
+# by Terraform, so this resource has to ADOPT that record set rather than add
+# alongside it — which is why `allow_overwrite` is true and why the Google
+# string is carried explicitly below.
+#
+# Dropping that string from this list would un-verify Search Console on the
+# next apply. It is a published DNS value, not a secret.
+resource "aws_route53_record" "apex_txt" {
+  zone_id = data.aws_route53_zone.main.zone_id
+  name    = var.domain_name
+  type    = "TXT"
+  ttl     = 300
+
+  records = [
+    var.google_site_verification,
+    # -all (hard fail) rather than ~all: SES is the only sender, so there is no
+    # legitimate mail from anywhere else to soft-fail. If another sender is
+    # added later, extend this include list in the SAME apply that enables it.
+    "v=spf1 include:amazonses.com -all",
+  ]
+
+  # Adopts the existing unmanaged record set. Without this the first apply
+  # fails with "record already exists"; with it, the record is replaced by the
+  # full list above — which is why both strings must be present.
+  allow_overwrite = true
+}
+
+# ── DMARC ────────────────────────────────────────────────────────────────────
+#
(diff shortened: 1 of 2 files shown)


onder-akkaya:
@dbrowneup this one needs your AWS account — it is a `tofu apply` against the prod Route 53 zone, so I cannot take it past review.

Re-checked live DNS just now, 2026-08-30, and nothing has changed since I opened it:

```
$ dig +short TXT archimedes-arc.com @8.8.8.8
"google-site-verification=nHeZsrl8SxRsJeKWIQx0kaSQkHOlzPDdfRZU_ZCUqk8"

$ dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8
(empty)
```

Still no SPF, still no DMARC. The domain is spoofable today, and `/privacy` and `/terms` publish privacy@archimedes-arc.com as the address for data-protection requests, which makes it a usable phishing pretext rather than a theoretical one.

Two things worth knowing before you look:

**The red detect-secrets row is a false positive.** Both hits are `nHeZsrl8SxRsJeKWIQx0kaSQkHOlzPDdfRZU_ZCUqk8`, the Google Search Console token — a published TXT record, as the `dig` above shows. It appears twice in the diff, once in a comment and once in the `records` list. It has to be carried explicitly because Route 53 holds one record set per (name, type), so the SPF string cannot be added alongside it without adopting the existing set. Dropping it would un-verify Search Console on the next apply.

**DMARC is `p=none` on purpose.** It publishes a policy and requests aggregate reports without changing how any receiver treats our mail, so it cannot silently drop sign-in or password-reset mail. Moving to quarantine and then reject is #1504, evidence-led, once the rua reports show every sending path is aligned.

Closes #1462, which is labelled `mainnet-gate`. 17 days to Sept 16.

PR #1506, opened 2026-08-29
[infra] Codify the privacy@ inbound mail path in Terraform (#1460)

onder-akkaya:
Addresses #1460. **Not applied and not imported — both need @dbrowneup's AWS account.** The issue's acceptance criteria are all post-import (`terraform plan` → *No changes*, `terraform state list`, two `dig` checks), so none of them can be met from here. `tofu validate` and `tofu fmt` pass; that is the whole of what I could check.

## Written to be imported, not applied

Every name matches the live resource exactly. That matters more than usual here: the address is published on `/privacy` and `/terms`, and **recreating the MX record drops inbound mail for it**. A plan that proposes to create rather than adopt is the failure mode, not a diff to accept.

```bash
cd infra
tofu import aws_sns_topic.privacy_inbox                arn:aws:sns:us-east-1:037613907429:archimedes-privacy-inbox
tofu import aws_sns_topic_policy.privacy_inbox         arn:aws:sns:us-east-1:037613907429:archimedes-privacy-inbox
tofu import aws_ses_receipt_rule_set.inbound           archimedes-inbound
tofu import aws_ses_active_receipt_rule_set.inbound    archimedes-inbound
tofu import aws_ses_receipt_rule.privacy_inbox         archimedes-inbound:privacy-inbox
tofu import aws_route53_record.mx                      Z0326924NBP1KIQXCIGO_archimedes-arc.com_MX
```

Then `tofu plan` — the issue's AC1. It should say `No changes.`

## The apex TXT is deliberately not here

This is the one place I've departed from the issue, and it's worth reading before review.

#1460 lists the Google Search Console TXT record in scope. It can't live here, because **#1462 has to manage that same record set** — Route 53 keeps one record set per (name, type), so the SPF string and the Google verification string must share a single `aws_route53_record`. Declaring it in both files would give Terraform two resources for one record and a permanent fighting diff, and the failure mode of getting that wrong is un-verifying Search Console.

So the apex TXT lives in `dns_email.tf` (PR #1503), carrying both strings, and this file covers the MX only. The two PRs are **order-independent** — whichever lands second doesn't need the other — but only one of them may ever declare that record.

## The subscription endpoint stays out of the repo

The live SNS subscription points at a personal inbox. That address doesn't belong in a public repository, so `privacy_inbox_email` defaults to `""` and the resource is `count`-gated on it — the exact `alarm_email` pattern already in `cloudwatch.tf`.

Consequence worth stating rather than glossing: **left unset, the live subscription stays unmanaged.** That's the status quo, not a regression, but it means AC1's *No changes* is partly true by omission for that one resource. Set it in `infra/terraform.tfvars` (gitignored, `.gitignore:66`) to bring it under management properly.

## Anti-goals respected

- Address, rule-set name and topic name unchanged.
- `tls_policy = "Optional"`, not `Require`. The issue is right that `Require` bounces senders on older TLS, and this is the one inbox that must not silently reject someone filing a data-protection request.
- No SPF/DMARC here — that's #1462 / PR #1503.
- The Google verification TXT is not deleted; it's declared in `dns_email.tf` with both strings, for the reason above.

## A correction `tofu validate` caught

My first draft added `data "aws_caller_identity" "current"` for the topic policy's `SourceAccount` condition. One already exists at `ecs.tf:66`, and Terraform rejects duplicate data-source names in a module. My grep for it had been truncated by a `head -5` that showed only the `aws_region` matches — I'd concluded "no such data source" from an output I'd cut short. Now reused, with a comment pointing at where it's declared.

Worth noting this is the second thing in two PRs that only `tofu validate` caught, and nothing in CI runs it — which is #1483.

## Known limitations, recorded not fixed

The issue lists these and I've not addressed them; they're real and belong in their own work:

- SES `SNSAction` caps at **150 KB**. A larger message is simply not delivered, and nothing alarms on it. For a privacy contact that is a real ceiling — someone attaching a document to a deletion request could vanish silently.
- Replies go out from the owner's personal address unless SES SMTP credentials are configured to send *as* `privacy@`.


onder-akkaya:
@dbrowneup same blocker as #1503 — this needs your AWS account. It is written to be **imported, not applied**: every resource name matches the live one, and the six `tofu import` lines are in the PR body.

Worth being deliberate on this one. The address is published on `/privacy` and `/terms`, and **recreating the MX record drops inbound mail for it**. A plan that proposes to create rather than adopt is the failure mode, not a diff to wave through. After the imports, `tofu plan` should say `No changes.` — that is the issue’s AC1.

One thing that is true by omission and I would rather flag than let you find: the live SNS subscription points at a personal inbox, which does not belong in a public repo, so `privacy_inbox_email` defaults to `""` and the resource is count-gated on it, following the `alarm_email` pattern in `cloudwatch.tf`. Left unset, that subscription stays unmanaged. Status quo rather than a regression, but it means `No changes.` is partly true because that resource is not declared. Set it in `infra/terraform.tfvars` (gitignored) to bring it under management.

This and #1503 are order-independent, but only one of them may ever declare the apex TXT record set, and that is #1503. Reasoning is in the body here under "The apex TXT is deliberately not here".

Closes #1460, labelled `mainnet-gate`.

dbrowneup:
Resolved the `variables.tf` conflict with #1503 (both PRs appended variables at the file tail — purely additive; both blocks kept) via a merge commit rather than a rewrite of the branch. `terraform validate` passes on the resolution. Verified as a union locally before either merged: all four queued PRs merged into one integration branch → full unit suite 3876 passed / 0 failed → fresh docker build healthy → UI smoke on the built stack.

Issue #1526, opened 2026-08-30
[chain][provenance] IPFS pinning: wire PINATA_JWT into prod or remove the dead path (split from #714)


dbrowneup:
Split out of #714 (2026-08-30) so that #714 narrows to the on-chain commit-reveal path only.

## State (verified against main today)

IPFS/Pinata provenance pinning is **code-complete** (`pinata_client.py`, `provenance_publisher.py`, CID wired into `reveal()`'s `storagePointer`) — #714 sub-task B. But `PINATA_JWT`:
- appears in `infra/scripts/setup-ssm-secrets.sh` only under "Forthcoming, as features land (roadmap T1.x)";
- is **absent from `infra/ecs.tf`'s backend `secrets{}`**.

So prod pinning is silently off: every reveal stores a `storagePointer` that nothing pinned. Fail-soft on a provenance claim — the exact pattern `docs/architectural-principles.md` § fail-soft forbids.

## Decision to make (owner)

Either **(a) wire it**: seed `/archimedes/prod/PINATA_JWT` in SSM, add the `secrets{}` entry to `ecs.tf`, apply, and verify one real reveal's CID resolves on a public gateway; or **(b) drop it deliberately**: remove the dead pinning path and let `storagePointer` carry the on-chain hash reference only, documented as such. What is not acceptable is the current middle state where the code implies pinning that never happens.

## Acceptance
- (a): `curl https://gateway.pinata.cloud/ipfs/<CID-from-a-real-reveal>` returns the trace bytes; `/health` or logs show pinning active. Or (b): `grep -rn PINATA backend/ infra/` returns only historical docs, and the spec notes the decision.
- Anti-goal: do not flip any other flag in the same PR.

Related: #714 (parent, now commit-reveal-only), #1353/PR #1403 (reveal reconciliation), `docs/specs/commit-reveal-trace-spec.md`.

Issue #714, opened 2026-06-24
T0.3 + T1.4 — Wire real on-chain commit-reveal + IPFS provenance (+ mnemonik scoping)


dbrowneup:
> **Cross-reference (added at filing, 2026-06-23):** This issue is the detailed implementation spec for the **backend commit-reveal wiring** that appears as bullet #3 of the open umbrella **#588** ("Redeploy hardened contract suite … + backend commit-reveal wiring"), plus the **T1.4 IPFS provenance pipeline** and the **mnemonik scoping memo**. Relationship to existing issues:
> - **#588** stays the redeploy / ABI-refresh umbrella; this issue owns the detailed Sub-task A backend wiring it references (one-line bullet there -> full spec here).
> - **#589** ("Enforce commit-before-trade at the Vault and bind commitments to a trade identifier") is the complementary **on-chain** enforcement -- a stronger follow-on needing a `Vault.sol` change + redeploy. This issue does **not** do that; it wires the *existing* `commit()`/`reveal()` from the backend (no contract change).
> - Triage action: link #588 / #589 / this issue and de-conflict the "backend wiring" bullet so it isn't done twice.

---

> **Owner:** Bogdan Sivochkin
> **Roadmap items:** T0.3 (real on-chain provenance), T1.4 (IPFS publishing pipeline), and mnemonik integration scoping (memo only)
> **Tier:** 0 (Claim Integrity) + 1 (Core Lepton vertical)

## Summary

[...]

PR #1529, opened 2026-08-30
[docs] Mnemonik integration scoping memo (#714 sub-task C): borrow patterns, do not integrate

dbrowneup:
Part of #714 — the long-open sub-task C deliverable (the umbrella itself is already closed; this memo completes its record). Verdict and full comparison table in the memo; evaluation provenance section included.

Docs-only. Three files: the memo, its `docs/README.md` register row, and a `mkdocs.yml` `exclude_docs` entry.

## Publication: internal, not published

**This memo is internal and is not on the docs site.** It names a third party and grades its maturity and maintainer — a decision record for us, not a page for readers. Per the owner's docs decision (Q9: internal material is excluded from the public site), it is **out of the nav and in `exclude_docs`**.

Nav-omission alone would not have been enough. mkdocs walks `docs_dir` and builds every file it finds; `validation.nav.omitted_files` defaults to `INFO`, which `--strict` ignores. Measured on the merge commit `2b55149d`, before this change (mkdocs 1.6.1 / material 9.7.7, the version `.github/workflows/docs-site.yml` pins):

```
$ mkdocs build --strict --site-dir _site
EXIT=0        WARNING lines: 0

$ ls -d _site/specs/mnemonik-integration-scoping
_site/specs/mnemonik-integration-scoping          # built and crawlable

$ grep -c mnemonik-integration-scoping _site/index.html
1                                                  # the docs/README.md register row
$ grep mnemonik-integration-scoping _site/index.html | grep -c md-nav__link
0                                                  # absent from the sidebar nav
```

For contrast, the same two commands on `54b39821` (the commit that added the nav row) return `2` and `1` — the register row plus a `class="md-nav__link"` entry in the sidebar.

So the page was **linked from Home but absent from the sidebar nav** — published-but-unlisted, the state the docs-site audit (#1816) catalogued. (An earlier revision of this description claimed that `grep -c` returned `0`. It did not; the true value was `1` in every commit of this branch, because `docs/README.md` is the site's Home page and the register row links the memo. The number was wrong and is corrected here.)

After the change, on `bc3fbaf3`:

```
$ mkdocs build --strict --site-dir _site
EXIT=0        WARNING lines: 0        pages: 197

$ ls -d _site/specs/mnemonik-integration-scoping
ls: No such file or directory                      # not published
```

**Caveat, stated rather than hidden.** Neither `exclude_docs` nor `backend/tests/test_mkdocs_exclude_docs_matches.py` exists on `main` — both arrive with #1816 and #1832, which are still open. So this PR introduces the block itself, in the shape #1816 uses (gitignore-style patterns, one per line, nothing after a pattern). Two consequences:

- The `docs/README.md` register row is still required by `.github/scripts/docs_index.py` ("a doc not listed here does not exist"), and its link now resolves nowhere on the site. mkdocs reports that at `INFO`, so `--strict` stays green. #1832 is the PR that fixes this class — its `published_uris()` hook rewrites a link into an excluded page to that file's GitHub blob URL.
- When #1816 or #1832 merges, `mkdocs.yml` will conflict on the `exclude_docs` block. The resolution is a union: keep their block, keep this one line in it.

## The corrections in this PR

An adversarial verification pass found three false claims. Each was re-checked line by line against the code on this branch.

**1. Both "do these instead" residuals were stale — both are closed.**

- *Old:* "3 legacy `trace_publisher.publish()` call sites in `backend/archimedes/chain/agent_runner.py` (~1554, 1652, 2101) still bypass the real commit/reveal path — the issue's own anti-goal grep is not yet 0."
  *New:* recorded as **done, 2026-08-30**, citing commit `7737fbd0` ("[chain] Retire the v1 publishTrace fallbacks from the agent tick (#714)"), which is an ancestor of this branch.
  *Verified:* `grep -n "\.publish(" backend/archimedes/chain/agent_runner.py` → no output; `graft callers trace_publisher.publish` → only `test_publish_circle_path` and `test_publish_no_signer_no_account` in `backend/tests/test_trace_publisher.py`.
- *Old:* "IPFS pinning prod gap split to #1526: … wire it or remove the path."
  *New:* recorded as **closed 2026-09-01 as outcome (b) — the path was removed, not wired**, citing `docs/adr/ipfs-pinning-not-live.md`, the deletion of `chain/pinata_client.py` / `chain/provenance_publisher.py`, and `backend/tests/test_ipfs_pinning_absent.py`.
  *Verified:* `gh issue view 1526` → `CLOSED` (2026-09-01); `sed -n 1577p backend/archimedes/chain/agent_runner.py` → `await trace_publisher.reveal(trace_id, trace, storage_pointer="")`.

**2. The commit-before-trade citation pointed at the wrong file entirely.**

- *Old:* "commit-before-trade enforced at `Vault.sol:422`".
  *Verified:* `sed -n 418,423p contracts/src/Vault.sol` → the `MIN_LIQUIDITY` inflation-attack-guard branch inside `previewDeposit`. Line 422 reads `// so a 1-wei first deposit (the classic attack setup) is impossible.` Nothing to do with trades.
  *New:* "enforced in the vault's `rebalance()` (`contracts/src/Vault.sol:446-453`): the trade id is recomputed from the exact swap parameters and `traceRegistry.executeTrade(tradeId)` reverts unless an earlier-block commitment binds that `(vault, trade)`. The registry is `immutable` and null-checked in the constructor (`contracts/src/Vault.sol:160`, `:220-227`)."
  *Verified:* `:452` = `bytes32 tradeId = keccak256(abi.encode(...))`, `:453` = `uint256 traceId = traceRegistry.executeTrade(tradeId);`, `:160` = `IReasoningTraceRegistry public immutable traceRegistry;`, `:222` = the `TraceRegistryNotSet` revert, `:227` = the assignment.

Every other `file:line` in the memo was re-verified the same way and tightened where it was loose:

| Claim | Old | New | Verified |
|---|---|---|---|
| registry range | `:108-209` (starts mid-`publishTrace`) | `:114-209` | `commit` opens at 114, `reveal` closes at 209 |
| `commit` time-lock | unpinned | `:138` | `require(claimedExecutionTime > block.timestamp, …)` |
| `executeTrade` binds `tradeId` | unpinned | `:163-164` | `_pendingTrade[msg.sender][tradeId]` + `require(traceId != 0, …)` |
| `executeTrade` block guard | unpinned | `:169` | `require(block.number > c.commitBlock, …)` |
| `reveal` committer check | unpinned | `:198` | `require(msg.sender == c.committer, …)` |
| `reveal` hash binding | unpinned | `:203` | `require(keccak256(fullTraceContent) == c.contentHash, …)` |
| `publishTrace`'s `metadata` | `publishTrace:103` | `:103-106` | param is at 106, signature opens at 103 |
| `commit`'s `tradeIntentSummary` | `tradeIntentSummary:114` | `:114-119` | the param is at **119**; 114 is `function commit(` |
| `Commitment.storagePointer` | `reveal:190` | `reveal` writes it at `:206` | `c.storagePointer = storagePointer;` |
| trace canonicalization | `models/trace.py`, no lines | `backend/archimedes/models/trace.py:77-113` | `_HASH_FIELDS` 77-91 → `canonical_json` 93-102 → `Web3.keccak` 112 |
| rerank helper | `paper_rag._rerank_with_embeddings` | `…/paper_rag.py:430` | `def _rerank_with_embeddings(` |

**3. `ipfs_publisher.py` was cited as a live sibling module. It does not exist.**

- *Old:* "a sibling of `ipfs_publisher.py` POSTs the same canonical public-provenance bytes…"
  *Verified:* `find . -name "*ipfs*" -not -path "./.git/*" -not -path "*__pycache__*"` returns exactly four paths — `docs/adr/ipfs-pinning-not-live.md`, `docs/specs/ipfs-reasoning-traces-design-note.md`, `backend/tests/test_ipfs_pinning_absent.py`, `ui/test/ipfs-pinning-copy.test.js`. No module. The pin path was `chain/pinata_client.py` + `chain/provenance_publisher.py`, both deleted by #1526.
  *New:* "a fail-soft step at the tail of `agent_runner._reveal_trace` (`backend/archimedes/chain/agent_runner.py:1519-1671`)".
- Also fixed: the memo's primary-sources line cited `origin/dbrowneup/debate-transcript-capture`, a branch that no longer exists (`git ls-remote --heads origin | grep -c debate-transcript-capture` → `0`). Replaced with the merged artifacts — `backend/archimedes/models/debate_transcript.py` and `backend/migrations/versions/5728d9ef1901_add_debate_transcripts.py`.
- Also disambiguated: `docs/WHITEPAPER.md`, `docs/how-it-works.md` and `docs/spec/memory-composition.md` are **Mnemonik's** docs, not ours; the memo read as if they were repo-local paths. Now explicitly attributed.

`updated:` moved 2026-08-30 → 2026-09-03 and the register row's summary now says the memo is internal.

## Verification (head `bc3fbaf3`)

| Check | Result |
|---|---|
| `git merge origin/main` | clean, no conflicts (main moved again since the last merge) |
| `.github/scripts/docs_links.py` | scanned 1889 in 189 files; **broken 0 (0.0%)** |
| `.github/scripts/docs_index.py` | 184 docs under `docs/`; **184 indexed, 0 orphaned** |
| `.github/scripts/docs_staleness.py` | 0 of 87 `current` docs older than 60 days (informational) |
| `mkdocs build --strict` | **exit 0, 0 WARNING**, 197 pages; memo **absent** from `_site/` |
| `pytest test_docs_site test_ipfs_pinning_absent test_canonical_repo_links` | **46 passed** |

No `.py`, `.sol` or `ui/` files touched. No AWS calls, no on-chain sends, no terraform.

**Adversarial pass on the new guard:** appending a trailing `# internal` to the `exclude_docs` pattern — the gitignore-semantics trap where the comment becomes part of the pattern — silently republishes the page (`_site/specs/mnemonik-integration-scoping/` reappears) while `mkdocs build --strict` stays exit 0. Reverted. That failure mode is exactly what #1832's `test_mkdocs_exclude_docs_matches.py` guards, and it is why the block carries the warning comment.


dbrowneup:
## Re-reviewed against the whitepaper — verdict shifted, memo updated (rebased on main)

Per owner direction, this went back through a **second evaluation (Opus)** focused on the protocol's *math*, because the value lives in the whitepaper — and the first pass had evaluated the wrong document.

**What the first pass missed:** it read `docs/how-it-works.md` (the Solana/Arweave/MCP *deployment*) and never read `docs/WHITEPAPER.md`, where the protocol is specified **chain-agnostically**. The dependency call was right; the significance call was not.

**The load-bearing idea (§4.3.1):** the quantized embedding vector is nested *inside* the canonical payload **before** the content hash is taken — `CID(A) = BLAKE3(S_canonical)`. So a semantic-recall hit is *provably* the same vector that was anchored: the index stays untrusted, the binding does not. **That is exactly the primitive for verifiable semantic memory over our reasoning traces** — the on-chain reasoning-memory vision. Our `models/trace.py` canonicalization is structurally the same shape, one field short.

**Verdict shift:** `borrow patterns, do not integrate` → **`adopt the protocol math, reimplement natively on Arc, take no dependency`**. Three ideas promoted to work items (integrity-bound embeddings; lineage-by-embedded-parent-hash — `consulted_paper_hashes` is already half that DAG; batch roots + quantized ranking), plus a phased path:
- **v0** — `embedding vector(384)` on traces + `debate_transcripts`, MiniLM local (cents of compute), recall only, no integrity claim.
- **v1** — `embedding_digest = keccak(model_id ‖ quantized_vector)` in `ReasoningTraceRegistry`'s existing `metadata` bytes — **no contract change**, and explicitly **not** in `_HASH_FIELDS` (would break `verifyTrace` parity for every anchored trace).
- **v2** — batch-root anchoring, κ-scoped cross-user query, `FRAME` injection-defense markers.

**Unchanged:** we take **no Mnemonik dependency** (Solana/Arweave chain mismatch, no Python SDK, unaudited single-maintainer, different core problem). We reimplement the ideas natively on Arc — keccak not BLAKE3, JSON not CBOR, SIWE not COSE.

Branch rebased on current `main` (batches 1–3). `docs-check` clean (0 broken / 0 orphaned).

danielscoffee (approved):
Docs-only memo; verdict is borrow patterns, do not integrate. CI clean. Approving as a record close for #714 sub-task C.

mnemonik-dev:
Author of Mnemonic (`mnemonik-xyz/monorepo`) here. Late to a merged PR, but the memo cites the whitepaper closely enough to deserve notes from the implementation side — one correction, two caveats that will bite at v1, and a confirmation.

## The verdict is right, and for the right reason

Mnemonic anchors **public-at-anchor-time**: Arweave holds the full artifact the moment the Solana memo lands. There is no hide-then-reveal in the protocol — that is not a gap we have not gotten to, it is excluded by design. For pre-trade commitment `ReasoningTraceRegistry.sol` is the correct primitive and a Mnemonic dependency would be strictly worse. "No Python SDK" is also accurate: `@mnemonik-xyz/sdk` is JS/TS only.

And since the plan is to reimplement: Apache-2.0, so borrowing the construction is fine without any further conversation.

## One number is inverted

§13.2 reports 4-bit quantization at an **87.5% memory-footprint reduction** — 12.5% of the f32 baseline, an 8x saving — not "at 87.5% of the memory":

> **4-bit Configuration:** Reduces the structural memory footprint by **87.5%** relative to raw 32-bit floating-point metrics. [...] retaining a Top-10 semantic retrieval accuracy index of **98.2%**.

This cuts in the memo's favor. The claim is that carrying a vector inside a hashed payload costs an eighth, not seven-eighths.

## The v1 digest is under-specified

`embedding_digest = keccak(model_id ‖ quantized_vector)` assumes the quantized vector is a function of the model and the text. In the reference implementation it is not — TurboQuant's projection is **seeded**, so `compress()` is a pure function of `(dim, bit_width, seed, embedding)`. We pin the seed as one named constant specifically because of what it does to hashes (`core/src/compress/mod.rs:109-118`):

> Changing this value re-keys every future artifact hash and makes existing rows unverifiable, so it lives here as one constant rather than a literal repeated at each construction site.

If you bind a quantized vector into an on-chain digest, pin the whole tuple `(model_id, dim, bit_width, seed)` into the same bytes and treat all four as immutable for the life of the registry. Bit width especially — our own standing rule is never to change it on an existing database, because old and new vectors stop being comparable for recall. A digest whose inputs you cannot reconstruct is decoration.

## Corollary: do not discard the raw f32

Our `verify` re-compresses the **stored raw embedding** to rebuild the exact compressed bytes that went into the hashed artifact. Quantized bytes are for the hash and the anchor; recall runs on uncompressed f32. If v1 computes the digest and then drops the f32 as redundant, the digest becomes permanently unverifiable — there is no path back to the input. Your v0 already stores `vector(384)`, so read this as "do not optimize that away at v1."

## Small corroboration on `model_id`

You reached for it independently; so did we. The memo describes the anchor payload as `{"h", "a", "v": 2}` — v3 is also live and adds `"m"`, the embed model id (`core/src/solana/mod.rs:169`, `:462`). Same conclusion reached the same way: a vector digest without its model id is not portable.

## Section pointers

The two "noted, not yet scoped" items are cited without refs. κ capability tokens are **§7.2** (cross-owner handshake in **§7.3**); `FRAME` is **§7.5**, Safe Injection (Context Framing). If v2 picks up only one, take §7.5 — it is cheap, has no on-chain surface, and injection-through-recalled-text becomes a live risk the moment v0 ships recall over agent-authored text.

## On the maturity row

Unaudited, effectively single-maintainer, roughly five months old — correct, and fair inputs to a decision under a claims-must-be-true constraint. I would have written the same row.

PR #1544, opened 2026-08-30
[infra][ci] Deploy verdicts measure the deployment, not a stopwatch (#1532)

### LINKED ISSUE #1532: [infra][ci] A slow ECS rollout fails the deploy job identically to a broken one — and skips the CloudFront invalidation
## Symptom

Two deploys failed today at the ECS rollout step while the rollout itself succeeded. Both stopped at the `aws ecs wait services-stable` cap, not on anything wrong with the revision.

| commit | deploy job | duration | outcome |
|---|---|---|---|
| `caf8eb0a` | 19:16:06 → 19:26:16 | **10m10s** | failure |
| `f0da8ec3` | 17:41:30 → 17:51:39 | **10m09s** | failure |
| `2382c35f` | 17:57:58 → 18:06:32 | 8m34s | success |
| `754df4ca` | 02:48:59 → 02:57:12 | 8m13s | success |
| `9dd1a708` | 16:20:54 → 16:28:37 | 7m43s | success |
| `41497944` | 15:28:05 → 15:32:12 | 4m07s | success |

Both failures land on the same number because that number is the waiter's budget, not the rollout's duration. `.github/workflows/deploy.yml:669` calls `aws ecs wait services-stable` with no override, so it is the default 40 attempts x 15s = 600s. The successes are not comfortable either — the most recent one finished with about 90 seconds of headroom.

The log says it plainly:

```
Waiting for the ECS service to reach steady state (deployment_circuit_breaker will auto-rollback a bad revision)...
aws: [ERROR]: Waiter ServicesStable failed: Max attempts exceeded
##[error]Process completed with exit code 255.
```

For `caf8eb0a` the rollout completed anyway. The circuit breaker did not roll back, prod ended up on the new revision, and `/health` returned 200 on 10 of 10 samples afterwards.

## Why this matters more than a red badge

**1. The signal cannot tell "slow" from "broken".** A deploy job that goes red because the rollout took eleven minutes looks exactly like one that went red because the image crashes on boot. Someone reading the badge has no way to tell without opening the log and checking whether the circuit breaker fired. That is a check being trusted for something it does not measure, and it is the shape our review rules already warn about — a green check that means less than it appears, in the other direction.

**2. A timeout silently skips the CloudFront invalidation.** The invalidation step sits after the waiter in the same job, so when the waiter exits 255 the invalidation never runs. Its own comment explains what that costs: CloudFront keeps serving the old cached `index.html`, which references the previous build's hashed `/assets/*` filenames, and those files are not in the new image. So a waiter timeout produces a real user-visible failure, not just a red badge.

Bounded, as it turns out — the html cache policy TTL is 60s, so it self-heals. Verified after `caf8eb0a`: `/` returns 200 and all six referenced assets resolve 200. But the coupling is accidental rather than designed, and the blast radius depends on a TTL nobody is thinking about when they change the waiter.

## What to change

Two parts, and the second matters more than the first.

- **Give the waiter a budget that reflects reality.** `--cli-read-timeout` / a `--waiter-config` with more attempts, sized off the measured 4–10 minute range with real headroom rather than the default that we have now outgrown.
- **Measure the rollout, not the clock.** Poll the service's `deployments[].rolloutState` until it reads `COMPLETED` or `FAILED`, and fail the job only on `FAILED` or on the circuit breaker rolling back. Then a slow rollout is slow and a bad one is red, which is the distinction the current check cannot make.
- **Decouple the invalidation** so a rollout that actually completed still gets its cache invalidated.

## Acceptance criteria

1. A deploy whose rollout takes longer than the current 600s but reaches steady state ends with the job **green**, and its CloudFront invalidation has run. Verify: `gh run view <id> --json jobs` shows both the deploy step and the invalidate step `success`.
2. A deploy whose revision genuinely fails to stabilise (circuit breaker rolls back) ends **red**. Demonstrate against a deliberately broken revision on a non-prod service, or by asserting on the `rolloutState` parse with a fixture — do not skip this, it is the half that makes criterion 1 safe.
3. `grep -n "wait services-stable" .github/workflows/deploy.yml` returns nothing, or returns a call with an explicit waiter-config whose numbers are justified in a comment.

## Anti-goals

- Do not paper over it by marking the step `continue-on-error`. That converts a real failure into silence, which is the direction our fail-soft rule forbids.
- Do not touch `deployment_circuit_breaker` in `infra/ecs.tf`. It behaved correctly here — it did not roll back a rollout that was merely slow.
- Do not fold in #1309's 502 window or #1346's merge-train cancellation. Same subsystem, separate causes.

## Related

#1309 (the ~2-minute 502/503 window per deploy — the outage this one lengthens), #1346 (deploy starvation from a fast merge train, which produced the two `cancelled` runs sitting between these failures).

Found while merging #1530; the deploy path was already failing this way at 17:41, before that merge.

dbrowneup:
Closes #1532. All three defects from Önder's evidence tables:

1. **Explicit rollout poll replaces `aws ecs wait services-stable`** — 15s polls of the PRIMARY deployment's `rolloutState` (selected by **status**, not array position: the draining ACTIVE deployment reports its own COMPLETED, and reading `deployments[0]` blindly can green a rollout that hasn't happened). Budget explicit: `DEPLOY_ROLLOUT_BUDGET_SECONDS=1200`, with a second-0 assertion that the job timeout leaves headroom. Slow-but-completed = green; circuit-breaker rollback / FAILED = red with the reason named; elapsed printed every poll so future budget changes come from measurements.
2. **CloudFront invalidation keyed to `rolloutState == COMPLETED`** via step output (`!cancelled()` guarded) — a slow good deploy never again ships stale edge content.
3. **Post-rollout answer probe** — 10 samples of `/api/health`, last 5 consecutive 200s under 5s, **and `.version` must equal the deployed SHA**. `services-stable` means ECS is content; this attests the app answers.

**Validation:** 13 adversarial fixtures via an `aws` stub (FAILED / rollback / reverted-PRIMARY / draining-COMPLETED trap / budget exceeded / version mismatch / all-200-but-slow = the 58e337dc case / trailing-window miss — every one red; clean and shipped-config cases green). Live read-only run against prod tracked a real rollout 0→464s through a mid-rollout task failure and recovery — and **the probe then caught #1532's second half live**: red while a target sat `Target.Timeout` unhealthy post-COMPLETED, green after ECS replaced it. Both directions demonstrated against production.

**Two flagged calls for review:** the 5s probe ceiling has thin margin against today's latency tail (kept strict deliberately — a red here is a finding about #1309/#1436, not a reason to loosen), and `timeout-minutes: 15→30` is a job-level edit required for the budget to be reachable.

**CI/CD wiring change — team-alignment PR per CLAUDE.md.** @onder-akkaya's #1532 diagnosis specified this fix; his review requested.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,7 +40,13 @@
 #                    definition revision pointing at the just-pushed <commit-sha>
 #                    image and force-redeploys the Fargate service behind the
 #                    existing ALB. NO docker build ever runs on a serving host.
-#                    Once the service reaches steady state, invalidates the
+#                    Then polls the service's own `rolloutState` to an explicit
+#                    verdict (issue #1532 — a slow rollout and a broken revision
+#                    used to produce the same red, and a green used to mean only
+#                    "ECS is content"), probes the live /api/health through
+#                    CloudFront to assert the app actually ANSWERS, and — keyed
+#                    to that rollout verdict rather than to job success —
+#                    invalidates the
 #                    CloudFront cache for the unhashed SPA entry points (/ and
 #                    /index.html) — without this, CloudFront keeps serving the
 #                    OLD cached index.html, which references hashed asset
@@ -150,6 +156,19 @@ env:
   NGINX_ECR_REPO: archimedes-nginx
   AUTH_ECR_REPO: archimedes-auth
   EC2_INSTANCE_ID: i-01803d3abc271d39b
+  # How long deploy-ecs will wait for the ECS rollout to reach COMPLETED before
+  # it calls the deploy red (issue #1532). This replaces the implicit 600s of
+  # `aws ecs wait services-stable` (40 attempts x 15s), which rollout duration
+  # had grown into: Önder's table on #1532 shows two runs nine seconds apart —
+  # caf8eb0a at 10m10s (red) and 58e337dc at 10m19s (green) — getting opposite
+  # verdicts from that stopwatch while both rollouts completed. Sized off the
+  # measured 4-10 min range with real headroom, and DELIBERATELY EXPLICIT so
+  # that raising it is a decision someone makes with the number in front of
+  # them. Exceeding it is still red — this is a budget, not a mute button; the
+  # error names the state so "slow" and "broken" stay distinguishable.
+  # Raising it past ~24 min also requires raising deploy-ecs's timeout-minutes
+  # (the poll step asserts that headroom and fails fast if it is gone).
+  DEPLOY_ROLLOUT_BUDGET_SECONDS: 1200
 
 jobs:
   build-and-push:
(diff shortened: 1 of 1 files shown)


dbrowneup:
@onder-akkaya — merged on the owner's call at 01:06 UTC before your review landed, so the ack is owed here. Your measurement chain (the waiter-budget signature, the 58e337dc silent outage) was the whole case for the fix, and it is already earning out: the first post-merge deploy failed the post-rollout answer probe on a real outage instead of reporting ECS's green rolloutState. #1532 closed with this PR. Your review request is stale, not ignored — retro-objections welcome.

PR #1567, opened 2026-08-31
[strategy] Real equal_weight/inverse_vol + realized_vol in the DSL interpreter, with audible rejections

dbrowneup:
Three defects of one shape — the DSL advertised something the interpreter did not do, silently:

1. **`equal_weight` / `inverse_vol` fell through an `else` into full-invest.** Verified against `origin/main`'s interpreter: `equal_weight` N=4 sized at 0.9900 of the account. Now real per-slot sizing.
2. **`realized_vol` validated but died at strategy construction.** Now a hand-written ddof=1 estimator pinned per-bar against the live twin at rel=1e-12 (the built-in indicator is ddof=0 — the wrong estimator).
3. **Margin-rejected orders vanished** — a silently flat run produced zero log lines. `notify_order` now names the strategy, side, size, and cash.

**Slot-invariance fixed properly (validator finding):** the old `min(weight, 1.0)` clamped *after* the slot multiply, making `inverse_vol` numerically identical to `volatility_target` on the graded sleeve path and breaking aggregate-exposure parity 2× between paths. The cap now applies to the **scale**, pre-slot; `volatility_target`'s published numbers are byte-preserved (same expression, same summation order). Honest cost, pinned not clamped away: a slot of 1.0 can request up to 2.0× — leverage the cash broker refuses, audible via `notify_order` (`test_the_slot_invariant_target_can_be_unfundable`).

Unknown keys inside `position_sizing` now reject loudly (`reference_vol` typos no longer run silently at the default). The tautological `test_spec_interprets` was replaced with a run-asserting test.

**Known open, documented not hidden:** the live evaluator (`strategy_signal_evaluator`) still returns weight=1.0 for both new types — the backtest↔live sizing divergence is pinned by an explicit test + named in the spec's Known-limitations, and the prompt guidance is adjusted; parity is the tracked follow-up (see the interpreter-unification plan's D3/D4 rows, marked in-flight).

Tests: full suite **4161 passed / 0 failed**; 219 across the seven touched suites; guards confirmed non-tautological by replay against `origin/main`'s interpreter.


dbrowneup:
## Union break with #1566, fixed on this branch

#1566's structural look-ahead audit AST-proves that every bar-indexed read in `dsl_to_backtrader.py` carries an offset `<= 0`. Its prover accepts a negative index only when the operand is **proven** non-negative: a non-negative int literal, a `range()`-bound variable whose range args are themselves proven, or a parameter with a dominating `if p < k: raise`. Anything else — a bare name, a call result, a module constant — is unproven, and unproven is treated exactly like disproved.

Three reads on this branch could not be proven. On the merged union that failed `verify_interpreter_surface()`, which fails *every* spec's LEAK leg — a total gate outage, 26 failures across `test_dsl_lookahead_audit.py` + `test_fusion_evaluator.py`:

```
line 184: self.data[...]       — unprovable bar offset [-i]
line 188: self.data[...]       — unprovable bar offset [-i]
line 462: self.data.close[...] — unprovable bar offset [-i]
```

Worth being clear that this was **not** a false positive. `line[-i]` is "i bars ago" only while `i >= 0`; for a negative `i` the same expression reads a bar that has not happened. Both loops were correct and neither was *checkable* — one bound its range on a call result, the other counted down through a range whose floor nothing checks (`range(n, -3, -1)` yields -1 and -2, and `close[1]` is the future).

### What changed (ba0f741a)

Each windowed read now takes its window length as a **parameter with a dominating precondition guard**, and counts upward from zero.

| site | before | after |
|---|---|---|
| `RealizedVolAnnualized.next` (184, 188) | `for i in range(n)` with `n = int(self.p.period)` — bound is a call result | loop moved to `_trailing_returns(self, period)` behind `if period < 1: raise DSLError(...)`; `i` then runs `0 … period-1`, so `-i` and `-i - 1` are bar `t` or earlier |
| `_sizing_realized_vol` (462) | `[self.data.close[-i] for i in range(_SIZING_VOL_LOOKBACK, -1, -1)]` — descending range over a module constant | `lookback` is now a guarded parameter (`if lookback < 0: raise`); the comprehension counts **up** over `range(lookback + 1)` and the result is reversed into the same chronological window |

Both guards are real preconditions, not audit theatre — they read correctly with the audit deleted.

**Behaviour is unchanged.** `_trailing_returns` walks the window newest-first in exactly the order the inline loop did, so the summation order (and with it the last float bit `test_realized_vol_parity_per_bar` pins per-bar) is untouched. `_sizing_realized_vol` returns the same 21-element oldest→newest list. Zero-price handling moved from an early `return` inside the loop to a `None` the caller turns into the same NaN. The reads stay on `self.data` / `self.data.close`, so **no new bar-access expression enters the module** — the audit's site inventory is byte-identical before and after this commit (12 sites, same 8 distinct expressions).

Tests, `/opt/homebrew/.../envs/archimedes/bin/pytest`, before vs after:

```
test_dsl_sizing_and_indicators + test_dsl_to_backtrader + test_dsl_momentum_convention
  + test_dsl_rsi_parity + test_strategy_dsl   90 passed  →  90 passed
test_interpreter_parity.py                    26 passed  →  26 passed
```

### Union verification

Throwaway branch off this one, merged `origin/dbrowneup/real-lookahead-audit` **and** `origin/main`:

```
pytest backend/tests/services/test_dsl_lookahead_audit.py \
       backend/tests/services/test_fusion_evaluator.py -q

before:  26 failed, 73 passed
after:   99 passed, 0 failed
```

### Adversarial: each guard demonstrably rejects something

Reverted each fix on the union and re-ran the verifier. The guards are load-bearing, not decorative:

```
CONTROL — fix in place
    sites=12  bar_access_verified=True   violations: none

A. `if period < 1: raise` REMOVED from _trailing_returns
    line 204: self.data[...] — unprovable bar offset [-i] — 'i' is not provably non-negative
    line 207: self.data[...] — unprovable bar offset [-i] — 'i' is not provably non-negative

B. `if lookback < 0: raise` REMOVED from _sizing_realized_vol
    line 513: self.data.close[...] — unprovable bar offset [-k] — 'k' is not provably non-negative

C. ORIGINAL pre-fix spellings restored at all three sites
    line 225: self.data[...] — unprovable bar offset [-i]
    line 229: self.data[...] — unprovable bar offset [-i]
    line 521: self.data.close[...] — unprovable bar offset [-i]     ← the exact three from the union
```

And the same at gate level — removing **just the one `lookback` guard**, nothing else:

```
guard removed :  25 failed, 74 passed
guard restored:  99 passed
```

### One line still owed, and it belongs on #1566

The union has a **second, independent break** that cannot be fixed from this branch, because it lives in a file this branch does not have. #1566's `test_verifier_finds_every_data_read_in_the_real_interpreter` pins a closed list of the distinct read expressions in the interpreter. `RealizedVolAnnualized` is new here and reads its own input line directly — inside an Indicator `self.data` **is** the line it was constructed with, so the read renders as `self.data[...]` with no `.close` suffix, and that string is not in the pin.

That is the pin working exactly as its docstring says it should ("if someone adds a data read the visitor does not model, this test says so"). It fails on the union **identically before and after this commit** — the inventory is `main`'s 7 entries + `self.data[...]` either way — so it is not something this change introduced or can remove. Whoever merges second applies:

```diff
             "self.data.volume[...]",
+            # RealizedVolAnnualized (#1567) reads its own input line directly:
+            # inside an Indicator `self.data` IS the line it was constructed
+            # with, so the read has no `.close` suffix to render.
+            "self.data[...]",
         ]
```

The `99 passed` above is with that one line applied on the throwaway union. Without it: `1 failed, 98 passed`, and the failure is the pin, not the audit — `bar_access_verified=True`, `violations=()`.

Issue #1556, opened 2026-08-30
[SECURITY] traces_routes read endpoints have no auth — every reasoning trace + portfolio is world-readable


dbrowneup:
**SECURITY — priority. Found by the 2026-08-30 red-team, verified at source.**

## The hole

`backend/archimedes/api/traces_routes.py` — the four READ routes have **no auth dependency at all**:
- `list_traces` (`:73`), `get_trace` (`:160`), verify (`:314`), `/canonical` (`:417`).
Only `publish_trace` (`:215`) is gated (`require_internal_agent_key`). Confirmed by reading the signatures: none carries `Depends(require_current_user)` or any owner check.

`vault_address` is a **filter query param, not a gate** — omit it and you enumerate every trace on the platform. `/canonical` returns the full hashed body including `portfolio_before` / `portfolio_after` (holdings) and `market_context`.

## Why it's not a live user-data leak *yet* — and why that's temporary

Today every trace belongs to the house agent (the agent-runner has 0 user vaults — see the agent-runner ground-truth report). So nothing user-owned leaks right now. **But** `POST /api/vaults/create` is user-facing and paper deployments are per-user: **the first user-owned vault that publishes a trace publishes its full portfolio and reasoning to the internet.** Nothing in this file would stop it. This is a landmine on the exact feature we're building tonight (user-reachable traces).

## Fix

1. Add `owner_user_id` to the off-chain trace record, written by `publish_trace`.
2. Gate `list`/`detail`/`verify`/`canonical` on `is_owner OR vault_is_public`.
3. **Ship-safe interim (do this first, tonight-cheap):** restrict `list_traces` to an allowlisted set of house vault addresses and require `vault_address` on every read, so enumeration is impossible before the owner model lands.

## Acceptance
- An anonymous `GET /api/traces/?limit=100` returns only house-public traces (or 401), never a user's.
- A non-owner cannot read another user's trace by id.
- Guard demonstrated: a seeded user-owned private trace is unreachable anonymously and by a different user.

Related: the reachable-traces work (dbrowneup/user-reachable-traces) MUST NOT merge before the interim gate; #1547-family visibility.

PR #1569, opened 2026-08-31
[backend][frontend] Make reasoning traces reachable from the strategy passport !minor

dbrowneup:
Reasoning traces become reachable from the strategy that moved money: the passport gains an on-chain traces panel with a scoped listing (`?strategy_id=`), the trace detail deep-links back to the passport, and three drifted UI anchor-state ternaries collapse into one `anchorState()` with four honest states.

**Built ON TOP of the #1556 gate, not beside it (this branch merged that work in before #1562 landed on main):** the widened detail body (`portfolio_before/after`, `market_context`, `ipfs_cid`) is served **inside** the ownership gate — `assert_strategy_visible` answers "may you know this strategy exists," `trace_visibility.can_read_trace` answers "may you read this row," and both run. Guard tests live in the gate's own test file, reusing its helpers, including the leaks-without-the-gate control (non-tautological) and the owner-reads-everything positive control.

Validator must-fixes, all closed with revert-demos:
- **Empty-string bypass dead:** `?strategy_id=` no longer skips the gate and returns the whole feed (min-length constraint; probe now a test).
- **`strategies_referenced` honesty:** the filter is scoped to decision traces with exact-match semantics (`trace_references_strategy()` documents the three real writers — agent-runner strategy ids vs paper anchors vs arXiv ids); bare-string substring-match and dict key-match latencies pinned closed; the panel copy now says exactly what is matched.
- Portfolio.jsx take-HEAD resolution subsumes main's #714 fix, pins its exact title string, and fixes its ordering bug; the stale "back to the Library" intro copy fixed with a test that actually covers it; the stale "no auth" comment corrected and the raw `fetch` moved to `apiGet`.
- `total` vs post-window filtering reconciled so the panel's "(n of m)" can't claim unreachable rows.

Tests: full suite **4190 passed / 0 failed**; ui **479 passed**; 57 across the scope/gate files; eslint clean.


dbrowneup:
## Union break with #1573 — fixed in `7129d75b`

`origin/main` is merged in, and the merge surfaced a collision with **#1573** (`backend/tests/services/test_trace_owner_cache.py`, the trace-owner memo guard suite). Two failures, one cause.

```
FAILED tests/services/test_trace_owner_cache.py::test_a_fifty_row_listing_resolves_each_distinct_vault_once
        - AssertionError: expected one database round trip across three listings, got 0
FAILED tests/services/test_trace_owner_cache.py::test_a_private_legacy_row_stays_private_on_the_warmed_listing
        - AssertionError: assert [] == ['trace-99']
```

### The break

#1573 landed that file on `main` with a fake `AgentStateStore.list_traces`:

```python
async def _list(vault_address=None, decision_type=None, limit=20, offset=0):
```

This branch inserted `strategy_id` into the real signature and passes it on **every** call, filtered or not:

```python
off_chain_traces, _ = await state.list_traces(
    vault_address=vault_address,
    decision_type=decision_type,
    strategy_id=strategy_id,   # ← added here
    limit=MAX_TRACE_SCAN,
    offset=0,
)
```

Neither side is wrong on its own. Both were green on their own branches — this is a textually-clean semantic conflict across a **shared function**, the exact shape CLAUDE.md rule 5 describes. (#1569 *did* update the sibling fake in `tests/api/test_traces_ownership_gate.py`; `test_trace_owner_cache.py` did not exist on this branch's base to update.)

### Why it failed the way it did — the interesting part

The `TypeError` never escaped. That call sits inside `except Exception:` which logs **"Redis unavailable — falling back to on-chain-only listing"**, so a stale fake does not error out — it silently re-routes every listing test onto the on-chain fallback, where there are no off-chain rows and `safe_resolve_vault_owners` is never consulted at all. Captured in the log of all five listing tests:

```
WARNING archimedes.api.traces_routes: list_traces: Redis unavailable — falling back to on-chain-only listing
TypeError: _store.<locals>._list() got an unexpected keyword argument 'strategy_id'
```

Both symptoms then read as *improvements*:

| test | reported | reads as | actually |
| --- | --- | --- | --- |
| round-trip economy | `0` lookups, expected `1` | a cache that got **better** | the resolver never ran |
| private legacy row | `[]`, expected `["trace-99"]` | a gate that got **stricter** | the listing never ran |

**On the privacy question specifically — there was no leak, and the direction matters.** `trace-99` is the *house-vault* row, the one that is supposed to be visible; the 50 private user rows are `trace-0…49`. The fallback returned **nothing**, so the house row vanished *along with* the private ones and `assert SECRET_REASONING not in resp.text` held throughout. The failure was fail-**closed** over-restriction, not disclosure. The #1562 model (write-time stamp → vault-owner lookup → `PUBLIC_TRACE_VAULTS` floor) was never reached, let alone weakened. Flagging that plainly because the opposite reading would have been the scarier one.

### The fix

`backend/tests/services/test_trace_owner_cache.py` only. **No assertion relaxed, no production code touched.**

1. The fake mirrors the real signature, and honours `decision_type` / `strategy_id` by **delegating to `trace_references_strategy`** rather than keeping a second copy of the provenance rule in a test double.
2. `_assert_mirrors(_list, AgentStateStore.list_traces)` runs at patch time and compares parameter **names and order** (order too — the route is free to call positionally). The next widening of that signature now fails with `the fake store has drifted from AgentStateStore.list_traces: [...] != [...]`, naming the missing parameter, instead of quietly emptying the page.

Point 2 is the durable half: the swallow-as-outage behaviour means *any* future signature change re-creates this silent hole, and a boundary double that only covers its own branch's calls is precisely the corollary CLAUDE.md rule 5 warns about.

### Adversarial evidence

**1 — revert the fix, the break reproduces.** `git show HEAD~1:…test_trace_owner_cache.py` over the fixed file:

```
E   assert 0 == 1                       # "expected one database round trip across three listings, got 0"
E   AssertionError: assert [] == ['trace-99']
TypeError: _store.<locals>._list() got an unexpected keyword argument 'strategy_id'   (×4, in the captured log)
2 failed, 11 passed
```

**2 — the new guard is not decorative.** Keep `_assert_mirrors`, drop *only* `strategy_id` from the fake:

```
E   AssertionError: the fake store has drifted from AgentStateStore.list_traces:
      ['vault_address', 'decision_type', 'limit', 'offset']
   != ['vault_address', 'decision_type', 'strategy_id', 'limit', 'offset']
```

Five listing tests fail with the drift **named**, instead of two failing with a misleading empty page. That is the whole point of the change.

**3 — restored, green.**

```
$ pytest tests/services/test_trace_owner_cache.py tests/api/test_traces_ownership_gate.py \
         tests/services/test_trace_ownership_stamp.py tests/test_api_routes.py -q
87 passed

$ pytest tests/test_traces_strategy_scope.py tests/test_traces_display_anchored_only.py \
         tests/api/test_traces_routes.py tests/services/test_redis_state.py \
         tests/services/test_redis_json_guard.py tests/services/test_construction_trace.py -q
135 passed

$ pytest -m "not integration" -q          # the CI command, from the repo root
4477 passed, 2 skipped, 2 deselected in 376.63s

$ ruff format --check + ruff check        # clean
```

Env: `/opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/pytest` (Python 3.12.13), per CLAUDE.md § Setup.

Issue #1522, opened 2026-08-30
[deps][ci] Split requirements-base.txt out so environment.yml can include it instead of restating 32 floors


onder-akkaya:
## Summary

One fact — *which lines in `backend/requirements.txt` are the CPU-only torch block* — is
written down in four places, and one file duplicates 32 version floors outright because it
cannot include the file that owns them.

`backend/requirements.txt:99` carries `--extra-index-url https://download.pytorch.org/whl/cpu`.
That single line is what forces every other consumer to work around the file rather than
read it:

| consumer | what it does today | why |
|---|---|---|
| `backend/Dockerfile:15-16` | installs the file as-is | wants the CPU wheel; correct |
| `quality-gate.yml:134` and `:313` | `grep -vE` strips the index URL, `torch`, `sentence-transformers` into `/tmp/req-ci.txt` | keeps the required check off the pytorch CDN. The same regex, copied twice |
| `environment.yml` pip block | restates 32 floors by hand | `- -r backend/requirements.txt` would drag the CPU wheel onto dev machines and lose MPS on Apple Silicon |
| `backend/tests/test_env_requirements_alignment.py` | asserts the two files agree | exists only because of the row above |

The duplication is not theoretical. When I added that guard in #1493, **18 of the 34 shared
packages had drifted, every one with the local floor below the deployed one** — including
`pandas>=2.2` against a production floor of `3.0.5`, a major version apart on the code path
that grades strategies. The guard converts that from silent to loud, which is worth having,
but it treats the symptom. Dependabot edits `requirements.txt` and only `requirements.txt`,
so a human has to mirror every bump by hand, forever.

Extracting the 32 non-torch packages into a file both consumers can include removes the
duplication instead of policing it, and deletes the copied grep from CI as a side effect.

## Scope

Do exactly this, nothing more.

1. **New `backend/requirements-base.txt`** — the 32 packages currently in
   `backend/requirements.txt` excluding `--extra-index-url`, `torch`, and
   `sentence-transformers`. Move the comments with their pins; do not reword them.
2. **`backend/requirements.txt`** becomes `-r requirements-base.txt`, then the
   `--extra-index-url` line, then `torch` and `sentence-transformers`. The Docker install
   path is unchanged in behaviour.
3. **`backend/Dockerfile`** — add `COPY backend/requirements-base.txt .` next to line 15.
   Without it the `-r` fails at build time.
[...]

PR #1572, opened 2026-08-31
[deps][ci] pymdown-extensions triage: direct pin infeasible, document + formally accept (#1490)

### LINKED ISSUE #1490: [deps] The permanently-red pip-audit row is pymdown-extensions, shipped in the backend image via titanoboa
## What the red pip-audit row actually is

Every PR carries `pip-audit (backend, python deps) ❌ not passed | 2` in the supply-chain
table. That `2` is not a package count and it is not the vyper pair — `quality-gate.yml`
already passes `--ignore-vuln PYSEC-2023-142 --ignore-vuln PYSEC-2025-33`, and the counter
sums vulns rather than packages. So the row is, today, exactly two advisories against one
package:

| Package | Version | Advisory | Fix |
| --- | --- | --- | --- |
| `pymdown-extensions` | 10.21.3 | PYSEC-2026-3609 | 11.0.0 |
| `pymdown-extensions` | 10.21.3 | PYSEC-2026-3654 | 11.0.1 |

## It is in the deployed image

```
circle-titanoboa-sdk[x402,wallets]   backend/requirements.txt:70 (pinned git dep)
  └── titanoboa 0.2.8
        └── mkdocs-material 9.5.41        <- a runtime dep of titanoboa, not a docs extra
              └── pymdown-extensions 10.21.3
```

`backend/Dockerfile:16` installs from `backend/requirements.txt`, so the whole mkdocs
toolchain ships in the backend image. My first read of this was that it was local docs
tooling outside the image; that was wrong, and I checked it rather than assuming — a control
run of `pip-audit -r` against a one-line requirements file reports nothing, so the finding
comes from this repo's own resolution and not from a polluted local env.

## Reachability

Low. Nothing in `backend/archimedes/` imports `pymdownx` or `markdown` — the package is
shipped but never loaded, so there is no path from a request to the vulnerable code. This is
a permanently red gate row and image bloat rather than a live exposure, which is why this is
an issue and not a hotfix.

## Reproduce

```bash
pip-audit -r backend/requirements.txt --progress-spinner off
pip show titanoboa | grep Requires        # mkdocs-material listed as a runtime dep
```

## Options, in the order I'd try them

1. **Direct transitive pin** in both `backend/requirements.txt` and `environment.yml`, the
   pattern already used for `starlette>=1.4.1`. Caveat worth checking before committing to
   it: mkdocs-material 9.5.x constrains `pymdown-extensions` to the 10.x line, so a pin to
   `>=11.0.1` may be unsatisfiable without also moving mkdocs-material. Resolve before
   pinning rather than after.
2. **Document the triage verdict** next to the dep, exactly as the vyper analysis is
   documented at `backend/requirements.txt:59-68`. If the conclusion is "shipped but
   unreachable, no fix path", that belongs written down where the next reader meets it —
   the vyper note is the reason nobody re-litigates those two advisories every week.
3. **Upstream**: `titanoboa` declaring `mkdocs-material` as a runtime rather than a docs
   extra is the root cause. Worth an upstream issue regardless of what we do locally, since
   it puts a docs toolchain in every consumer's production image.

## Why this is worth closing rather than muting

A gate row that is always red teaches everyone to ignore it, and the next real CVE arrives
into a row nobody reads. Either fix or a written verdict restores the signal. Adding another
`--ignore-vuln` without the analysis would be the one option that does not.

Found while confirming the red rows on #1489 were not introduced by that PR (they were not).

dbrowneup:
Closes #1490 — and the fix is deliberately NOT the one the issue proposed, with the receipts.

**The starlette-pattern direct pin is infeasible, confirmed empirically:** `mkdocs-material` 9.5.41's wheel metadata declares `pymdown-extensions~=10.2` (<11), and `titanoboa` 0.2.8 — what `circle-titanoboa-sdk`'s floor actually resolves to — declares `mkdocs-material==9.5.41`, an **exact** pin. A real `pip install` in a throwaway venv produces the resolver conflict; neither the leaf nor its parent can move without titanoboa changing its own pin. Also confirmed: titanoboa's 51 source files never import mkdocs/pymdownx — it's a mis-declared docs-toolchain dep riding into the runtime image, unreachable at runtime.

**The smallest honest fix:** the full triage verdict documented as a comment block (mirroring the vyper block above it), and the two advisories added to quality-gate's pip-audit `--ignore-vuln` list beside the existing vyper pair — turning the permanently-red supply-chain row green as *assessed-and-accepted*, not ignored. Upstream note recorded: mkdocs-material 9.7.7 has since uncapped, so the real fix is titanoboa dropping its exact pin — worth an upstream nudge.

**2026-09-01 refresh:** this branch was 2 days stale against `main` and had gone `CONFLICTING` — #1522 (merged after this PR's last push) split the 32 shared floors out of `backend/requirements.txt` into a new `backend/requirements-base.txt` that both `backend/requirements.txt` and `environment.yml` now pull in with `-r`. The triage comment block moves with it: it now lives in `backend/requirements-base.txt` (next to the titanoboa pin it explains), not `backend/requirements.txt`, and there is no longer a separate mirror note needed in `environment.yml` — that file no longer restates any shared floor by hand (per `backend/tests/test_env_requirements_alignment.py`'s "only torch/sentence-transformers may be duplicated" rule; confirmed the merged `environment.yml` is now byte-identical to `main`'s). Re-verified the finding itself is unchanged, not just stale prose: re-ran `pip-audit` against current `main`'s resolved requirements (same 4 advisories / 2 packages), re-downloaded `titanoboa==0.2.8` and `mkdocs-material==9.5.41` wheels and re-read their `Requires-Dist` metadata (same exact pin / same `~=10.2` cap), and re-ran the resolver-conflict repro in a fresh venv (same `ResolutionImpossible`). `mkdocs-material` latest on PyPI is still 9.7.7 — no upstream movement since the original triage.

⚠️ **Ask-before-acting disclosure:** this PR touches `.github/workflows/quality-gate.yml` — on the CLAUDE.md ask list. (`environment.yml` is no longer touched by this PR after the #1522 rebase — see above.) This PR *is* the ask: merging it is the owner formally accepting the two-CVE ignore on an unreachable dep.

## Verification

Baseline (current `main`, `backend/requirements.txt` resolved through the new `-r requirements-base.txt`): `pip-audit -r backend/requirements.txt` reproduces the issue's table exactly — 4 vulns / 2 packages (`pymdown-extensions` PYSEC-2026-3609 + PYSEC-2026-3654, `vyper` PYSEC-2023-142 + PYSEC-2025-33).

With this PR's 4 `--ignore-vuln` flags: exit 0, computed count 0.

Adversarial check on the ignore list itself: dropped `PYSEC-2026-3609` from the flags and re-ran — exit 1, exactly `pymdown-extensions: PYSEC-2026-3654` reported. The ignore list is not a blanket suppression; each ID gates independently.

`backend/tests/test_env_requirements_alignment.py`: 6/6 pass against the merged tree (comment-only addition, no floor lines touched, so the parser sees no change).

Full unit suite (`pytest -m "not integration" -q"`): 5556 passed / 3 skipped / 2 deselected, 1 failure (`test_health_always_answers.py::...test_a_stalled_local_read_still_answers_inside_the_budget[regime_detector]`, a `/health` latency-budget assertion) under heavy concurrent CPU load from 3 other full-suite runs sharing this box at the time. Re-ran that single parametrized test in isolation immediately after: 6/6 pass, the `regime_detector` case at 0.81s (vs. the reported 4.32s under load). This PR's diff touches zero executable code — only comments in `backend/requirements-base.txt` and two `--ignore-vuln` strings + a path reference in `quality-gate.yml` — so there is no mechanism by which it could affect `/health` timing; the failure is a load-induced flake in a pre-existing test, not a regression from this change.

The two advisories stay suppressed until titanoboa upstream drops its exact `mkdocs-material==9.5.41` pin (or moves it to a docs extra) — there's no repo-local fix left to try. If that's not an acceptable standing state, the alternative is vendoring/forking titanoboa's `mkdocs-material` requirement, which is a materially bigger change than this PR and wasn't attempted.


danielscoffee (approved):
Small docs/CI change: documents why a direct pymdown-extensions pin is infeasible and formally accepts the pip-audit row. No runtime image change. CI clean.

PR #1576, opened 2026-08-31
[perf][backend] Memoize trace ownership resolution — the proof surface was loading in 21s (#1573)

### LINKED ISSUE #1573: [perf][security] #1562's per-row ownership resolution makes GET /api/traces/ take ~21s on the public proof surface
Measured live 2026-08-31 right after the #1562 deploy: `GET /api/traces/?limit=50` → 200 in **21.2s** (health: 2.8s). Cause: `trace_visibility`'s legacy-row owner resolution (vault_metadata + identity_events lookups, sync SQLAlchemy inside the async route) runs **per row**, uncached — 50 rows over only 5 distinct vault addresses. /reasoning and /quant-lab render this listing anonymously, so the public proof surface now loads in ~21s. Fix shape: memoize owner resolution per vault_address (in-process TTL cache; the mapping is stable — a vault's owner doesn't change), resolve each distinct vault once per request at minimum, and move the sync session off the event loop (asyncio.to_thread) if it stays. Guard: a test asserting the resolver is called once per distinct vault, not once per row. Non-goal: no weakening of the gate — same verdicts, fewer lookups. Found during post-deploy verification; fix builder dispatched.

dbrowneup:
Closes #1573.

## Evidence

`GET /api/traces/?limit=50` measured live at **21.2s** right after #1562 deployed (health: 2.8s), over only **5 distinct vault addresses**. `/reasoning` and `/quant-lab` render that listing **anonymously**, so the public proof surface — the page the product's central claim rests on — loaded in ~21s.

The read gate re-derived the vault → owner mapping on **every request**: a synchronous `vault_metadata` query plus, for any vault it could not resolve, a `MAX_TRACE_SCAN` (2000) row `identity_events` scan materialized as ORM objects — executed **inside the async event loop**, where it stalls every other in-flight request. The listing batched within a request but nothing survived the request boundary; the three detail routes were worse, opening a session per unstamped read (`can_read_trace`).

Local timing, tmp SQLite seeded with 2000 `vault_created` events and 5 unresolvable vaults (the expensive branch), throwaway script, not committed:

```
uncached avg :    23.04 ms   <- pre-#1573: paid on every request, on the event loop
warm avg     :   0.0025 ms   <- post-#1573: cache hit
```

SQLite understates it by a lot — on Aurora the same call is a network round trip plus `pool_pre_ping`, and it was serialized against the event loop rather than a worker thread.

## What changed

| | |
| --- | --- |
| `safe_resolve_vault_owners` (**read** path) | memoized per lowercased vault for `VAULT_OWNER_CACHE_TTL_SECONDS = 300`. Only vaults that MISS reach the database. |
| Negative results | cached too — an unresolvable vault is what triggers the `identity_events` scan, so caching only successes would leave exactly the slow path uncached. |
| `resolve_vault_owners` (**write** path) | deliberately **left uncached**. It backs `save_trace`'s ownership stamp, which is permanent: a stale answer there is written into the record rather than expiring. |
| `TraceOwnerLookupIncomplete` | keeps "this vault has no owner" distinguishable from "the database did not answer", so a DB outage is never negative-cached and cannot outlive itself by a TTL. |
| `asyncio.to_thread` | wraps the resolution in both `list_traces` branches and in `_assert_can_read`, so the synchronous session no longer blocks the loop. |
| `invalidate_vault_owner` | called from the two writes that CREATE ownership — `POST /api/vaults/create`'s `vault_created` event and `POST /api/vaults/metadata`'s commit. |

## Why the TTL cannot weaken the gate

Same verdict for every row — the memo returns the mapping the uncached resolver would return. The staleness analysis, by direction:

- **A stamped row never consults the resolver at all.** Every trace written since #1556 carries its owner on the record, so the entire population written from here on is outside the cache's blast radius. Guarded by `test_a_stamped_listing_never_touches_the_resolver` and `test_can_read_trace_skips_the_resolver_for_a_stamped_row`.
- **Stale positive** (owner A → owner B): the row stays private, to A rather than B. Private-to-someone, never public. Owner changes are essentially write-once anyway — a vault gets its owner at deploy and does not change.
- **Stale negative** (unowned → owned) is the only widening direction, and it is the one closed at the write: `invalidate_vault_owner` drops the entry the moment either ownership-establishing write commits. The TTL is the backstop for that write landing in a sibling Fargate task, not the primary mechanism. With `PUBLIC_TRACE_VAULTS` armed the floor makes even that window fail **closed** (an unowned row outside the list is private).

## Guards — demonstrated to fail

Per CLAUDE.md § "A test that passes against the unfixed code proves nothing". 13 new tests in `backend/tests/services/test_trace_owner_cache.py`. Neutralizing **only** the memo (`VAULT_OWNER_CACHE_TTL_SECONDS = 0`, i.e. exactly the pre-#1573 behaviour: always re-resolve) and re-running:

```
6 failed, 7 passed
FAILED test_a_fifty_row_listing_resolves_each_distinct_vault_once
       - AssertionError: expected one database round trip across three listings, got 3
FAILED test_repeated_detail_reads_of_a_legacy_row_resolve_its_vault_once
       - AssertionError: expected one database round trip across ten detail reads, got 10
FAILED test_the_unresolvable_vault_is_negative_cached
       - AssertionError: the negative result was not cached: 3 round trips
FAILED test_only_the_uncached_vaults_reach_the_database
       - AssertionError: warm vaults were re-resolved
FAILED test_a_cache_hit_returns_the_same_verdict_as_a_cold_resolve
FAILED test_becoming_owned_invalidates_the_memo
       - AssertionError: sanity: the memo is real, or the next line proves nothing
```

With the memo restored: **13 passed.** The spy (`_count_db_lookups`) wraps the one function that actually opens a session and delegates to the real implementation, so it counts the database path rather than replacing it. `test_a_cache_hit_returns_the_same_verdict_as_a_cold_resolve` computes every verdict cold, then makes the DB path raise so the second pass can only be answered from the memo, and asserts the two verdict vectors are identical for owner / non-owner / anonymous — with `assert any(cold)` and `assert not all(cold)` so a degenerate all-same vector cannot pass it.

`_clear_vault_owner_cache` was added as an autouse fixture in `backend/tests/conftest.py`, mirroring the existing `_clear_rigor_cache`: several test files reuse the same vault addresses against different per-test SQLite databases, and without the reset an ownership *verdict* would depend on suite order.

`test_stamped_record_stays_private_when_the_identity_db_is_down` now patches `_resolve_vault_owners_uncached` as well — patching only `resolve_vault_owners` would have left the read path talking to a live database and quietly stopped simulating the outage.

## Tests

- `backend/tests/services/test_trace_owner_cache.py` → **13 passed** (new)
- `backend/tests/api/test_traces_ownership_gate.py` → **20 passed** (the #1556 gate, unchanged verdicts)
- Full unit suite, `pytest -m "not integration" -q` → **4350 passed, 2 skipped, 5 failed**

The 5 failures are **pre-existing on `origin/main`** and untouched by this branch — `test_trace_ownership_stamp.py`'s five `save_trace` cases fail with `TypeError: object MagicMock can't be used in 'await' expression` (its fake Redis lacks the `srem`/`hdel` the #1353 reconcile-index maintenance awaits). Verified by running that file on a clean checkout of `origin/main`: same 5 failures. Fixed by #1565 (`dbrowneup/stamp-test-union-fix`), which is not on `main` yet. The new tests here are deliberately independent of that file.

## Not fixed here — the other term in the 21.2s

Flagging honestly rather than claiming the page is now fast: `AgentStateStore.list_traces` (`services/redis_state.py`) does `zrevrange(index, 0, -1)` and then **one sequential `GET` per trace hash in the whole index**, windowing only afterwards. That is O(index size) awaited Redis round trips against TLS ElastiCache on every listing, and it predates #1562 (`list_recent_traces` was given an index-level bound for exactly this reason in #1276, but `list_traces` was not). Depending on how many traces are in the index it is plausibly the larger share of the 21.2s. Worth a follow-up issue and a live re-measurement after this deploys, which will separate the two terms.

### DIFF
--- a/backend/archimedes/api/traces_routes.py
+++ b/backend/archimedes/api/traces_routes.py
@@ -128,19 +128,27 @@ def _offchain_trace_response(t: dict, fallback_id: str = "") -> TraceResponse:
     )
 
 
-def _assert_can_read(trace: dict, request: Request) -> None:
+async def _assert_can_read(trace: dict, request: Request) -> None:
     """404 unless the caller may read this trace (#1556).
 
     404 rather than 403 deliberately: a 403 on someone else's trace id
     confirms the id exists, which is half of the enumeration the gate is here
     to prevent.
+
+    ``can_read_trace`` may open a synchronous SQLAlchemy session (only for an
+    unstamped legacy row — a stamped one is answered from the record), so it
+    runs in a worker thread: a blocking DB call made directly here stalls the
+    whole event loop, every other in-flight request included (#1573).
     """
+    import asyncio
+
     from fastapi import HTTPException
 
     from archimedes.services.trace_visibility import can_read_trace
 
     caller_user_id, caller_wallet = _caller_identity(request)
-    if not can_read_trace(trace, caller_wallet, caller_user_id=caller_user_id):
+    allowed = await asyncio.to_thread(can_read_trace, trace, caller_wallet, caller_user_id=caller_user_id)
+    if not allowed:
         raise HTTPException(status_code=404, detail="Trace not found")
 
 
@@ -159,6 +167,8 @@ async def list_traces(
     stays a *filter* — it narrows the visible set, it never widens it — so
     dropping it can no longer enumerate the platform.
     """
+    import asyncio
+
     from archimedes.services.redis_state import AgentStateStore
     from archimedes.services.trace_visibility import (
         MAX_TRACE_SCAN,
@@ -188,12 +198,18 @@ async def list_traces(
         if off_chain_traces:
             # One batched ownership lookup for the whole page, and only for rows
             # that lack an on-record stamp — a stamped row needs no DB at all.
-            owners = safe_resolve_vault_owners(
+            # `safe_resolve_vault_owners` memoizes per vault, so the handful of
+            # distinct vaults behind a 2000-row candidate set costs at most one
+            # lookup each per TTL rather than one per request; `to_thread` keeps
+            # the synchronous session it may still open off the event loop,
+            # where it was blocking every other in-flight request (#1573).
+            owners = await asyncio.to_thread(
+                safe_resolve_vault_owners,
                 {
                     str(t.get("vault_address") or "")
                     for t in off_chain_traces
                     if not (t.get("owner_user_id") or t.get("owner_wallet"))
-                }
+                },
             )
             visible = [
                 t
@@ -238,7 +254,7 @@ async def list_traces(
             # still names the vault and the anchor — gate it on exactly the
             # same predicate so a caller cannot enumerate another user's
             # vault's trace ids by taking Redis out of the picture.
-            owners = safe_resolve_vault_owners({str(d["vault"]) for _, d in on_chain_details})
+            owners = await asyncio.to_thread(safe_resolve_vault_owners, {str(d["vault"]) for _, d in on_chain_details})
             for trace_id, detail in on_chain_details:
                 view = trace_owner_view({"vault_address": detail["vault"]}, owners)
                 if not is_trace_visible(view, caller_wallet, caller_user_id=caller_user_id):
@@ -272,7 +288,7 @@ async def get_trace(trace_id: str, request: Request):
             logger.warning("get_trace: Redis unavailable — falling back to on-chain-only lookup", exc_info=True)
             off_chain = None
         if off_chain:
-            _assert_can_read(off_chain, request)
+            await _assert_can_read(off_chain, request)
             return _offchain_trace_response(off_chain, fallback_id=trace_id)
 
         try:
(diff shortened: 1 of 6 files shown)


dbrowneup:
CI status: the only red check is **Backend — unit tests**, and it is the *pre-existing* `main` breakage, not this branch.

CI reports `5 failed, 4355 passed` — the five are `backend/tests/services/test_trace_ownership_stamp.py`'s `save_trace` cases failing with `TypeError: object MagicMock can't be used in 'await' expression`. Same five fail on a clean checkout of `origin/main` with this branch nowhere in the picture. **#1565** (`dbrowneup/stamp-test-union-fix`, OPEN + MERGEABLE) is the fix; merge it first, then re-run this PR's checks against the regenerated base.

Every test this PR adds or touches is green: `test_trace_owner_cache.py` 13 passed, `test_traces_ownership_gate.py` 20 passed.

PR #1591, opened 2026-08-31
[backend][provenance] Wire paper-trading decisions into the commit-reveal trace pipeline

### LINKED ISSUE #1575: [backend][provenance] Wire paper-trading decisions into the commit-reveal trace pipeline (the MVP-completeness gap)
The product's core claim is auditable reasoning behind every move. Tonight's ground truth: the agent-runner ticks against zero live vaults, and **paper trading — the surface users actually touch — produces no reasoning traces at all** (it's a separate replay engine writing `paper_daily_returns`; #1568's intraday marks re-price and deliberately never re-decide). So a user's paper deployment moves (paper) money with no hashed, owner-stamped, verifiable trace. This is the biggest MVP-completeness gap (2026-08-30 assessment) and the heart of "get the strategy generation and execution systems working."

## What
1. **Short design doc first** (docs/plans/): where a paper "decision" is born — the settle-path rebalance boundary (NOT marks; marks never decide) — and what its trace carries (spec-derived decision, inputs consulted, portfolio before/after, the same shape the house agent publishes).
2. **Publish through the existing `save_trace` choke point** so #1556's owner stamps land for free (`owner_user_id`/`owner_wallet` from the deployment), and #1569's passport panel reaches them for free (`decision_type` conforming, `strategies_referenced` = the strategy id).
3. **Hashing per the commit-reveal spec** (`docs/specs/commit-reveal-trace-spec.md`): canonical body + keccak, off-chain verifiable. On-chain anchoring for paper traces is a design decision to argue in the doc (cost + consent), default OFF behind the existing anchor path.
4. **Guards, adversarial per house rules**: a paper decision with no trace must fail loudly somewhere visible (fail-soft rule: absence must be loud); a trace that doesn't re-derive its hash must be detectable; non-owner reads blocked (reuses #1556 tests' shape).

## Acceptance (machine-checkable)
- A seeded paper deployment advancing one settle produces exactly one trace: `pytest backend/tests/test_paper_trace_pipeline.py -q` → 0 failed, and the trace round-trips `verify` (hash re-derives).
- `GET /api/traces/?strategy_id=<the strategy>` (as owner) returns it; anonymous → 404-shape.
- Guard demo in the PR body: pipeline with publishing disabled → the loud failure, shown.

## Anti-goals
- No live-execution path changes; PAPER only. No schema fork of the trace record. No new on-chain writes by default. Do not slow the settle path unboundedly (publish is async-tolerant, but its failure is never silent).

Refs: #714 (anchors), #1556/#1562 (ownership), #1568 (marks), #1569 (reachability), docs/specs/commit-reveal-trace-spec.md, docs/plans/2026-08-30-intraday-paper-trading.md.

dbrowneup:
Implements the design in #1578 (`docs/plans/2026-08-30-paper-trading-reasoning-traces.md`), build steps 1–6.

**Stacked on the design branch.** Base is `dbrowneup/1575-paper-trace-design` (PR #1578, docs-only). Merge #1578 first, or retarget this at `main` after it lands — the only code-relevant carry-over is the design doc itself, which this PR also amends (§9.3 now names the real revision id and the `provenance` column the drift check needs).

Closes #1575.

## Evidence

```
$ pytest -m "not integration" -q
4386 passed, 3 skipped, 2 deselected in 380.12s

$ pytest backend/tests/test_paper_trace_pipeline.py -q       # the issue's named acceptance file
17 passed, 1 skipped        (the skip is G5's #1569 half — see Sequencing)

$ ruff format --check .   →  641 files already formatted
$ ruff check <changed>    →  All checks passed!
$ make docs-check         →  0 broken links of 1306; 151/151 indexed, 0 orphaned
$ alembic heads           →  ['d7c41f9b2e58']   (single head)
```

## The three acceptance criteria

**1. One settle advance → exactly one trace.** `test_one_settle_advance_produces_exactly_one_trace`: 1 decision, 1 trace, **3 ledger rows**. Keying on the *date* is the point — tracing per ledger row would manufacture ~21 fake decisions per real one for a monthly strategy, and per symbol leg would fork one move into several.

**2. The hash round-trips.** `test_the_trace_round_trips_its_hash_through_the_canonical_route` fetches `/api/traces/{id}/canonical` as the owner and re-runs the keccak: it equals the stored `trace_hash`. `/verify` is asserted in the same test to report `is_verified: false` / "not published on-chain" — an unanchored paper trace returning a green check would be the exact fabrication this issue exists to remove.

**3. Owner-gated reads.** `test_owner_reads_the_paper_trace_and_nobody_else_does`: owner gets 200 and the row in their list; a different signed-in user and an anonymous caller get 404 on detail/canonical and an empty list with `total: 0`. Asserted on *content* (a distinctive symbol string) as well as status codes.

## Guard demo — publishing disabled (the issue's explicit ask)

```
$ PAPER_TRACE_PUBLISH=off  →  one settle advance
result            {'decisions': 1, 'published': 0, 'disabled': 1, 'appended': 3}
store             {}                                    # nothing published
paper_decision_traces  [('disabled', None)]             # durable row, no trace id
log (WARNING)     "...NOT traced — PAPER_TRACE_PUBLISH is off. The deployment's
                   provenance claim has a hole in it for this decision..."
trace_coverage    {'status': 'disabled', 'decisions': 1, 'published': 0,
                   'disabled': 1, 'first_gap_at': '2026-08-04'}
```

`appended: 3` is deliberate: the ledger keeps advancing. Freezing every user's paper history behind a switch trades a visible gap for an invisible stall. Boot logs an ERROR once, and `test_a_disabled_gap_heals_on_the_next_settle` shows the gap closing itself without double-publishing.

## Adversarial pass — 12 mutations, each shown to fail

Per CLAUDE.md § "A guard must be shown to reject something", every guard was run against the input that should break it. Each mutation was applied to `main`-side source, the named tests run, then reverted.

| # | Mutation | Result |
|---|---|---|
| M1 | `vault_address` → the zero-address sentinel | FAILS |
| M2 | owner stamp dropped from the save payload | FAILS |
| M3 | `decision_type` → `"paper_rebalance"` | FAILS |
| M4 | coverage `status` from a bare `published > 0` | FAILS |
| M5 | publishing-disabled reports success | FAILS |
| M6 | idempotency key lookup disabled | FAILS |
| M7 | coverage accounting identity does not raise | FAILS |
| M8 | `cash_before` via `executed.value` (the sell bug) | FAILS |
| M9 | store outage swallowed as success | FAILS |
| M10 | unownable deployment publishes anyway | FAILS |
| M11 | provenance dropped from the hashed context | FAILS |
| M12 | trace id back to `uuid4()` | FAILS |

**Two of these did not fail on the first run, and both were real holes in my tests, not in the mutation:**

- **M6** — the idempotency test asserted "one distinct trace id in the store". Because the trace id is derived from the decision key, that is true *even when the writer republishes every settle*: a tautology. Now the fake store counts `save_trace` **calls** and the test asserts `store.saves == 1`.
- **M7** — the miscount test accepted `KeyError` alongside `PaperTraceCoverageError`, so it passed on a `counts[status]` blow-up without the identity ever running. The publisher's status is now bucketed with an explicit membership check so an unrecognised status falls *through* to the identity, and the test pins `PaperTraceCoverageError` with a message match.

## Two bugs the adversarial pass surfaced in the implementation

1. **`order.executed.value` is not signed by direction.** On a *closing* order backtrader reports the *opened* position's value — positive, at the entry price — so `cash_before = cash_after + value + comm` overstated a sell's pre-trade cash by roughly the whole position. Corrected to the signed `size * price`, verified against a real run, and pinned by `test_cash_and_position_bracket_every_leg` (which asserts both directions are exercised, so it cannot pass on buys alone).
2. **`trades_executed` keyed on `"side"` 500s `GET /api/traces/`.** `TradeExecutedResponse` requires `symbol`/`direction`/`amount`. A leg the wire schema cannot validate makes the trace unreachable just as surely as a non-conforming `decision_type` — same conformance argument, louder failure. Legs now emit the conforming keys with the paper-specific numbers alongside (inside the hash; ignored by the wire schema). `value_usdc` is deliberately left unset — paper legs are not denominated in USDC.

## Design decisions carried through

- **`vault_address=""`, never `UNBOUND_VAULT`.** `test_the_zero_address_sentinel_would_leak_the_same_body` rewrites the identical record onto the sentinel with the stamp removed and asserts it **is** readable anonymously — the leaks-without-the-gate control that makes the ownership test a demonstration rather than a tautology.
- **`decision_type="rebalance"`**, pinned against the live `list_traces` regex read out of the route signature, with `"paper_rebalance"` shown to fail it.
- **Paper-ness and provenance are inside the hash.** `trigger`, `market_context.venue` and `trace_provenance` are all hashed, so a paper trace cannot be laundered into a live one and a backfill cannot be relabelled real-time.
- **Anchoring default OFF, two gates** (`PAPER_TRACE_ANCHOR` **and** `PaperDeployment.anchor_traces`). `reveal()` writes `portfolio_before/after` on-chain permanently; anchoring a private simulation by default would defeat #1556 on purpose, irreversibly. The both-gates-on branch currently logs a WARNING that the anchor path is not wired rather than claiming an anchor — `arc_tx_hash` stays null and `is_verified` stays false.

## Deltas from the design worth flagging

- **One replay, not two.** §9.2 read as "add a sibling replay function"; that would have doubled the settle-path backtest cost and — worse — let the trace and the ledger row for a date come from two different runs of the same spec. `_replay(..., decisions=bool)` produces both in one pass; `replay_spec` keeps its exact signature and semantics.
- **Deterministic trace id.** §9.4 asks for "hash stable across two builds of the same decision", which `uuid4()` cannot give since `id` is a hashed field. The id is now `uuid5` over the decision key — which also makes a republish overwrite the same record rather than mint a duplicate.
- **`provenance` is a column.** Drift detection re-derives a published trace's hash, and provenance is hashed, so without persisting it every re-replay would read as drift. One `String(16)` column; design doc amended.
- **Backfill overflow is recorded, not dropped.** Decisions past `PAPER_TRACE_BACKFILL_MAX` get a durable retryable row. An untraced decision with no row would be invisible to `trace_coverage` — the silent zero §7 exists to prevent.

## Two pre-existing tests updated (both were pinning something they should not)

- `test_account_deletion_cascade.py::test_migrated_schema_is_what_is_under_test` hard-coded `85ca5310b7a1` as head, so **every** future migration breaks it for a reason unrelated to what it guards. Now resolves head from the script directory (also asserting the chain stays serial) and separately asserts the cascade-policy revision is in the applied chain — strictly more, not less.
- `test_alembic_migrations.py::_seed_phase1_fixture` seeded `paper_deployments` through the ORM against a *pre-migration* schema, so any later column broke it. Now inserts through the table as it exists at that revision, values still from the model.
- `test_paper_trading.py::test_advance_all_isolates_failures` patches the settle path's actual seam (`replay_spec_with_decisions`). Same law, same assertions.

## Sequencing / known limits (stated, not papered over)

- **`GET /api/traces/?strategy_id=` depends on #1569**, which is not on `main`. `test_it_conforms_to_1569s_matcher_when_that_lands` imports `trace_references_strategy` and **skips with a reason** today; it becomes live the day #1569 merges, with no edit. The conformance that *can* be checked today (the `list_traces` regex) is checked.
- **SKIP decisions are not traced in v1.** A rebalance-eligible bar where the condition did not fire produces no order, so an order observer cannot see it. Surfaced at the API as `trace_coverage.kinds: ["rebalance"]` rather than hidden; inferring skips would mean a second cadence implementation that can silently disagree with the first.
- **Anchoring is opt-in and the writer path is a follow-up** (design §6.4). Both gates on today logs a WARNING and publishes off-chain only.
- **UI is the follow-up PR** (design step 7): `trace_coverage` rendering and the honest "not anchored (paper)" state in `anchorState()`.

## Anti-goals, verified

- No live-execution path changes — `chain/agent_runner.py`, `chain/executor.py` and `contracts/` are untouched (`git diff --stat`).
- No schema fork of the trace record — same `ReasoningTrace`, same `_HASH_FIELDS`, same canonical JSON, same keccak.
- No new on-chain writes by default — two gates, and the reachable branch does not anchor.
- Settle is not slowed unboundedly — one keccak + one publish per *decision* (a monthly strategy makes ~12/year), one replay pass, bounded backfill, and a publish failure never blocks the ledger.


dbrowneup:
CI note, so the green row is not read for more than it says: `quality-gate.yml` triggers only on `pull_request: branches: [main]`, so **the blocking gate (`backend-tests` + `ruff-gate`) has not run on this PR** — only Docs Gate and import-guard did. That is a property of the stacked base (`dbrowneup/1575-paper-trace-design`), not of the change.

The equivalent was run locally with the env's interpreter, using CI's own command from the repo root:

```
pytest -m "not integration" -q   →  4386 passed, 3 skipped, 2 deselected  (0 failed)
ruff format --check .            →  641 files already formatted
ruff check <changed files>       →  All checks passed!
```

Per CLAUDE.md § "a stale PR's green check describes a base that no longer exists": re-verify against current `main` once #1578 lands and this is retargeted — that retarget is also what makes the blocking gate fire.

PR #1599, opened 2026-08-31
[strategy] Delete the self-declared look_ahead_safe flag; the LEAK verdict is the audit's alone

dbrowneup:
Stacked follow-up to #1566 (merged tonight), now rebased onto current main and union-tested.

#1566 made the LEAK leg a real structural audit but left the LLM's self-declared `look_ahead_safe` flag in the schema and vocabulary. This PR deletes the flag end to end and unifies the four-state verdict vocabulary so the only LEAK signal that exists is the one the AST proof derives — a self-declaration that no longer feeds anything is a claim surface waiting to be trusted by accident.

Union check on current main (post-#1567/#1591 trains): lookahead + fusion-evaluator + selection-bias suites green locally; fresh CI on this push is the authoritative run.

Part of the 2026-08-31 closure plan (handover §3, staged follow-on to #1566).


onder-akkaya:
Read this one properly rather than skimming it, since it changes what a LEAK verdict means. The change itself is right. The base it sits on is what blocks it.

## What's right

Four-state `pass`/`fail`/`pending`/`degenerate` reuses the vocabulary `RigorGateVerdict` already has instead of inventing a third one, and keeping `pending` apart from `degenerate` is the correct reading of § fail-soft: "we did not look at this strategy" and "we looked and the instrument gave no reading" have different fixes.

The floor is fail-closed where it counts. `la_passed = False` is the default at `rigor_evaluator.py:1450` and only an explicit `look_ahead_audit_passed=True` flips it, so a caller that forgets the parameter blocks rather than admits.

`render_state` is gone with nothing left dangling. `git grep -c look_ahead_render_state` on this branch returns nothing tree-wide, and the UI never read the key, so there is no frontend break hiding in the removal.

`_note_legacy_fields` not mutating the caller's dict is right, and so is refusing to re-honour a legacy `look_ahead_safe: false`. Both are argued in the docstring instead of left for a reader to reconstruct.

The tests are guards shown rejecting something, not assertions that the code does what it does: `test_degenerate_blocks_the_gate`, `test_no_state_but_pass_is_ever_deployable`, `test_neither_gate_admits_what_the_other_blocks`, `test_self_attested_is_never_written_again`.

## What blocks it

**The base is 71 commits behind main, and 13 of the 27 files have moved on main since the branch cut.** The green checks above describe a base that no longer exists.

It does not merge, and both conflicts are the kind where resolving them is worse than rebasing:

**`docs/specs/strategy-dsl-spec.md`** — main took this from 200 to 383 lines after the cut, and the new content is the whole `inverse_vol` documentation (15 mentions of it). This branch's copy is 199 lines. Take this branch's side and 183 lines of newer spec disappear. Take main's side and 6 `look_ahead_safe` mentions survive in the one file the PR exists to clean.

**`backend/archimedes/agents/strategy_fusion.py`** — same shape in the generation prompt. Main's version gained `inverse_vol`, `reference_vol_annual`, and the line *"position_sizing accepts NO other keys — a key outside that list is a hard validation error, not an ignored field, so do not invent one."* This branch's prompt predates all three.

Neither is a conflict to resolve. The flag deletion has to be re-applied on top of what main carries now.

## Union test

Merged locally onto `main@bad63978`, resolved both conflicts toward main's newer content, re-applied the two prompt edits by hand, and ran the CI command in a clean worktree:

```
$ pytest -m "not integration" -q          # repo root, clean worktree, env's pytest

main@bad63978          26 failed, 4995 passed, 3 skipped, 2 deselected, 4 errors  (266s)
union (#1599 + main)   26 failed, 5019 passed, 3 skipped, 2 deselected, 4 errors  (271s)
```

The failure sets are identical: 30 shared entries, nothing on either side alone. Every one is `test_alembic_migrations.py` or `test_account_deletion_cascade.py` dying on `FileNotFoundError: 'alembic'`, which is a bare binary resolved off ambient PATH and has nothing to do with either branch. The union adds 24 passing tests and no failures, so once the two conflicts are re-applied rather than resolved, this merges clean.

**A false alarm I raised and then killed, because the trap is waiting for whoever rebases this.** 13 test files on main mention `look_ahead_safe` and this PR touches none of them. Run them against the union and 6 fail in `test_strategy_fusion.py` (corpus loader, `assert 10000 == 4`). They are not this branch's. The same file on clean main **in the same directory** fails the identical 6, and in a fresh worktree it passes 41. Something about a working tree that has been used makes the loader ignore `ARCHIMEDES_CORPUS_MANIFEST`. Worth its own issue; nothing to do with this PR. Anyone re-running these locally should use a fresh worktree or they will chase it too.

## One thing worth fixing in the same PR

`ui/src/components/Landing.jsx:31-38` is the provenance comment for the LEAK criterion the landing page renders, and this PR makes all three of its citations false:

```
//   LEAK  — strategy_dsl.py:192-193 rejects look_ahead_safe=false at validation;
//           generation_pipeline.py:1892 persists look_ahead_audit_source=
//           "self_attested"; rigor_evaluator.look_ahead_audit (the real AST pass)
//           runs only against cited library source.
```

The rendered copy underneath is fine. It already says "not on the generator's own say-so", which this PR makes more true rather than less. It is the receipt that goes stale, and a receipt for a public claim that cites line numbers no longer doing what it says is worth the two minutes while the context is open.

## Process note

11 lines of body for +1267/−534 across 27 files, and none of it is the verification. The tests do demonstrate the guards; the body shows none of them. Worth pasting the four transcripts in, because a reviewer who does not open the test file has no way to tell this apart from a rename pass.

Nothing here touches `contracts/`, so one review clears it.

dbrowneup:
@onder-akkaya — thank you for this; it raced the merge by ~55 minutes but everything in it got acted on:

**The merge did what you prescribed, verified.** The resolution re-applied the deletion ON TOP of main's newer content rather than resolving toward either side: strategy_fusion's prompt kept main's inverse_vol/reference_vol_annual/no-other-keys rules with the do-not-emit line re-applied ([`bdb93547`](https://github.com/a-apin/archimedes/commit/bdb93547)); the spec kept main's full 385 lines. Your union numbers matched our independent run (+24 passed, zero new failures). The 26 local reds you saw are the bare-`alembic` PATH trap from CLAUDE.md's toolchain table — CI and the env-pytest run are clean.

**Your Landing receipt catch was real and is fixed** — plus a second instance of the same hazard you predicted: main's spec rewrite had re-introduced `look_ahead_safe` as a *live requirement* in four sections. Both fixed in **#1639** (spec sections → retired-field documentation of the `LEGACY_IGNORED_FIELDS` tolerance; Landing LEAK receipt → cites the #1566 structural audit).

**Your worktree/corpus-loader trap is now **, quoted with credit — good catch, that would have burned hours.

**Process note taken**: the four guard transcripts should have been in this body. They're pasted in #1639's evidence section going forward, and your #1631 template will force the discipline anyway. One more honest note: the body's original 11 lines were mine to fix before merging — the finisher beat your review in, which is a sequencing failure on our side, not yours.

dbrowneup:
Correction to the comment above — a shell bug ate the issue number: your worktree/corpus-loader trap is **#1640**.

PR #1618, opened 2026-08-31
[frontend] Fix Explore price-chart x-axis dates (timezone shift, tick density, clipping)

### LINKED ISSUE #1602: [frontend] Explore price-chart x-axis dates render wrong and are hard to read
Owner report 2026-08-31: on the Explore page, clicking a card/asset opens the time-series price chart, and the x-axis dates "look messed up and not accurate, hard to read."

**Task:** fix the x-axis date formatting/scaling on the Explore price charts — correct dates from the series index (no off-by-timezone or index-as-label bugs), sensible tick density at every range, readable at mobile width. While in there, apply the light Explore cleanup the owner asked for alongside the #1218 disclosure work.

**Acceptance:** dates on the rendered chart match the underlying series dates exactly (unit test on the tick-formatter with a fixed fixture series); no overlapping/rotated-unreadable labels at default width; `cd ui && npm test` green.

**Anti-goals:** no charting-library swap; no data-fetch changes (that's #1218's lane).

dbrowneup:
## What

Fixes the Explore price chart's x-axis: dates that did not match the series, labels with no year on multi-year ranges, repeated clock times on the 1D range, edge labels clipped by the viewBox, and ~4.5px axis text on a phone.

x-axis tick logic moves out of `PriceHistoryChart.jsx` into a new pure module `ui/src/chartTicks.js`. That is the `statUtils.js` precedent and it is load-bearing here: `npm test` is `node --test`, which cannot parse JSX, so logic living inside a `.jsx` file can only ever be *text-scanned* by the UI suite, never executed. Moving it out is what makes the formatter testable at all.

## Why — the actual defect

The old code read every timestamp back through the **viewer's local calendar**:

```js
const parsed = Date.parse(tsStr.replace(' ', 'T'))
const d = new Date(parsed)
label = `${String(d.getMonth() + 1).padStart(2, '0')}-${String(d.getDate()).padStart(2, '0')}`
```

`Date.parse` on an offset-bearing stamp yields an absolute instant; `getMonth()`/`getDate()` then re-home that instant in whatever zone the browser is in. A crypto daily close stamped `2026-05-25 00:00:00+00:00` is UTC midnight, which is `2026-05-24 20:00` in `America/New_York` — so the axis labelled the bar **`05-24`**, a date that appears nowhere in the series.

It was also *inconsistent*, which is why it shipped. Per ECMA-262 a tz-naive date-**time** string parses as local wall time (never shifted) but a date-**only** string parses as UTC (shifted by the local getters). Reproduced against the pre-fix code:

```
$ TZ=America/New_York node repro.mjs      # series date is 2026-05-25 in every case
  2026-05-25 00:00:00+00:00    -> "05-24"   <-- wrong (crypto daily close)
  2026-05-25                   -> "05-24"   <-- wrong (resampled weekly/monthly bar)
  2026-05-25 00:00:00-04:00    -> "05-25"   <-- correct (US equity daily bar)
  2026-05-25 00:00:00          -> "05-25"   <-- correct
$ TZ=UTC node repro.mjs
  ...all four -> "05-25"                     <-- correct, which is why this was never caught
```

Half the ranges rendered correctly and half were a day early, in the same chart — exactly "look messed up and not accurate."

**The fix is to never convert.** A bar's label must report the bar's own wall clock, which the series already carries in the string. `parseSeriesTimestamp` reads the calendar fields **lexically** and no `Date` is ever constructed for display, so the rendered label equals the series timestamp by construction, in every timezone. (Re-zoning would also be wrong on the merits: a US equity daily close is the May 25 *session* and must not be relabelled "May 24" because the viewer is in Tokyo.)

Also fixed, all from the same report:

| | before | after |
|---|---|---|
| 5Y / 10Y / MAX | `05-25` — no year at all across a decade | `2016 … 2026` |
| 1D (served as period `2d`) | `09:30 … 09:30` repeating across two sessions | `May 25 09:30 … May 26 09:40` |
| window straddling New Year | `Dec 31`, `Jan 2` — which year? | `Dec 31, 2025`, `Jan 2, 2026` |
| first/last label | centred on the plot edge → hung into the y-gutter / clipped by the viewBox | anchored `start` / `end` |
| tick count | always 4, fixed | derived from available width, never overlapping |
| axis text on a 360px phone | ~4.5px (9 units in a fixed 720 viewBox) | true 11px (viewBox sized to the measured container) |
| intraday detection | `points[0]` vs `points[1]` — one weekend gap flipped the whole axis | median step |

## Test commands + output

```
$ cd ui && npm ci && npm test
ℹ tests 633
ℹ suites 0
ℹ pass 633
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1425.889375
```

633 = 614 on `main` before this change + 19 new. New file only:

```
$ node --test test/chart-ticks.test.js
ℹ tests 19
ℹ pass 19
ℹ fail 0
```

```
$ npx eslint .          # whole ui/, clean, no output
$ npm run build
✓ built in 2.04s
```

Tests are hermetic: pure functions, no DOM, no network, no DB, no `.env`, no clock read. The only environment they touch is `process.env.TZ`, which they set and restore themselves.

## Demonstration 1 — the regression test fails with the fix reverted

Reverted `parseSeriesTimestamp` to the pre-#1602 `Date` round-trip, changing nothing else:

```js
export function parseSeriesTimestamp(ts) {
  if (typeof ts !== 'string') return null
  const parsed = Date.parse(ts.replace(' ', 'T'))
  if (Number.isNaN(parsed)) return null
  const d = new Date(parsed)
  return { year: d.getFullYear(), month: d.getMonth() + 1, day: d.getDate(),
           hour: d.getHours(), minute: d.getMinutes() }
}
```

```
$ TZ=America/Los_Angeles node --test test/chart-ticks.test.js
✖ #1602: a UTC-stamped daily close labels its OWN date in every timezone
✔ #1602: every tick label equals the calendar date of the sample it marks
✖ #1602: a date-only stamp is not shifted either (Date.parse treats it as UTC)
✖ #1602: an intraday bar keeps the exchange clock it was stamped with
✖ parseSeriesTimestamp reads each pandas stringification the backend emits
✖ parseSeriesTimestamp rejects junk instead of inventing a date
✔ formatTickLabel falls back to the raw stamp rather than NaN for junk
✔ medianStepMs is robust to one gap (the old first-pair heuristic was not)
✔ pickTickFormat picks resolution from the bar interval and span
✖ a 1D chart spanning two sessions does not repeat the same clock times
✔ a decade chart labels years, not bare month-days
✖ a New-Year-straddling window carries the year on every label
✔ chooseTickCount never lets two labels touch, at any width
✔ tick density actually adapts: a phone gets fewer ticks than a desktop
✔ chooseTickCount degrades safely on a nonsense width
✔ buildXTicks always marks the first and last sample
✔ buildXTicks anchors the edge labels inward so neither is clipped
✔ buildXTicks indices are strictly increasing and inside the series
✖ buildXTicks handles degenerate series without duplicating a tick
ℹ pass 11
ℹ fail 8
```

The headline guard, verbatim:

```
✖ #1602: a UTC-stamped daily close labels its OWN date in every timezone
  AssertionError [ERR_ASSERTION]: x-axis labels drifted in America/Los_Angeles
  + actual - expected

    [
  +   'May 17',
  +   'May 19',
  +   'May 21',
  -   'May 18',
  -   'May 20',
      'May 22',
  +   'May 24',
  +   'May 26'
  -   'May 23',
  -   'May 25',
  -   'May 27'
    ]
```

Restoring the fix returns `19 pass / 0 fail`.

**The reverted tally is itself timezone-dependent — read the failing-test names, not the count.** The transcript above is `TZ=America/Los_Angeles`. Two of those eight tests (`a New-Year-straddling window…`, `buildXTicks handles degenerate series…`) use date-only stamps and inherit the runner's ambient zone, so they only break west of UTC:

```
$ for tz in UTC America/Chicago America/Los_Angeles Asia/Tokyo; do TZ=$tz node --test test/chart-ticks.test.js; done
UTC                  -> 13 pass / 6 fail
America/Chicago      -> 11 pass / 8 fail
America/Los_Angeles  -> 11 pass / 8 fail
Asia/Tokyo           -> 13 pass / 6 fail
```

**Six tests fail under revert in every zone tested, including UTC** — those are the real guards, and they bite because they set `TZ` themselves rather than inheriting it:

```
✖ #1602: a UTC-stamped daily close labels its OWN date in every timezone
✖ #1602: a date-only stamp is not shifted either (Date.parse treats it as UTC)
✖ #1602: an intraday bar keeps the exchange clock it was stamped with
✖ parseSeriesTimestamp reads each pandas stringification the backend emits
✖ parseSeriesTimestamp rejects junk instead of inventing a date
✖ a 1D chart spanning two sessions does not repeat the same clock times
```

The **committed** suite is zone-independent, which is the property that actually matters — `parseSeriesTimestamp` is lexical and constructs no `Date`, so there is nothing left for the ambient zone to move. Verified across the full UI suite:

```
$ for tz in UTC America/New_York America/Los_Angeles Asia/Tokyo Pacific/Kiritimati Australia/Sydney; do TZ=$tz npm test; done
UTC                  -> 633 pass / 0 fail
America/New_York     -> 633 pass / 0 fail
America/Los_Angeles  -> 633 pass / 0 fail
Asia/Tokyo           -> 633 pass / 0 fail
Pacific/Kiritimati   -> 633 pass / 0 fail
Australia/Sydney     -> 633 pass / 0 fail
```

Note that one test — "every tick label equals the calendar date of the sample it marks" — **passes both ways**. It compares `buildXTicks`'s label against `formatTickLabel` on the same timestamp, so both sides move together under the reverted code. It is a useful internal-consistency check but it is **not** a regression guard for this bug, and I am not counting it as one (CLAUDE.md rule 3).

### End-to-end, not just the module

The pure module is only half the claim, so I bundled the real `PriceHistoryChart.jsx` with the project's own bundler and server-rendered it, then read the `<text>` elements the browser actually receives. Fixture series spans **2026-05-18 → 2026-05-27**:

```
### WITH THE FIX
America/New_York       "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"
UTC                    "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"
Asia/Tokyo             "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"
Pacific/Kiritimati     "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"

### WITH THE FIX REVERTED
America/New_York       "May 17" "May 19" "May 21" "May 22" "May 24" "May 26"   <-- whole axis a day early
UTC                    "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"
Asia/Tokyo             "May 18" "May 20" "May 22" "May 23" "May 25" "May 27"
```

Rendered layout attributes confirm the anchoring and sizing fixes:

```
viewBox: 0 0 720 259 | height attr: 259
   x=   56.0  anchor=start   "May 18"      <-- was anchor=middle, hung into the y-gutter
   x=  199.6  anchor=middle  "May 20"
   x=  343.1  anchor=middle  "May 22"
   x=  414.9  anchor=middle  "May 23"
   x=  558.4  anchor=middle  "May 25"
   x=  702.0  anchor=end     "May 27"      <-- was anchor=middle, overhung the 720 viewBox
```

## Demonstration 2 — the guards reject bad input

Each guard was fed the input it is supposed to catch, and confirmed to fail.

**2a. Tick-density guard, bounds half.** Cranked `MAX_TICKS` 6 → 12, the plausible "let's show more ticks" edit:

```
$ node --test --test-name-pattern "never lets two labels touch"
✖ chooseTickCount never lets two labels touch, at any width
  AssertionError [ERR_ASSERTION]: time@400: 8 outside [2,6]
ℹ pass 0
ℹ fail 1
```

**2b. Tick-density guard, spacing half.** 2a fired on the bounds assertion, so the *spacing* assertion is separately un-vacuous-ed by making the formula under-estimate label width (`CHAR_WIDTH_RATIO` 0.62 → 0.20, `TICK_GAP` 16 → 0):

```
$ node --test --test-name-pattern "never lets two labels touch"
✖ chooseTickCount never lets two labels touch, at any width
  AssertionError [ERR_ASSERTION]: time@120: 6 ticks gives 24.0px spacing for a 34.1px label
ℹ pass 0
ℹ fail 1
```

**2c. Index-as-label** — the exact anti-bug named on the issue. Replaced the label with the array index:

```js
-    label: formatTickLabel(points[index] && points[index].ts, fmt),
+    label: String(index),
```

```
$ node --test test/chart-ticks.test.js
ℹ pass 12
ℹ fail 7
```

**2d. Junk timestamps.** `parseSeriesTimestamp` is asserted to return `null` — not a fabricated date — for `null`, `undefined`, `42`, `''`, `'not-a-date'`, `'2026-13-01'`, `'2026-05-32'`, `'2026-05-25 25:00'`, `'May 25 2026'`, and `formatTickLabel` is asserted never to emit `NaN` (the old code rendered `"NaN-NaN"` on unparseable input).

All modifications above were reverted; the committed tree is `633 pass / 0 fail`.

## Files

- `ui/src/chartTicks.js` — **new**, pure tick logic
- `ui/test/chart-ticks.test.js` — **new**, 19 tests
- `ui/src/components/PriceHistoryChart.jsx` — consume the module; measured viewBox; edge anchors; axis font 9→11 (x) and 9→10 (y)
- `ui/eslint.config.js` — give `test/**` Node globals (`npm test` *is* `node --test`; the timezone matrix needs `process.env.TZ`). Repo lint was clean before and after; the only 4 errors this introduced were the ones this line fixes.

## Not done here

- **No data-fetch changes and no charting-library swap** — both anti-goals on the issue. `AssetModal`/`AssetGroupModal` fetch code is untouched; the chart is still hand-rolled SVG.
- **The "light Explore cleanup alongside #1218" mentioned on the issue is NOT done.** I scoped this to the chart the report was about. The issue does not say what that cleanup is, and I did not want to guess at owner intent in a PR whose testable claim is about dates. Worth a follow-up issue with specifics.
- **No visual/screenshot regression test.** The repo has no DOM or snapshot test infra for `ui/` and adding one is well beyond this issue. The end-to-end render above was a one-off verification run, not committed — the committed guard is the pure-module suite.
- **`CHAR_WIDTH_RATIO = 0.62` is an estimate**, not a measured text advance. It is deliberately conservative (real sans digits/letters at 11px average below this), and the guard asserts the resulting spacing clears the widest label each format can produce. A browser `measureText` would be exact but needs a canvas at render time.
- **Plot height is now `clamp(220, width*0.36, 320)`** rather than a fixed 260-unit viewBox with `maxHeight: 320`. At the 720px width the old code assumed this is 259 vs 260, and in the widest modal (860px) it is 310 vs the old ~320 — visually equivalent. Called out because it is a rendered-size change beyond the x-axis.
- **`pickTickFormat` thresholds (12h / 180d / 1100d) are judgement calls**, tested at the boundaries the real ranges produce but not tuned against every possible series.

Closes #1602

### DIFF
--- a/ui/eslint.config.js
+++ b/ui/eslint.config.js
@@ -62,4 +62,12 @@ export default defineConfig([
     files: ['postcss.config.js', 'vite.config.js', 'eslint.config.js'],
     languageOptions: { globals: { ...globals.node } },
   },
+  // `npm test` is `node --test`, so the suite executes in Node and may use
+  // Node globals (#1602 needs `process.env.TZ` to re-home the local calendar
+  // and prove the chart's date labels do not move with the viewer's timezone).
+  // Browser globals stay in scope too — these files also assert on UI source.
+  {
+    files: ['test/**/*.js', 'scripts/**/*.mjs'],
+    languageOptions: { globals: { ...globals.browser, ...globals.node } },
+  },
 ])
--- a/ui/src/chartTicks.js
+++ b/ui/src/chartTicks.js
@@ -1,0 +1,215 @@
+// x-axis tick logic for the Explore price charts (#1602).
+//
+// Extracted as a plain module (same precedent as statUtils.js) so the tick
+// formatter is directly unit-testable — `node --test` cannot parse JSX, so
+// logic that lives inside PriceHistoryChart.jsx can only ever be text-scanned,
+// never executed, by the UI suite.
+//
+// ── Why this exists ──────────────────────────────────────────────────────────
+//
+// The previous implementation round-tripped each timestamp through a Date
+// object and then read it back with LOCAL calendar accessors:
+//
+//     const d = new Date(Date.parse(ts.replace(' ', 'T')))
+//     `${d.getMonth() + 1}-${d.getDate()}`
+//
+// That is an off-by-one-day bug for every viewer west of UTC. A crypto daily
+// close stamped "2026-05-25 00:00:00+00:00" is UTC midnight; read back through
+// America/New_York's calendar it is 2026-05-24 20:00, so the axis labelled the
+// bar "05-24" — a date that does not appear anywhere in the series. Worse, it
+// was inconsistent: ECMA-262 parses an offset-bearing or date-ONLY string as an
+// absolute instant (shifted by local getters) but a tz-naive date-TIME string as
+// local wall time (not shifted). So "2026-05-25 00:00:00-04:00" rendered
+// correctly while "2026-05-25 00:00:00+00:00" rendered a day early, in the same
+// chart, depending on which range the user picked.
+//
+// The fix: never convert. A bar's label must report the bar's own wall clock,
+// which the series already carries in the string. We read the calendar fields
+// LEXICALLY and never construct a Date for display, so the rendered label is
+// the series timestamp by construction, in every timezone.
+//
+// (Re-zoning would also be wrong on the merits: a daily close for a US equity
+// is the May 25 *session*, and must not be relabelled "May 24" because the
+// viewer happens to be in Tokyo.)
+
+/** Leading calendar fields of an ISO-ish timestamp. The trailing UTC offset,
+ * if any, is deliberately NOT captured — see the module note above. Accepts
+ * "2026-05-25", "2026-05-25 00:00:00", "2026-05-25T09:35:00+00:00". */
+const TS_RE = /^(\d{4})-(\d{2})-(\d{2})(?:[T ](\d{2}):(\d{2}))?/
+
+const MONTHS = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
+
+const MS_PER_DAY = 24 * 60 * 60 * 1000
+
+/**
+ * Read the wall-clock calendar fields out of a series timestamp, lexically.
+ * Returns null for anything unparseable, so callers can fall back honestly
+ * instead of rendering "NaN-NaN".
+ */
+export function parseSeriesTimestamp(ts) {
+  if (typeof ts !== 'string') return null
+  const m = TS_RE.exec(ts.trim())
+  if (!m) return null
+  const year = Number(m[1])
+  const month = Number(m[2])
+  const day = Number(m[3])
+  if (month < 1 || month > 12 || day < 1 || day > 31) return null
+  const hour = m[4] === undefined ? 0 : Number(m[4])
+  const minute = m[5] === undefined ? 0 : Number(m[5])
+  if (hour > 23 || minute > 59) return null
+  return { year, month, day, hour, minute }
+}
+
+/** Wall-clock fields → a comparable scalar. Date.UTC is used purely as
+ * calendar arithmetic on fields we already hold; because every stamp in a
+ * series goes through the same basis, differences are exact. No timezone
+ * conversion happens here and the result is never formatted for display. */
+function toComparable(f) {
+  return Date.UTC(f.year, f.month - 1, f.day, f.hour, f.minute)
+}
+
+/** Median gap between consecutive samples, in ms. Median (not the first pair,
+ * which is what the old `isIntraday` heuristic used) so one weekend gap or one
+ * missing bar cannot flip a whole axis into the wrong format. */
+export function medianStepMs(fields) {
+  if (!fields || fields.length < 2) return null
+  const steps = []
+  for (let i = 1; i < fields.length; i += 1) {
+    steps.push(Math.abs(toComparable(fields[i]) - toComparable(fields[i - 1])))
+  }
+  steps.sort((a, b) => a - b)
+  const mid = Math.floor(steps.length / 2)
+  return steps.length % 2 === 0 ? (steps[mid - 1] + steps[mid]) / 2 : steps[mid]
+}
+
+/**
+ * Choose the label format for a whole series.
+ *
+ * Resolution comes from the bar interval; the year/day qualifier comes from the
+ * span, so a label is never ambiguous about which day (or year) it marks:
+ *
+ *   intraday bars, one calendar day    → 'time'     "09:35"
+ *   intraday bars, several days        → 'dayTime'  "May 25 09:35"
+ *   daily+ bars, < 180d, one year      → 'day'      "May 25"
+ *   daily+ bars, < 180d, spans a NY    → 'dayYear'  "May 25, 2026"
+ *   daily+ bars, < ~3y                 → 'month'    "May 2026"
+ *   longer                             → 'year'     "2026"
+ *
+ * The 'dayTime' case is not hypothetical: the 1D range is served as period
+ * "2d" (asset_market_service._HISTORY_RANGE_MAP), so a 1D chart normally holds
+ * two calendar days of 5-minute bars. Labelling those with bare HH:MM repeats
+ * the same clock times twice across the axis.
+ */
+export function pickTickFormat(points) {
+  const fields = (points || []).map(p => parseSeriesTimestamp(p && p.ts)).filter(Boolean)
+  if (fields.length === 0) return 'day'
+
+  const step = medianStepMs(fields)
+  const intraday = step !== null && step < 12 * 60 * 60 * 1000
+
+  const first = fields[0]
+  const last = fields[fields.length - 1]
+  const spanMs = Math.abs(toComparable(last) - toComparable(first))
+
+  if (intraday) {
+    const sameDay =
+      first.year === last.year && first.month === last.month && first.day === last.day
+    return sameDay ? 'time' : 'dayTime'
+  }
+  if (spanMs < 180 * MS_PER_DAY) {
+    return first.year === last.year ? 'day' : 'dayYear'
+  }
+  if (spanMs < 1100 * MS_PER_DAY) return 'month'
+  return 'year'
+}
+
+const pad2 = n => String(n).padStart(2, '0')
+
+/**
+ * Render one series timestamp in the given format. Unparseable input falls
+ * back to the raw leading date text rather than inventing a date.
+ */
+export function formatTickLabel(ts, format) {
+  const f = parseSeriesTimestamp(ts)
+  if (!f) return typeof ts === 'string' ? ts.slice(0, 10) : ''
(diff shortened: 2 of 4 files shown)


dbrowneup:
**Independent verification pass** (fresh worktree off this branch, `npm ci` from clean).

Re-ran every claim in the description rather than trusting it. All the substantive ones hold:

- `npm test` → **633 pass / 0 fail**
- `npx eslint .` → clean, exit 0
- `npm run build` → succeeds (only the pre-existing chunk-size warning)
- Revert demo bites: reverting `parseSeriesTimestamp` to the `Date` round-trip fails the guards, with the headline assertion reproducing verbatim (`x-axis labels drifted in America/Los_Angeles`, `May 17/19/21…` vs `May 18/20/22…`)
- Guard demo 2a reproduces verbatim: `time@400: 8 outside [2,6]`
- Guard demo 2b reproduces verbatim: `time@120: 6 ticks gives 24.0px spacing for a 34.1px label`
- Guard demo 2c (index-as-label) reproduces: 12 pass / 7 fail
- Premise confirmed against the tree: `Explore.jsx:324` `setOpenAsset` → `AssetModal` → `PriceHistoryChart`, so this is the chart the owner report is about

**One correction, now applied to the description.** Demonstration 1 reported `ℹ pass 12 / ℹ fail 7`, but the transcript underneath it lists **eight** `✖` marks — internally inconsistent, and the number reproduced in no timezone I tried (UTC and Tokyo give 13/6; Chicago and Los Angeles give 11/8). Root cause is benign: two of the reverted failures use date-only stamps and inherit the runner's ambient zone, so the counterfactual tally moves with `TZ` while the committed suite does not.

Corrected in the body to the verified `TZ=America/Los_Angeles` run (11/8), with the per-zone matrix and the six guards that fail under revert **in every zone including UTC** spelled out — those are the ones doing the actual work.

Also added a claim the description was implicitly resting on but had not shown: the committed suite is zone-independent. Full UI suite under six zones:

```
UTC / America/New_York / America/Los_Angeles / Asia/Tokyo / Pacific/Kiritimati / Australia/Sydney
  -> 633 pass / 0 fail  (all six)
```

That is the property that matters for the fix — `parseSeriesTimestamp` is lexical and constructs no `Date`, so there is nothing left for the ambient zone to move.

No code changed in this pass; the tree is byte-identical to what was already pushed. Description edit only.

Issue #1646, opened 2026-08-31
[frontend] Strategy Passport redesign: dense viewport-fit layout, papers table, visible DSL, tight Library linkage


dbrowneup:
**Why (owner, 2026-08-31 product review):** boxes are "squished and weird, information spread out" — needs to be dense, clean, and fit the viewport; every paper feeding the strategy should appear in a TABLE with citations in the abstract; the generated DSL code should be cleanly and clearly visible; Recently-Generated and Library passport pages need tight linkage.

## Summary

`ui/src/components/StrategyPassport.jsx` (1126 lines) currently renders each source paper as a stacked, vertically-growing card (`.passport-panel`, lines 538-600 for the multi-paper/fusion case, 601-630 for the single-paper case) — title, an arxiv link, a contribution sentence, separated by `border-bottom` dividers inside one card. Not a table; no fixed columns; grows unbounded with paper count.

**The generated DSL code is not rendered anywhere on this page today** — verified structurally, not assumed. `backend/archimedes/models/strategy.py:129-134` and `strategy_store.py:70-158` show the field exists at the model layer (`StrategyPassport.strategy_spec: dict | None`, the "raw validated DSL spec JSON"), but `backend/archimedes/api/schemas.py`'s `StrategyResponse` (the payload this page actually renders, lines 168-227+) has **no `strategy_spec` field at all** — it's stripped before the API boundary. The only places `strategy_spec`/DSL are mentioned client-side are error strings for paper-trading gating (`"no_strategy_spec"`, `"invalid_strategy_spec"`), never a display.

**Recently-Generated → Library linkage is a real, small gap, not a large one.** `ui/src/components/GenerationStatus.jsx` (the "Recent generations" table on the Generate page) fetches `/api/generate/jobs?limit=20`, and the backend's `_job_summary` (`generate_routes.py:1002-1023`) **already returns `best_strategy_id`** on every job. But `GenerationStatus.jsx` never reads that field — clicking a `done` row calls `onDrillIn(j.job_id)`, which reopens the generation-stream SSE replay, not the resulting strategy's passport. `Generate.jsx` already has `onNavigate` in scope (used at line 708 for `GenerationStream`) but never threads it into `<GenerationStatus>` (line 1368). `Strategies.jsx:774` already has the exact call this needs: `onNavigate('strategy', { strategyId })`.

## Scope

[...]

Issue #1637, opened 2026-08-31
[provenance][frontend] End-to-end paper association: selection→passport contract, real papers table, grounded abstract citations


dbrowneup:
**Why**

A paper association is currently a **string in three incompatible shapes**, and the shape is inside the content hash.

- Writers emit `{arxiv_id,title,authors}` (`main.py:294`), `{arxiv_id,title:""}` (`debate_engine.py:533`), `{arxiv_id,sha256:""}` (`strategies_routes.py:2086`); the column comment claims a fourth (`strategy_store.py:53`). `_compute_content_hash` (`:210-233`) hashes the **whole dicts** — so the *same* paper set through two writers produces two `content_hash` values, two ids, two "different" strategies. Dedup and the paper→strategy back-index (`:369-386`) both degrade.
- **The selection-time list is never recorded.** `strategy_fusion.py:722` and `debate_engine.py:271` are locals. Nothing anywhere says *"these 6 were considered, these 3 were cited."* The rerank cap hit is invisible; `_critic_prov`'s drops (`:426-431`) are log-only.
- **The trace is worse than incomplete.** `agent_runner.py:226` writes `consulted_paper_hashes` of length **1** with `content_hash = ss.strategy_id`. It is in `_HASH_FIELDS` (`trace.py:90`) and is signed. `verify_source_papers` (`source_tracker.py:44-88`) has **zero production callers**; `/verify` re-hashes the body and never checks the papers. The "trace-verify button" isn't one button — the verifier is unreachable.
- **The passport promises a table it cannot fill.** `StrategyPassport.jsx:548-592` renders title + arXiv link + `contribution`. `contribution` is **never written by any path** — and is silently absent from `PassportPaperRef.to_dict()` (`strategy_passport_record.py:255-264`). Authors/venue/year/DOI/citation-count are structurally NULL by construction (`generation_pipeline.py:1683`, `:2132`). The data to fill `contribution` **already exists** as `fusion_reasoning` — the prompt defines it as "what mechanism EACH cited paper contributes" (`strategy_fusion.py:838`) — and is dropped at `debate_engine.py:535`.
- **Enrichment is guaranteed to be wiped.** `ingest_passport(force_update=True)` does `DELETE FROM passport_paper_refs` then rebuilds from id+title only (`passport_loader.py:113-115`), and runs on every real-returns refresh (`generation_pipeline.py:2136`, `:2159`).
[...]

Issue #1359, opened 2026-08-20
[frontend][backend] Commit-reveal: the UI denies a guarantee the contract enforces, and "Verify hash on-chain" recomputes nothing


dbrowneup:
## Summary

Two claim-integrity defects live in `ui/src/components/Reasoning.jsx`, and they point in opposite directions.

**(1) The app denies a guarantee the contract actually enforces.** Commit-before-trade *is* enforced on-chain: `Vault.rebalance()` derives a `tradeId` and calls `ReasoningTraceRegistry.executeTrade()`, which reverts without a matching commitment from a strictly earlier block. `/architecture` says exactly that. `/app/reasoning` says the opposite — it heads the panel "Block Order Check **(off-chain)**" and tells the reader the ordering is "not yet enforced on-chain — commit-reveal wiring is on the roadmap". Because `TraceResponse` coerces `temporal_binding_valid` to `None` unless `temporal_binding_source == "chain"`, and that sentence renders only when `temporal_binding_valid` is truthy, **the sentence is wrong 100% of the times a user can see it**. Two public surfaces contradict each other, and the product under-claims its strongest provenance guarantee. `temporal_binding_source` — the field that exists precisely to carry this provenance to the UI — has **zero readers in `ui/src`**.

**(2) "Verify hash on-chain" verifies less than the copy claims.** The intro prose promises the button will "recompute and check against the on-chain anchor". The endpoint compares two already-stored strings and never re-hashes canonical bytes, so tampering with the stored trace body is undetectable. Worse, the no-off-chain-record branch — which **a Redis outage also falls into** — returns `is_verified=True` with *zero hashes compared*, and the UI paints the identical green tick while rendering `details` only when `!is_verified`, i.e. suppressing the one honest caveat in exactly the case it exists to qualify. `GET /api/traces/{id}/canonical` already serves the exact bytes needed to do this honestly and is not wired to the button.

All line references below were re-read on `main` @ `c0e6400f`.

## Evidence

**The contract enforces the ordering.**

- `contracts/src/Vault.sol:452-453` — `bytes32 tradeId = keccak256(abi.encode(tokensIn, amountsIn, tokensOut, amountsOut));` / `uint256 traceId = traceRegistry.executeTrade(tradeId);` inside `rebalance()` (`:440`).
[...]

PR #1688, opened 2026-08-31
[strategy][provenance] assoc/v1 paper associations: one shape, one identity, traces bound to it, verifier wired

dbrowneup:
## ⚠️ Owner review required — this touches the strategy-passport / reasoning-trace data flow

`CLAUDE.md` § *When to ask before acting* names this surface explicitly. Three things in here are decisions, not just code, and I am flagging them for **Dan** rather than assuming them:

1. **`strategy_store.content_hash` changes definition, and a data migration re-stamps existing rows.** The hash now sees sorted, de-duplicated `(handle, role)` pairs instead of the whole association dicts. Migration `b41c7e0d95a2` recomputes each row's *historical* hash from that row's own stored columns and rewrites `content_hash` **only when the recomputation reproduces what is actually stored** — so it is exact and self-verifying, and rows it cannot reproduce (the provider examples, seeded with a domain-separated `sha256("example:" + id)`) are skipped by construction. `id` is never recomputed: it is an FK from four other tables. **Downgrade does not reverse the re-stamp** (the per-writer shapes are gone), and that cost is written into the migration docstring rather than hidden.
2. **`consulted_paper_hashes` written by the agent runner changes content**, so newly-anchored traces hash differently. Old traces are **not** re-anchored — per the issue's anti-goal, they keep their (wrong, signed) single-paper field and are legacy by construction.
3. **`PaperRefResponse.title` and `StrategyResponse.paper_title` now return `null` instead of `""`**, and `_resolved_title` no longer falls back to the bare arXiv id. UI-visible; the renderers already have their own fallbacks (`{p.title || p.arxiv_id || "—"}`) and `ui` tests are green, but it is a payload change #1646 will build on.

### One acceptance criterion went away under me — please sanity-check the call

`main` moved 34 commits while this was in flight. **#1595 deleted the fusion bypass route** (`POST /api/strategies/generate` → `_run_fusion_job`) and added `test_sole_generation_route_guard.py` to keep it gone. That route held the construction-trace writer acceptance 7 asked me to fix, and `services/construction_trace.py` had already been deleted as a zero-caller surface — so on merged `main`, `git grep '"decision_type": "construction"' backend` returns **nothing**.

So I took main's deletion in full and **did not** admit `"construction"` to `STRATEGY_REFERENCE_DECISION_TYPES`. The case for admitting it was "fix that writer so the type becomes honest"; with no writer, widening a provenance filter for a decision type nothing produces buys nothing and costs the one promise that constant exists to make. The rule that *did* survive — **papers go in `consulted_paper_hashes`, never in `strategies_referenced`** — is now pinned by a source guard over all three live trace writers, with the deleted writer's own line as the input it must reject. Rationale is recorded on the constant itself, and in the merge commit, so nobody re-derives it.

**If you'd rather have `"construction"` admitted pre-emptively, say so and I'll add it** — I read the anti-goal against speculative provenance widening as pointing the other way.

Not merging. `Refs #1637` (not `Closes`) — see the not-done list at the bottom.

---

## What this is

`Refs #1637` — the **contract + trace-binding + verifier** half. The passport's **render** half is #1646's; nothing here touches layout.

**Principle from the issue: an association is a record, not a string.**

Before: four writers, **three incompatible key sets**, and the *shape* was inside the content hash.

| writer | key set |
| --- | --- |
| `main.py:294` | `{arxiv_id, title, authors}` |
| `debate_engine.py:533` | `{arxiv_id, title}` (always `""`) |
| `generation_pipeline.py:798` | `{arxiv_id, title}` — same key set as debate |
| `strategies_routes.py:2086` | `{arxiv_id, sha256}` (always `""`) |
| `strategy_store.py:53` (column comment) | `{arxiv_id, sha256}` — documented a shape nothing actually emitted |

`_compute_content_hash` hashed the **whole dicts**, so the same paper set through two writers produced two hashes, two ids, two "different" strategies. Backfilling a title would have forked one strategy into two again.

### 1. `assoc/v1` — one shape (NEW `backend/archimedes/models/paper_assoc.py`)

`{arxiv_id, role, content_hash, title, authors[], year, venue, doi, contribution, selection_rank, semantic_score, schema}` — every key always present, absent facts `None`. `""` normalizes to `None` everywhere ("null is honest; `""` is not"). `role` is closed to `cited` | `considered` and **raises** on anything else rather than coercing. Every writer emits it; `upsert_strategy` normalizes defensively as well.

**One deviation from the issue's literal wording, and why.** The issue says "one id space (`arxiv_id`)". Taken strictly that drops every curated strategy:

```
$ grep -h "PAPER_ARXIV_ID" analytics-engine/strategies/*.py | grep -c None
34          # out of 34 strategy files
```

All 34 declare no arXiv id and identify their paper by title (sometimes DOI). So `assoc_handle()` is `arxiv_id → doi → case-folded title`, and only an association naming *nothing at all* is dropped — that one being the `PaperRef(title=strategy_name)` placeholder, which is a fabricated citation. `passport_loader._ref_key` now delegates to the same function, so the store and the passport cannot drift into two ideas of "the same paper". The honest cost is stated at the definition: for a paper with neither arXiv id nor DOI, the title *is* the identifier, so a title backfill moves that one association's identity. Every arXiv-identified paper — i.e. every generated strategy's — is unaffected.

### 2. Identity, not shape, is what the hash sees

`assoc_identity()` reduces an association list to sorted, de-duplicated `(handle, role)` pairs, and that projection — nothing else — reaches the canonical JSON. Enrichment cannot move a strategy's id.

### 3. Trace binding

- `StrategySignals` gains `paper_arxiv_ids: list[str]` (the full cited set). `StrategyPassport.paper_arxiv_id` keeps a docstring that **forbids provenance use**.
- `agent_runner._paper_hashes_from_signals` records **every** cited paper as `f"{arxiv_id}:{content_hash or ''}"`. It used to emit one entry *per strategy*, with `content_hash = ss.strategy_id`, and — when the strategy had no arXiv id — the *paper id itself* was the strategy id. That field is in `_HASH_FIELDS` and is signed.
- `_route_helpers.persist_trace_off_chain` silently dropped `consulted_paper_hashes` — a hashed field, so a trace persisted through it could never rebuild its own canonical bytes. It has no callers today, which is exactly why the omission survived; fixed so wiring it later cannot inherit the trap.
- The fusion job's construction trace was going to be fixed here too; #1595 deleted the whole route first. See the merge note above.

Both new corpus reads run through `asyncio.to_thread` at their async call sites (`verify_trace`, `_commit_trace`): a blocking SQLAlchemy session on the event loop stalls every other in-flight coroutine — the same rule `_assert_can_read` follows since #1573.

### 4. Verifier wiring

`verify_source_papers` had **zero production callers**. `GET /api/traces/{id}/verify` now checks the trace's cited set against the live corpus and reports it **independently** of the on-chain check, tri-state:

```
papers_verified: true | false | null
source_paper_verification: {mode, checked, verified, missing[], hash_mismatch[]}
mode: "checked" | "no_papers_claimed" | "corpus_unavailable"
```

`null` means *not checked*. A corpus outage is neither a pass nor a failure — the same failure class #1359 fixed on the on-chain half. Per the issue's anti-goal, an **empty claimed hash asserts nothing** (corpus `content_hash`/`pdf_sha256` are NULL until #1091), so this is an existence check today and says so; nothing synthesizes a hash on either side.

### 5. Passport projection (exposed, not rendered — #1646 renders it)

- `PassportPaperRef` gains `role` / `selection_rank` / `semantic_score` / `content_hash`; `to_dict()` gains `contribution`, which it had been silently omitting while `StrategyPassport.jsx` rendered a column for it.
- `ingest_passport(force_update=True)` **merges by `(passport_id, assoc_handle)`** — the issue says `(passport_id, arxiv_id)`; same rule, widened by the same curated-papers reason as above. It used to `DELETE FROM passport_paper_refs` and rebuild — and it runs on every real-returns refresh from a caller that has **id + title only**, so enrichment could not survive by construction. Three rules: match don't replace; never overwrite a value with an absence; never blind-delete (an empty incoming list is ignorance, not a deletion instruction).
- `PaperRefResponse` carries the new fields. `title` and `paper_title` return `null` instead of `""`, and `_resolved_title` no longer falls back to the bare `arxiv_id` — an id presented as a title is a small fabrication, the renderer already composes `{p.title || p.arxiv_id || "—"}` itself, and this aligns the passport with `_resolve_source_papers`, whose `resolved_title` has always stayed `None` here.

---

## Files

**New**
- `backend/archimedes/models/paper_assoc.py` — the `assoc/v1` contract
- `backend/migrations/versions/b41c7e0d95a2_assoc_v1_paper_associations.py` — projection columns + the one-time re-stamp
- `backend/tests/test_paper_assoc_contract.py` — 47 cases

**Changed** — `models/{strategy_store,strategy_passport_record,paper_ref,strategy}.py` · `services/{passport_loader,source_tracker,paper_trace,redis_state,strategy_signal_evaluator}.py` · `api/{traces_routes,strategies_routes,schemas,_route_helpers}.py` · `chain/agent_runner.py` · `agents/{debate_engine,generation_pipeline}.py` · `main.py` · 3 existing test files updated (each pinned behaviour this issue deliberately changes — itemised below)

Three commits: the work, then two merges of `origin/main` (it moved 35 commits while this was in flight). **Read the first merge's message** — its conflict resolution is a decision, not a mechanical fix. The second picks up #1673, which fixed a `test_cloudwatch_alarms` failure that was already red on the `main` I first merged (verified by running that test on a detached checkout of `b359f324`) and is unrelated to this work.

---

### Full unit suite (the exact CI command), from the repo root

```
$ pytest -q -m "not integration"
5090 passed, 3 skipped, 2 deselected, 297 warnings in 1250.73s (0:20:50)
```

```
$ ruff format --check .
683 files already formatted
$ ruff check --select E9,F63,F7,F40,F82 .
All checks passed!
```

Hermetic gate (CLAUDE.md § Testing conventions):

```
$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest \
    backend/tests/test_paper_assoc_contract.py -q
47 passed

$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest \
    backend/tests/test_alembic_migrations.py -k assoc \
    backend/tests/services/test_xia_protocols.py -q
4 passed, 70 deselected, 1 warning in 19.21s
```

### Every regression test revert-demo'd

Each fix was reverted in place, the test run, and the tree restored. **A test that passes against the unfixed code proves nothing** (CLAUDE.md rule 3).

```
### R1 — revert the hash to whole-dict hashing (pre-#1637)
$ pytest backend/tests/test_paper_assoc_contract.py::TestContentHashIgnoresEnrichment
FAILED ...::test_enrichment_does_not_change_the_hash
FAILED ...::test_all_three_legacy_shapes_hash_to_one_value
FAILED ...::test_order_and_duplicates_do_not_change_identity
FAILED ...::test_a_doi_identified_paper_keeps_its_identity_under_enrichment
4 failed, 4 passed

### R2 — revert passport merge to the blind DELETE-then-rebuild
$ pytest backend/tests/test_paper_assoc_contract.py::TestReingestPreservesEnrichment
FAILED ...::test_backfilled_fields_survive_an_id_and_title_only_refresh
FAILED ...::test_a_real_new_value_still_overwrites
FAILED ...::test_an_empty_paper_list_is_ignorance_not_a_deletion
3 failed, 1 passed

### R3 — revert to_dict() to the version that omits `contribution`
$ pytest backend/tests/test_paper_assoc_contract.py::TestContributionRoundTrips
FAILED ...::test_contribution_reaches_the_column_the_dict_and_the_response
1 failed, 2 passed

### R4 — WITHDRAWN (it demonstrated the "construction" scope widening, which
###       the merge with main removed; R11/R12 below replace it)

### R5 — un-wire verify_source_papers from /verify (its pre-#1637 state: zero callers)
$ pytest backend/tests/test_paper_assoc_contract.py::TestVerifySourcePapersIsWired
FAILED ...::test_absent_paper_reports_papers_verified_false_and_names_it
FAILED ...::test_present_papers_report_verified_true
FAILED ...::test_a_claimed_hash_that_disagrees_is_a_mismatch
FAILED ...::test_no_claimed_papers_is_not_checked_rather_than_passed
FAILED ...::test_a_corpus_outage_is_not_a_provenance_failure
5 failed, 1 passed

### R6 — revert _paper_hashes_from_signals to the strategy-id-as-content-hash version
$ pytest backend/tests/services/test_xia_protocols.py::TestSourceTrackingRuntimeWiring
FAILED ...::test_records_every_cited_paper_never_the_strategy_id
FAILED ...::test_suffix_is_the_corpus_hash_when_the_corpus_has_one
FAILED ...::test_corpus_outage_still_records_the_papers
3 failed, 2 passed

### R7 — reintroduce the debate engine's legacy {arxiv_id, title: ''} shape
$ pytest backend/tests/test_paper_assoc_contract.py::TestAssocSchemaIsSingleShape
FAILED ...::test_no_legacy_paper_shape_survives_at_any_writer
1 failed, 10 passed

### R8 — revert to_dict()'s title to '' instead of None
$ pytest ...::TestContributionRoundTrips::test_to_dict_never_prints_an_empty_title_as_a_title
FAILED ...::test_to_dict_never_prints_an_empty_title_as_a_title
1 failed

### R9 — remove the collision guard from the migration (naive UPDATE)
$ pytest backend/tests/test_alembic_migrations.py::test_assoc_migration_restamps_only_rows_that_reproduce_their_stored_hash
FAILED ...::test_assoc_migration_restamps_only_rows_that_reproduce_their_stored_hash
1 failed

### R10 — revert paper_title to '' for a zero-paper row, and _resolved_title to the arxiv-id fallback
$ pytest backend/tests/test_paper_assoc_contract.py::TestSingleAndZeroPaperPassportsAreHonest
FAILED ...::test_zero_paper_row_reports_a_null_title_not_empty_quotes
1 failed, 1 passed

### R11 — route papers back into strategies_referenced (the deleted writer's habit)
$ pytest backend/tests/test_paper_assoc_contract.py::TestPapersNeverLandInTheStrategyField
FAILED ...::test_no_live_writer_puts_an_arxiv_id_in_strategies_referenced
1 failed, 3 passed

### R12 — drop consulted_paper_hashes from persist_trace_off_chain again
$ pytest backend/tests/test_paper_assoc_contract.py::TestPapersNeverLandInTheStrategyField
FAILED ...::test_every_trace_writer_carries_the_hashed_paper_field
1 failed, 3 passed
```

All twelve were re-run against the post-merge tree; the transcripts above are that run.

R10 is worth a note on method. My first version of that test asserted
`(passport.paper_title or None) is None` — a **tautology**: it recomputes the
fix inside the assertion, so it passes identically against unfixed code. That is
CLAUDE.md rule 3 ("passing the same literal to both sides of the boundary the
bug lives on"), and I wrote it anyway. Caught it on re-reading the file before
demoing. It now drives the real `_passport_to_strategy_response`, and the demo
above shows it failing when reverted.

### Every guard shown rejecting

**R9 is the important one, and it is a bug I caught by running the guard rather than by reading the code.** `strategy_store` carries `UNIQUE(content_hash)`. The whole point of the new identity is that rows differing only by writer shape now *agree* — so on a database that actually holds the split-brain this issue describes, two rows compute the same value and a naive `UPDATE` aborts the migration on a constraint violation. First re-stamp wins (ordered by `created_at`); a colliding row keeps its legacy hash and is **named in a WARNING**. Merging is deliberately not done: `strategy_store.id` is an FK from the generator, marketplace, billing and deployment tables, so collapsing two rows is an ownership call, not a schema step.

Contract guards, run against the inputs they must reject:

```
$ python -c "<see backend/tests/test_paper_assoc_contract.py>"
make_assoc("2301.1", role="probably")
  -> rejected: ValueError: assoc role must be one of ['cited', 'considered'], got 'probably'
normalize_assoc({"arxiv_id":"2301.1","role":"maybe-cited"})
  -> rejected: ValueError: assoc role must be one of ['cited', 'considered'], got 'maybe-cited'
assert_assoc({"arxiv_id":"2301.1","sha256":""})   # the fusion job's legacy shape
  -> rejected: ValueError: assoc/v1 key mismatch — missing=['authors', 'content_hash',
     'contribution', 'doi', 'role', 'schema', 'selection_rank', 'semantic_score',
     'title', 'venue', 'year'] extra=['sha256']
assert_assoc({"arxiv_id":"2301.1","title":""})    # the debate engine's legacy shape
  -> rejected: ValueError: assoc/v1 key mismatch — missing=[...] extra=[]
assert_assoc(valid | {"schema": "assoc/v2"})      # tampered schema tag
  -> rejected: ValueError: assoc/v1 schema tag must be 'assoc/v1', got 'assoc/v2'

accepted (control): make_assoc("2301.1", role="considered") -> considered
is_assoc(legacy fusion shape) -> False
is_assoc(make_assoc("2301.1"))  -> True
```

Guards whose adversarial companion lives **inside the same test**, so the guard can never pass while being unable to fail:

| guard | the input it was shown rejecting |
| --- | --- |
| legacy-shape source scan over the four writers | the literal reintroduction of each legacy shape (and R7, in real code) |
| no-semantic-claim scan (acceptance 13, precedent #778) | `"papers found by embedding similarity"` → hit; `"sourced from the Knowledge Graph"` → hit |
| `/verify` corpus-outage tri-state | `CorpusUnavailable` → `papers_verified: null`, `mode: "corpus_unavailable"` — **not** `false` |
| papers-never-in-the-strategy-field scan over the three live trace writers | the deleted `_run_fusion_job` line verbatim: `"strategies_referenced": result.source_arxiv_ids,` → hit |
| `trace_references_strategy` whole-string match | a trace listing `["2301.00001", …]` matches no strategy id, and `"2301.0000"` does not substring-match |
| `verify_source_papers` empty claimed hash | asserts nothing → existence-checked, not counted as a mismatch |

---

## Acceptance criteria — status

| # | criterion | status |
| --- | --- | --- |
| 1 | `test_assoc_schema_is_single_shape` | ✅ `TestAssocSchemaIsSingleShape` (11 cases, incl. an end-to-end "three writers → one stored key set") |
| 2 | `test_content_hash_ignores_enrichment` | ✅ `TestContentHashIgnoresEnrichment` (8 cases, incl. the DOI-identified curated case) |
| 3 | `test_selection_record_written_for_every_candidate` | ❌ **not done** — `paper_selections` table not built. See not-done list. |
| 4 | `test_full_cited_set_survives_to_passport` (**the pin**) | ✅ `TestFullCitedSetSurvivesToPassport` (2 cases) — the pin passes on unmodified `main` too; that is the point |
| 5 | `test_trace_records_every_cited_paper` | ✅ `TestTraceRecordsEveryCitedPaper` (4) + `TestSourceTrackingRuntimeWiring` (5, rewritten — see the not-done list) |
| 6 | `test_verify_source_papers_is_wired` | ✅ `TestVerifySourcePapersIsWired` (6 cases) |
| 7 | `test_construction_trace_cites_papers_in_the_paper_field` | ⚠️ **moot — the writer was deleted upstream by #1595.** Replaced by `TestPapersNeverLandInTheStrategyField` (4 cases), which pins the surviving rule across all three live writers. See the merge note. |
| 8 | `test_reingest_preserves_enrichment` | ✅ `TestReingestPreservesEnrichment` (4 cases) |
| 9 | `test_contribution_round_trips` | ⚠️ **plumbing only** — round-trips column → `to_dict()` → `PaperRefResponse`. Nothing *writes* a per-paper contribution yet. See not-done list. |
| 10 | `test_abstract_citations_are_grounded` | ❌ **not done** — no `[arxiv:ID]` markers. See not-done list. |
| 11 | `test_proposal_joins_to_strategy` | ❌ **not done** — `StrategyProposal` still has no `strategy_id`. |
| 12 | `test_single_and_zero_paper_passports_render_honestly` | ⚠️ **backend half** — `TestSingleAndZeroPaperPassportsAreHonest`; the `> 1` → `>= 1` render condition is #1646's |
| 13 | `test_no_semantic_claim_on_provenance_surfaces` | ✅ `TestNoSemanticClaimOnProvenanceSurfaces` (3 cases), with the guard shown firing |

**Anti-goals — checked, not assumed**

- ✅ No passport layout/visual work. Zero files under `ui/` in this diff.
- ✅ Nothing fabricated. `null` authors/DOI/venue/year stay `null`; no "unknown" placeholders; no inferred contributions. The normalizer turns `""` into `None` rather than the reverse.
- ✅ No verification claimed before the verifier had a caller — and the endpoint's honest-empty behaviour is explicit: `corpus_content_hashes` returns what the corpus actually has, empty claimed hashes assert nothing, and no hash is synthesized to make a checkmark green.
- ✅ Retrieval untouched: no change to `select_candidates`, the 150 rerank cap, or corpus size.
- ✅ No existing on-chain trace re-anchored. Fix-forward only; old traces keep their signed field.

---

## Honest not-done list

**Out of scope for this PR** (my assignment scoped me to the contract + trace binding + verifier; the render half is #1646):

1. **`paper_selections` table (issue item 1, acceptance 3).** Not built. The `assoc/v1` contract *supports* it — `role="considered"`, `selection_rank`, `semantic_score` all exist and are plumbed through to `passport_paper_refs` — but **nothing writes a `considered` association today.** `strategy_fusion.py:722` still discards its selection list. "Considered but not cited" is still unanswerable; the schema is now ready for the answer.
2. **`contribution` is plumbed but never written (acceptance 9).** The full round-trip works and is tested. What is missing is the *source*: `fusion_reasoning` is a single free-text blob, so splitting it per-paper needs a prompt change (`strategy_fusion.py:836`) that emits structure. I did not fabricate a split — inferring "what paper X contributed" from prose is exactly the kind of invented provenance this issue's anti-goals forbid.
3. **Abstract `[arxiv:ID]` citations (issue item 6, acceptance 10).** Not done — prompt + post-parse validator, same `strategy_fusion` surface as (2).
4. **`StrategyProposal.strategy_id` (acceptance 11).** Not done. Proposal rows remain unjoinable except through `extra.candidate_id`.
5. **Frontend (issue item 7).** Not touched, by design → **#1646**. Specifically still open there: linkifying `[arxiv:ID]`, the `length > 1` → `>= 1` papers-table condition, and the bare-quotes render at `StrategyPassport.jsx:604`/`:607`. The backend now hands them `null` instead of `""`, plus `role` / `selection_rank` / `semantic_score` / `content_hash` / `contribution` on every `PaperRefResponse`.

**Optional follow-up, not done here:** `docs/CONVENTIONS.md` routes a frozen schema to `docs/specs/` with a `docs/README.md` row. `assoc/v1` arguably qualifies. I left it out to keep this one logical change — say the word and I'll add the spec doc + index row.

**Known limits of what IS here:**

6. **Pre-existing duplicate rows are disclosed, not merged.** If two rows already collapse to one identity, the oldest takes it and the others keep their legacy hash and are logged by id. Reconciling them needs an FK/ownership decision I am not taking in a migration.
7. **`build_consulted_hashes` and `paper_trace.resolve_paper_hashes` now differ on purpose.** The association path records *every* cited paper with an empty suffix when no hash resolves (the issue's item 4); `paper_trace` still drops unresolved ids, because on that path the bare ids are separately recorded in `market_context.source_arxiv_ids`. Both policies are documented at their definitions. If you'd rather they converge, say which way.
8. **`_paper_hashes_from_signals` now does a corpus read per trace.** Same query `paper_trace` already runs per settle, and an outage degrades to "ids without hashes" rather than raising — but it is new I/O on the agent tick path.
9. **Three existing test files were updated because they pinned the old behaviour.** Each is a deliberate change, not a test bent to fit:
   - `test_xia_protocols.py::TestSourceTrackingRuntimeWiring` asserted `any("abc123" in h)` where `abc123` was a **strategy id** — it was guarding the defect in place.
   - `test_strategy_store.py` asserted the stored column equals the caller's literal legacy input.
   - `test_multipaper_passport.py` asserted `title == arxiv_id` for an unresolvable title.

   (`test_traces_strategy_scope.py` was edited and then **reverted to main's version** during the merge, once the `"construction"` widening was withdrawn. It is unchanged in the final diff.)
10. **Two second-order bugs I introduced and then caught by running things, not by reading them** — recorded because the review value is in the pattern, not the fix: (a) the migration's naive `UPDATE` aborted on `UNIQUE(content_hash)` (R9); (b) `paper_arxiv_ids=[p.arxiv_id for p in strategy.papers]` crashed `evaluate_strategies` on a duck-typed strategy with no `papers`, killing a whole market scan — now `getattr(..., None) or []`.
11. **Branch is level with `origin/main`** (0 behind at the time of writing) and the suite result above is the *merged* result, not this branch in isolation — CLAUDE.md rules 1, 2 and 5.
12. **Not verified against a live database.** All migration evidence is SQLite via the repo's own alembic harness. The re-stamp's behaviour on prod Aurora — row count, how many rows reproduce, how many collide — is unknown until someone runs it; the WARNING is there so it is not silent.


danielscoffee (changes requested):
## REQUEST_CHANGES — owner surface, irreversible restamp, live DB unknown

CLAUDE.md names strategy-passport / reasoning-trace as ask-before-acting. This PR is honest about that and does not merge itself. I am not rubber-stamping the data migration or the hash-definition change.

The contract work (`assoc/v1`, identity-only hashing, merge-not-delete passport refs, `/verify` paper check) is real and well-tested. That is not enough to approve: the restamp is not reversible, it has not been run against Aurora, and the owner questions in the PR body are unanswered (no human review/comment in-thread; only github-actions nags).

### Owner questions still open — please answer in-thread before merge

**Q1. Accept the `content_hash` redefinition + irreversible restamp?**
Migration `b41c7e0d95a2` rewrites `strategy_store.content_hash` when `_compute_content_hash_v0` reproduces the stored value. `id` is left alone (FK from generator / marketplace / billing / deployment). **Downgrade drops the new `passport_paper_refs` columns and does not restore the old hashes or the per-writer `source_papers` shapes.** After restamp, existing `id` values are no longer `content_hash[:16]` (new rows still are). Anything that treated `content_hash` as the on-chain / cross-system identity desyncs; FKs on `id` stay valid.

**Q2. Accept `consulted_paper_hashes` changing for newly-anchored traces?**
`agent_runner._paper_hashes_from_signals` now records every cited arXiv id as `{arxiv_id}:{corpus_hash or ''}`. That field is in `_HASH_FIELDS` and is signed. Old traces are not re-anchored (matches the issue anti-goal). New traces hash differently from old ones by construction.

**Q3. Accept `PaperRefResponse.title` / `StrategyResponse.paper_title` returning `null` instead of `""`, and dropping the arXiv-id-as-title fallback?**
UI-visible payload change. Author says renderers already do `{p.title || p.arxiv_id || "—"}` and that this is for #1646. Confirm that is acceptable on the live passport.

**Q4. Admit `"construction"` to `STRATEGY_REFERENCE_DECISION_TYPES` pre-emptively?**
Author recommends **no** — the writer was deleted by #1595; widening a provenance filter for a type nothing produces is speculative. I agree with leaving it out unless you want the constant ready for a future writer.

**Q5. Converge `build_consulted_hashes` vs `paper_trace.resolve_paper_hashes`?**
They now differ on purpose (empty suffix vs drop unresolved ids). Author asks which way, if at all.

**Q6. Land an `assoc/v1` spec under `docs/specs/`?**
Optional; author will add it if you say so.

**Q7. Duplicate-row ownership?**
Colliding rows are **not** merged. Oldest takes the new hash; others keep the legacy hash and are WARNING-logged by id. Reconciling them is an FK/ownership call this migration correctly refuses to make. Who owns that follow-up?

**Q8. Is existence-only `papers_verified: true` acceptable?**
Corpus `content_hash`/`pdf_sha256` are NULL until #1091. `/verify` treats an empty claimed suffix as existence-only and can still return `papers_verified: true`. Honest relative to the anti-goal (nothing is synthesized), but it is a green check without a hash comparison. Confirm the button copy can live with that.

**Q9. `assoc_handle` deviation from the issue's "one id space (`arxiv_id`)"?**
Handle is `arxiv_id → doi → case-folded title` so 34 curated strategies with `PAPER_ARXIV_ID = None` are not dropped. Cost: a title backfill **moves identity** for papers with neither arXiv id nor DOI, and therefore can move `content_hash`. Generated (arXiv-identified) strategies are unaffected.

### Merge blockers (not nits)

1. **No Aurora dry-run.** Author states this explicitly (PR not-done #12). Tests seed SQLite with `_compute_content_hash_v0`, so they cannot tell us how many prod rows reproduce. If v0 matches few/no Aurora rows, **hashes are not restamped**, `source_papers` is still normalized, and post-deploy `upsert_strategy` (no v0 fallback — by design) **inserts duplicates** — the split-brain this PR exists to remove. Do not apply this migration until someone reports: row count, restampable, collisions, skipped (examples + non-reproducing). If reproduce-rate is ~0, abort; do not ship a silent JSON rewrite.

2. **Deploy order is load-bearing and undocumented.** New writers must not run against un-restamped rows. If the new code deploys before the migration, upserts insert new-hash rows; the later restamp then hits `UNIQUE(content_hash)` and leaves the originals on legacy hashes. Permanent split-brain. Require: migrate first (or atomically with the release), and do not roll the app back independently after restamp (downgrade cannot rebuild v0 hashes).

3. **Leftover split-brain is the success path for collisions and examples.** Example rows (domain-separated `sha256("example:" + id)`) and colliding duplicates keep legacy hashes forever. Next upsert of that content will not attach to those rows. That is disclosed; it is still an owner call, not a default.

### What I did confirm (not blocking)

- `GET /api/traces/{id}/verify` actually calls `verify_source_papers`; tests hit the ASGI route (`TestVerifySourcePapersIsWired`, 6 cases). Corpus outage → `papers_verified: null` / `mode: "corpus_unavailable"`, not `false`. Independent of the on-chain check. Claim matches code.
- `_compute_content_hash_v0` is a byte-faithful copy of `main`'s current hasher (sorted whole dicts + `sort_keys=True`). Self-verify is the right shape **if** prod was written with that function.
- Collision path exists and is tested (`test_assoc_migration_restamps_only_rows_that_reproduce_their_stored_hash`). Naive UPDATE would abort on `uq_strategy_content_hash`.
- Passport reingest merges by handle; empty incoming list is not a delete (`TestReingestPreservesEnrichment`).
- Verifier / hash / writer guards have revert-demos in the PR body. I did not re-run them.

### Out of scope (accepted as #1637 remainder, not this PR)

`paper_selections`, per-paper `contribution` writer, `[arxiv:ID]` abstract markers, `StrategyProposal.strategy_id`, frontend (#1646). `Refs` not `Closes` is correct.

Please answer Q1–Q9 and paste Aurora dry-run counts. Then this can be re-reviewed. Until then, do not merge.

dbrowneup:
## Owner answers — Q1–Q9, plus where I want this PR to go

Answering in-thread as requested. Daniel's three concerns (owner surface, irreversible restamp, live DB unknown) are the right ones; my call on the third changes the shape of the PR.

**Q1. "Accept the `content_hash` redefinition + irreversible restamp?"**
Yes to the redefinition, not yet to the restamp. Identity-only hashing is correct and is the whole point of #1637 — enrichment must not fork a strategy. But `id` is the identity every other table joins on, and under #1792 the passport row keyed by `id` is now the rigor verdict of record, so nothing downstream reads `content_hash` as an identity. That makes the restamp a dedup-hygiene step, not an identity migration, and a dedup-hygiene step does not get to be irreversible on an unmeasured table. Restamp lands only after the dry-run in the recommendation below.

**Q2. "Accept `consulted_paper_hashes` changing for newly-anchored traces?"**
Yes, the new content is the honest one — recording every cited paper instead of one entry per strategy hashed with the strategy id was a straight bug. But #1800 is about to add the raw-completion hash to the same trace preimage, so I do not want two re-anchoring boundaries a week apart. Sequence this field change to land with or after #1800's preimage change so there is exactly one "traces before this date hash differently" line to explain.

**Q3. "`title`/`paper_title` → `null`, and drop the arXiv-id-as-title fallback?"**
Accepted. An arXiv id printed in a title slot is a small fabrication and this is a validation product; `null` is honest and the renderers already compose their own fallback. Ship it, and note it in #1646 so the render half does not reinvent the fallback server-side.

**Q4. "Admit `construction` to `STRATEGY_REFERENCE_DECISION_TYPES` pre-emptively?"**
No — agreed with both of you. #1595 deleted the writer; widening a provenance filter for a decision type nothing emits costs the one promise the constant makes. The source guard over the three live writers is the right residue. If a construction writer returns, it admits itself in that PR.

**Q5. "Converge `build_consulted_hashes` vs `paper_trace.resolve_paper_hashes`?"**
Converge on the trace side's rule — record the id with an empty suffix rather than dropping an unresolved id. Dropping loses the claim entirely; an empty suffix preserves "this paper was claimed, no corpus hash existed," which is exactly the distinction #1800 is trying to protect elsewhere. File it as a follow-up, not a blocker on this PR.

**Q6. "Land an `assoc/v1` spec under `docs/specs/`?"**
Yes, in this PR — it is a contract other writers must conform to, and a contract that lives only in a merged diff gets re-derived wrong. Keep it under the docs-site exclusions; this is internal engineering reference, not published material.

**Q7. "Duplicate-row ownership?"**
Mine. The migration is right to refuse it — collapsing two rows with FKs from generator/marketplace/billing/deployment is an ownership decision, not a schema step. Open a follow-up issue that takes the WARNING-logged ids as its input; it cannot be written before the dry-run tells us whether the set is empty.

**Q8. "Is existence-only `papers_verified: true` acceptable?"**
Acceptable, with the copy fixed. Corpus `content_hash`/`pdf_sha256` are NULL until #1091, so the check is real but weaker than a hash comparison. The button must not say "verified" unqualified — it says the paper exists in the corpus, and the tri-state `mode` must be surfaced, not just returned. Nothing here touches money: execution is paper trading, and #1240 closed the mainnet cutover, so this is a truthfulness question, not a settlement one.

**Q9. "`assoc_handle` deviating from one id space?"**
Accepted as written. `arxiv_id → doi → case-folded title` keeps all 34 curated strategies; strict arXiv-only would drop every one of them. The stated cost — a title backfill can move identity for a paper with neither id nor DOI — is real and bounded to curated rows, and #1792's stored verdict keys off `id`, which never moves. Keep the cost documented at the definition.

### What #1688 needs to become mergeable

Not superseded by #1800 — they are complementary, and the contract work here is the better-tested half. But split it:

1. **PR-1 (this branch, restamp removed):** `assoc/v1`, identity-only hashing, merge-not-delete passport refs, `/verify` wiring, the spec doc, the schema-only migration (new `passport_paper_refs` columns + `source_papers` normalization). Reversible, no hash rewrite. Needs a merge from main first — the branch is currently conflicting.
2. **PR-2 (restamp, gated):** blocked on a read-only Aurora dry-run. The hash is keccak256 via web3, so this cannot be a SQL query — it is a read-only script importing the migration's own `_compute_content_hash_v0`, run against the reader endpoint or a snapshot restore, reporting: total rows, examples, reproduced, collision groups with ids, and non-reproducing non-example ids. If the reproduce rate is near zero, we abort and fix forward instead. PR-2 also documents the deploy order (migrate first or atomically; no independent app rollback after restamp) — Daniel's blocker 2, which is correct and currently undocumented.
3. **PR-3:** the `consulted_paper_hashes` change, riding #1800's preimage boundary.
```

🌱 graft saved ~12,440 tokens this turn

dbrowneup:
Superseded by the owner's split (comment of 2026-09-03): PR-1 (the contract half, schema-only) landed as #1845 on 2026-09-04; PR-2 (the re-stamp, gated on a read-only dry-run) and PR-3 (consulted_paper_hashes, riding #1800) are the remaining follow-ups tracked on #1637. Closing this branch unmerged.

Issue #1691, opened 2026-08-31
[auth][security] Better Auth rate limits collapse to one global bucket in production — no client IP resolves behind CloudFront/ALB/nginx


dbrowneup:
## Summary

Better Auth's production rate limiter has **no client IP to key on**, so every bucket it
maintains is global per-path rather than per-caller. Three signups from anywhere in the
world exhaust `/sign-up/email` for everyone for ten minutes; ten sign-in attempts exhaust
`/sign-in/email` for everyone for a minute; and `/request-password-reset` and
`/send-verification-email` are three-per-minute **globally**. That last pair is exactly the
traffic the `EMAIL_VERIFICATION_ENFORCED` flip creates, which is why this is being filed
now rather than as a general hardening item.

This also makes two shipped claims untrue, which is the CLAUDE.md §1 problem rather than
only a availability one:

- `auth/auth.js`'s `rateLimit.customRules` comment: *"3 signups / 10 min per Better Auth's
  rate key"*.
- `docs/account-authentication.md` § "Signup friction and abuse bounds": *"Better Auth's
  production rate limiter: `/sign-up/email` at 3 per 10 minutes"*, listed as the first of
  three independent abuse layers.

Neither is per-caller in production. The nginx `limit_req` zone (layer 2) and the per-IP
daily generation cap (layer 3) are unaffected — this is layer 1 only — but layer 1 is the
one that guards the mail endpoints, and mail endpoints are what SES production access is
judged on.

### Mechanism, verified against the installed library

`@better-auth/core/dist/utils/ip.mjs`:

- `createRateLimitKey(ip, path)` (:225) builds every bucket key as `${ip}|${path}`.
- `getIp` (:201) resolves that IP from `x-forwarded-for` and, with no
  `advanced.ipAddress.trustedProxies` configured — `auth/auth.js` sets no
  `advanced.ipAddress` block at all — trusts the header **only when it carries exactly one
  value**: `if (forwardedIps.length !== 1) return null` (:189). Every other token in a chain
  is spoofable, so refusing is the correct library behaviour.
- When it returns `null`, `api/rate-limiter/index.mjs:288` substitutes the literal
  `"no-trusted-ip"` and logs, once: *"Rate limiting could not determine a client IP and is
  falling back to a single shared per-path bucket."*

In production the request reaches the auth container through CloudFront → ALB → nginx and
each hop appends to the header: CloudFront sets the client, the ALB appends the edge it saw,
[...]

PR #1653, opened 2026-08-31
[docs][agent-native] Agent-native onboarding design + owner decision list

dbrowneup:
Closes #1644

A design doc and an owner decision list. **No production code changes** — the diff is three files: the new doc, its `docs/README.md` index row, and one `mkdocs.yml` nav line.

`docs/specs/agent-native-onboarding-spec.md` — how a CLI / agent-skill / (possible) MCP caller creates an account, links a wallet, and pays; what is measured about it; how it reconciles with the 3-free-generation gate in #1643; and **six decisions only you can close**.

Written as **deltas** from `docs/agent-quickstart.md` and `docs/agent-api.md`, not as a second copy of them. Those two remain the recipe and the reference; this doc names only the places the agent path *differs* and the places a decision is *missing*.

---

## What you have to decide

The decision table is § 7 of the doc. Six rows, each with an option set and what breaks under each, in the format of #1432's "What you have to decide". Summarised so you can triage before opening the file:

| # | Question | Why it can't be answered without you |
| --- | --- | --- |
| **D1** | Do agents get the same **3 free generations** as a browser account? | This is the #1643 reconciliation. Either answer is fine; *not* answering it is the drift this doc exists to prevent. |
| **D2** | Does an **MCP** server ship at all? | No MCP server exists in this repo. `architectural-principles.md` calls "MCP-native" part of the pitch — if D2 is "no", that line needs softening in the same breath. |
| **D3** | Is there a **non-cookie credential** for agents (API key)? | There is no bearer token, no PAT, no API key today. Every CI job re-implements a cookie jar and re-sends a password on a 7-day cycle. |
| **D4** | Do we want **per-agent attribution**, or is per-account enough? | See finding 1 below — today the answer is structurally "we cannot see it". |
| **D5** | What happens at the **faucet wall** (#1294)? | Money and keys. Not delegable. |
| **D6** | Does the **CLI** grow to cover the money path? | `archimedes` 0.1.0 can't finish a production generation. Also decides the PyPI question. |

There is also an **"Already decided — do not reopen"** table (account always required per your #1643 directive; credit-not-refund per its ADR; `headless` provider for API callers; quota-before-paywall; ERC-8004 `registration_pending`), so a follow-up PR doesn't relitigate them.

---

### 1. An agent stops being an agent the moment it signs in

[...]

Issue #1643, opened 2026-08-31
[product][auth] Free path: account required, wallet optional for the first 3 generations, then wallet-gate


dbrowneup:
**Why (owner, 2026-08-31 product review):** account required, wallet optional at first — 3 generations on a mere account, then wallet-gate. No wallet-only-without-account path.

## Summary

This **reverses a 12-day-old explicit policy**, and the issue must say so rather than pretend it's new ground: `backend/archimedes/services/generation_payment.py`'s module docstring, dated to Dan's own 2026-08-19 directive, currently reads *"Generation REQUIRES wallet connection + payment — for humans and agents alike, no free path."* In production (`GENERATION_PAYMENT_REQUIRED="true"`), `generate_routes.py`'s `start_generation` (line ~394) requires `require_current_user` (account — already correct) and then, if no wallet is linked, refuses with `409 wallet_link_required` (line ~457-475) on the caller's **very first** generation. There is no free-generation allowance today; `services/generation_quota.py`'s `GenerationQuota` is a rolling 36-hour rate cap (10/account/day, 20/IP/day per `.env.example`), not a lifetime "first N are free" counter — it gates volume, not the wallet requirement. `docs/agent-quickstart.md` (dated 2026-08-30, one day before this review) documents the current no-free-path behavior at Step 1 and must be updated in lockstep once this ships.

## Scope

- New durable, lifetime (not daily-resetting) per-account free-generation counter — modeled on `backend/archimedes/models/generation_credit.py`'s durability discipline (a real DB row, not a Redis-only counter that a cache flush could reset generosity on). Simplest correct shape: a new small table or a counter column reachable from `auth_users.id`.
- `backend/archimedes/api/generate_routes.py`'s `start_generation` — before the `if not linked_wallet:` 409 branch (line ~458), check the account's free-generation count: if `< 3`, skip both the wallet-link requirement and the paywall (`_paywall_with_credit`) for this call, and increment the counter. At `>= 3`, today's wallet-gate + paywall behavior is unchanged.
- `backend/archimedes/api/account_usage_routes.py`'s `GET /api/account/usage` — surface `free_generations_remaining` alongside the existing daily-cap fields, matching its existing `DailyCapUsage`-style honesty (`None`/an explicit error field on a backend failure, never a fabricated `0` or `3`).
[...]

PR #1695, opened 2026-08-31
[auth] Audit and harden the email-verification and password-reset flows before the enforcement flip

### LINKED ISSUE #1693: [auth][security] Validate the email-verification and password-reset flows before EMAIL_VERIFICATION_ENFORCED flips — audit, hermetic harness, runbook
## Summary

SES production access is approved and live, but neither transactional mail flow Better Auth
owns has been validated end to end. Verification mail has been sending on every signup since
#1194, and password reset has been wired since #1323 — but the whole time the AWS account was
in the **SES sandbox**, where a send to any address that is not an individually-verified
destination identity is rejected by SES and then deliberately swallowed fail-soft by
`auth/auth.js` (loud log line, signup still returns 200). So the honest state of the world is
that *no signal has ever reached a user either way*, and `EMAIL_VERIFICATION_ENFORCED`
(`infra/ecs.tf`, flip-list row 3) is queued to flip on the strength of an untested path.

Password reset has never been run against a real inbox at all.

Flipping enforcement is a task-definition env change plus `terraform apply` — cheap to do and
cheap to revert — but it turns `/sign-in/email` into a hard refusal for every account with
`emailVerified = false`, and it promotes two endpoints that have never carried real traffic
(`/send-verification-email`, `/request-password-reset`) into the critical path. This issue is
the work that has to happen *before* that, in three parts:

1. **Code truth.** Map both flows as actually built — what triggers a send, what the token
   lifetimes really are, what is single-use and what is replayable, what the flag does and
   does not gate — and write the findings down where a reviewer can check them against the
   source rather than against prose.
2. **Hermetic tests.** The half a human with a mailbox cannot cover cheaply or repeatably:
   expiry, replay, revocation, and the information the request endpoints leak. Mocking stops
   at the mailer boundary, per CLAUDE.md § "Mock at boundaries, not internals" — everything
   else is the real library.
3. **A runbook** for the half that structurally cannot be automated: that a message left AWS,
   survived the receiving provider's filter, and that the link works when a human clicks it.

**No live sends are in scope here.** This issue produces code, tests, and a procedure; the
procedure is executed later, by the owner, against a real inbox.

## Scope

Do exactly this, nothing more.

**Audit (output: the findings table, in the PR body and § 7 of the runbook).** Read
`auth/auth.js`, `auth/mailer.js`, `infra/ecs.tf` (auth container env + `ecs_task_ses_send`),
`infra/dns_email.tf`, `nginx/nginx.conf`, and the *installed* `better-auth@1.6.25` under
`auth/node_modules` — the installed source, not the changelog and not memory (the precedent
is the `accountLinking` comment block in `auth.js`, which is built entirely on file:line
reads of the installed library). Cover at minimum: what triggers each send; token lifetimes
and where they come from; single-use vs replayable; whether reset revokes sessions; what the
enforcement flag gates and what it leaves alone; and how the flip interacts with the agent
lane (PR #1653 finding 3 — an agent cannot click an inbox link) and with the free-generation
gate (#1643).

**Fixes — only those the audit proves, each with its own test.** Anything larger gets its own
issue rather than riding along.

**Tests: a new `auth/test/email-flows.test.js`.** `auth/test/auth.test.js` already covers
"does the happy path work and is the mail sent" for both flows; this file covers what happens
to a token that should be **refused**, which is where the security claim lives. Same
conventions as its sibling: real Better Auth endpoints, in-memory sqlite via
`node:sqlite` + `getMigrations`, a capturing mailer as the only double.

**Runbook: `docs/runbooks/email-verification-validation.md`,** plus its row in
`docs/README.md` and `docs/runbooks/README.md` in the same commit (CLAUDE.md: a doc not in the
index does not exist). It must contain the exact manual steps — read-only pre-flight checks,
what to click, expected outcome at each step, what to screenshot — for the verification pass,
a **password-reset rehearsal**, the SES sandbox-vs-production gotchas, and the flip's blast
radius and rollback.

## Acceptance criteria

- [ ] `cd auth && npm ci && npm test` → `fail 0`, and the count is strictly greater than the
      59 cases on `main` at `848673e0`.
- [ ] `auth/test/email-flows.test.js` covers, each driving the real endpoint: verification
      token round trip; **expired** verification token refused; **expired** reset token
      refused; reset token **single-use** (replayed with a *different* password, and that
      password proven not to work — a status-code assertion alone cannot distinguish
      "refused" from "accepted but idempotent"); reset revokes pre-existing sessions, or, if
      it does not, a finding says so.
- [ ] Every guard added has a mutation transcript in the PR body: the change reverted, the
      test run, the failure pasted (CLAUDE.md § "A guard must be shown to reject something").
- [ ] `docs/runbooks/email-verification-validation.md` exists and is reachable from both
      indexes: `python3 .github/scripts/docs_index.py` → `0 orphaned`;
      `python3 .github/scripts/docs_links.py` → `broken 0`.
- [ ] The runbook states, for the enforcement flip: the exact preconditions, the read-only
      command that answers each one, what the flip does **not** change, the rollback, and the
      query that counts who gets locked out.
- [ ] `"$ENV_BIN/pytest" -m "not integration" -q` → `0 failed`; `"$ENV_BIN/ruff format
      --check ." ` → all formatted.
- [ ] Any finding not fixed in this PR is filed as its own issue with a reproduction, not
      left as a sentence in a PR description.

## Verify

Clean clone, no docker, no `.env`, no AWS credentials — everything here must pass cold:

```bash
cd auth && npm ci && npm test          # → fail 0
python3 .github/scripts/docs_links.py  # → broken 0
python3 .github/scripts/docs_index.py  # → 0 orphaned
"$ENV_BIN/pytest" -m "not integration" -q
```

The runbook itself is **not** verified by this issue — it is verified the day the owner runs
it against a real inbox, which is the point of writing it down.

## Anti-goals

- **Do NOT send live email**, and do not add a test that can. Every test here must pass with
  no AWS credentials and no network.
- **Do NOT change `EMAIL_VERIFICATION_ENFORCED`'s deployed value**, in `infra/ecs.tf` or
  anywhere else. This issue prepares the flip; it does not perform it.
- **Do NOT touch `nginx/nginx.conf`, `docker-compose*.yml`, or `infra/*.tf` deploy config**
  without separate alignment (CLAUDE.md § "When to ask before acting"). If the audit finds
  something there, file it.
- **Do NOT weaken any existing guard to make a new test pass**, and do not relax
  `minPasswordLength`, `revokeSessionsOnPasswordReset`, `disableImplicitLinking`,
  `allowDifferentEmails`, or `session.freshAge`.
- **Do NOT change token lifetimes to new values.** Pinning an inherited default at exactly
  its current effective value is in scope (it makes the value auditable without changing
  behaviour); choosing a *different* number is a product decision and is not.
- Do not duplicate coverage that already exists in `auth/test/auth.test.js` — a second
  assertion of the same property is not coverage, it is drag.

## Precedent

- `auth/test/auth.test.js` — the whole file. In particular `testAuthWithMailer` (the mailer
  double), the `freshAge is pinned, not an inherited default` test (the shape for pinning a
  library default), the source-pinned `link-account.mjs` / `callback.mjs` reads, and the
  `reset-request timing does not leak account existence` test (the shape for a timing
  assertion that is robust to jitter).
- `docs/runbooks/runner-ec2-wedge.md` — the most recent runbook, and the house style: a
  symptoms table, read-only diagnosis before any action, and an explicit "not in scope".
- `docs/operations/feature-flag-fliplist.md` row 3 — the flip's recorded preconditions.
- `docs/account-authentication.md` — the existing prose description of both flows; the audit
  should correct it where it has drifted rather than writing a second copy.

## Context

Owner directive, 2026-08-31: identity and API management must be *"strict, sane, secure,
safe"*, auditable by anyone reading the code, exposing no sensitive information. Security is
trust.

Related: PR #1653 (agent-native onboarding — finding 3 names this flag as a latent kill
switch for the programmatic path, decision D3 "is there a non-cookie credential for agents"
is open); #1643 (free-generation gate).

dbrowneup:
Closes #1693

SES production access is live, but neither mail flow had been validated end to end and the
password-reset path had never been run against a real inbox at all. This is the pre-flip
pass for `EMAIL_VERIFICATION_ENFORCED`: code truth, a hermetic harness, and the runbook for
the half a test suite cannot reach.

**No live email was sent. No env flag's deployed value changed. No deploy config touched.**

---

## 1. The flows as built

| Question | Answer, read off the code |
|---|---|
| What sends verification mail | `emailVerification.sendOnSignUp` — **every** signup, whatever the enforcement flag says. Plus `POST /api/auth/send-verification-email`, which is reachable **with no session** |
| What sends reset mail | `POST /api/auth/request-password-reset`, only for an address that actually exists — an unknown one never reaches the callback |
| Transport | AWS SDK v3 `SESv2Client` / `SendEmailCommand` — **not SMTP**. Credentials come from the ECS task role; nothing is stored in the image. `EMAIL_MAILER` defaults to `console`, so a deploy that forgets the var degrades to a visible log line rather than a silent SES failure |
| Identity / from | domain identity `archimedes-arc.com`; `EMAIL_SENDER = no-reply@archimedes-arc.com`; `ecs_task_ses_send` scopes `ses:SendEmail` to that one ARN. SPF `-all`, three DKIM CNAMEs, DMARC `p=none` — all live (`infra/dns_email.tf`) |
| Verification token | a **stateless JWT**. No stored row, so nothing can revoke it early. **1 h** (was a library default parameter; now pinned) |
| Reset token | a real `auth_verifications` **row**, consumed on first use (`consumeVerificationValue`), identifier stored **hashed**. Single-use *and* revocable. **1 h** (was `\|\| 3600 * 1` inside the library; now pinned) |
| Reset revokes sessions | **yes** — `revokeSessionsOnPasswordReset: true`, every session for that user |
| What the flag gates | exactly one line: `sign-in.mjs:312`, the credential sign-in handler. Signup, OAuth sign-in, already-live sessions, and password reset are all untouched |
| `emailVerified` downstream | `models/account.py:40` maps the column; `api/account_auth.py:63` exposes it; `AccountSettings.jsx:505` renders the badge and the resend button. **Nothing in the Python backend gates on it** — the only gate is Better Auth's sign-in refusal |

### Findings

| ID | Finding | Severity | Status |
|---|---|---|---|
| **EV-1** | Behind CloudFront → ALB → nginx, `X-Forwarded-For` is multi-hop and Better Auth trusts a forwarded header only when it carries exactly one value. **No client IP resolves, so every rate-limit bucket is global per-path.** `/sign-up/email` = 3 per 10 min *for the whole internet*; `/sign-in/email` = 10/min globally; `/request-password-reset` and `/send-verification-email` = 3/min globally — the exact endpoints the flip promotes into the critical path | **high** | **open, filed as #1691.** Not fixed here: the good fix touches `nginx/nginx.conf`, which CLAUDE.md says to align on first. Pinned by a matched test pair |
| **EV-2** | Both token lifetimes were library defaults — invisible to a reader of `auth.js`, free to move on a version bump. Same defect class as the `session.freshAge` round-2 blocker | medium | **fixed here**, pinned at their current effective 3600 s. Behaviour identical |
| **EV-3** | `/send-verification-email` was an **account-existence oracle**. It is session-less, and Better Auth defends it with a 500 ms constant-time floor; `sendVerificationEmail` awaited the SES round trip, which walks straight through the floor | medium→high at the flip | **fixed here**, fire-and-forget with the same `.catch` shape `sendResetPassword` already used |
| **EV-4** | A completed reset proves mailbox control but does **not** set `emailVerified`, so under enforcement a successful reset still ends in 403 | medium | **open, owner decision.** Pinned by a test so the runbook's expected outcome is checkable. The fix is an `onPasswordReset` hook; it is a security-semantics call, not a bug fix |
| **EV-5** | No SES configuration set and no event destination → **no bounce/complaint telemetry**, no suppression handling. Production access is *kept* on bounce < 5% / complaint < 0.1%, computed over mail to addresses typed by strangers | medium | **open** — runbook § 9. Note `mailer.js` sends no `ConfigurationSetName`, so adding one in the console alone changes nothing |
| **EV-6** | **An agent cannot click an inbox link** (PR #1653 finding 3). `scripts/agent_journey.py` signs up then immediately signs in (:49-76); after the flip that second call is a 403 and the programmatic path — including the dogfooding harness — stops. There is no API-key or bearer lane (D3, open) | blocks the flip | **open, owner decision.** Three options in the runbook § 7 |

### How the flip interacts with the agent lane and the free-generation gate

`require_current_user` is the first gate on `POST /api/generate/start`, and the only way an
agent obtains that session is `POST /api/auth/sign-in/email`. That is the one call the flag
turns into a 403. So the flip does not *degrade* the agent path, it **ends** it — and it ends
it before #1643's three-free-generations gate is ever consulted, because the account cannot
get a session to spend them with. Whatever D1 decides about free generations for agents is
moot until D3 (a non-cookie credential) is decided, and D3 is now on the critical path for
this flip rather than being a nice-to-have. Nothing here decides that; it is named so the
flip is not made without confronting it.

---

## 2. The harness — `auth/test/email-flows.test.js`, +14 cases (59 → **73**)

`auth/test/auth.test.js` already covers "does the happy path work and is the mail sent" for
both flows. This file covers what happens to a token that should be **refused**, which is
where the security claim lives. Real Better Auth endpoints against in-memory sqlite; the
mailer is the only double, per CLAUDE.md § "Mock at boundaries, not internals".

Round trip · expired verification token · expired reset token · reset token single-use ·
the verification link's one-shot auto-sign-in · reset does not verify (EV-4) · resend-path
timing · resend-path enumeration · what the flag does not gate (source-pinned) · no
auto-resend on a refused sign-in · EV-1 in three tests.

Two deliberate choices worth reviewing:

- **Single-use is proven with a *different* password on the replay.** A status-code
  assertion alone cannot distinguish "refused" from "accepted but idempotent", so the test
  asserts the replayed password does not sign in and the first one does.
- **Expired tokens are minted, not mangled.** The verification case uses the library's own
  `createEmailVerificationToken` with `expiresIn: -60` — a correctly-signed token whose only
  defect is age. The reset case ages the real stored row. A hand-mangled string would be
  refused for being malformed and would prove nothing about expiry.

---

### EV-3 · revert to an awaited send

```
$ MUTATION APPLIED: sendVerificationEmail awaits mailer.send() again (pre-fix shape)
$ node --test test/email-flows.test.js

✖ the anonymous resend path does not leak account existence through response time (2528ms)
  AssertionError: resend timing distinguishes a real unverified account (906ms) from an
  unknown address (507ms) against a 900ms mailer — the send is being awaited again
```

Same probe against the fixed code: unknown **507 ms**, known-unverified **506 ms** — both on
the library's floor.

### EV-2 · remove both TTL pins

```
$ MUTATION APPLIED: both TTL pins removed — the values drift back to the library's inherited defaults
$ node --test test/email-flows.test.js

✖ the verification link's lifetime is pinned in auth.js and is what the mailed token actually carries
  AssertionError: Expected values to be strictly equal
✖ the reset link's lifetime is pinned in auth.js and is what the stored row actually carries
  AssertionError: Expected values to be strictly equal
```

### EV-1 · the matched pair, both directions

The finding test asserts a **429**, so on its own it could be satisfied by a limiter that
always refuses. Its control sits next to it and differs **only in the hop count**:

```
single-hop XFF   A1 200  A2 200  A3 200  A4 429   B1 200   <- B unaffected by A
multi-hop  XFF   A1 200  A2 200  A3 200           B1 429   <- B's FIRST EVER request
```

And the candidate remediation flips it, which is what proves the test tracks the finding
rather than the limiter's existence:

```
$ MUTATION APPLIED: candidate remediation — advanced.ipAddress.trustedProxies covers the nginx/ALB and CloudFront hops
$ node --test test/email-flows.test.js

✖ FINDING: with the multi-hop header production actually sends, every client shares ONE bucket
  AssertionError: client B was NOT limited by client A — the IP now resolves and finding
  EV-1 is fixed; update this test
✖ FINDING EV-1, mechanism: a multi-hop forwarded header resolves to no client IP without trustedProxies
```

*(`trustedProxies` is the weaker of the two candidate fixes — it needs CloudFront's changing
edge ranges maintained. #1691 prefers having nginx set one trusted single-valued header.
This mutation demonstrates the mechanism, not the recommendation.)*

### Existing guards still hold

The full 73-case suite passes, including `auth.test.js`'s three anti-enumeration tests on the
reset path and its own timing test — the fire-and-forget change is on the sibling callback
and does not disturb them.

---

## 4. Runbook — `docs/runbooks/email-verification-validation.md`

For the owner to run once the SES program guidelines land. Human-only, and it says so:
every step is a person with a browser and a mailbox.

Read-only pre-flight (sandbox state, identity + DKIM, **the suppression list** — the most
common "verification is broken" false alarm, DNS, two log greps including the one that
confirms EV-1 live) · the count of accounts the flip would lock out, as SQL · a local
console-mailer rehearsal including the `docker-compose.override.yml` recipe for testing
enforcement (compose does not plumb the flag) · the **sandbox-safe production test** —
verify a plus-addressed alias you own as a destination identity and prove the real SES path
with zero risk to anyone else · a **password-reset rehearsal** across two sessions, checking
all four properties in order, ending with the replay · the flip's exact blast radius,
what it does **not** change, and the rollback · a 12-row evidence table · the SES gotchas.

One gotcha worth surfacing here: **corporate link scanners prefetch URLs.** With
`autoSignInAfterVerification` on, Outlook Safe Links opening a verification link consumes
the one auto-sign-in it carries, and the human's own click then does nothing. That is the
first hypothesis for "the link didn't work", not a token bug.

---

## Validation

```
cd auth && npm ci && npm test        →  tests 73, pass 73, fail 0   (was 59)
$ENV_BIN/pytest -m "not integration" →  5037 passed, 3 skipped, 0 failed
$ENV_BIN/ruff format --check .       →  674 files already formatted
python3 .github/scripts/docs_links.py →  scanned 1536, broken 0
python3 .github/scripts/docs_index.py →  164 indexed, 0 orphaned
```

`ruff check .` reports 47 pre-existing findings, unchanged — no Python file is touched by
this PR.

## Review notes

- **The one behaviour change is EV-3.** `sendVerificationEmail` returns before the mail is
  dispatched. That matches what `sendResetPassword` has always done; failures still log the
  same single line; signup still returns 200 on a sandboxed SES.
- **The TTL pins are behaviour-identical by construction** — 3600 is what both flows already
  produced. If live validation shows 1 h is too short for real inbox latency, changing it is
  now a one-line explicit decision instead of an invisible default.
- **EV-4's test asserts today's behaviour, not the desirable one**, and its failure message
  says so. If you decide a completed reset should verify the address, that PR updates this
  test — which is the point of pinning it.
- Cross-lane: EV-1's preferred fix touches `nginx/nginx.conf` (infra). Deliberately left to
  #1691 rather than folded in here.


dbrowneup:
## CI caught a regression this PR caused — fixed in b0a42f23

`ui/test/verification-resend.test.js` source-reads `auth/auth.js` and pinned
`sendVerificationEmail`'s literal `try { await mailer.send(...) } catch { console.error }`
spelling. Finding EV-3 replaced that with fire-and-forget, so **the pin failed on the fix it
should have been guarding**. Worth flagging rather than quietly fixing: it is the one guard
in the tree that already had an opinion about this code path, and it had pinned the
spelling rather than the property.

Re-pinned on the correct shape *and* strengthened — it now asserts all three properties
instead of two:

| | before | after |
|---|---|---|
| the send failure is caught | ✔ | ✔ |
| the failure is logged, not swallowed | ✔ | ✔ |
| **the send is not awaited** | — | ✔ |

The third is the one that matters: the await *was* the defect.

**Adversarial demo — the re-pinned guard rejects the shape it exists to reject:**

```
$ MUTATION APPLIED: sendVerificationEmail awaits mailer.send() again (pre-fix shape)
$ cd ui && node --test test/verification-resend.test.js

✖ the mailer failure catches stay fail-soft (no throw) and stay loud (console.error)
  AssertionError: The input did not match the regular expression
  /sendVerificationEmail:\s*async[\s\S]{0,120}mailer\.send\(\{[\s\S]{0,450}\}\)\.catch\([\s\S]{0,200}console\.error/
```

Restored → `fail 0`.

**Re-validated after the fix:**

```
cd ui   && npm ci && npm test  →  tests 680, pass 680, fail 0
cd ui   && npm run lint        →  clean
cd auth && npm ci && npm test  →  tests  73, pass  73, fail 0
```

The backend suite (`5037 passed, 0 failed`) and `ruff format --check` are unaffected — this
commit touches one `ui/test` file and nothing else.

dbrowneup:
SES program email received from the owner (2026-08-31, production access confirmation). Facts of record for this PR's runbook + the sending code:
- **Quota: 50,000 messages/day · max send rate 14/sec · out of sandbox · us-east-1**, effective immediately.
- Obligations we must comply with: AWS AUP + SES Service Terms; send only high-quality mail to recipients who expect it; **a bounce/complaint handling process is required** (currently not evidenced in this audit — fold into #1691 or the runbook's follow-ups if not already); **use the SES mailbox simulator for testing** (the runbook's live-validation section should prefer simulator addresses where the scenario allows, real inbox only for the final end-to-end); apply for higher limits before needed; pay-per-use.
The runbook's manual-validation steps should cite these constraints so the first live validation stays inside the program rules.

PR #1702, opened 2026-08-31
[perf][backend] Batch the Library's per-row DB reads: one query for cohort returns, one for publish rights

### LINKED ISSUE #1662: [perf][backend] Batch get_all_daily_returns — stop reading 349 KB per strategy, twice per Library load
### I‑3 — Batch `get_all_daily_returns`: stop reading 349 KB per strategy, twice per Library load

> **Summary.** `backend/archimedes/services/backtest_repository.py:189-202` loops `get_daily_returns` per strategy; each issues an un-projected single-row query and `json.loads` an **~349 KB** `artifact_json` blob (`:300-302`). Library runs it over 34 curated strategies at `strategies_routes.py:365-375`, and `/api/selection-bias/gate` runs the **identical** read over the **identical** ids at `selection_bias_routes.py:337-339`, concurrently, from one page load: **~24 MB off Aurora and 68 blob deserializations per Library mount, on the event loop.** Measured: `/api/strategies/` p50 **12.85 s**, `/api/selection-bias/gate` p50 **15.62 s** with p95 pinned at the 30 s nginx ceiling and 4/20 requests 5xx. The shared rigor cache cannot help — the returns *are* the cache key (`strategies_routes.py:387`), so a cache **hit** still pays the full read. `_passport_responses`' docstring at `strategies_routes.py:1355-1363` already documents this defect in prose.
>
> **Scope.** `services/backtest_repository.py:149-202` only. Rewrite `get_all_daily_returns` as **one** windowed query using the `ROW_NUMBER() OVER (PARTITION BY strategy_id ORDER BY created_at DESC, id DESC)` pattern already at `:270-285`, projecting only `strategy_id, artifact_json, equity_curve_json`. **The per-row `equity_curve` fallback branch at `:178-186` must be preserved exactly** — it is the reason this cannot be a naive `IN` query. Callers unchanged. **Do exactly this, nothing more.**
>
> **Acceptance criteria.**
> - [ ] `<env>/pytest backend/tests/ -k "daily_returns or backtest_repository" -q` → `N passed, 0 failed`.
> - [ ] A new test asserts the SQLAlchemy session issues **exactly 1** query for `get_all_daily_returns(session, [10 ids])`, counted at the session boundary → `1 passed`. **Adversarial: revert, confirm it reports 10 and fails.**
> - [ ] A new test covers the mixed case — some strategies with `artifact_json` returns, some falling through to `equity_curve` derivation, some with neither — and asserts the returned dict is **byte-identical** to the pre-change implementation for the same fixture → `1 passed`.
> - [ ] `<env>/pytest -m "not integration" -q` from the repo root (the exact CI command) → `0 failed`. *(Four callers depend on this boundary: `/api/strategies/`, `/api/selection-bias/gate`, `/api/strategies/passports`, `services/live_rigor_gate.py:276`.)*
> - [ ] Post-deploy: `/api/strategies/` p50 **< 1.5 s** and `/api/selection-bias/gate` p50 **< 1.5 s** over 24 h of ALB access logs.
>
> **Verify.** The four commands above, from the repo root, clean clone, no Docker, no env vars.
>
> **Anti-goals.**
> - Do **not** change any caller. `git diff --name-only` must contain **only** `backend/archimedes/services/backtest_repository.py` and test files.
> - Do **not** defer or drop `equity_curve_json` — `api/risk_routes.py` consumes it off the provider cache (`backtest_repository.py:301-305` explains why the #1543 deferral deliberately spared it). `grep -n "defer(BacktestResultRecord.equity_curve_json" backend/` → no hits.
> - Do **not** add a cache in front of this read. The read produces the cache key; caching it inverts the dependency.
> - Do **not** add a new column or migration in this PR. Persisting `daily_returns` at write time is the *right* long-term fix and belongs in its own issue — `git diff --name-only` must not touch `backend/migrations/versions/`.
>
> **Precedent.** `backtest_repository.py:270-305` — the windowed-query-plus-explicit-projection pattern from #1543, in this same file, with its reasoning written out.

---

---
_Provenance: 2026-08-31 serving-latency evidence sprint (ALB-log per-route latency + CloudWatch + code profile, three-agent measured pass). Numbers in this issue are from that window; re-measure before quoting elsewhere. Sequencing and the full decision package live with the owner._

### LINKED ISSUE #1663: [perf][backend] wallet_can_publish N+1: one IN query, not one per row
### I‑4 — `wallet_can_publish` N+1: one `IN` query, not one per row ①

> **Summary.** `strategies_routes.py:624-630` and `:695` call `wallet_can_publish` per response row; each is a single `.first()` (`models/strategy_generators.py:62`). At 34 curated rows that is 34 extra sequential round trips, each paying a `pool_pre_ping` `SELECT 1` (`db.py:114`). It short-circuits on anonymous callers — so **visitors do not pay it and the signed-in owner does**, which is backwards for a demo.
>
> **Scope.** `api/strategies_routes.py:624-630` and `:695` only. Build one `set[strategy_id]` with a single `IN` query before the loop. **Do exactly this, nothing more.**
>
> **Acceptance criteria.**
> - [ ] `<env>/pytest backend/tests/ -k "strategies_routes or can_publish" -q` → `0 failed`.
> - [ ] A test asserts one query for a 20-row window → `1 passed`. **Adversarial: revert, confirm it reports 20.**
> - [ ] Anonymous and wallet-linked responses are byte-identical to pre-change for the same fixture → `1 passed`.
>
> **Verify.** As above, repo root, no Docker.
>
> **Anti-goals.** Do **not** change who is allowed to publish. `git diff backend/archimedes/models/strategy_generators.py` must be empty — the env-var parsing there is shared with `PLATFORM_ADMIN_WALLETS` (see I‑8) and the two must not drift. Do **not** relax the anonymous short-circuit.

---

---
_Provenance: 2026-08-31 serving-latency evidence sprint (ALB-log per-route latency + CloudWatch + code profile, three-agent measured pass). Numbers in this issue are from that window; re-measure before quoting elsewhere. Sequencing and the full decision package live with the owner._

dbrowneup:
## What

The Library serving path stops issuing per-row queries. Two reads on the same route surface, one logical change:

- `get_all_daily_returns` is now **one windowed query** for the whole cohort instead of a Python loop over `get_daily_returns` (N un-projected single-row reads, each deserializing a ~349 KB `artifact_json`).
- `wallet_can_publish` is no longer called once per response row on either Library listing; a new `_publishable_strategy_ids` helper answers the whole window with one `IN` query.

## Why

From the 2026-08-31 serving-latency evidence sprint. One Library mount ran the identical persisted-returns read twice — `/api/strategies/` and `/api/selection-bias/gate`, concurrently, over the identical ids — for a measured ~24 MB off Aurora and 68 blob deserializations on the event loop. `/api/strategies/` p50 **12.85 s**, `/api/selection-bias/gate` p50 **15.62 s**, p95 pinned at the 30 s nginx ceiling, 4/20 requests 5xx.

The shared rigor cache cannot help, and this is the part that makes the fix have to live in the read itself: **the returns ARE the cohort cache key** (`rigor_cache.cohort_key`), so a cache *hit* still pays the full read. Caching in front of this function would invert the dependency. `_passport_responses`' own docstring already documented the defect in prose; that prose is updated here rather than left standing as a claim the code no longer matches.

On top of that, `wallet_can_publish` per row added 34 extra sequential round trips on a 34-row curated page, each paying a `pool_pre_ping` `SELECT 1` — and because the per-row call short-circuits on anonymous callers, **visitors paid nothing and the signed-in owner paid all 34**, which is backwards for a demo.

Query counts, from the measured per-call numbers below (arithmetic over a 34-strategy curated page, not a separate live measurement):

| Route | before | after |
| --- | --- | --- |
| `GET /api/strategies/` | 34 returns reads + 34 publish reads = **68** | 1 + 2 = **3** |
| `GET /api/selection-bias/gate` | **34** | **1** |
| `GET /api/strategies/generated` (20-row page) | **20** publish reads | **1** |

## Issues

Closes #1662
Closes #1663

## Verification

Env: `/opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/pytest` (Python 3.12.13), repo root, isolated worktree, no Docker, no env vars. Rebased onto `origin/main` @ `848673e0` and re-run there — `main` moved (#1673) while this was in flight, and a frozen test-merge ref describes a base that no longer exists (CLAUDE.md § Before you approve a merge, rule 1). The two commits `main` gained touch only `backend/tests/test_cloudwatch_alarms.py`, which this PR does not.

### CI command

```
$ pytest -m "not integration" -q
5074 passed, 3 skipped, 2 deselected, 296 warnings in 434.65s (0:07:14)
```

**0 failed.** (Two earlier full runs showed unrelated failures, both chased down rather than waved off: `test_cloudwatch_alarms::TestAntiGoals::test_the_oracle_stale_alarm_is_untouched` reproduced on a bare `origin/main` checkout and was fixed on `main` by #1673 — it is gone after the rebase; and a pair of wall-clock budget assertions in `test_health_always_answers.py` failed only while two full suites ran concurrently on one machine, including in a bare-`origin/main` control run with zero changes applied.)

### Scoped

```
$ pytest backend/tests/ -k "daily_returns or backtest_repository" -q       # issue #1662's command
40 passed, 5039 deselected, 2 warnings in 19.60s

$ pytest backend/tests/ -k "strategies_routes or can_publish or publishable" -q   # issue #1663's command
65 passed, 5014 deselected, 2 warnings in 27.32s
```

### ruff on touched Python

```
$ ruff check backend/archimedes/services/backtest_repository.py \
             backend/tests/test_backtest_repository.py \
             backend/tests/test_publishable_strategy_ids.py \
             backend/tests/test_strategies_routes.py
All checks passed!

$ ruff format --check <the five touched files>
5 files already formatted
```

`strategies_routes.py` still reports one pre-existing `SIM105` at `:1211` (`_rollback_quietly`), untouched by this PR — confirmed identical on `origin/main`:

```
$ git stash && ruff check backend/archimedes/api/strategies_routes.py
help: Replace `try`-`except`-`pass` with `with contextlib.suppress(Exception): ...`
Found 1 error.
```

### Anti-goal checks

```
$ grep -rn "defer(BacktestResultRecord.equity_curve_json" backend/
   no hits

$ git diff --stat backend/archimedes/models/strategy_generators.py
   empty

$ git status --porcelain | grep "migrations/versions"
   none
```

No cache was added in front of `get_all_daily_returns` — the read produces the key. No new column, no migration; persisting `daily_returns` at write time is the right long-term fix and belongs in its own issue.

---

## Adversarial demos

Every behavioural claim above, reverted, run, and shown failing. Source files were restored byte-for-byte after each (`diff -q` against the pre-revert copy → `SOURCES RESTORED CLEAN`).

### A — revert the batching to the pre-#1662 per-strategy loop

```
REVERTED get_all_daily_returns to the pre-#1662 per-strategy loop
$ pytest backend/tests/test_backtest_repository.py -q \
    -k "issues_exactly_one_query or projects_only_the_columns or hydrates_no_orm_rows"

E   AssertionError: get_all_daily_returns issued 10 queries for 10 strategies; the reader is
    still looping per strategy. statements=['SELECT backtest_results.id AS ...
E   AssertionError: assert 4 == 1
E   AssertionError: batched read hydrated 6 BacktestResultRecord objects it never uses

FAILED ... ::test_get_all_daily_returns_issues_exactly_one_query
FAILED ... ::test_get_all_daily_returns_projects_only_the_columns_it_decodes
FAILED ... ::test_get_all_daily_returns_hydrates_no_orm_rows
3 failed, 23 deselected in 5.39s
```

The 10 captured statements are the full un-projected `SELECT backtest_results.id, ..., backtest_results.artifact_json ...` — i.e. the counter sees the real N+1, so "exactly 1" is not a tautology. `test_per_strategy_loop_reports_ten_queries_for_ten_ids` keeps that control permanently in the suite by running the verbatim pre-change code through the same counter.

### B — delete the equity-curve fallback branch (the naive-`IN`-query mistake #1662 warns about)

```
REVERTED: equity-curve fallback branch DELETED
$ pytest backend/tests/test_backtest_repository.py -q -k "matches_the_pre_change or cohort_cache_key"

E   AssertionError: assert {'s_artifact'...0.022, 0.033]} == {'s_artifact'...4851485], ...}
E   AssertionError: assert '5f7fb8bd0373...4cb1ebb0a037e' == 'c49e064c2114...88b81effa470f'

FAILED ... ::test_get_all_daily_returns_matches_the_pre_change_implementation_exactly
FAILED ... ::test_get_all_daily_returns_does_not_change_the_cohort_cache_key
2 failed, 24 deselected in 4.12s
```

The second failure is the cache-key subtlety made concrete: dropping one decode branch moves `cohort_key`, which would silently invalidate every shared rigor-cache entry and put the ~6 s cohort pass back on every request. `test_cohort_key_control_detects_a_perturbed_series` is the paired control — it perturbs one series by `1e-9` and confirms the key moves, so "the key is unchanged" is a real measurement and not a statement about a key that ignores its input.

### C — flip the window `ORDER BY` to ASC (oldest row wins)

```
REVERTED: window ORDER BY flipped to ASC — the OLDEST backtest row now wins
$ pytest backend/tests/test_backtest_repository.py -q -k "reads_the_latest_row"

FAILED ... ::test_get_all_daily_returns_reads_the_latest_row_per_strategy
    - AssertionError: assert {'s1': [0.0]} == {'s1': [3.0]}
1 failed, 25 deselected in 7.80s
```

### D — revert `_publishable_strategy_ids` to the per-row `wallet_can_publish` loop

```
REVERTED _publishable_strategy_ids to the pre-#1663 per-row wallet_can_publish loop
$ pytest backend/tests/test_publishable_strategy_ids.py -q

FAILED ::test_generated_window_issues_exactly_one_query
    - AssertionError: issued 20 queries for 20 rows; still looping per row.
FAILED ::test_curated_window_query_count_is_constant_not_linear
    - AssertionError: 5 rows → 5 queries; expected a constant 2.
FAILED ::test_curated_window_for_an_admin_wallet_issues_one_query - assert 0 == 1
FAILED ::test_repeated_ids_do_not_change_the_answer_or_the_query_count - assert 16 == 1
4 failed, 20 passed in 2.18s
```

Note that the 20 *parity* tests still pass under the revert — correct, and the point: the answers are unchanged, only the query count moved.

### E — drop the `PLATFORM_ADMIN_WALLETS` override (the naive IN-query-only version)

```
REVERTED: PLATFORM_ADMIN_WALLETS override DROPPED
$ pytest backend/tests/test_publishable_strategy_ids.py -q

FAILED ::test_curated_window_query_count_is_constant_not_linear - 5 rows → 1 queries; expected 2
FAILED ::test_curated_window_for_an_admin_wallet_issues_one_query - assert set() == {'strat_000',...}
FAILED ::test_matches_the_per_row_gate_for_an_admin_wallet[True] - assert set() == {'strat_000',...}
FAILED ::test_admin_may_publish_a_curated_row_it_did_not_generate - assert set() == {'strat_000',...}
4 failed, 20 passed in 2.39s
```

### F — relax the anonymous short-circuit (grant visitors publish rights)

```
REVERTED: anonymous short-circuit RELAXED (visitors granted publish rights)
$ pytest backend/tests/test_publishable_strategy_ids.py backend/tests/test_strategies_routes.py -q \
    -k "anonymous or per_row_gate or matches_the_per_row"

FAILED test_publishable_strategy_ids.py::test_anonymous_caller_issues_no_query_at_all
    - AssertionError: assert {'strat_000',...} == set()
FAILED test_publishable_strategy_ids.py::test_matches_the_per_row_gate_exactly[None-0-True]
FAILED test_publishable_strategy_ids.py::test_matches_the_per_row_gate_exactly[None-0-False]
FAILED test_strategies_routes.py::test_library_response_identical_to_per_row_gate_for_an_anonymous_visitor
    - AssertionError: assert {'degraded': ..., 'total': 34} == {'degraded': ..., ...}
4 failed, 12 passed, 48 deselected in 5.12s
RESTORED
```

### G — neuter the hermeticity fixture and hand the suite a hostile ambient env

`test_publishable_strategy_ids.py` pins `PLATFORM_ADMIN_WALLETS` empty unless a test sets it. Without that, a developer's `.env` naming the test's own stranger wallet would flip the curated-page query count from 2 to 1 and quietly turn the non-admin assertions into a different test than CI runs.

```
hostile ambient PLATFORM_ADMIN_WALLETS=0xc3c3…c3   (= the test's _STRANGER)

--- fixture NEUTERED ---
$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend PLATFORM_ADMIN_WALLETS=$HOSTILE \
      python -m pytest backend/tests/test_publishable_strategy_ids.py -q
FAILED ::test_curated_window_query_count_is_constant_not_linear
    - AssertionError: 5 rows → 1 queries; expected a constant 2.
1 failed, 23 passed in 1.72s

--- fixture RESTORED, same hostile env ---
24 passed in 1.23s
```

---

- **#1663's acceptance asks for "one query for a 20-row window"; the curated route costs a constant *two*.** The generated route (`is_example=False`) is exactly one. The curated route pays one extra indexed lookup to resolve the `PLATFORM_ADMIN_WALLETS` override *through `wallet_can_publish` itself*, rather than re-parsing that env var in the route. Re-parsing would create a third copy of the parsing (`models/strategy_generators.py` and `api/metrics_private_routes.py` already hold two) and a copy that drifts silently changes who may publish — which the issue's anti-goal pins the model file shut to prevent. The probe is skipped when it cannot change an answer, and an actual admin pays zero extra queries (the override returns before the row lookup). If you would rather have the literal one query, the call is "add a shared `is_platform_admin()` helper", and that belongs with I-8's consolidation, not here.
- **This PR touches `strategies_routes.py`, which #1662's anti-goal scopes out.** That is #1663's scope, merged into one PR by instruction because it is the same route surface and the same logical change. #1662's own file, `services/backtest_repository.py`, changed exactly as specified; `git diff --name-only` shows no migration and no other caller.
- **One narrow behaviour change inside `get_daily_returns`.** The equity-curve fallback now reads `json.loads(equity_curve_json or "[]")` directly instead of going through `row.to_backtest_result()`. Identical output — `BacktestResult` is a frozen dataclass with no coercion, and it applies exactly that expression to the same column. The one difference: a row with a *malformed `monthly_returns_json`* used to raise out of the fallback branch and now does not. That column is irrelevant to daily returns, so this is a strictly narrower failure surface, but it is a difference and it is deliberate — say so if you want it preserved.
- **`_daily_returns_from_columns` is deliberately not hardened.** The artifact parse still catches only `JSONDecodeError`/`KeyError`; the equity-curve parse still catches nothing. Widening either would change which malformed rows raise into callers' `except` blocks — a behaviour change wearing a perf fix's clothes.
- **The bytes are only half-fixed.** The round trips are gone; each winning row's `artifact_json` is still transferred and `json.loads`-ed. Making this genuinely cheap needs `daily_returns` persisted as its own column at write time, which is out of scope here by anti-goal and wants its own issue.
- **The post-deploy acceptance is not met by this PR.** `/api/strategies/` p50 < 1.5 s and `/api/selection-bias/gate` p50 < 1.5 s over 24 h of ALB logs can only be confirmed after deploy; re-measure before quoting.
- **One existing test changed.** `test_single_strategy_endpoint_numeric_fields_equal_live_gate` stubbed only `get_daily_returns` and relied on `get_all_daily_returns` delegating to it internally. It now stubs both — the DB boundary its sibling tests already use. It fails without that edit (`assert None == 1.0 ± 1.0e-09`), which is the test being coupled to the old implementation, not a served-value regression: the two route-level parity tests added here compare full response JSON against the per-row gate and are identical.


dbrowneup:
Review threads dispositioned before merge (per the resolve-before-merge rule): the **empty-except** thread is fixed in the pushed commit — the deliberate fall-through to the equity-curve fallback now carries its explanatory comment. The three **mixed import** threads are resolved as working-as-intended: `import archimedes.api.strategies_routes` + `from … import <name>` in the same test file is this repo's established patch-at-module idiom (the module reference is what `patch.object` targets; the names are what the assertions call) — collapsing to one form would break the patching pattern the sibling test files use.

Issue #1080, opened 2026-07-09
Vault holdings API returns corrupted weight_pct for unpriceable synths


onder-akkaya:
## Summary

GET /api/vaults/{address} returns corrupted `weight_pct` / `value_usdc` for synth holdings whenever the oracle price for a token can't be read. Example on the live deploy, vault `0xA3b077e16C208cD794581db46b559FDC9619ada7`: USDC 100.0%, sSPY 12.0% (plausible), but sNKY **96,543,640.8%** and sGOLD **3,070,572,228.2%**. Weights also don't sum to 100 — USDC alone reads 100%.

Root cause (verified numerically against the live payload): two defects in `ChainExecutor.read_portfolio` / `_token_to_usdc` (`backend/archimedes/chain/executor.py`):

1. When the oracle price call fails for a synth (or the token isn't in `synth_addresses`), `_token_to_usdc` falls through to `return amount` — the **raw 18-decimal base-unit amount** — which the caller then divides by 1e6 as if it were a USDC value. Net inflation: 10^12. For sNKY and sGOLD, `value_usdc * 1e6` equals the raw on-chain token amount byte-for-byte. sSPY is sane because its oracle answered ($747.40).
2. Holding weights are computed against on-chain `totalAssets()`, which for this vault counts only USDC (19.389) — so USDC reads 100% and the weights can never sum to 100, even with correct per-token values.

The same raw-amount fallthrough also poisons the off-chain NAV fallback in `_safe_total_assets` (it sums `_token_to_usdc` results).

## Repro

```
curl -s https://archimedes-arc.com/api/vaults/0xA3b077e16C208cD794581db46b559FDC9619ada7 | jq '.holdings'
```

## Acceptance criteria

- [ ] Every holding `weight_pct` in [0, 100]; weights sum to ~100.
- [ ] `value_usdc` correct against on-chain balances with per-token decimals; a token whose price genuinely can't be read is valued at 0 with a loud warning log — never at its raw base-unit amount.
- [ ] `pytest backend/tests/chain/test_executor_reads.py` → 0 failed, with regression tests for the unpriceable-synth case and the weights-sum case.

## Notes

- Related prior work: #942 / PR #1026 fixed the same hardcoded-decimals class of bug on the contract side.
- `ui/src/components/QuantLab.jsx` has a client-side sanity gate (skips vaults with weight_pct > 150) that can be relaxed once this lands.

PR #1704, opened 2026-09-01
Paper trading — drive paper deployments with the agent tick loop (executor interface, two venues)

### LINKED ISSUE #1410: APIN - Paper Trading - Drive paper deployments with the agent tick loop (vault-mechanics validation without chain risk)
## Why

The oracle+agent runner currently serves only the vault execution path — and zero vaults exist, so the most-MVP piece of the vault feature (the agent's signal→target-weights→rebalance loop) runs against nothing and gets no validation. Meanwhile paper deployments are advanced by a backend scheduler loop (`services/paper_trading.py`) that replays strategy signals but exercises none of the agent's decision loop.

Adapting the agent tick loop to drive paper deployments validates the core vault mechanic end-to-end (signal FSM → target weights → executed rebalance → bookkeeping) with zero chain risk, and upgrades paper trading from replay to a live-loop simulation.

## What

- Extract the agent runner's per-tick decision core (read signals → compute target weights → diff against current positions → produce trades) behind an executor interface with two implementations: the existing chain executor (vaults) and a new paper executor that records trades into the existing paper-trading position/valuation tables.
- Run the paper executor from the existing backend scheduler (same cadence as the current paper advance loop) — **no new runner process, no new infra**. The EC2 runner box (#1402) is not a dependency.
- Signal semantics must be the unified F2/F3 replay-derived position FSM (#1320, parity-pinned) — the paper path and the vault path must share the same signal interpretation code, not a copy.
- Provenance: each agent-driven paper trade records which signal/tick produced it.

## Acceptance

- `pytest -m "not integration"` → 0 failed.
- A parity test proves the paper executor and chain executor produce identical target weights from identical signal input (same fixture through both, assert equal weights; chain side mocked at the chain-client boundary).
- A paper deployment advanced N ticks shows agent-attributed trades with tick provenance.
- Grep guard: no second copy of the signal-FSM logic (`grep -rn` for the FSM entry symbol shows one definition).

## Anti-goals

- Do NOT touch the chain executor's live behavior or the vault contracts.
- Do NOT add a new daemon/process/container; backend scheduler only.
- Do NOT change paper-trading valuation math — only who drives the trades.

Precedent to copy: the scheduler wiring of the existing paper advance loop in `services/paper_trading.py`; the executor boundary in `backend/archimedes/chain/executor.py`.

dbrowneup:
Closes #1410

## What

Extracts the agent runner's per-tick decision core behind an executor interface with two venues — the existing chain executor (vaults) and a new paper venue — and drives paper deployments with it from the existing paper advance loop.

The vault feature's most-MVP piece — the agent's *signal FSM → aggregated target weights → diff against current positions → executed trades → bookkeeping* loop — had **zero vaults to run against and therefore zero validation**. Meanwhile paper deployments were advanced by a replay that exercises none of it. Pointing the same core at a paper deployment runs every link of the chain, on the same code, with no chain and no money.

| File | Role |
| --- | --- |
| `backend/archimedes/execution/core.py` | `DRIFT_THRESHOLD`, `weights_to_targets`, `compute_trades`, `targets_from_signals`, `TickDecision`. Lifted **verbatim** out of `StrategyRunner`; `_weights_to_targets` / `_compute_trades` are now one-line delegates. No I/O. |
| `backend/archimedes/execution/venue.py` | `ExecutionVenue` protocol + `ChainVenue`, a forwarding adapter over `chain_executor`. |
| `backend/archimedes/execution/paper_venue.py` | `PaperVenue`. Positions are the fold of an append-only agent-trade ledger. |
| `backend/archimedes/services/paper_agent_execution.py` | `advance_agent_execution` — one agent tick per active deployment. |
| `backend/archimedes/models/paper_store.py` + migration `c5e81a4f7b32` | `paper_agent_trades`. |
| `backend/archimedes/services/strategy_signal_evaluator.py` | `TICKER_TO_SYNTH` / `synths_for_universe` lifted out of `evaluate_strategies`'s body so the paper path shares the table rather than copying it. |
| `backend/tests/test_execution_venue_parity.py` | The issue's acceptance file, G1–G8 (21 cases). |
| `backend/tests/test_alembic_migrations.py` | Round-trip + `create_all()` parity for the new table. |

## Acceptance, against the issue's list

| # | Criterion | Where |
| --- | --- | --- |
| 1 | `pytest -m "not integration"` → 0 failed | see below |
| 2 | Parity test: paper and chain venues produce identical target weights from identical signal input, chain side mocked at the chain-client boundary | `TestTargetWeightParity` — `ChainVenue(executor=AsyncMock(), usdc_address=…, synth_addresses=…)`; no web3, no `.env`. Plus `test_the_live_runner_method_agrees_with_the_paper_venue`, which invokes the **real** `StrategyRunner._weights_to_targets` / `._compute_trades` so the claim pins the live vault method and not only the shared function |
| 3 | A paper deployment advanced N ticks shows agent-attributed trades with tick provenance | `TestPaperTickProvenance::test_n_ticks_produce_attributed_trades` — 3 ticks (enter / hold / rotate), every row asserted for `tick_id`, `decided_at`, `spec_hash`, `signal_strategy_id`, `signal_state`, `signal_reason`; the aligned tick trades nothing |
| 4 | Grep guard: one definition of the FSM entry symbol | `TestSingleFsmDefinition` — `grep -rn "def _replay_position_state"` over `backend/` returns exactly 1 |

## Anti-goals — held literally

1. **Do NOT touch the chain executor's live behavior or the vault contracts.** `ChainVenue` forwards to `chain_executor` and reimplements nothing; `StrategyRunner` still calls `chain_executor` directly and is *not* rerouted through the protocol by this PR. `_weights_to_targets` / `_compute_trades` bodies moved unchanged (same #1080 unpriced guard, same `0.15` threshold, same rounding), and the venue is constructed from the *same snapshotted* address maps the runner has always used. `test_chain_venue_forwards_without_reinterpreting` asserts the trade list reaches the executor unaltered. No `contracts/` file is touched.
2. **Do NOT add a new daemon/process/container.** `advance_agent_execution` is called from the existing `paper_advance_loop` in the web tier. Nothing in `infra/`, `docker-compose*.yml` or `deploy-runners.yml` changes. The EC2 runner box (#1402) is not a dependency.
3. **Do NOT change paper-trading valuation math.** `paper_daily_returns` and `paper_marks` are not read or written by any new code — grep the diff: the only new table is `paper_agent_trades`. `advance_all` is untouched (the new pass sits beside it in the loop, not inside it), so the replay ledger's semantics are byte-identical.

## Two decisions the issue's wording did not survive contact with, stated plainly

**"records trades into the existing paper-trading position/valuation tables"** — there is no trade table, and neither existing table can host this data without breaking its own law. `paper_daily_returns` is the append-only **track record** produced by the graded engine; writing agent-decided rows into it would mean a user's track record was produced by something other than the engine that graded the verdict they paid for. `paper_marks` is documented as "a decoration with a TTL", safe to delete wholesale, with a retention policy that ends in `DELETE` — trade provenance must not inherit that. So: one new table, and anti-goal 3 holds literally.

**Weights, never dollars.** `PaperDeployment` has no notional/capital column — there is no deployed capital amount anywhere in this system (`PaperMark`'s docstring already refuses to render `$10,347` for exactly this reason). The paper book is therefore an **index**: `total_value_usdc == 1.0`, and a trade is stored as the portfolio fraction that moved (`prior_weight` → `target_weight`, signed `weight_delta`). Nothing in the paper venue can render a dollar figure because nothing in it knows one.

**Known limit, disclosed rather than hidden.** The folded position is a *signal-state* book, not a marked-to-market one: between ticks the recorded weights do not move with prices, so the drift this venue can see is drift the **signals** created, never drift a **price move** created. The paper venue therefore exercises signal→weights→diff→trade faithfully and does **not** exercise price-drift-triggered rebalancing. Closing that would mean this table growing a second opinion about what a paper portfolio is worth, alongside `paper_marks` — a deliberate non-goal, documented in `PaperAgentTrade`'s docstring and in `paper_venue`'s.

**No regime snapshot on the paper path.** The vault runner classifies an exogenous market regime each tick from a VIX/S&P snapshot; doing that in the web tier would add live market I/O to a loop that has no need of it. The pass passes `regime=None`, which takes `PortfolioConstructor`'s own documented conservative default — the same branch a vault tick takes whenever its regime classification fails. Consequence: the paper book is sized as a vault with an unavailable regime would be sized, not as one with a healthy detector.

### 1. The FSM single-definition guard (acceptance #4) — live, on the real tree

Planted a second `def _replay_position_state` in `backend/archimedes/services/`, ran the guard, deleted the probe:

```
tests/test_execution_venue_parity.py F                                   [100%]
E   AssertionError: expected one FSM definition, found 2:
E     …/backend/archimedes/services/_adversarial_fsm_copy.py:4:def _replay_position_state(spec, prices, max_period):
E     …/backend/archimedes/services/strategy_signal_evaluator.py:1211:def _replay_position_state(
E   assert 2 == 1
FAILED tests/test_execution_venue_parity.py::TestSingleFsmDefinition::test_exactly_one_definition_exists
```

After `rm`: `18 passed`. The same probe is reproduced hermetically in `test_a_planted_second_copy_is_caught` (two files in `tmp_path`, same grep, expects 2) so the demonstration is permanent, not just a one-off.

### 2. The venue-parity guard (acceptance #2) — live, on the real tree

The realistic way this regresses is not a whole forked module — it is a "harmless" venue-specific tweak inside the shared core. Patched `core.weights_to_targets` with `weight * 0.9 if venue.name == "paper"`, ran the parity tests, reverted:

```
E   AssertionError: assert {'USDC': 0.44...SPY': 0.37331} == {'USDC': 0.44...7899999999997}
E     Differing items:
E     {'sSPY': 0.37331} != {'sSPY': 0.33597899999999997}
E     {'sQQQ': 0.18669} != {'sQQQ': 0.168021}
…/test_execution_venue_parity.py:176: AssertionError

E   AssertionError: assert [('USDC', <Tr...'buy'>, 0.28)] == [('USDC', <Tr...buy'>, 0.252)]
E     At index 1 diff: ('sQQQ', <TradeDirection.BUY: 'buy'>, 0.28) != ('sQQQ', <TradeDirection.BUY: 'buy'>, 0.252)
…/test_execution_venue_parity.py:252: AssertionError
```

After restoring `core.py` (`grep -c "ADVERSARIAL PROBE" → 0`): `18 passed`. Two in-suite variants keep the demonstration permanent — `test_a_forked_copy_of_the_core_breaks_parity` (a drifted USDC floor, the divergence shape `marketplace/service.py` already exhibits) and `test_a_second_drift_threshold_breaks_parity` (a book within the shared 15% of target trades nothing; a forked 0.01 threshold trades).

### 3. The tick-provenance guard (acceptance #3)

`tick_id` is `NOT NULL`, so provenance is not a convention that can be skipped. `test_a_trade_with_no_tick_is_refused` writes a row with `tick_id=None` straight past the venue and asserts `IntegrityError` on flush. The migration test additionally asserts the string `tick_id VARCHAR(32) NOT NULL` is in the migrated DDL, so the guard is checked on the *migrated* schema and not only on `create_all()`.

### 4. The idempotent-re-apply guard

`test_same_tick_twice_writes_once` — second `execute_trades` for the same `TickDecision` returns `[]` and the row count is unchanged. `test_a_duplicate_row_is_rejected_by_the_constraint` bypasses the venue's cheap skip and writes a duplicate `(deployment, tick, symbol)` directly, asserting `IntegrityError` — the constraint, not the skip, is what makes the guarantee real.

### 5. The kill switch

`test_the_kill_switch_stops_the_pass` — with `PAPER_AGENT_EXECUTION=0` the pass returns `{"enabled": False, …}` and writes zero rows. The flag defaults **ON**: the pass is additive, self-idempotent and fail-soft, so a default-off flag would be an operator ritual that exists only to be forgotten. It is a kill switch, not an enable switch.

### 6. The broken-spec gate

`_spec_signal` **never raises** on a broken spec — it logs and returns FLAT. So an unvalidated bad spec does not arrive as an error; it arrives as a full set of legitimate-looking flat signals, aggregates to cash, and has the agent **sell the whole book**. `_tick_deployment` therefore calls `validate_strategy_spec` *before* evaluating.

`test_without_the_gate_a_broken_spec_would_liquidate_the_book` builds exactly that input — a spec whose `entry` references `not_an_indicator`, run through the **real** evaluator with injected prices (hermetic) — and asserts (a) the evaluator produces signals rather than complaining, (b) every one is `FLAT`/weight 0, and (c) the resulting targets produce a liquidating `SELL` against a fully-invested book. `test_an_invalid_spec_is_failed_and_trades_nothing` then shows the gate holding: `failed == 1`, zero rows for that deployment, and the healthy deployment in the same pass still ticks.

### 7. Failure isolation

`test_one_broken_deployment_does_not_stall_the_others` — a deployment with `spec_json = "{not json"` alongside a healthy one: `failed == 1`, `ticked == 1`, trades written. Same isolation rule `advance_all` follows.

## A migration bug this caught, worth recording

The first draft declared the unique constraint as a follow-up `op.create_unique_constraint`. SQLite has no ALTER for constraints, so `alembic upgrade head` raised `NotImplementedError` and **every** test in `test_alembic_migrations.py` plus `test_account_deletion_cascade.py` failed — none of which touch paper trading. Fixed by declaring the constraint inside `create_table` (the same placement `uq_paper_decision_traces_dep_date` uses), and `test_alembic_paper_agent_trades_table_added_and_removed` now asserts the constraint exists on the migrated table so the regression cannot come back quietly.

## Residuals, named rather than glossed

- **`StrategyRunner.tick()` still composes the shared steps inline** (aggregate → construct → `_weights_to_targets`) rather than calling `targets_from_signals`. Collapsing it would be behaviour-preserving but would force a rewrite of `test_agent_runner_tick`'s module-level `patch("archimedes.chain.agent_runner.strategy_evaluator")` — editing tests to fit a refactor, on the live-money path, for no acceptance-criterion gain. `test_the_live_runner_method_agrees_with_the_paper_venue` covers the gap that leaves by pinning the runner's actual methods against the paper venue.
- **`marketplace/service.py` is untouched.** See below.

## Not in this PR, deliberately

`backend/archimedes/marketplace/service.py:95` carries a hand-ported copy of `_compute_trades` (its own comment says so) that has **already drifted**: it lacks the #1080 unpriced-holding guard and it keeps its own `_DRIFT_THRESHOLD = 0.15`. Unifying it onto `execution.core` would change the marketplace's live rebalance semantics, which is neither in #1410's acceptance list nor covered by its anti-goals. Flagging it here rather than doing it silently — it is the strongest argument that this extraction was worth doing, and it wants its own issue.

## Validation

From the **repo root** (the CI command — `backend/` as cwd silently changes what is collected and fails two path-relative tests that have nothing to do with this branch):

```
$ /opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/python -m pytest -m "not integration" -q -p no:cacheprovider
5059 passed, 3 skipped, 2 deselected, 296 warnings in 440.36s (0:07:20)
```

```
$ ruff check   <12 touched files>   → All checks passed!
$ ruff format --check <same>        → 12 files already formatted
$ alembic upgrade head → downgrade → upgrade, tmp SQLite → clean
```

22 of those cases are new: 21 in `test_execution_venue_parity.py`, 1 in `test_alembic_migrations.py`. No `.tf` file is touched, so `terraform fmt` is not applicable.


danielscoffee (commented):
Research review of the #1410 claim, not a code-style pass.

This does what the issue asked: one decision core, two venues, paper driven from the existing advance loop, no new process, valuation tables untouched. The disclosed refusals (new `paper_agent_trades` rather than writing into `paper_daily_returns` / `paper_marks`; weights, not dollars) are the right integrity calls.

What it does **not** validate, even though the issue framed this as vault-mechanics validation without chain risk:

1. **Price-drift rebalancing.** The paper book is a signal-state index. Holdings do not mark to market between ticks, so `DRIFT_THRESHOLD` can only fire on a signal change, never on a price move. Live vaults rebalance on both. This path exercises signal → weights → diff → trade faithfully and leaves the price-drift half of the vault mechanic untested.

2. **Regime-conditioned sizing.** Paper always passes `regime=None` (constructor conservative default). A vault with a working VIX detector will size differently. Paper is a regime-unavailable vault simulation, not a healthy-detector one.

3. **Live composition.** `StrategyRunner.tick()` still composes aggregate → construct → `_weights_to_targets` inline. `test_the_live_runner_method_agrees_with_the_paper_venue` pins the extracted functions; it does not pin the live orchestration.

Two hazards to keep named after merge:

- `paper_daily_returns` (the graded track record) and `paper_agent_trades` (this ledger) are now two books. Correct for not contaminating the gate. Easy to overclaim that paper trading validates the vault if they stay conflated in the product.
- `marketplace/service.py` already carries a drifted copy of `_compute_trades` (missing the #1080 unpriced-holding guard, own `0.15` threshold). Flagged here and left alone. That is the strongest argument this extraction was worth doing, and it wants its own issue before anyone treats marketplace rebalances as running this core.

Not blocking. The PR already names most of these. I want them to stay named after merge, not get rounded up into paper path = vault path.

danielscoffee:
**Citable split (MVP). Two books, neither is vault validation.**

| Book | What it is |
| --- | --- |
| `paper_daily_returns` | Graded track record. This is what the rigor gate sees. Untouched by this PR. |
| `paper_agent_trades` | New executor ledger. Signal-state index, not marked-to-market. Drift does not fire on price. |

**Proves:** the executor interface, tick loop, and two venues (chain adapter + paper ledger) can drive paper deployments from the existing advance loop, sharing the signal FSM and 15% drift *filter*.

**Does not prove:** vault mechanics. `paper_agent_trades` is not a vault book. Regime is always `None`. Live `StrategyRunner` still composes inline. Marketplace `compute_trades` is still a drifted fork (missing #1080).

Do not round this up into “paper trading validates the vault.”

dbrowneup:
Review response pushed (96005c9f) — and Dan weighed in directly on the framing, so recording it here as the product position:

**Paper trading is not vault validation, and was never intended to be.** Archimedes is two engines — a strategy *generation* engine (debate society + rigor gate) and a strategy *execution* engine (the agent tick loop). Paper trading is where the execution engine deploys and runs for real; it is the execution engine's product surface in its own right, not a rehearsal for vaults. Vaults are roadmap; when the chain venue ships it validates the vault mechanic on its own risk budget.

What changed on the branch:
- `execution/__init__.py` — the "WHY THIS EXISTS" block no longer says the paper venue "validates the vault mechanic end to end"; it now states the two-engine framing above.
- `execution/paper_venue.py` — "WHAT THIS VALIDATES, PRECISELY" is now "WHAT THIS RUNS, PRECISELY", and "WHAT THIS IS NOT" gained a first bullet making your points 1–2 structural and permanent: the signal-state book cannot fire price-drift rebalances by construction, and `regime=None` means sizing never exercises a healthy detector. They're now named *in the module doc*, so they survive the merge — your ask.
- Your point 3 (live composition still inline in `tick()`, parity pinned only on extracted functions) stands as accepted residual risk — the #1410 anti-goal was exactly not to ride a live-money refactor into this PR. Migrating `StrategyRunner.tick()` onto the venue protocol is follow-up work with its own risk budget.
- Your hazard 2 (the drifted marketplace `_compute_trades`) got its issue and its fix while you were reviewing: #1719 → PR #1727. Given this PR moves the canonical body into `execution.core.compute_trades`, #1727 will be reworked after this merges so the marketplace adapter calls that same core — one implementation, three call sites, zero forks.
- Quality-bot nits: the redundant `paper_store` side-effect import is dropped (the named import on the next line performs the same registration). The `_DRIFT_THRESHOLD` "unused global" finding is a false positive — `test_agent_runner_tick.py:969` imports it, which is what the re-export comment says. The `venue.py` "statement has no effect" trio are the `...` bodies of Protocol stubs — idiomatic and intentional.

Issue #1632, opened 2026-08-31
[prod][chain] Backend segfaults (exit 139) under RPC distress — abandoned web3 calls' session teardown suspected


dbrowneup:
**Recurred on a converged service, so it gets its own issue** (condition set in the #1594 thread). Two occurrences, different revisions:

- 2026-08-31 ~08:0xZ: task-def **:170**, backend exit 139, during the RPC-429 incident.
- 2026-08-31 ~19:2xZ: task-def **:177** (current tip), backend exit 139, killing the desired=1 singleton (~2-3 min of 503s until replacement).

**Ruled out:** task-level OOM — memory peaked 1,031 MB against the 3,072 MB task ceiling in the surrounding 3 hours.

**Crash context (final app-log lines of the :177 task):** aiohttp `TimeoutError` inside `web3/_utils/http_session_manager.py`, oracle probes blowing their 1.2s budgets ('last completed read 51.6s ago'), slow-request warnings, and — last line before death — `DEADLINE_EXCEEDED: eth-rpc eth_chainId did not answer within 3.00s — abandoning the call (5 in flight)`.

**Working hypothesis:** the #1593 abandonment pattern (BoundedAsyncHTTPProvider abandons calls that blow their deadline while the underlying aiohttp/web3 work keeps running) can leave native-extension code touching a session/connection that later teardown frees — use-after-free → SIGSEGV. Five abandoned calls in flight at crash time fits.

**Acceptance:**
1. **Immediately: enable `faulthandler`** in the backend entrypoint (`python -X faulthandler` or `faulthandler.enable()` at startup) so the NEXT crash logs the native-frame traceback instead of dying silent — cheap, ships in one line, makes this diagnosable.
2. Audit the abandonment path: an abandoned call must never share a session object whose lifecycle the abandoner controls — options: per-call sessions for RPC, a dedicated never-closed session for abandoned strays, or draining strays before teardown.
3. A fault-injection test (mock RPC that answers after the deadline) demonstrating the abandonment path no longer crashes — and pinning whatever invariant the fix establishes.
4. Zero exit-139 stops across one week of prod task history after the fix deploys.

**Anti-goals:** don't remove the deadline/abandonment behavior itself (#1593's fix is correct and load-bearing — the health endpoint must keep answering); don't mute or widen anything in the health path.

Related: #1594 (RPC 429s are the trigger environment), #1593 (the abandonment machinery), MinCapacity=2 (mitigates the blast radius of any singleton crash).

[...]

PR #1726, opened 2026-09-01
Swap psycopg2-binary for source-built psycopg2, and guard the dual-OpenSSL class

dbrowneup:
Part of #1632.

`#1725`'s flag is the tourniquet. This is the surgery: the backend image ships
**source-built `psycopg2` instead of `psycopg2-binary`**, so exactly one OpenSSL
lives in the process, and a permanent guard makes any future second copy fail
the build.

## The symptom this is aimed at

Prod backend aborts ~10 minutes after every boot, on the paper-replay tick,
inside psycopg2's `do_executemany` during the OHLCV cache-write commit:

```
Fatal Python error: Aborted
```

Nothing else. No Python traceback, no glibc `*** buffer overflow detected ***`,
no C++ `terminate called`, no OpenSSL error string — a bare C `abort()`.

## Honest status of the mechanism

**Dual-OpenSSL is the leading hypothesis, not a proven root cause.** Lane 1 owns
the A/B that confirms or kills it, and this PR does not wait on that result or
claim it.

**The swap is correct regardless of how the A/B lands.** psycopg2's own
installation docs say it plainly: the binary package "is a practical choice for
development and testing but in production it is advised to use the package built
from sources", because its bundled libraries can clash with other extension
modules linked against a different build. This image loads ~230 C extensions
(ckzg, greenlet, uvloop, cryptography, …) alongside web3/aiohttp on the
interpreter's OpenSSL. If Lane 1 kills the hypothesis, this PR is still the
configuration psycopg2 tells us to ship, and the guard is still the thing that
would have made the condition visible from day one instead of after a
ten-minute-crashloop investigation.

### 1. The check fails inside the real image when the binary wheel comes back

This is the deploy.yml step, run against `archimedes-backend` built from this
branch, with `psycopg2-binary` force-installed over `/deps`:

```
$ docker run --rm -u root archimedes-backend:1632 sh -c \
    'pip install -q --target=/deps --force-reinstall psycopg2-binary; \
     python -m archimedes.scripts.check_single_openssl; echo EXIT=$?'

check_single_openssl: FAIL
  3 distinct libcrypto builds mapped:
    - /deps/psycopg2_binary.libs/libcrypto-6aa7cfbd.so.1.1.1k
    - /deps/psycopg2_binary.libs/libcrypto-a90fc9c6.so.3
    - /usr/lib/aarch64-linux-gnu/libcrypto.so.3
  2 distinct libssl builds mapped:
    - /deps/psycopg2_binary.libs/libssl-ad00b19a.so.3
    - /usr/lib/aarch64-linux-gnu/libssl.so.3
EXIT=1
```

**Read the third line again.** The binary wheel does not merely add a second
OpenSSL 3 — it also carries a vendored **OpenSSL 1.1.1k**. So the shipped image
had OpenSSL **1.1.1 and 3.x resident in one address space**, and 1.1.1 vs 3.x
export the same symbol names across an incompatible ABI (opaque-struct changes,
different `EVP_*` / `*_CTX` layouts, a different init path). That is about as
close to a purpose-built `abort()` generator as a dependency can get, and it
raises my confidence in the mechanism above where it was when I started — while
still leaving the confirmation to Lane 1's A/B.

Reproduced independently on a bare `python:3.12-slim` with nothing but
`pip install psycopg2-binary aiohttp`, so this is a property of the wheel, not of
our image layout.

### 2. The same check passes on the image this PR builds

```
$ docker run --rm archimedes-backend:1632 python -m archimedes.scripts.check_single_openssl
check_single_openssl: OK — exactly one OpenSSL build mapped.
  /usr/lib/aarch64-linux-gnu/libcrypto.so.3
  /usr/lib/aarch64-linux-gnu/libssl.so.3

$ ldd /deps/psycopg2/_psycopg.cpython-312-*.so | grep -E 'libpq|libssl|libcrypto'
  libpq.so.5    => /lib/aarch64-linux-gnu/libpq.so.5
  libssl.so.3   => /lib/aarch64-linux-gnu/libssl.so.3
  libcrypto.so.3 => /lib/aarch64-linux-gnu/libcrypto.so.3
```

### 3. The source-level backstop was shown failing too

Mutated `requirements-base.txt` back to `psycopg2-binary>=2.9.12` and re-ran:

```
FAILED backend/tests/test_single_openssl.py::test_requirements_do_not_reintroduce_the_binary_wheel
AssertionError: backend/requirements-base.txt installs psycopg2-binary again. ...
Found: ['psycopg2-binary>=2.9.12']
```

Restored, re-ran, green. Every parser test in `TestParser` likewise asserts on
captured two-OpenSSL text that must be reported as a failure — the guard is not
a check that has only ever seen the passing case.

## What changed

| File | Change |
| --- | --- |
| `backend/requirements-base.txt` | `psycopg2-binary>=2.9.12` → `psycopg2>=2.9.12`, with the incident written down next to it |
| `backend/Dockerfile` (builder) | `gcc libpq-dev git` → `gcc libc6-dev libpq-dev git` — psycopg2 now compiles here |
| `backend/Dockerfile` (runtime) | new `libpq5` layer; the distroless comment at the top of stage 2 no longer says "psycopg2-binary" |
| `environment.yml` | conda-level `libpq` so `conda env create` can build the same source package |
| `backend/archimedes/scripts/check_single_openssl.py` | new — the runtime check |
| `backend/tests/test_single_openssl.py` | new — parser regression tests + live check + source backstop |
| `.github/workflows/deploy.yml` | new step: run the check inside the built image, right after the `/health` gate |
| `.github/workflows/quality-gate.yml` | `libpq-dev` before the two `pip install -r requirements-base.txt` steps |

### Two things a reviewer would otherwise have to discover the hard way

**`gcc` alone is not enough.** The first build of this branch failed with
``Python.h:23:12: fatal error: stdlib.h: No such file or directory`` —
`python:3.12-slim` ships no libc headers, so `apt-get install gcc` gets you a
compiler that cannot compile. `libc6-dev` is the missing piece
(`build-essential` also works; this is the same minus g++/make/dpkg-dev).

**`libpq5` at runtime is not optional.** The source-built extension links libpq
dynamically instead of vendoring it. Without the runtime layer the image builds
clean and then dies at import with `libpq.so.5: cannot open shared object file`.

### Image size

Measured on the built images rather than estimated — **the swap makes the image
~11 MB smaller**, because the binary wheel's vendored libraries were larger than
the shared libpq we now install:

| | KB |
| --- | ---: |
| removed — `psycopg2-binary` wheel total | 16,164 |
| &nbsp;&nbsp;of which vendored libpq + two OpenSSL builds | 15,340 |
| added — source-built `psycopg2` package | 1,564 |
| added — `libpq5` + its krb5/sasl chain in the runtime layer | 3,498 |
| **net** | **−11,102** |

Final image: 2.69 GB (dominated by torch + the baked MiniLM, unchanged by this PR).
The builder stage grows by `libc6-dev`, which never reaches the runtime image.

## The guard, and why it is shaped this way

The check is on the **runtime property** — "one OpenSSL is mapped into this
process" — not on the string `psycopg2-binary`. Any future wheel, from any
dependency, can vendor an OpenSSL without the word psycopg2 appearing anywhere
in the diff. A lockfile assertion would miss that; `/proc/self/maps` after
importing `ssl` + `psycopg2` + `aiohttp` catches the class.

Four layers, in increasing order of how much they prove:

1. `test_requirements_do_not_reintroduce_the_binary_wheel` — plain text read,
   runs everywhere including macOS, fails at review time.
2. `TestParser` — the parser against captured `/proc/self/maps` text, including
   the real two-OpenSSL text that **must** be reported as a failure. This is the
   layer that carries correctness on macOS, where layer 3 skips.
3. `test_this_process_maps_at_most_one_openssl` — the live check against the
   interpreter running the suite. Green on the CI runner; skipped where
   `/proc/self/maps` does not exist.
4. The deploy.yml step — the same check **inside the shipped artifact**, which is
   the only place the real runtime linkage exists. This is the one that matters.

Not vacuous by construction: an empty result is a **failure**, not a pass —
importing `ssl` must map an OpenSSL, so zero means the parser stopped matching
rather than that the process is clean.

## Verification

All run on this branch, off a fresh `origin/main` (`e43abe9f`), in a dedicated
worktree.

**Unit suite — the exact CI command, from the repo root:**
```
$ pytest -m "not integration"
5565 passed, 4 skipped, 2 deselected in 737.21s
```
0 failed. The one new skip is `test_single_openssl.py`'s live check on macOS
(`/proc/self/maps` is Linux-only); it runs for real in CI and in the image.

**The alignment guard stays green** — all 6 cases in
`test_env_requirements_alignment.py` pass, untouched. It needed no special-casing:
`environment.yml` reaches the psycopg2 floor through its existing
`-r backend/requirements-base.txt`, so the swap propagated with zero duplication,
and the new `libpq` line is a **conda-level** entry that the guard's pip-block
parser deliberately never sees. No second home for a version was created, so
`test_only_torch_and_sentence_transformers_are_still_written_down_twice` still
holds. (The torch precedent was read first — it did not apply, because unlike the
CPU-wheel index there is nothing here that must be kept out of `environment.yml`.)

**Image builds:** `docker build -f backend/Dockerfile .` → exit 0.

**Image boots against compose Postgres**, following deploy.yml's own boot-validate
pattern (throwaway `postgres:18-alpine` on a user-defined network):

- `python -m archimedes.scripts.alembic_migrate_preflight` → full migration chain
  applied, FK validation clean
- `python -m archimedes.scripts.check_single_openssl` → OK
- `GET /health` → 200, and the three rich-health assertions the deploy gate makes
  all hold: `regime_detector: "live"`, `strategy_count: 34`, `version: "local1632"`
  (the `GIT_SHA` build-arg). Overall `"degraded"` is correct here — no chain RPC,
  no LLM creds, empty corpus in a throwaway DB.
- **The crashing code path, exercised for real:** 20,000-row `cur.executemany(...)`
  + `conn.commit()` against live Postgres through the new psycopg2 — the same shape
  as the OHLCV cache write that aborts in prod. Completed, 20,000 rows read back.
  That is a smoke test, not a repro of a ~10-minute-uptime crash, and I am not
  claiming it as proof the bug is dead.

**Lint:** `ruff check` + `ruff format --check` clean on both new files.

## Needs an explicit ack before merge

`environment.yml` is an env contract everyone rebuilds against, and CLAUDE.md §
"When to ask before acting" lists editing it as ask-first. The change is one
conda-level `libpq` line, required so `conda env create` can build the source
package that production now uses — but it does mean **every dev needs a C
toolchain** (Xcode CLT on macOS, `build-essential` on Linux) on their next env
create/update, where the binary wheel previously needed none. Flagging rather
than assuming.

Also worth a second opinion: the two `libpq-dev` apt steps added to
`quality-gate.yml` are CI wiring, which is the other ask-first item.


danielscoffee (commented):
## Review (not an approval)

This is correctly-scoped hygiene under the #1725/#1728 tourniquet, not a proven #1632 fix. I am **not approving**. Do **not** treat merge as license to set `PAPER_ADVANCE_ENABLED` back to true.

### What looks right
- Builder `gcc` + `libc6-dev` + `libpq-dev` and runtime `libpq5` match a source-built `psycopg2` (the stdlib.h / missing `libpq.so.5` failure modes are real).
- `psycopg2>=2.9.12` replaces `psycopg2-binary>=2.9.12` in the single floor file; conda only adds `libpq` for `pg_config`.
- The guard is non-vacuous (empty maps → fail), dedups segments/`(deleted)`, and `deploy.yml` runs it on the built image **before** ECR push.
- This diff does **not** touch `PAPER_ADVANCE_ENABLED`. The existing deploy pin must stay `false`.
- PR description is honest that Lane 1 A/B owns confirmation, and the 20k `executemany` is a smoke test, not a 10-minute repro.

### Do not merge as “the surgery that kills #1632”
#1729 already reported dual-OpenSSL **present** in the prod image and **not sufficient** to abort in 30 minutes. Error strings in this PR still say the abort “was caused by psycopg2-binary”. That overclaim is how the tourniquet gets lifted too early.

This swap also changes **every** DB connection, not just the paper-replay tick: Debian `libpq5` + system OpenSSL (CA bundle, ciphers, krb5/sasl) instead of the wheel-vendored client. Image size/start time are fine (~11 MB smaller; compile is build-time only).

### Guard gaps (not merge-blockers, but don’t over-read a green check)
- Image check lives only in `deploy.yml` (`push: main`), so **this PR’s CI never built the artifact**. First honest run is deploy; a fail there should stop the push (good), not be “CI was green so ship and flip the flag”.
- Check imports only `ssl` + `psycopg2` + `aiohttp`. Production loads cryptography/web3/etc. It will catch `-binary` coming back; it is not a full-process dual-OpenSSL proof.
- Parser fixture is two OpenSSL 3 copies, not the 1.1.1k+3.x trio described in the PR body (parser would still match `libcrypto-*.so.1.1.1k`).

### Rollback
Revert **the whole PR**. Putting `psycopg2-binary` back without reverting the guard/`test_single_openssl.py` will fail unit tests and deploy. Operational rollback for the abort remains `PAPER_ADVANCE_ENABLED=false`, which this PR must not change.

### CI
PR checks are green (backend unit tests included). Deploy/OpenSSL image gate is not in PR CI. Supply-chain/detect-secrets comments are informational.

**Merge only if we accept this as production-packaging hygiene while the tick stays off.** Soften the “caused by” copy in the guard/deploy comments before anyone uses a green check as evidence to re-enable paper-advance.

dbrowneup:
Review response pushed. Agreed on every point — and to say it plainly for the record: **this PR is packaging hygiene under the tourniquet, not the #1632 fix; merging it changes nothing about `PAPER_ADVANCE_ENABLED`, and nothing about this PR is evidence for re-enabling the tick.** The root cause remains unproven after #1729's A/B (co-residency reproduced, abort not).

What changed on the branch:
- `check_single_openssl.py`'s INCIDENT text no longer says the abort "was caused by" the wheel — it now states the leading-but-UNPROVEN hypothesis, cites #1729's non-repro, and keeps the part that is true regardless: one process, one OpenSSL.
- `deploy.yml`'s comment says CAN produce (not "produce"), cites the A/B, and carries an explicit "a green run here is not license to re-enable PAPER_ADVANCE_ENABLED" warning at the exact spot someone would read for that license.
- The parser fixture now carries the vendored `libcrypto-1.1.1k` alongside the 3.x pair — the mixed-major trio #1729 measured — and the assertion pins all three (your "fixture doesn't match the incident" catch).

Your guard-gap notes (image check runs only at deploy; imports three consumers, not full-process proof) are accepted as stated limits — both are documented in the script docstring rather than fixed here, since the honest first run *is* the deploy gate and it fails closed.

Re-enable protocol stays owner-scheduled: after this deploys and soaks, Dan decides whether to flip the tick on inside #1728's isolated child and watch for child aborts — that soak (or an amd64 repro) is what closes #1632, not this merge.

dbrowneup:
Post-merge record correction: rev 214 (this PR's image) aborted 3× at cold boot with exactly one OpenSSL mapped — **the dual-OpenSSL hypothesis is falsified**, exactly the outcome your review guarded against overclaiming. The real mechanism (concurrent corpus-load session-teardown race via abandoned /health probes) is documented on #1632 with the faulthandler traceback, and #1740 carries the fix. This PR's guard remains correct packaging hygiene — one OpenSSL per process — but it is now provenance for a hygiene property, not for the #1632 fix. The softened language shipped here means nothing needs rewording.

PR #1729, opened 2026-09-01
Reproduction harness + A/B rig for the #1632 psycopg2 abort

dbrowneup:
## Lane 1 — reproduction + A/B verdict for #1632

The #1725 flag is the tourniquet. This is the rig for the surgery: a harness that drives the real crash path in the real image, plus the A/B against the leading mechanism.

**Headline: the dual-OpenSSL condition is CONFIRMED PRESENT in the prod image, but this harness did NOT show it to be sufficient. Neither variant aborted in 30 minutes.** That is a non-repro, reported as a result — the hypothesis is neither confirmed nor killed.

### What the harness drives

The shipped code, not a mock of it:

```
paper_advance_loop            asyncio.to_thread → DEFAULT executor thread
  └─ replay → fetch_real_panel → _fetch_one
       └─ CachingMarketDataProvider.get_daily_ohlcv       ← real, shipped
            └─ read miss → inner fetch → _write_cached_ohlcv
                 └─ session.commit()  →  psycopg2 do_executemany  →  ABORT
```

Only the market-data vendor is substituted (a recorded 2600-bar OHLCV frame). Runs inside the image built from `backend/Dockerfile` — never bare-metal, because the bundled libssl only exists in the pip wheel — against a **TLS-enabled** `postgres:18-alpine`, concurrent with aiohttp TLS churn on a uvloop event loop.

Three fidelity points that were checked, not assumed:
- `pg_stat_ssl` confirms the libpq connection is real TLS: `ssl=True, TLSv1.3, TLS_AES_256_GCM_SHA384`. Prod's `DATABASE_URL` (`infra/outputs.tf`) carries no `sslmode`, so libpq's default `prefer` negotiates TLS against Aurora — the bundled OpenSSL is live, not merely resident. The repo's `localdb` postgres runs `ssl=off`, which would have left it untouched; hence the separate compose file.
- 287 C extensions loaded (prod: ~230), via `import archimedes.main`.
- `uvloop` installed, as `uvicorn[standard]` does in prod.

### Evidence 1 — how many libssl copies each variant loads

`openssl_inventory.py`, in-image, `/proc/self/maps` + `OpenSSL_version()` per mapping:

| | A: `psycopg2-binary` (prod) | B: psycopg2 from source |
| --- | --- | --- |
| libssl/libcrypto mappings | **5** | **2** |
| distinct OpenSSL builds | **2** | **1** |

Variant A, in one address space:

| library | build | origin |
| --- | --- | --- |
| `/usr/lib/*/libssl.so.3` + `libcrypto.so.3` | OpenSSL 3.5.6 | interpreter `_ssl` → aiohttp, web3, httpx, boto3 |
| `psycopg2_binary.libs/libssl-ad00b19a.so.3` + `libcrypto-a90fc9c6.so.3` | OpenSSL 3.5.6, *separate build, separate global state* | the wheel's bundled libpq → the Aurora TLS session |
| `psycopg2_binary.libs/libcrypto-6aa7cfbd.so.1.1.1k` | **OpenSSL 1.1.1k FIPS (2021)** | the wheel's bundled krb5/ldap stack |

So the condition psycopg2's docs warn about is real here, and a 2021-vintage OpenSSL 1.1.1 is riding along inside the wheel. Variant B loads only the system OpenSSL — which is what makes it a usable control.

### Evidence 2 — the A/B, 30 minutes each

Base `e43abe9f` (pre-#1725), `linux/arm64`:

| | A: `psycopg2-binary` | B: psycopg2 from source |
| --- | --- | --- |
| wall clock | 1823 s of 1800 s | 1815 s of 1800 s |
| cache writes | 3 664 | 4 794 |
| **`do_executemany`** | **1 685** (4 381 000 param rows) | **2 206** (5 735 600 param rows) |
| statement | `UPDATE asset_daily_bars` ×1685 | `UPDATE asset_daily_bars` ×2206 |
| `do_execute` | 13 415 | 17 546 |
| aiohttp TLS requests | 290 051 | 311 955 |
| errors | 0 | 0 |
| **aborted?** | **no** | **no** |

**Verdict: non-repro on both halves.** 1 685 passes through the exact frame prod aborted in, on a TLS libpq connection, concurrent with 290 k interpreter-side TLS requests, did not abort. What this *does* kill: the idea that the co-activity alone is sufficient on arm64 against a local postgres. The A/B cannot yet say whether the source build fixes anything, because there was nothing to fix in the control.

### The finding that narrows the search

The harness counts `do_executemany` and `do_execute` separately, and it changed the picture:

| `_write_cached_ohlcv` branch | what SQLAlchemy 2.0 emits | psycopg2 entry point |
| --- | --- | --- |
| new rows (`session.add`) | `INSERT … VALUES (…),(…) RETURNING id` via *insertmanyvalues* | `do_execute` |
| existing rows | `UPDATE asset_daily_bars SET … WHERE id = %(id)s` batched by PK | **`do_executemany`** |

Measured, 40 s: `REPRO_TICKER_MODE=rotate` (all-INSERT) → `do_executemany: 0`. Default `mixed` → 41 calls, all `UPDATE asset_daily_bars`.

**An insert-only loop never enters the crash frame at all.** So prod's abort is on the warm-cache **UPDATE** branch — re-writing a cache entry that already exists, not priming a cold one. Whoever carries this forward should aim there.

### Adversarial pass on the guards

Per the repo rule, each guard was shown to reject something before pushing.

1. **`do_executemany == 0` ⇒ NON-RESULT (exit 3), not "no crash."** Built the input that should trip it (`REPRO_TICKER_MODE=rotate`): harness logged `!!! do_executemany has NOT been called` and exited 3. **This guard caught two real defects in my own work** — (a) a `DefaultDialect.do_executemany` monkeypatch that never fired because `PGDialect_psycopg2` overrides the method, so the counter would have read a silent permanent zero while the run reported a reassuring "no crash"; (b) the INSERT/UPDATE split above.
2. **`configure_mitigation()` actually flips the shipped mitigation.** Mean `executemany` batch size is the observable: `=1` (default) → **2600 rows/call** (= `REPRO_BARS`, one batch per frame, pre-#1725 shape); `=0` → **433 rows/call**, under the 500-row cap. **First attempt failed the check**: `REPRO_DISABLE_MITIGATION` was not in `docker-compose.repro.yml`'s `environment:` block, so it never reached the container and the run reported the default while looking like it honoured the setting. Fixed, then re-verified in both directions.
3. **`libssl copies` in the verdict was over-reporting 3×** (it counted path lines across all three fingerprint dumps, printing 15 instead of 5). Now reads the count the harness itself computed.
4. **Stale verdict files could be reported as this run's.** A run that dies before writing one left the previous file in place. `run_variant` now deletes its own verdict first and stamps each with a UTC finish time.

### Fidelity gaps — state these alongside the non-repro

- **Architecture.** `linux/arm64` here; prod Fargate is `linux/amd64`. The wheel bundles OpenSSL on both, but the binaries differ. This is the biggest gap and the first thing to close.
- **Server.** Local `postgres:18-alpine` vs Aurora 18.3 across a VPC — different TLS peer, latency, idle-timeout behaviour.
- **Load shape.** No FastAPI request traffic, no LLM calls, no on-chain RPC.
- **The B image is not a pure one-variable swap.** Its `apt-get` layer also moves system OpenSSL 3.5.6 → 3.5.7 and libpq 170009 → 170011 — patch-level, but it would need ruling out before attributing any future difference to the psycopg2 flavour alone.

### Note for whoever writes the fix

`backend/Dockerfile`'s builder stage installs `gcc libpq-dev`, which is enough only because `psycopg2-binary` is a prebuilt wheel that never compiles. Switching the requirement to source-built psycopg2 without also adding **`libc6-dev`** fails the image build with `Python.h:23: fatal error: stdlib.h: No such file or directory`. Found the hard way; recorded in `Dockerfile.psycopg2-source`.

### Verification

- Unit suite on this branch: **5557 passed, 3 skipped, 2 deselected, 0 failed** (`pytest -m "not integration"`, repo root, env's pytest 9.0.3 / Python 3.12.13).
- `ruff check` + `ruff format --check` clean on `scripts/repro/issue-1632/`.
- `docker compose -f scripts/repro/issue-1632/docker-compose.repro.yml config -q` valid; `bash -n run.sh`, `sh -n initdb-ssl.sh` clean.
- Rebased onto current `main` (`131149aa`) after #1728 landed the mitigation in the function under test; the default `REPRO_DISABLE_MITIGATION=1` reproduces the pre-mitigation shape, confirmed by the 2600 rows/call batch size matching the original runs exactly.

### Not in this PR

No production code touched — this is `scripts/repro/issue-1632/` only. No merge; the real fix is a separate call.

Part of #1632

### DIFF
--- a/docs/operations/feature-flag-fliplist.md
+++ b/docs/operations/feature-flag-fliplist.md
@@ -191,6 +191,7 @@ require them.
 | `TESTING` | test harness | Forces hermetic paths (fixture pipeline, `refresh_enabled()` off). Never set in a deployed environment. |
 | `PAPER_TRACE_BACKFILL_MAX` | tunable | Default 500. |
 | `REVENUE_SWEEP_INTERVAL_S` / `REVENUE_SWEEP_MIN_USDC` | tunable | Only consulted once `REVENUE_SWEEP_ENABLED=true`. |
+| `REPRO_DISABLE_MITIGATION` | diagnostic harness | Read **only** by [`scripts/repro/issue-1632/repro_1632.py`](../../scripts/repro/issue-1632/repro_1632.py), the #1632 reproduction rig, and injected only by its own `docker-compose.repro.yml`. Default `1` monkeypatches the #1725/#1728 OHLCV write chunking + lock off inside the harness process so it drives the pre-mitigation crash shape; `0` leaves them on. No deployed service reads it, and nothing in `infra/` sets it — flipping it in production would do nothing. Delete this row with the harness once #1632 has a proven cause. |
 
 ---
 
--- a/scripts/repro/issue-1632/.gitignore
+++ b/scripts/repro/issue-1632/.gitignore
@@ -1,0 +1,2 @@
+.certs/
+.logs/
--- a/scripts/repro/issue-1632/Dockerfile.psycopg2-source
+++ b/scripts/repro/issue-1632/Dockerfile.psycopg2-source
@@ -1,0 +1,43 @@
+# The B half of the #1632 A/B: the SAME prod image, with psycopg2 rebuilt from
+# source against the system libpq/libssl instead of the `psycopg2-binary` wheel
+# that bundles its own OpenSSL.
+#
+# Derived from the already-built image rather than re-running the full
+# `pip install -r requirements.txt`, so the ONLY difference between the two
+# variants is the psycopg2 flavour — torch, web3, aiohttp and the rest are the
+# byte-identical layers of the A half. That is what makes the comparison mean
+# something.
+ARG BASE_IMAGE=archimedes-repro1632:binary
+FROM ${BASE_IMAGE}
+
+# Must match the psycopg2-binary version in the base image (run.sh reads it off
+# the A image and passes it in) so the A/B differs in flavour, not in version.
+ARG PSYCOPG2_VERSION=2.9.12
+
+USER root
+
+# libpq-dev pulls libpq5, which the compiled extension needs at RUNTIME too —
+# so this image keeps the toolchain installed. It is a diagnostic image, never
+# a deploy artifact; the production Dockerfile already has gcc + libpq-dev in
+# its builder stage, so the real fix costs no runtime bloat there.
+# NOTE FOR THE FIX PR: `gcc libpq-dev` alone is NOT enough — python:3.12-slim
+# ships gcc without the libc headers, so compiling psycopg2 dies on
+# `Python.h:23: fatal error: stdlib.h: No such file or directory`. libc6-dev is
+# required. backend/Dockerfile's builder stage currently installs only
+# `gcc libpq-dev` and gets away with it because psycopg2-BINARY is a prebuilt
+# wheel that never compiles; switching that requirement to source-built
+# psycopg2 without adding libc6-dev will break the image build.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends gcc libc6-dev libpq-dev \
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /deps/psycopg2 /deps/psycopg2_binary.libs /deps/psycopg2_binary-*.dist-info \
+    && pip install --no-cache-dir --no-binary psycopg2 --target=/deps "psycopg2==${PSYCOPG2_VERSION}"
+
+# Fail the build rather than silently ship a still-bundled variant: if
+# psycopg2_binary.libs survived, the A/B would be comparing an image with
+# itself.
+RUN test ! -d /deps/psycopg2_binary.libs \
+    && ldd /deps/psycopg2/_psycopg*.so | grep -q '/lib/.*libpq.so.5' \
+    && echo "verified: psycopg2 links the system libpq"
+
+USER nonroot
--- a/scripts/repro/issue-1632/README.md
+++ b/scripts/repro/issue-1632/README.md
@@ -1,0 +1,222 @@
+# Repro rig — issue #1632 (backend SIGABRTs in `psycopg2` `do_executemany`)
+
+A local, prod-shaped harness for the crash where the production backend dies
+with a bare `Fatal Python error: Aborted` — no glibc message, no C++
+`terminate`, no OpenSSL error string — inside `psycopg2`'s `do_executemany`
+during the OHLCV cache-write commit on the paper-replay tick.
+
+The flag added in #1725 is the tourniquet. This directory is the diagnosis: it
+drives the exact crash path in a loop and A/Bs the leading mechanism.
+
+## The hypothesis under test
+
+The shipped image installs `psycopg2-binary` (`backend/requirements-base.txt`),
+the wheel flavour that bundles its own OpenSSL — the one
+[psycopg2's own documentation warns against for production][psycopg2-docs] for
+precisely this reason. The result is **three different OpenSSL builds in one
+address space**:
+
+Measured by `openssl_inventory.py` against the image built from
+`backend/Dockerfile` (2026-09-01, `linux/arm64`) — **5 mappings, 2 distinct
+OpenSSL builds**:
+
+| library | build | who uses it |
+| --- | --- | --- |
+| `/usr/lib/*/libssl.so.3` + `libcrypto.so.3` | OpenSSL 3.5.6 | the interpreter's `_ssl` → aiohttp, web3, httpx, boto3 |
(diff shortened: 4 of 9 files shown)


dbrowneup:
CodeQL flagged one new high-severity alert on the first push: `py/insecure-protocol` on the harness's loopback TLS server — `ssl.SSLContext(PROTOCOL_TLS_SERVER)` still permits TLS 1.0/1.1. Fixed by pinning `ctx.minimum_version = ssl.TLSVersion.TLSv1_2`, which is also the more faithful setting: the libpq side of this harness negotiates TLS 1.3 against postgres, so a 1.0-capable loopback server was exercising OpenSSL paths prod never touches.

Also documented why the *client* side runs `verify_mode=CERT_NONE` — the peer is this process's own throwaway self-signed loopback server, created so the harness needs no network. That is safe here and a genuine trust bypass if `REPRO_TLS_URLS` is pointed at a real host, so the comment says so at the call site.

Re-verified after the change: 40 s run, 6 229 aiohttp TLS requests completed, 40 `do_executemany` calls at 2 600 rows/call (the pre-#1725 shape), 0 errors. `ruff check` + `ruff format --check` clean.

dbrowneup:
CI caught one my local run had predated: `test_feature_flag_fliplist_drift.py::test_every_discovered_flag_is_on_the_fliplist` failed on `{'REPRO_DISABLE_MITIGATION': {'scripts/repro/issue-1632/repro_1632.py'}}`. My green local suite ran *before* the rebase that added that knob — rule 2 in the book, and the guard did its job.

Fixed by adding a row to `docs/operations/feature-flag-fliplist.md` under `## Not flags`, alongside the `TESTING` precedent: diagnostic harness only, read by no deployed service, set by nothing in `infra/`, delete with the harness. I did not reshape the read to dodge discovery — the knob is a real env switch and the checklist is entitled to know about it.

Adversarial check on the fix, per rule 4: removed the row → `test_every_discovered_flag_is_on_the_fliplist` FAILS with the same assertion; restored it → passes. So the row is what closes it, and the guard is not being satisfied vacuously.

Full suite re-run on the current tree with the CI command (`pytest -m "not integration"`, repo root): **5607 passed, 3 skipped, 2 deselected, 0 failed** (9m19s).

danielscoffee (commented):
This is a diagnostic landing only — it is not the #1632 fix, does not prove dual-OpenSSL caused the abort, and must not be treated as license to re-enable paper advance.

Diff is nine files: `scripts/repro/issue-1632/*` plus one `docs/operations/feature-flag-fliplist.md` row for `REPRO_DISABLE_MITIGATION` under “Not flags”. `backend/Dockerfile` copies `backend/`, analytics-engine, corpus, and ABIs only, so this never enters the Fargate image or the uvicorn runtime. `PAPER_ADVANCE_ENABLED` / `deploy.yml` are untouched — keep the kill switch pinned false; live path stays leftovers #1730 then #1704 with paper trading off.

Claims vs evidence are mostly honest (neither variant aborted in ~30 min; `do_executemany==0` is a NON-RESULT; UPDATE-vs-INSERT routing is a real narrowing), but “CONFIRMED PRESENT in the prod image” is a locally built `linux/arm64` `backend/Dockerfile` image, not the amd64 Fargate artifact — the README already names that gap, so don’t over-read the headline. `REPRO_DISABLE_MITIGATION` is read only by the harness process; flipping it in prod is a no-op.

Do not merge as a fix; landing as a harness is fine once Backend unit tests go green. Not approving while that required check is still pending, and so this is not misread as closing #1632 or blessing a paper-advance re-enable.

dbrowneup:
Closing per the owner (Q6, 2026-09-02): "close if #1740 really fixes the problem — verify personally first and close with evidence."

**Verified by the owner's session, 2026-09-02 04:55Z:**
- #1740 (merged, live since task-def 215 at 2026-09-01 ~21:30Z) closed the cold-boot mechanism (the corpus-load race). Since then the fleet has rolled through revisions 216 → 230 across nine deploys tonight with **zero** `exit 139`, `Segmentation`, `core dumped` or `Aborted` events in `/archimedes/app` backend streams (filter run over the whole window) and **zero** stopped tasks with a non-zero backend exit code.
- The other suspected frame — the paper-advance replay reaching the OHLCV cache write — was exercised for real at 03:32Z after #1778 lifted the tourniquet: the isolated child ran a full replay of 4 deployments (4 ok, 3 rows appended) and exited cleanly; the web process stayed HEALTHY; CPU peaked at 82 % during the replay and returned to ~8 %. No child exit, no abort, 50+ minutes of healthy observation.

Daniel's framing on this PR was right: it was a diagnostic harness, not the fix, and it never proved the dual-OpenSSL hypothesis. With the fixed mechanism identified elsewhere (#1740) and the tick frame now proven in production, the harness has no remaining purpose on main. The branch stays for the record. Related: #1632, #1728, #1740, #1778.

Issue #1714, opened 2026-09-01
[infra][deploy] Rolling update drains old tasks before new ones are ready — 503 windows on every task-def replace


dbrowneup:
Observed twice tonight (2026-09-01, ECS events): during a task-def replacement the service began 'draining connections' and deregistered old targets while the two NEW tasks were still `pending` — running=0/pending=2 → front-page 503 for ~90s (monitor transitions logged 04:0x and 04:4x). MinCapacity=2 keeps two tasks RUNNING in steady state but does not govern replacement ordering.

Fix shape: set the deployment's `minimumHealthyPercent=100` / `maximumPercent=200` on the service (start new alongside old, drain only after new are healthy) in infra/ecs.tf — verify the current values first and record them in the PR. Acceptance: a task-def-replacing deploy on quiet traffic serves 200 continuously (monitor shows no 5xx transition); ECS events show new tasks HEALTHY before any 'draining' event. Anti-goal: don't raise desiredCount as a workaround.

Part of #1594.

danielscoffee:
Looked at this against `main` and against closed #1309.

The filed fix is already in `infra/ecs.tf`:

```
deployment_minimum_healthy_percent = 100
deployment_maximum_percent         = 200
```

A PR that only sets those two would be a no-op.

This is the same drain-too-soon window as #1309 (old target deregistered ~9s after the new task started, despite 100/200). #1395 landed the nginx container healthCheck and `infra/scripts/verify-zero-downtime-deploy.sh`, then #1309 was closed on that merge. The PR itself said that merge does not close the window: the check is inert until a terraform apply plus one CI deploy, and the probe has never been run against a live rollout.

MinCapacity=2 (autoscaling) keeps two tasks RUNNING in steady state. It does not govern replacement ordering. Last night's ECS events (running=0 / pending=2, ~90s front-page 503 around 04:0x and 04:4x UTC) are consistent with that.

Sibling, not this issue: #1718 moved the corpus-manifest seed off the startup path after the 18,907-row uncap was killing new tasks before they bound. That was a different 502 flap. It does not explain the drain-before-healthy sequence.

What is still unverified (needs AWS, which this look did not have):

1. Live `deploymentConfiguration` vs terraform. CI `deploy.yml` clones the currently registered task definition and rewrites image tags. `deploymentConfiguration` is not in the service's `lifecycle.ignore_changes`, so drift is possible until the next apply.
[...]

Issue #1594, opened 2026-08-31
[ops][chain] Oracle health probe intermittently exceeds its budget — VPC→RPC latency needs a look


dbrowneup:
Post-incident residual (2026-08-31 ~04:30Z): serving is healthy — /health answers ~1.2s, targets green under the real /health check, oracle_probe_state=live on request — but HEALTH_ORACLE_STALE literals are still logging ~30/12min across the 3 tasks, meaning a meaningful fraction of the bounded 1.2s oracle probes time out. #1593 makes this harmless to serving (cached last-known + honest probe_timeout state), which is exactly why the alarm now measures something real instead of taking the site down: the VPC→RPC path (fck-NAT egress → rpc.testnet.arc.network) is intermittently slow. Same signature as tonight's incident trigger. TO INVESTIGATE (daylight): per-call RPC latency histogram from inside a task; whether the two fck-NAT instances or the RPC's rate limiting of our egress IP is the source; whether the oracle RUNNER (same NAT) sees push latency; consider a second RPC endpoint (the back-pocket adapters, #794 stays open) or per-endpoint failover. ALARM NOTE: archimedes-oracle-stale will sit in ALARM while this persists — that is now signal, not noise; do not mute it, fix the latency or re-scope the alarm to distinguish probe_timeout from data-stale. Refs #1592 (closed), #1528, #794.

dbrowneup:
**Root cause found, 2026-08-31 ~07:40Z: the Arc RPC is rate-limiting us.** Live `/archimedes/app` logs (stream `ecs/backend/559322ec…`) show repeated `aiohttp.client_exceptions.ClientResponseError: 429, message='Too Many Requests', url='https://rpc.testnet.arc.network'`. Everything this issue tracked is downstream of that:

- The "VPC→RPC intermittent latency" = 429s + client retries, not network path.
- The oracle-stale alarm firing = honest signal of the same throttling (do-not-mute call stands, vindicated).
- **New tonight — it also breaks deploys:** `/health` does a real RPC round trip, so a cold task warming against the throttled endpoint blows the ALB's 5s×3 budget, the target group kills it, and the deployment circuit breaker declared three consecutive *good* revisions bad (#1597/#1567/#1569 rollouts). Prod was briefly held by a single surviving task of a "failed" deployment. Recovery: TG budget relaxed live to 10s×5 + force-deploy of :171; **PR #1600** codifies the TG values so terraform doesn't revert them.

[...]

PR #1733, opened 2026-09-01
[infra][deploy] Warm request-path caches before the task is an ALB target

### LINKED ISSUE #1713: [infra][deploy] Warm the caches before target registration — every deploy ships a cold fleet
Tonight's measured pattern (2026-09-01): warm Library path 0.75–0.85s, but the FIRST hits after every roll are 12–13s (`/api/strategies/`) and up to 24s (`/api/selection-bias/gate`) while caches fill — and with near-continuous deploys the fleet was perpetually cold, which is most of what the CloudWatch dashboards showed as 'unhealthy'. Two concrete consequences: the strict post-rollout answer probe (5 consecutive <5s) fails whole deploys on single cold samples (three failures tonight → deploy-drift ALARM with prod behind main), and the first visitor after every deploy eats the 13s page.

Fix shape: a warm-up step in the container entrypoint (or an ECS health-check grace + warm-up hook) that primes the known-heavy caches — cohort returns, rigor cache, explore assets — BEFORE the task registers with the target group; alternatively/additionally a short warm-up grace window in the deploy probe (first N samples advisory). Acceptance: a deploy on quiet traffic passes the answer probe 10/10 with no sample >5s; first post-deploy request to /api/strategies/ <2s. Anti-goals: do not loosen the 5s bar itself; do not mark the task healthy before the app can actually serve.

Part of #1594.

danielscoffee:
## What

New Fargate tasks prime the Library page's heavy caches — cohort daily returns, the `strategies_list` rigor memo, and the `selection-bias/gate` rigor memo — in the FastAPI lifespan **before** `yield`. Uvicorn is not listening, so the ALB cannot mark the target healthy, until those caches are populated.

Review follow-up on this branch (do not merge):

1. The 60s budget is real: sync primes run in `asyncio.to_thread` so `wait_for` can fire. A hung `time.sleep(3)` in the worker raises `RequestPathWarmupTimeout` in &lt;2s at a 0.05s budget.
2. Timeout is fail-closed. `RequestPathWarmupTimeout` is **not** swallowed by lifespan. The process never binds; ECS does not get a healthy cold target. That was the #1713 bug.
3. CI rewrite pins backend `healthCheck.startPeriod=90` (same path as `PAPER_ADVANCE_ENABLED=false`) so a cloned live task-def cannot leave startPeriod at 30 against a 60s warmup budget. Existing healthCheck command/interval/timeout/retries are preserved. `infra/ecs.tf` backend startPeriod is 90 for the next terraform apply; **`deployment_minimum_healthy_percent=100` / `200` are untouched.**
4. Tests cover hung/slow sync prime does not listen cold, and the budget is actually interruptible.

Explore-assets is still armed in the background, not awaited.

## Why

Part of #1713. Does not close it. Measured 2026-09-01: warm Library is 0.75–0.85s, but the first hits after every roll were 12–13s (`/api/strategies/`) and up to 24s (`/api/selection-bias/gate`) while caches filled. Near-continuous deploys kept the fleet perpetually cold — which is most of what CloudWatch showed as unhealthy — and the post-rollout answer probe (5 consecutive &lt;5s) failed whole deploys on single cold samples.

`#1718` already moved the corpus seed off the boot path (the acute ALB-window killer). This is the remaining request-path class: `rigor_cache` could not absorb the blob decode because the returns **are** that cache's key, so a rigor HIT still paid `get_all_daily_returns` on every new task.

## Issues

Part of #1713. Does not close #1713 or any issue.
Part of #1594.

## Verification

Rebased onto `main` `38a48530` (#1745) at `68aad0f7`. The only rebase conflict was the duplicate `test_loader_env_override` isolation; kept #1745's `isolated_empty_sqlite` (still asserts `len(load_corpus()) == 4`). This PR no longer diffs `test_strategy_fusion.py`. Warmup lifespan isolation stays in `test_request_path_warmup.py`. Production DB-first `load_corpus()` is untouched.

Hermetic (no AWS, no Redis, no `.env`):

```
env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend \
  python3 -m pytest \
  backend/tests/test_request_path_warmup.py \
  backend/tests/test_strategy_fusion.py::test_loader_env_override -q
```

**12 passed** (warmup file then the fusion env-override test, CI order).

### Prove — a test that fails with the fix reverted

**1. Drop `to_thread`, run `_prime_sync` on the loop.** Hung `time.sleep(3)` (not `asyncio.sleep`) is no longer interruptible.

**2. Restore the lifespan `except Exception` swallow.** Timeout is logged non-fatal and the context yields (listen / register cold).

**3. Cloned startPeriod 30 is overwritten.** `test_live_start_period_30_is_overwritten_command_kept`.

`/health` stays 200 once we listen. We do not listen until warmup finishes. **`PAPER_ADVANCE_ENABLED` remains pinned `false`.** Do not re-enable the tick. Do not raise `desiredCount`.

## What this proves

- Lifespan calls `arm_request_path_warmup` **before** `yield`, does **not** name `evaluate_rigor_gate` or `BacktestResultRecord`, and does **not** wrap warmup in `except Exception`.
- A hung **sync** prime is interrupted inside the budget; lifespan does not yield on timeout.
- CI rewrite pins `PAPER_ADVANCE_ENABLED=false` **and** `healthCheck.startPeriod=90`.
- `REQUEST_PATH_WARMUP` is on the feature-flag fliplist as LIVE (code default ON; `REQUEST_PATH_WARMUP=0` is the emergency cold-serve kill switch). Not pinned in the ECS task-def.
- `test_loader_env_override` still asserts `len(load_corpus()) == 4`. Isolation of that fixture is on main via #1745. This PR only isolates the warmup lifespan drive onto tmp sqlite.

## What this does not prove

- **Vaults.** Untouched.
- **`paper_agent_trades`.** Untouched. `PAPER_ADVANCE_ENABLED` remains `false`.
- **A live deploy.** The issue's acceptance is a production measurement after merge.
- **`deployment_minimum_healthy_percent=100`.** Already in `ecs.tf` (#1714). This PR does not terraform-noop 100/200.
- **Marketplace, Landing.jsx, #1704, #1722.** Untouched.

- Explore-assets is **armed, not awaited**.
- Cancelling `to_thread` does not kill the worker; the decision not to listen fires on the budget.
- `REQUEST_PATH_WARMUP=0` is an emergency kill switch that still listens cold. It is not in the ECS task-def.

**Do not merge. Do not close #1713. Do not re-enable paper-advance. Do not raise desiredCount.**


danielscoffee (commented):
## Review — #1733 / #1713 (cold-fleet warmup)

Not an approve. Deploy/infra that changes when a task becomes an ALB target is high caution, and this PR’s safety valve does not do what the comments claim on the actual hot path.

Happy path is the right shape: `arm_request_path_warmup` runs in FastAPI lifespan **before** `yield`, so uvicorn is not listening, backend `/health` cannot pass, nginx (`dependsOn: backend HEALTHY`) does not start, and the ALB target (nginx:8080) cannot go healthy until Library caches are primed. That is not user-data and not an ECS health-check hook; it is listen-gating. `PAPER_ADVANCE_ENABLED` is untouched (`.github/scripts/ecs_rewrite_task_def.py`, `deploy.yml`, `infra/ecs.tf` are not in the diff). `#1714` rolling (`minimumHealthyPercent=100` / `maximumPercent=200`) is untouched.

### Where warmup runs
- **Container entrypoint:** no
- **ECS health check / ALB target group / user-data:** no
- **FastAPI lifespan, `backend/archimedes/main.py` step 2b, awaited before `yield`:** yes
- Mechanism: delay listen → delay backend HEALTHY → delay nginx → delay ALB healthy. Old tasks keep serving under 100/200, so this should not recreate the #1714 drain 503s. It **does** make new tasks slower to join.

### Delay registration vs log-and-hope
On success, this is real delay-until-warm, not log-and-hope.
On timeout / exception, it is fail-soft: log, `yield`, **register cold**. That is pre-#1713 behaviour and can still fail the post-roll answer probe — the bug this issue exists to stop.

### Timeout / failure mode (the load-bearing gap)
`asyncio.wait_for(_prime, 60s)` cannot interrupt the work that actually costs 12–24s. `_prime` is sync on the event loop for `list_strategies`, `init_db` + `get_all_daily_returns` (blob decode), and `_live_rigor_results_for_strategies` (cohort gate). `wait_for` only fires at the next `await` (`evaluate_rigor_gate` / explore arm). `test_warmup_budget_does_not_block_boot_forever` only hangs on `asyncio.sleep`, so it does not prove the advertised 60s cap.

If that sync work runs long:
- it does **not** fail the task (good for “don’t never join”, bad for “don’t register cold”)
- it **can** hang boot past backend `healthCheck.startPeriod=30s` (live cloned task-def) and ALB `health_check_grace_period_seconds=90`
- nginx will not start until backend `/health` answers; a slow warmup delays the whole task, not just the app container
- with retries `3 × 30s` a task can still become HEALTHY after ~120s, but the 90s ALB grace is tighter once nginx is the registered target (`healthy_threshold=2`, interval 30s)

Authors already note “if warmup regularly exhausts 60s, bump `startPeriod` later — not this PR.” Shipping path is `deploy.yml` cloning the live task-def, so an `ecs.tf` bump would not take effect anyway. Leaving 30s `startPeriod` against a 60s budget plus remaining lifespan (marketplace, RPC, rehydrate, scheduler — all still after warmup, before `yield`) is the deploy risk.

Explore-assets is armed, not awaited. Documented; Library AC still holds; first `/api/explore/assets` after roll can still pay the yfinance rebuild.

### What is solid
- Cohort-returns memo in `get_all_daily_returns`, TTL 600s, copy-out, invalidated from `insert_backtest_if_missing` next to `rigor_cache.clear()`. Tests prove second call issues no SELECT, reversed ids share the key, insert of `[9.0]` does not serve `[0.0]`.
- Lifespan does not name `evaluate_rigor_gate` / `BacktestResultRecord` (OOM frame pin). `evaluate_rigor_gate(strictness=DEFAULT_LEVEL)` is explicit, so it should not hit the Query-object boot bug.
- `TESTING` skips warmup; `REQUEST_PATH_WARMUP=0` kill switch exists (not pinned in the task-def — a broken warmup still needs a deploy to disable).
- No secrets added. Kill switch `PAPER_ADVANCE_ENABLED` remains false on the path that ships.

### Ask before merge
1. Run the sync primes in a worker thread / `asyncio.to_thread` so the 60s budget can actually cancel, **or** drop the claim that hung warmup cannot pin boot.
2. Either pin a larger backend `startPeriod` on the **rewrite** path (the path that ships) so 30s vs 60s is not a live footgun, or measure boot-to-listen on a prod-shaped cohort and show it stays inside 30s + retries and the 90s ALB grace.
3. Wait for **Backend — unit tests** (still in progress on this PR).

Not a merge until SUPER BOSS says merge. I would not approve this revision as-is.

danielscoffee (commented):
Re-review on `e52bc198` — the three blockers I raised. All three are addressed. Not an APPROVE yet: Backend unit tests are still in progress on this head, and those are the tests that prove the cap and fail-closed path. Do not merge. `PAPER_ADVANCE_ENABLED` stays false. `deployment_minimum_healthy_percent` 100/200 untouched.

### 1. 60s cap is real
`arm_request_path_warmup` now does `asyncio.wait_for(asyncio.to_thread(_prime_sync), timeout=WARMUP_BUDGET_SECONDS)`. TimeoutError becomes `RequestPathWarmupTimeout`. `test_warmup_budget_interrupts_a_sync_prime` hangs on `time.sleep(3)` (not `asyncio.sleep`) at a 0.05s budget and requires `RequestPathWarmupTimeout` in &lt;2s. `test_arm_uses_to_thread_so_the_budget_can_fire` pins the source. Cancelling `to_thread` still does not kill the worker; the decision not to listen fires on the budget. That is disclosed and acceptable.

### 2. Timeout is fail-closed
Lifespan awaits `arm_request_path_warmup` before `yield` with no `except Exception` around it. `test_timed_out_warmup_does_not_yield_lifespan` drives the real lifespan and asserts `yielded == []` even if the exception is caught. Source grep forbids the old “failed (non-fatal)” swallow in the lifespan frame. Per-step failures inside `_prime_sync` still fail-soft; only budget exhaustion aborts boot. That is the right split.

### 3. `startPeriod=90` is on the clone path
`.github/scripts/ecs_rewrite_task_def.py` pins `BACKEND_HEALTHCHECK_START_PERIOD = 90` on the backend container (existing command/interval/timeout/retries kept; missing healthCheck filled). Same rewrite still pins `PAPER_ADVANCE_VALUE = "false"`. Tests: live startPeriod 30 overwritten, last-good clone with no healthCheck gets 90, budget ≤ pin, nginx/auth not rewritten. `infra/ecs.tf` 90 is next terraform apply; the load-bearing pin is the rewrite. Terraform fmt job on this head succeeded.

### Still not merge, still not “certainly okay” until CI
- Backend unit tests **in_progress** on this head. Hold the GitHub APPROVE until they pass — this PR’s mutation tests are the proof of (1) and (2).
- Explore-assets still armed, not awaited (disclosed).
- `REQUEST_PATH_WARMUP=0` still listens cold and is not in the task-def (disclosed emergency hatch).
- Do not flip paper-advance. Do not raise `desiredCount`.

danielscoffee:
Not merging. Backend unit tests on `e52bc198` failed 2/5633:

1. `test_every_discovered_flag_is_on_the_fliplist` — `REQUEST_PATH_WARMUP` has no row in `docs/operations/feature-flag-fliplist.md`. Add it as LIVE (default on; `REQUEST_PATH_WARMUP=0` is the kill switch).
2. `test_loader_env_override` — `load_corpus()` returned 18752 papers instead of the fixture 4. Likely the same class as #1740. Don't delete the test.

PAPER_ADVANCE stays false. Not closing #1713.

danielscoffee:
Fliplist row for `REQUEST_PATH_WARMUP` (LIVE) is in on `c2d7e2a`. Still not merging.

Backend tests: 1 failed / 5643 passed — `test_loader_env_override` (`len(load_corpus())` 18752 vs 4). Not deleting that test. Next push should isolate the fixture so warmup/lifespan does not leak the full corpus into it. Related to #1740, but this PR stays red until that assertion holds.

Does not close #1713. PAPER_ADVANCE stays false.

dbrowneup:

**Verdict: FIX-FIRST** — the fail-closed design is right and the container-healthcheck pin is the correct half; the arithmetic that is supposed to make it safe does not close, and two body claims are wider than the diff.

**1. `infra/ecs.tf:504` + `.github/scripts/ecs_rewrite_task_def.py:42-55` — the 90s window is compared against the wrong quantity, and the PR's own comment says so.**
The pin is justified as `startPeriod(90) >= WARMUP_BUDGET_SECONDS(60)`. The real requirement is `grace >= boot-before-warmup + 60 + nginx dependency chain + ALB healthy-threshold`. Nothing in this diff bounds the pre-warmup portion (imports, `init_db`, migrations, the step-1/2 seeds in `main.py` before the new `await arm_request_path_warmup(_app)` at `backend/archimedes/main.py:409-420`). Worse, `infra/ecs.tf:876+` in this same diff keeps the note that nginx `dependsOn backend condition=HEALTHY` with its own `startPeriod=15`, and that with backend at **30s** "its earliest possible verdict lands closer to T+45s". Shift backend to 90 and that same sentence gives ~T+105s — past the 90s `health_check_grace_period_seconds` the comment claims it matches, with `deployment_minimum_healthy_percent=100` holding the old tasks. Exact edit: either raise `health_check_grace_period_seconds` (and the nginx-side budget) to cover `boot + 60 + nginx`, or drop `WARMUP_BUDGET_SECONDS` (`backend/archimedes/services/request_path_warmup.py:301`) to a value that provably fits inside 90 with the measured pre-warmup boot time — and put that measured boot number in the body. Do not merge on `90 > 60`.

**2. `backend/archimedes/services/backtest_repository.py:24-40, 306-312, 350-355` — claim wider than diff: the new memo is process-local, so the invalidation is not "the same" as `rigor_cache`'s.**
The docstring says the memo is "invalidated on the same writer path as `rigor_cache.clear()`". `rigor_cache` has a shared Redis layer (your own warmup docstring names it); `_cohort_returns_store` is a module-level dict behind a `threading.Lock`. A write on task A leaves task B serving stale decoded returns for up to 600s — and since the returns **are** `cohort_key`, task B also serves a stale rigor verdict keyed on stale data. Under Fargate `desiredCount>1` that is a user-visible split-brain on a just-graded strategy, exactly the surface the stored-verdict model cares about. Exact edit: either move the memo behind the same Redis layer, or amend the docstring and the PR body to say **per-task, up to 600s cross-task staleness after a write**, and say why that is acceptable.

**3. PR body does not prove the 503 window shrinks — and the mechanism can widen it.** Every number quoted (12–13s `/api/strategies/`, ~24s `/api/selection-bias/gate`, warm 0.75–0.85s) is *request latency*, not 503 duration; the body concedes "A live deploy … acceptance is a production measurement after merge." Directionally this change makes each new task listen **later** by up to 60s, so the 503/`unhealthy` window is traded, not obviously reduced — it only shrinks if old tasks are still draining, which depends on item 1 holding. Exact edit: retitle the claim to "removes cold first-hit latency after a roll", and state the 503 claim as untested, or land a before/after ALB 5xx count.

**4. Nit, `backtest_repository.py:306` — `cache_key = tuple(sorted(ids))` is not dedup'd,** so a caller passing a duplicate id keys a separate entry that serves the same payload. Harmless today, one `set()` away from correct: `tuple(sorted(set(ids)))`.

**What is right:** `TimeoutError` is raised, not swallowed, and lifespan does not catch it (`main.py:409-420`) — that is the #1713 fail-closed inversion, correctly done. `asyncio.to_thread` is genuinely load-bearing for `wait_for` and the comment says so honestly. Warmup does not hit `/health` (nothing is listening yet), so the container probe cannot time out *against a live listener* — the exposure is item 1's grace window, not the probe's 5s×3. Body's "What this does not prove" section is unusually honest.

**What I ran (exit 0 both):** `gh pr view 1733 --json title,body,files,mergeable,mergeStateStatus` → `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`; `gh pr diff 1733`. Mergeable state is clean; that is not a reason to merge — the body says do not merge, and I agree.

**NOT checked, for lack of time:** I did not run the suite or read `backend/tests/test_request_path_warmup.py` / `test_ecs_paper_advance_deploy_pin.py`, so the three "fails with the fix reverted" proofs are **unverified by me** — a reviewer must see each one red on the reverted input before this lands (guards-shown-red rule). I did not read `infra/alb.tf` to confirm the actual `health_check_grace_period_seconds` value the comment claims is 90 — item 1 rests on that number, check it first. I did not check `evaluate_rigor_gate`'s signature for FastAPI `Depends` defaults being passed positionally by the `asyncio.run(...)` call in `_prime_sync`.

danielscoffee (commented):
Re-review on rebased head `c70586ae`. The three blockers I raised still hold. Holding GitHub APPROVE until **Backend — unit tests** go green on this head (still pending). Do not close #1713. Do not merge from me.

### Blockers — still intact
1. **60s cap is real.** `arm_request_path_warmup` still uses `asyncio.wait_for(asyncio.to_thread(_prime_sync), …)` and raises `RequestPathWarmupTimeout` on budget exhaustion. (Rebase note: strategies-list prime now calls `stored_passports_for` instead of the deleted live-gate helper — that matches main’s verdict-of-record ADR and does not weaken the budget.)
2. **Timeout is fail-closed.** Lifespan still awaits the helper before `yield` without swallowing `RequestPathWarmupTimeout`.
3. **`startPeriod=90` on the clone path.** `ecs_rewrite_task_def.py` still forces `BACKEND_HEALTHCHECK_START_PERIOD = 90` / `rewritten["startPeriod"] = …`. `infra/ecs.tf` backend `startPeriod = 90`. `deployment_minimum_healthy_percent = 100` untouched.

### PAPER_ADVANCE — flag, not introduced by this PR
You said PAPER_ADVANCE stays false. On this head **and on `main`**, `PAPER_ADVANCE_VALUE = "true"` in `.github/scripts/ecs_rewrite_task_def.py` and `infra/ecs.tf` already has `{ name = "PAPER_ADVANCE_ENABLED", value = "true" }`. This PR’s rewrite diff only adds the startPeriod pin; it does **not** flip the tick. If the kill switch must still be off for this merge, that is a separate main/ops decision — not a #1733 regression, but also not “stays false” on the ship path today.

### Next
When Backend unit tests pass on `c70586ae`, I will record APPROVE (self-approve may be blocked; then COMMENT-as-approve). No issue close.

PR #1741, opened 2026-09-01
Lift the paper-advance tourniquet (HELD: after #1740 deploys clean + owner word)

dbrowneup:
**DO NOT MERGE YET.** Sequenced after #1740 (the actual #1632 fix) is deployed and cold-boot deploys run clean, and on Dan's explicit word — he asked for the lift-and-observe experiment.

Part of #1632 — the lift half of its closure. #1740 is the fix half.

## What flips

| site | was | now |
| --- | --- | --- |
| `.github/scripts/ecs_rewrite_task_def.py` `PAPER_ADVANCE_VALUE` | "false" | "true" (the pin that ships on every CI deploy) |
| `infra/ecs.tf` env | "false" | "true" (apply-time twin — **Dan runs terraform apply after merge**) |
| code default | "false" | unchanged — unset still cannot tick |

Guard tests that exist to resist this change are re-aimed at the armed state (`test_ecs_paper_advance_deploy_pin`, `test_paper_advance_kill_switch`), the kill-switch test renamed honestly, and the fliplist row now records the lift: cause found (concurrent corpus-load session-teardown race), fixed (#1740), dual-OpenSSL falsified, tick confined to the #1728 child interpreter so a residual abort kills a child and never `/health`. 53 guard tests green.

## The observation plan after this deploys

Tick fires at +240s then daily; it runs in the child. Watch: task stability (no exit 139 on the web tier), child-abort logs if any, paper ledgers advancing again (the product claim this restores), and prod latency during the tick. Rollback = revert this PR (one commit, three sites move back together).


dbrowneup:
Adversarial pass before merge (two independent Opus refuters, 2026-09-01 ~21:45Z). **Verdict: FIX-FIRST — do not merge as-is.** The findings change the order of operations, so recording them here.

**Blocking**
1. **The lift is inert under the current fleet.** `backend/archimedes/main.py:602-632`: the paper-advance arming (`arm_paper_advance_for_web_tier`) is nested *inside* `if refresh_enabled():`. Task-def 216+ pins `BACKTEST_REFRESH_ENABLED=false` (#1760 mitigation), so that branch logs "backtest refresh scheduler disabled" and never reaches the paper arming. Flipping `PAPER_ADVANCE_ENABLED=true` today would produce zero evidence in either direction. Fix = hoist the arming into its own try/except (the revenue-sweep block is the model) + a source-inspection test that the call site is not nested under `refresh_enabled()`. The #1760 freeze PR (in flight) deletes the scheduler and must carry that hoist; this PR then rebases on it.
2. **The child interpreter drops INFO.** `services/paper_trading.py:1234-1243` `_module_main` never calls `logging.basicConfig`, so the one line that proves a tick worked (`paper advance: {...}`) cannot reach CloudWatch. Add basicConfig(INFO, stdout) + a subprocess test.
3. **Branch is behind main and does not contain #1740**, the precondition the body names; `docs/operations/feature-flag-fliplist.md` conflicts with main (the row still writes `a-apin`). Merge main in and re-run.
4. **Guard regression:** `test_already_false_is_not_duplicated` had its input changed from `"false"` to `"true"`, so the exact production input on the first post-merge deploy (live task-def says false → rewrite must overwrite once) is no longer tested. Restore the input, keep the `== "true"` assertions, rename.
5. **Comments/tests now assert the opposite of the code:** `infra/ecs.tf:558-594` ("TEMPORARY … MUST GO BACK TO true", "keep this line false") above `value = "true"`, and `test_paper_advance_kill_switch.py:219` *requires* the word TEMPORARY to stay; `.github/scripts/ecs_rewrite_task_def.py:16-22,54-59` still say "pins false / do not flip it here"; `deploy.yml:375-377,713` still instruct not to re-enable; fliplist `:53` and the deployed-value cell at `:77` still say pinned false, and the Notes cell now asserts two incompatible causes; `paper_trading.py:73-84` still asserts the retracted OHLCV-write cause; three test names/docstrings assert the old value. Commit subject "all three pins" — two moved (code default stays false).
6. **Terraform interlock:** `BACKTEST_REFRESH_ENABLED` is in no `.tf` and not in the rewrite; the service ignores task_definition changes, but the NEXT CI deploy reads the family's latest ACTIVE revision, which after a terraform apply is the terraform-authored one — the 216+ pin would evaporate. Moot once the freeze PR deletes the loop; until then, no apply.

**Not blocking, must be honest about**
7. The tick's own crash frame (`fetch_real_panel` → `_write_cached_ohlcv`) is not the path #1740 serialized; the #1728 process boundary is the whole protection. Say that instead of "fixed (#1740)". The in-process `_OHLCV_CACHE_WRITE_LOCK` no longer serializes child-vs-parent writes; with ≥2 tasks the concurrency variable is back — a Postgres advisory lock around the commit is the real fix.
8. Two tasks tick at +240s within seconds of each other; `PaperDailyReturn(deployment_id, date)` is unique and `_run` commits the whole cycle in one transaction, so the loser's `IntegrityError` rolls back its entire cycle (the winner's stands). Functionally survivable, noisy, and the "idempotent appends" docstring is idempotent-by-constraint-violation. Fix: one ticker per fleet (advisory lock) or per-deployment commits.
9. No `.cancel()` on `app.state.paper_advance_task` at shutdown: a draining task's child keeps writing while its replacement's child starts.

**Sequence now:** freeze PR (#1760, hoists the arming) → rework this PR on top of it with 1–9 → merge → observe with the refuter's 17-point checklist (armed lines at +0–3 min, first tick at +4–6 min, no exit 139/134, `paper_daily_returns.appended_at` moving within 30 min).

danielscoffee:
Independent check from the polish lane (paper-trading UX) on the adversarial FIX-FIRST comment.

**Verified on `main` today**
1. **Lift is inert under the current fleet.** In `backend/archimedes/main.py`, `arm_paper_advance_for_web_tier()` sits inside `if refresh_enabled():`. With task-def 216+ pinning `BACKTEST_REFRESH_ENABLED=false`, flipping `PAPER_ADVANCE_ENABLED` alone cannot arm the tick. Agree with the revenue-sweep-shaped hoist + a source-inspection test that the call site is not nested under `refresh_enabled()`.
2. **Child drops INFO.** `_module_main` in `services/paper_trading.py` still has no `logging.basicConfig`, so the proving `paper advance: {...}` line cannot reach CloudWatch.
3. **This branch cannot merge as-is.** `mergeStateStatus=DIRTY` / conflicting with `main` (and #1740 is already merged — the named precondition).

**Sequencing note:** #1760 is still an **open issue** (cold-boot backtest-refresh storm); there is no freeze PR open yet. Happy to keep the hoist + child INFO config out of this lift PR so observation can work once the pin flips — shipping that as a small follow-up on `main`. Will not approve #1741 until the hoist is in, the honesty/guard regressions in findings 4–5 are cleaned, and the branch is rebased.

Not closing. Not merging.

dbrowneup:
Superseded by #1778 — the same lift reworked on top of the freeze PR #1766 with every FIX-FIRST item from the adversarial pass addressed (unconditional arming, child INFO logging, one ticker per fleet via the advisory lock, honest fliplist/guard wording, first-deploy overwrite test). Daniel's polish-lane check on #1778 is positive. Closing this one so there is a single PR to vet; nothing merges without the owner's word.

Issue #1636, opened 2026-08-31
[strategy] Papers per strategy: ask for 5, allow 30, keep 2 as the honest floor — and put the papers into the debate


dbrowneup:
**Why**

Generated strategies cite **two papers** most of the time. That is not a truncation bug; it is three compounding design choices:

- **`max_papers` is retrieval width *and* fusion budget.** `select_candidates` returns *exactly* `paper_budget` (`strategy_fusion.py:722`). At the API default of 5, the model is shown five abstracts, total. There is no "retrieve 30, fuse 5" separation.
- **The prompt anchors on two.** `strategy_fusion.py:818-820`: *"You MUST fuse AT LEAST TWO … You may use a subset."* Nothing asks for more; nothing penalizes two. `max_papers_to_fuse` is passed in the payload (`:881`) but only `min_papers_to_fuse: 2` is enforced anywhere.
- **The gate is `>= 2`, so 2 is a pass, not a signal.** `strategy_fusion.py:1014`, `:777-779`. No shortfall log, no budget-vs-used record, no UI surfacing.

Defaults also disagree across entry points — **5** (`generate_schemas.py:47`, `Generate.jsx:104`) vs **4** (`strategy_fusion.py:523`, `strategies_routes.py:1842`, `:1985`) — so the two generation routes already behave differently.

Second half of the same problem: even when the proposer *does* use papers, the **debate stage receives none**. `debate_engine.py:328` sends a `"; "`-joined list of strategy *names* as the entire user message for all four turns. A judge reading "multi-agent debate over 10,000 papers" is being told something the live path does not do.

**What**

**A. Split the overloaded knob.** `MIN_PAPERS = 2` currently plays *three* roles (corpus-viability precheck `debate_engine.py:230,238`; per-steer evidence gate `:271-273`; pre-LLM gate `strategy_fusion.py:971-980`; post-hoc acceptance `:1014`). Raising it to 5 converts retrieval thinness into `GENERATION_UNAVAILABLE`. Instead:

```python
MIN_PAPERS       = 2   # KEEP — hard validity floor (fusion ≠ extraction)
FUSE_TARGET_MIN  = 5   # NEW  — what we ASK the model to fuse
FUSION_MAX_PAPERS = 30 # RAISE — retrieval width + prompt budget
```

[...]

PR #1744, opened 2026-09-01
[strategy] Multi-paper utilization: make the paper→mechanism link explicit, validated, and durable

### LINKED ISSUE #1739: [strategy][generation] Multi-paper utilization is unverified: a citation list is a self-report with no link to the mechanism the spec trades
## Summary

A generation run puts up to 80 paper-slots in front of the proposers (`max_papers` default 8 × `DEBATE_POOL_MAX` 10 steers — `strategy_fusion.py:78-91`, `debate_engine.py:322`), and the debate now genuinely argues over named papers (#1636). But **nothing connects the papers a strategy cites to the mechanism it trades.** `FusionProposal.is_actionable` counts list length (`strategy_fusion.py:863-872`); `source_arxiv_ids` is the model's own claim, filtered only for set membership (`strategy_fusion.py:1142-1146`); `validate_strategy_spec` accepts `source_arxiv_ids: []` and never compares the spec's list to the proposal's (`strategy_dsl.py:250-256`); and `_canonical_spec_hash` explicitly excludes citations from the behavioural hash (`debate_engine.py:234`) — the codebase's own statement that the papers do not touch behaviour.

Demonstrated (env python, from `backend/`):

```
validate_strategy_spec -> ['sma_200']      # one mechanism, source_arxiv_ids: []
_dsl_conformance_ok    -> True
is_actionable          -> True | is_shortfall -> False   # claims 5 papers
hash(2 papers) == hash(1 totally different paper) -> True
```

A single-`sma_200` crossover claiming five papers passes every check and is hash-identical to the same spec citing one unrelated paper. On the passport, citation count reads as evidence depth — the exact laundering #1636 named as "strictly worse than an honest 2".

Four things that *would* make it checkable already exist and are dropped:

| value | computed at | lost at |
| --- | --- | --- |
| `papers_offered` (budget-vs-used) | `strategy_fusion.py:1167, :1237` | `_make_entry` never copies it — `debate_engine.py:814-839` |
| `fusion_reasoning` (per-paper mechanism prose) | `strategy_fusion.py:1229` | write-only on `_CandidateResult.reasoning` (`debate_engine.py:821`); `_persist_candidate`/`upsert_strategy` have no parameter (`generation_pipeline.py:1694-1723`, `strategy_store.py:240-258`) |
| `debate_paper_verdicts` (per-paper tally) | `debate_engine.py:1046, :1114-1116` | "NOT persisted… in-process + SSE-visible only" (`generation_pipeline.py:439-442`); SSE carries a summary string only (`debate_engine.py:1049-1054`) |
| paper titles | `evidence_by_id` (`debate_engine.py:354-367`) | hardcoded `"title": ""` (`debate_engine.py:819`) |

Also in scope because it ships on every proposer call: a merge-conflict marker `>>>>>>> origin/main` at `strategy_fusion.py:912`, inside `_SPEC_CONTRACT`, therefore inside both `_SYSTEM_PROMPT` (`:973`) and `_SPEC_REPAIR_SYSTEM` (`:982`). Verified present in both at runtime. Introduced by `bdb93547`.

## The substrate this builds toward (owner vision — direction, NOT this diff's scope)

Dan, 2026-09-01: the KB pipeline is all about **how the agents use the papers** — a knowledge graph that helps agents retrieve, relate, and understand papers, plus the ability to read parts or the whole text, inside the strategy-generation process. The human analogy: a mental model of all known papers → scan abstracts → find relevant information → pull multiple papers together into a synthesis → read some fully, scan bits of others. **The highly recombinatorial nature — many potential composites from various subsets of the corpus — is what feeds novelty and strategy creation.** It is a key part of the intelligence underlying the generation engine, not a deferrable side feature.

This issue is the first honest rung of that ladder: before the substrate can *improve* multi-paper composition, the pipeline must be able to *measure and record* it. The scope below makes the paper→mechanism mapping explicit, validated, and durable on today's lexical corpus. The KB substrate (#1090–#1093, #778's residue) is consolidated here as this issue's roadmap — those closures do not retire the vision; they point at it living here.

## Scope

Do exactly this, nothing more.

1. **`backend/archimedes/agents/strategy_fusion.py`**
   - Delete line 912 (`>>>>>>> origin/main`).
   - Extend the output schema in `_SYSTEM_PROMPT` (`:949-974`) and `_SPEC_CONTRACT` with `paper_mechanisms: [{"arxiv_id": <from source_arxiv_ids>, "mechanism": <str>, "spec_elements": ["<indicator alias used in entry/exit>", …]}]`. State that an id whose mechanism cannot be named must be **omitted from `source_arxiv_ids`**, not padded in.
   - Add `paper_mechanisms: list[dict[str, Any]] = field(default_factory=list)` and `distinct_mechanism_papers: int = 0` to `FusionProposal` (beside `papers_offered`, `:860`).
   - In `propose` (after the `valid_ids` filter, `:1142-1146`): drop `paper_mechanisms` entries whose `arxiv_id` is not in `source_ids`; drop `spec_elements` entries that are not an indicator alias present in the validated spec's `indicators`. Reuse the check shape at `strategy_dsl.py:269-274`. Set `distinct_mechanism_papers = len({e["arxiv_id"] for e in kept if e["spec_elements"]})`. Log a WARNING when `distinct_mechanism_papers < len(source_ids)`.

2. **`backend/archimedes/agents/generation_pipeline.py`** — add to `_CandidateResult` (beside `:442`): `papers_offered: int = 0`, `distinct_mechanism_papers: int = 0`, `fusion_reasoning: str = ""`.

3. **`backend/archimedes/agents/debate_engine.py` `_make_entry` (`:800-839`)** — populate the three new fields from the proposal, and build `source_papers` as
   `[{"arxiv_id": a, "title": evidence_by_id.get(a, {}).get("title", ""), "mechanism": …, "spec_elements": […]} for a in proposal.source_arxiv_ids]`.
   Thread `evidence_by_id` into `_make_entry` and `build_leaderboard`.

4. **`backend/archimedes/agents/generation_pipeline.py` `_persist_candidate` (`:1694-1723`)** — no schema change: the enriched dicts ride the existing `source_papers` JSON text column (`strategy_store.py:334`), and `_resolve_source_papers` does `entry = dict(paper)` (`api/strategies_routes.py:1171`) so the extra keys reach the API response unchanged. Attach `debate_paper_verdicts` and `fusion_reasoning` to the persisted `debate_transcripts` row in `_persist_debate_transcripts` (`:1011-1061`) — that table already exists and already stores a JSON turn list.

5. **`backend/tests/test_multi_paper_utilization.py`** (new, hermetic) — the tests below.

**No migration. No new route. No new LLM call.**

## Acceptance criteria

Environment: clean checkout, no `.env`, no Docker/Redis/DB, no network. Use the env's pytest explicitly (`CLAUDE.md` § Setup).

- [ ] **The conflict marker is gone from the live prompts.**
  ```bash
  cd backend && /opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/python -c "
  import sys; sys.path.insert(0,'.')
  from archimedes.agents.strategy_fusion import _SYSTEM_PROMPT, _SPEC_REPAIR_SYSTEM
  assert '>>>>>>>' not in _SYSTEM_PROMPT and '>>>>>>>' not in _SPEC_REPAIR_SYSTEM
  print('OK')"
  ```
  Expected: `OK` (run the check on unmodified `main` first and confirm it raises).

- [ ] **`test_unattributed_citation_does_not_count_as_a_mechanism_paper`** — a stub citing 5 valid ids but naming `spec_elements` for only 2 yields `len(source_arxiv_ids) == 5` **and** `distinct_mechanism_papers == 2`, plus a WARNING log line containing `"2 of 5"`.
- [ ] **`test_spec_element_outside_the_spec_is_dropped`** — a `paper_mechanisms` entry whose `spec_elements` names `bollinger_20` (not in the spec's `entry`/`exit`) is dropped; that paper contributes 0 to `distinct_mechanism_papers`. Mirrors `strategy_dsl.py:269-274`.
- [ ] **`test_single_mechanism_spec_claiming_five_papers_is_labelled_not_hidden`** — the adversarial spec above (`entry`/`exit` on `sma_200` only, `source_arxiv_ids` of length 5) yields `is_actionable is True` **and** `distinct_mechanism_papers == 1`. It is **labelled, not rejected** — #1636's honest-shortfall rule.
- [ ] **`test_papers_offered_and_mechanism_count_survive_to_the_candidate`** — `_make_entry` output has `papers_offered == 30`, `distinct_mechanism_papers == 2`, non-empty `fusion_reasoning`. Revert the `_make_entry` copy and confirm it fails.
- [ ] **`test_persisted_source_papers_carry_title_and_mechanism`** — after `_persist_candidate` against tmp-sqlite (copy the `_use_tmp_db` fixture from `backend/tests/test_generated_citation_truth.py:41-53`), `json.loads(record.source_papers)` has, for every entry, a non-empty `title` and a `mechanism` key; `_resolve_source_papers` preserves both.
- [ ] **`test_debate_paper_verdicts_are_durable`** — the persisted `debate_transcripts` row for a run carries the tally; assert every retrieved id appears, `unused` rows included. Precedent: `backend/tests/test_debate_engine.py:944-990`.
- [ ] **Suite green.** `/opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/pytest -m "not integration"` from the repo root → `0 failed`. Baseline for the two touched files today: `pytest --collect-only -q backend/tests/test_strategy_fusion.py backend/tests/test_debate_engine.py` → `84 tests collected`; the count must go **up**, and all 84 must still pass.
- [ ] **Adversarial pass in the PR body** (guards-need-an-adversarial-pass): paste the run of `test_single_mechanism_spec_claiming_five_papers_is_labelled_not_hidden` and `test_spec_element_outside_the_spec_is_dropped` against **unmodified `main`**, showing both fail there.

## Verify

```bash
cd <repo-root>
PYTEST=/opt/homebrew/Caskroom/mambaforge/base/envs/archimedes/bin/pytest
$PYTEST -q backend/tests/test_multi_paper_utilization.py     # → 0 failed
$PYTEST -q backend/tests/test_strategy_fusion.py backend/tests/test_debate_engine.py
$PYTEST -m "not integration"                                  # → 0 failed
grep -rn '>>>>>>>' backend/archimedes/                        # → no output
```

## Anti-goals (for THIS diff — the roadmap section above is direction, not scope)

Written so a `grep` can prove them absent.

- **No knowledge-graph, no embeddings, no semantic-retrieval implementation in this diff.** Retrieval stays lexical (`strategy_fusion.py:772-802`). Do not touch `services/paper_rag.py`, do not raise `_RERANK_MAX_TEXTS`, do not add `sentence-transformers`/SPECTER2, do not add a `kg_entities`/`kg_relations`/`cluster_id` read or write, and do not use the words "semantic", "embedding", or "knowledge graph" in any user-facing string — the substrate arrives via its own specs when scheduled, not as a side effect here.
- **No new pipeline and no new route.** The debate society stays the sole generation path (`docs/adr/debate-society-sole-generation-pipeline.md`); `backend/tests/test_sole_generation_route_guard.py` must stay green.
- **No curated-library changes.** Do not touch `analytics-engine/`, the 34 curated fixtures, or anything in `docs/audits/2026-07-09-curated-consolidation.md`.
- **Do not turn `distinct_mechanism_papers` into a gate.** No new reject path, no `raise`, no `DebateUnavailable`. It labels. An honest single-mechanism strategy ships with an honest count.
- **Do not change `MIN_PAPERS` (2), `FUSE_TARGET_MIN` (5), `FUSION_MAX_PAPERS` (30), or `DEFAULT_MAX_PAPERS` (8)** (`strategy_fusion.py:78-91`), and do not touch `api/generate_schemas.py:29-31` — `backend/tests/test_generate_schemas_depth_drift.py` pins them equal.
- **No Alembic migration.** No file added under `backend/alembic/versions/`.
- **Do not add `source_arxiv_ids` back into `_HASH_IGNORED_SPEC_KEYS`'s exclusion** — the dedup at `debate_engine.py:234` is correct and #893 depends on it. This issue makes the mapping visible, it does not make citations behavioural.
- **No full-text ingestion in this diff** (that is the substrate roadmap, was #1091).

## Precedent to copy

- `strategy_dsl.py:269-274` — `parameter_variants` keys are validated to reference an indicator alias actually used in `entry`/`exit`. `paper_mechanisms.spec_elements` is the same check with a different key set; reuse `all_indicators`.
- `strategy_fusion.py:1140-1146` — the `valid_ids` anti-hallucination filter shape.
- `debate_engine.py:467-501` — keep-the-claim-but-strip-the-id honesty pattern: an unattributable paper is *recorded as unattributed*, never deleted and never laundered.
- `backend/tests/test_generated_citation_truth.py:41-53` — the hermetic tmp-sqlite DB fixture.

## References — the closed issues consolidate HERE

Bulk-closed 2026-09-01 as out-of-MVP. Per the owner, those closures do not retire the KB vision — it lives in this issue's substrate/roadmap section, and future substrate work supersedes from here rather than reopening them piecemeal.

- **#1090** real KB pipeline invocation → substrate roadmap (retrieve/relate/understand layer).
- **#1091** full-text hydration (668/10,000) → substrate roadmap (read-parts-or-whole capability).
- **#1092** cluster/topic/KG entity backfill → substrate roadmap (the relate layer).
- **#1093** Bedrock KB bridge → substrate roadmap (optional retrieval bridge, per the hybrid build-vs-buy decision).
- **#778** "10,000 papers" claim-integrity → honest-surface half shipped; the RAG/KG residue is this issue's substrate roadmap. `docs/specs/multi-agent-debate-spec.md:193, :423, :440, :457` still name it as the quality lift for proposer divergence.
- **#1636** papers-into-the-debate (completed) — this issue is its direct successor: #1636 fixed retrieval width and debate attribution; it left the paper→mechanism link unmade, `papers_offered` as a log line, and `debate_paper_verdicts` in-process only (`generation_pipeline.py:439-442`).

dbrowneup:
## What & why

`Closes #1739`

A generation run puts up to 80 paper-slots in front of the proposers, and since #1636 the debate genuinely argues over named papers — but **nothing connected the papers a strategy cited to the mechanism it traded.** `is_actionable` counted list length, `source_arxiv_ids` was filtered only for set membership, `validate_strategy_spec` accepted `source_arxiv_ids: []`, and `_canonical_spec_hash` deliberately excludes citations. A single-`sma_200` crossover claiming five papers passed every check and was hash-identical to the same spec citing one unrelated paper. On the passport, citation count reads as evidence depth — the exact laundering #1636 named as "strictly worse than an honest 2".

This makes the mapping explicit, validated, and durable, on today's lexical corpus. **It labels; it never gates.**

## What changed

**`strategy_fusion.py`**
- Deleted the `>>>>>>> origin/main` merge-conflict marker at `:912`. It sat inside `_SPEC_CONTRACT`, therefore inside **both** `_SYSTEM_PROMPT` and `_SPEC_REPAIR_SYSTEM`, and shipped on every proposer call (introduced by `bdb93547`).
- Output schema gains `paper_mechanisms: [{arxiv_id, mechanism, spec_elements}]`, with the hard rule stated: an id whose mechanism cannot be named is **omitted from `source_arxiv_ids`**, not padded in. The `_SPEC_CONTRACT` half states that `paper_mechanisms` is a top-level proposal field, *not* a spec key (so the repair prompt, which emits only the spec, is not misled) and that `spec_elements` follow the same alias rule `parameter_variants` keys do.
- `FusionProposal` gains `paper_mechanisms` and `distinct_mechanism_papers`.
- `propose()` filters both halves server-side: entries whose `arxiv_id` is not in `source_ids` are dropped (the `valid_ids` anti-hallucination shape, right beside it); `spec_elements` that are not an indicator alias in the **validated** spec's `indicators` are dropped (reusing `all_indicators`, `strategy_dsl.py:269-274`). The entry is kept with its claim and its id — only the unsupported link is stripped, the keep-the-claim/strip-the-id pattern from `debate_engine.py:467-501`. `distinct_mechanism_papers = len({e["arxiv_id"] for e in kept if e["spec_elements"]})`, with a WARNING when it is below `len(source_ids)`.

**`generation_pipeline.py`** — `_CandidateResult` gains `papers_offered`, `distinct_mechanism_papers`, `fusion_reasoning`. `_persist_debate_transcripts` attaches the run's `debate_paper_verdicts` tally and `fusion_reasoning` to the persisted row as one extra entry on the JSON list the `transcript_json` column already holds (no migration, no new table). The stale `# NOT persisted…` comment on `debate_paper_verdicts` is corrected rather than left to read false.

**`debate_engine.py`** — `_make_entry` populates the three fields and builds `source_papers` as `{arxiv_id, title, mechanism, spec_elements}`, with the title coming from `evidence_by_id` (previously the literal `""`). `evidence_by_id` is threaded through `build_leaderboard` as an optional keyword, so the pure ranking tests keep calling it unchanged; it never influences ranking, culling, or the abstain paths.

**`backend/tests/test_multi_paper_utilization.py`** — new, hermetic (no `.env`, no network, no Redis, tmp-file sqlite; DB fixture copied from `test_generated_citation_truth.py:41-53`).

No migration. No new route. No new LLM call. No UI change.

## Acceptance criteria

**Conflict marker gone from the live prompts** — run against unmodified `main` first, as the issue asks:

```
# unmodified main
$ cd backend && .../python -c "...assert '>>>>>>>' not in _SYSTEM_PROMPT and ... _SPEC_REPAIR_SYSTEM; print('OK')"
Traceback (most recent call last):
  File "<string>", line 4, in <module>
AssertionError
$ grep -rn '>>>>>>>' backend/archimedes/
backend/archimedes/agents/strategy_fusion.py:912:>>>>>>> origin/main

# this branch
$ cd backend && .../python -c "... print('OK')"
OK
$ grep -rn '>>>>>>>' backend/archimedes/
(no output)
```

**The six named tests** — plus one added in review (the sanitizer guard, below)

```
$ .../pytest -q backend/tests/test_multi_paper_utilization.py
collected 7 items
backend/tests/test_multi_paper_utilization.py .......                    [100%]
7 passed, 1 warning in 2.42s
```

**Suite green + the pinned baseline goes up, with all 84 still passing**

```
$ .../pytest --collect-only -q backend/tests/test_strategy_fusion.py backend/tests/test_debate_engine.py
84 tests collected                       # unchanged — no existing test was edited
$ .../pytest --collect-only -q <those two> backend/tests/test_multi_paper_utilization.py
91 tests collected                       # 84 -> 91

$ .../pytest -q <those two> + the 4 guard files named in the anti-goals
127 passed, 7 warnings in 16.53s

$ .../pytest -m "not integration"
5632 passed, 4 skipped, 2 deselected, 327 warnings in 545.64s (0:09:05)     # exit code 0
```

Guard files named in the anti-goals, run explicitly and green inside the 127 above: `test_sole_generation_route_guard.py`, `test_generate_schemas_depth_drift.py`, `test_debate_transcript_persistence.py`, `test_generated_citation_truth.py`.

**Anti-goals, checked mechanically**

```
$ git diff -U0 | grep '^+' | grep -niE "semantic|embedding|knowledge graph|kg_entities|kg_relations|cluster_id|sentence-transformers|SPECTER"
(no output — none of that vocabulary is introduced anywhere in the diff)
$ git status --porcelain backend/migrations/
(empty — no Alembic revision added)
$ git diff --stat
 backend/archimedes/agents/debate_engine.py       |  59 ++-
 backend/archimedes/agents/generation_pipeline.py | 120 ++++-
 backend/archimedes/agents/strategy_fusion.py     |  95 +++-
 backend/archimedes/models/debate_transcript.py   |  59 ++-
```
`services/paper_rag.py`, `_RERANK_MAX_TEXTS`, `analytics-engine/`, `api/generate_schemas.py`, `_HASH_IGNORED_SPEC_KEYS` and the four paper constants are all untouched. `distinct_mechanism_papers` has no `raise` and no reject path — grep the diff for `DebateUnavailable`: absent.

**Lint** — `ruff format --check` clean on all five files; `ruff check` clean on the four I touched with new code. The two `SIM905` hits `ruff check` reports on `generation_pipeline.py` are **pre-existing on `main`** (verified against the unmodified checkout: `Found 2 errors`) and are in word-list constants my diff does not touch.

## Adversarial pass

Per *guards-need-an-adversarial-pass*: every guard was run against code that should fail it, and does.

### a. Every new test against unmodified `main`

```
$ cd <main worktree> && .../pytest -q backend/tests/test_multi_paper_utilization.py
collected 6 items
FFFFFF                                                                   [100%]

test_unattributed_citation_does_not_count_as_a_mechanism_paper
E   AttributeError: 'FusionProposal' object has no attribute 'distinct_mechanism_papers'
test_spec_element_outside_the_spec_is_dropped
E   AttributeError: 'FusionProposal' object has no attribute 'paper_mechanisms'
test_single_mechanism_spec_claiming_five_papers_is_labelled_not_hidden
E   AttributeError: 'FusionProposal' object has no attribute 'distinct_mechanism_papers'
test_papers_offered_and_mechanism_count_survive_to_the_candidate
E   TypeError: _make_entry() got an unexpected keyword argument 'evidence_by_id'
test_persisted_source_papers_carry_title_and_mechanism
E   TypeError: _CandidateResult.__init__() got an unexpected keyword argument 'distinct_mechanism_papers'
test_debate_paper_verdicts_are_durable
E   TypeError: _CandidateResult.__init__() got an unexpected keyword argument 'distinct_mechanism_papers'

6 failed, 1 warning in 1.80s
```

Those are *absence* failures, which prove the feature is new but not that the checks bite. So each guard was also mutated **on this branch**, one at a time, and the mutation reverted after:

### b. MUT-A — accept any `spec_elements` value (drop the `all_indicators` check)

`if isinstance(el, str) and el in valid_elements` → `if isinstance(el, str)`

```
E   assert ['bollinger_20'] == []
    "an alias the spec never uses must not survive"
1 failed, 1 passed, 4 deselected
```

### c. MUT-B — count every mechanism entry, attributed or not

`len({… if e["spec_elements"]})` → `len({… for e in paper_mechanisms})`

```
E   AssertionError: assert 5 == 2   test_unattributed_citation_does_not_count_as_a_mechanism_paper
E   AssertionError: assert 5 == 1   test_single_mechanism_spec_claiming_five_papers_is_labelled_not_hidden
2 failed, 4 deselected
```

That is the exact laundering the issue describes — five papers reported as five mechanisms — and it is now a red test.

### d. MUT-C — revert the `_make_entry` copy (the issue asks for this one by name)

Restored `source_papers=[{"arxiv_id": a, "title": ""} …]` and removed the three field copies.

```
E   AssertionError: assert 0 == 30
    assert entry.papers_offered == 30
1 failed, 5 deselected
```

### e. MUT-D — persist the bare transcript (drop the attribution record)

`transcript=_transcript_with_paper_record(c)` → `transcript=c.debate_transcript`

```
E   assert 0 == 1
    "exactly one attribution record per row"
1 failed, 5 deselected
```

### f. MUT-E — read path drops the enrichment

`_resolve_source_papers`' `entry = dict(paper)` → a fixed `{arxiv_id, title}` projection.

```
assert [p["mechanism"] for p in resolved] == ["200-day trend filter", "carry"]
1 failed, 5 deselected
```

## Review follow-up (2nd commit `b70da92`)

Three things came out of review. All three are in the second commit; the first commit is unchanged.

### 1. BLOCKER — unsanitized model prose reached the durable transcript

`_transcript_with_paper_record` scrubbed `fusion_reasoning` at the call site but wrote `debate_paper_verdicts` **verbatim**. Those rows carry `discard_reasons`, which `_aggregate_paper_verdicts` copies straight out of the **raw** `_debate_round` turns — so the identical bear sentence that gets scrubbed on its way in as `discard[].reason` landed unscrubbed under the new `paper_verdicts` key. `sanitize_transcript` only knew `verdict` / `claims` / `discard`.

That breaks the contract `debate_transcript.py:26-30` states outright ("a raw read of `transcript_json` is already safe"), and it is exactly the regression class the `_sanitize_claim` comment at `:82-89` names: a **shape change** re-opening the leak the write-time sanitizer exists to close. It is served to owners by `GET /strategies/{id}/debate`, which has no read-time net, by design.

Fixed where the contract lives — in `sanitize_transcript`, not at the one call site, so any future writer of that shape is covered too:

* `paper_verdicts` rows get their `discard_reasons` scrubbed (`_sanitize_paper_verdict`).
* `fusion_reasoning` is scrubbed there rather than at the attach point.
* `title` is deliberately **not** scrubbed: it is corpus metadata, not model prose, and `Phase-\w*` would redact a real paper ("Phase-Locked Cycles …" → "[redacted] Cycles …"). Same provenance reason validated `arxiv_ids` are left alone in `_SCRUBBED_CLAIM_KEYS`.

New guard: **`test_persisted_paper_verdicts_do_not_leak_internal_jargon`** — the same jargon sentence on both paths, asserted against the RAW `transcript_json` string.

**MUT-F — the pre-fix behaviour** (`sanitize_transcript` does not know `paper_verdicts`):

```
E   AssertionError: a raw read of transcript_json must already be safe (module docstring)
E   assert 'T3.5' not in '[{"claims":...y trades."}]'
1 failed, 6 deselected
```

**MUT-I — over-correct the other way** (scrub every string in the row, titles included):

```
E   AssertionError: assert '[redacted] C...odity Futures' == 'Phase-Locked...odity Futures'
1 failed, 1 passed, 5 deselected
```

### 2. The link never reached the surface #1739 named (reviewer's call)

`GET /strategies/{id}` serves the passport — `strategy_passports.paper_refs` — not `StrategyRecord.source_papers`, and the debate panel carrying the attribution summary is **owner-only**, so a public passport never showed it. All three passport constructors built `PaperRef(arxiv_id=…, title=…)` and nothing else, so `contribution` (the column `StrategyPassport.jsx` renders as "per-paper contribution") had **no writer at all** and every generated row was an em-dash.

One `_passport_paper_refs(c)` helper now feeds all three sites. It writes the named mechanism into `contribution` **only when `spec_elements` survived validation**, and leaves it `None` otherwise — the em-dash *is* the unattributed record; writing unvalidated prose there would launder the same claim one column over. The column already exists (`PassportPaperRef.contribution`), `_build_paper_refs` already passes it through, and `PaperRefResponse` already serves it: **no schema change, no migration, no UI change**. Titles now reach the passport too (they used to be `""`).

`test_persisted_source_papers_carry_title_and_mechanism` was extended past `StrategyRecord` to assert the passport row, per the review.

**MUT-G — the old constructor** (`contribution=None`):

```
E   AssertionError: assert None == '200-day trend filter'
1 failed, 6 deselected
```

**MUT-H — write the mechanism regardless of whether `spec_elements` survived**:

```
E   AssertionError: assert 'carry' is None
1 failed, 6 deselected
```

### 3. The `spec_elements` guard was not falsifiable

`validate_strategy_spec` rebuilds `indicators` from the entry/exit conditions (`indicators=sorted(all_indicators)`) and **ignores** the `indicators` list the model declares. But the test fixture declared exactly `["sma_200"]`, so validating the *declared* list instead of the *derived* one passed the test anyway — the guard could not tell the two apart, which is the same self-report problem this issue is about, one field over.

The fixture now declares a `bollinger_20` it never trades. Checking the model's own list is now red:

**MUT-J — `valid_elements = set((strategy_spec or {}).get("indicators") or [])`**:

```
E   assert ['bollinger_20'] == []
    "an alias the spec never uses must not survive"
1 failed, 6 deselected
```

Every earlier mutation (MUT-A→E) and the fails-on-`main` run were re-run against the final tree: 7 failed on `436cc1a0` (`AttributeError` / `TypeError` — absence), and each mutation still turns its guard red.

## Reviewer notes

- **`distinct_mechanism_papers` is not a gate, deliberately.** `test_single_mechanism_spec_claiming_five_papers_is_labelled_not_hidden` asserts `is_actionable is True` *and* `distinct_mechanism_papers == 1`. An honest single-mechanism strategy ships with an honest count; turning this into a reject would fail generation instead of improving it, which #1636 ruled out.
- **One judgement call worth naming:** the tally rides `transcript_json` as an extra list entry with `role: "attribution"` and a plain-sentence `verdict`. `StrategyReasoning.jsx` maps every entry to a card, so a keyless entry would render as a blank one; giving it a real summary sentence means the existing renderer shows something honest with **zero UI changes**. Both of its model-prose keys are scrubbed inside `sanitize_transcript` itself — see the review follow-up below, which is where the first version of this got it wrong.
- **Residual the spec does not cover** (flagging, not fixing — out of scope here): a model that names the *same* alias for all five papers would still score `distinct_mechanism_papers == 5`. The check as specified validates that each claimed element is real, not that the five claims are distinct from each other. Worth its own issue if it shows up in live output.

### DIFF
--- a/backend/archimedes/agents/debate_engine.py
+++ b/backend/archimedes/agents/debate_engine.py
@@ -797,14 +797,41 @@ def _rigor_verdict_dict(ev: Any) -> dict[str, Any]:
     }
 
 
-def _make_entry(candidate_id: str, proposal: Any, ev: Any, *, regime: str) -> _CandidateResult:
+def _make_entry(
+    candidate_id: str,
+    proposal: Any,
+    ev: Any,
+    *,
+    regime: str,
+    evidence_by_id: dict[str, dict[str, str]] | None = None,
+) -> _CandidateResult:
     """One leaderboard entry — a fully-populated ``_CandidateResult``.
 
     ``has_real_rigor=True`` (carries C-rigor's real backtest) so the downstream
     ``_patch_pbo`` and buy-and-hold gather correctly SKIP it (keyed on
     ``has_real_rigor``), preserving the CSCV PBO from ``evaluate_fusion_spec``.
+
+    ``evidence_by_id`` (#1739) is the ``{arxiv_id: {"title", "published"}}`` map
+    ``_propose_pool`` already built — the only place in the run that knows a
+    cited paper's TITLE. Without it every ``source_papers`` entry shipped
+    ``"title": ""`` and the Library card had nothing but an id to print.
+    Defaults to ``None`` (→ blank titles, the old behaviour) so the pure
+    ranking tests can call ``build_leaderboard`` without an evidence map.
     """
     spec = proposal.strategy_spec or {}
+    # (#1739) The server-filtered paper→mechanism map, keyed for the join
+    # below. First entry per id wins; a cited paper with no entry is carried
+    # with an empty mechanism, which is the honest "cited but unattributed"
+    # record (never dropped — dropping it would hide the shortfall).
+    mechanisms = getattr(proposal, "paper_mechanisms", None) or []
+    mech_by_id: dict[str, dict[str, Any]] = {}
+    for entry in mechanisms:
+        if not isinstance(entry, dict):
+            continue
+        arxiv_id = str(entry.get("arxiv_id", "") or "")
+        if arxiv_id and arxiv_id not in mech_by_id:
+            mech_by_id[arxiv_id] = entry
+    titles = evidence_by_id or {}
     # Real per-bar returns for the live gate (#788/#818, mirrors the fusion
     # path): without return_series, _persist_real_returns SKIPS the winner and
     # the passport reads "pending" forever even though C-rigor just ran a real
@@ -816,7 +843,15 @@ def _make_entry(candidate_id: str, proposal: Any, ev: Any, *, regime: str) -> _C
         strategy_name=proposal.strategy_name or "Debate candidate",
         thesis=proposal.thesis,
         asset_universe=list(spec.get("asset_universe", []) or []),
-        source_papers=[{"arxiv_id": a, "title": ""} for a in proposal.source_arxiv_ids],
+        source_papers=[
+            {
+                "arxiv_id": a,
+                "title": (titles.get(a) or {}).get("title", ""),
+                "mechanism": str((mech_by_id.get(a) or {}).get("mechanism", "") or ""),
+                "spec_elements": list((mech_by_id.get(a) or {}).get("spec_elements", []) or []),
+            }
+            for a in proposal.source_arxiv_ids
+        ],
         weights={},  # debate emits a DSL spec, not a static weight vector
         reasoning=proposal.fusion_reasoning or proposal.novelty_rationale or "",
         rigor_verdict=_rigor_verdict_dict(ev),
@@ -836,6 +871,11 @@ def _make_entry(candidate_id: str, proposal: Any, ev: Any, *, regime: str) -> _C
         # live agent runner later loads it to evaluate a deployed generated
         # strategy under its own signal rule instead of silently skipping it.
         strategy_spec=spec or None,
+        # (#1739) The budget-vs-used pair and the per-paper prose, carried onto
+        # the candidate instead of ending at a log line / a write-only field.
+        papers_offered=int(getattr(proposal, "papers_offered", 0) or 0),
+        distinct_mechanism_papers=int(getattr(proposal, "distinct_mechanism_papers", 0) or 0),
+        fusion_reasoning=str(getattr(proposal, "fusion_reasoning", "") or ""),
     )
 
 
@@ -935,6 +975,7 @@ def build_leaderboard(
     base_id: str,
     regime_force_abstain: bool = False,
     regime_reason: str = "",
+    evidence_by_id: dict[str, dict[str, str]] | None = None,
 ) -> list[_CandidateResult]:
     """Deterministic C-regime gate → C-null cull + rank → top-N leaderboard.
 
@@ -943,6 +984,11 @@ def build_leaderboard(
     market regime structurally overrides candidate consensus (Hierarchy-of-Truth).
     Otherwise: C-null cull → rank → leaderboard (leader keeps ``base_id``,
     alternatives get ``base_id_alt{n}``). Pure + deterministic — directly tested.
+
+    ``evidence_by_id`` (#1739) is threaded through to ``_make_entry`` purely so
+    a cited paper's TITLE reaches ``source_papers``; it never influences
+    ranking, culling, or the abstain paths. Optional, so the pure ranking tests
+    keep calling this with rigor results alone.
     """
     if regime_force_abstain:
         return [
@@ -966,7 +1012,13 @@ def build_leaderboard(
     # candidate set can't balloon when the pool is large.
     survivors = survivors[: _leaderboard_max()]
     return [
-        _make_entry(base_id if i == 0 else f"{base_id}_alt{i}", p, ev, regime=regime)
+        _make_entry(
+            base_id if i == 0 else f"{base_id}_alt{i}",
+            p,
+            ev,
+            regime=regime,
+            evidence_by_id=evidence_by_id,
+        )
         for i, (p, ev) in enumerate(survivors)
     ]
 
@@ -1102,6 +1154,7 @@ async def _run_debate_leaderboard(
         base_id=candidate_id,
         regime_force_abstain=regime_gate["force_abstain"],
         regime_reason=regime_gate["reason"],
+        evidence_by_id=evidence_by_id,
     )
     # Stamp the ONE transcript this run produced onto every entry (winner AND
     # the tail Considered Alternatives) — build_leaderboard is pure and knows
--- a/backend/archimedes/agents/generation_pipeline.py
+++ b/backend/archimedes/agents/generation_pipeline.py
@@ -1022,6 +1022,40 @@ async def _persist_generation_cost(
         logger.warning("cost record: durable persist failed for job %s (non-blocking)", job_id, exc_info=True)
 
 
+def _passport_paper_refs(c: _CandidateResult) -> list[Any]:
+    """The passport's paper rows, with the attributed mechanism in ``contribution`` (#1739).
+
+    ``_persist_candidate`` and its two sibling passport rebuilds all built
+    ``PaperRef(arxiv_id=…, title=…)`` and nothing else, so ``contribution`` —
+    the column ``StrategyPassport.jsx`` renders as "per-paper contribution" —
+    had no writer at all and every generated row showed an em-dash. The
+    passport is what ``GET /strategies/{id}`` serves (``paper_refs``, not
+    ``StrategyRecord.source_papers``), and unlike the owner-only debate panel it
+    is public, so this is the only path on which the paper→mechanism link
+    reaches a reader of a published strategy.
+
+    ``contribution`` stays **None** for a cited paper whose ``spec_elements``
+    did not survive validation. The em-dash IS the unattributed record: writing
+    the model's unverified mechanism prose there would reinstate exactly the
+    laundering this issue removes, one column to the right.
+    """
+    from archimedes.models.paper_ref import PaperRef
+
+    refs: list[Any] = []
+    for p in c.source_papers or []:
+        if not isinstance(p, dict):
+            continue
+        mechanism = str(p.get("mechanism", "") or "").strip()
+        refs.append(
+            PaperRef(
+                arxiv_id=p.get("arxiv_id"),
+                title=p.get("title", ""),
+                contribution=mechanism if (mechanism and p.get("spec_elements")) else None,
+            )
+        )
+    return refs
+
+
 def _transcript_with_paper_record(c: _CandidateResult) -> list[dict[str, Any]]:
     """The candidate's transcript plus one trailing paper-attribution entry (#1739).
 
@@ -1037,12 +1071,15 @@ def _transcript_with_paper_record(c: _CandidateResult) -> list[dict[str, Any]]:
     turns sees an honest summary line rather than a blank card, and carries the
     machine-readable tally alongside it.
 
-    The prose is scrubbed here because ``sanitize_transcript`` only knows the
-    turn keys it was written for — writing unsanitized model text into this
-    table would re-open the leak its write-time sanitizer exists to close.
+    Both new keys carry MODEL prose — ``fusion_reasoning`` directly, and each
+    ``paper_verdicts`` row's ``discard_reasons``, which
+    ``_aggregate_paper_verdicts`` copies out of the raw debate turns. They are
+    scrubbed by ``sanitize_transcript``, which ``record_debate_transcript``
+    applies to everything written to this table; teaching the sanitizer the new
+    shape (rather than scrubbing at this one call site) is what keeps the
+    table's stated contract true — a raw read of ``transcript_json`` is already
+    safe, whoever wrote the row.
     """
-    from archimedes.models.debate_transcript import sanitize_debate_text
-
     transcript = list(c.debate_transcript or [])
     verdicts = c.debate_paper_verdicts or []
     reasoning = (c.fusion_reasoning or "").strip()
@@ -1061,7 +1098,7 @@ def _transcript_with_paper_record(c: _CandidateResult) -> list[dict[str, Any]]:
             "round": None,
             "verdict": summary,
             "paper_verdicts": verdicts,
-            "fusion_reasoning": sanitize_debate_text(reasoning),
+            "fusion_reasoning": reasoning,
         },
     ]
 
@@ -1795,13 +1832,10 @@ async def _persist_candidate(
 
             # Also write to the unified strategy_passports table (Issue #160)
             try:
-                from archimedes.models.paper_ref import PaperRef
                 from archimedes.models.strategy import StrategyPassport, StrategyStatus
                 from archimedes.services.passport_loader import ingest_passport
 
-                papers = [
-                    PaperRef(arxiv_id=p.get("arxiv_id"), title=p.get("title", "")) for p in (c.source_papers or [])
-                ]
+                papers = _passport_paper_refs(c)
                 # Map candidate regime to passport regime_tag
                 _regime_tag_map = {"bull": "bull", "bear": "bear"}
                 _regime_tag = _regime_tag_map.get(c.regime, "regime_neutral")
@@ -1915,14 +1949,13 @@ def _refresh_passport_real_metrics(
     """
     from datetime import date as _date
 
-    from archimedes.models.paper_ref import PaperRef
     from archimedes.models.strategy import StrategyPassport, StrategyStatus
     from archimedes.models.strategy_store import StrategyRecord
     from archimedes.services.passport_loader import ingest_passport
 
     record = session.query(StrategyRecord).filter_by(id=strategy_id).first()
     status_val = StrategyStatus(record.status) if record and record.status else StrategyStatus.CANDIDATE
-    papers = [PaperRef(arxiv_id=p.get("arxiv_id"), title=p.get("title", "")) for p in (c.source_papers or [])]
+    papers = _passport_paper_refs(c)
     regime_tag = {"bull": "bull", "bear": "bear"}.get(c.regime, "regime_neutral")
     passport = StrategyPassport(
         id=strategy_id,
@@ -2180,7 +2213,6 @@ async def _backtest_and_persist(c: _CandidateResult, strategy_id: str, emit: _Em
         from datetime import date as _date
 
         from archimedes.db import get_session
-        from archimedes.models.paper_ref import PaperRef
         from archimedes.models.strategy import StrategyPassport, StrategyStatus
         from archimedes.models.strategy_store import StrategyRecord
         from archimedes.services.backtest_mapper import canonical_artifact_hash
@@ -2266,7 +2298,7 @@ async def _backtest_and_persist(c: _CandidateResult, strategy_id: str, emit: _Em
             #    in-place update.
             record = session.query(StrategyRecord).filter_by(id=strategy_id).first()
             status_val = StrategyStatus(record.status) if record and record.status else StrategyStatus.CANDIDATE
-            papers = [PaperRef(arxiv_id=p.get("arxiv_id"), title=p.get("title", "")) for p in (c.source_papers or [])]
+            papers = _passport_paper_refs(c)
             _regime_tag_map = {"bull": "bull", "bear": "bear"}
             _regime_tag = _regime_tag_map.get(c.regime, "regime_neutral")
             passport = StrategyPassport(
--- a/backend/archimedes/agents/generation_pipeline.py
+++ b/backend/archimedes/agents/generation_pipeline.py
@@ -436,10 +436,24 @@ class _CandidateResult:
     # verdict}]`` over every paper that entered a proposer prompt. Computed by
     # ``debate_engine._aggregate_paper_verdicts`` (0 tokens) and stamped onto
     # every entry, same as the transcript. ``None`` on every non-debate path.
-    # NOT persisted: the transcript table's column holds the turn list and this
-    # issue explicitly carries no migration, so today this is an in-process +
-    # SSE-visible record only. Durable storage is follow-up work.
+    # (#1739) Now DURABLE: ``_persist_debate_transcripts`` appends it to the
+    # same ``debate_transcripts.transcript_json`` list the turns ride, so no
+    # migration was needed to stop it dying with the request.
     debate_paper_verdicts: list[dict[str, Any]] | None = None
+    # (#1739) The budget-vs-used pair, carried off the proposal instead of
+    # being left in a log line. ``papers_offered`` is how many papers were put
+    # in front of the proposer; ``distinct_mechanism_papers`` is how many of
+    # the CITED ones name a mechanism tied to an indicator the spec actually
+    # trades. Together they make ``len(source_arxiv_ids)`` readable: "2 of 30
+    # offered, 2 attributed" is a different claim from "5 cited, 1 attributed".
+    # Both 0 on every non-fusion path (fixture / buy-and-hold), which never saw
+    # a paper — 0 there means "no papers were offered", not "attribution failed".
+    papers_offered: int = 0
+    distinct_mechanism_papers: int = 0
+    # (#1739) The proposal's per-paper mechanism prose. ``reasoning`` above
+    # already falls back to ``novelty_rationale``, so it cannot be read as "the
+    # model named each paper's contribution"; this field is that text or "".
+    fusion_reasoning: str = ""
 
 
 def _is_deployable(c: _CandidateResult) -> bool:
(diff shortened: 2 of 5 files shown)


backend/archimedes/agents/debate_engine.py
```diff
@@ -816,7 +843,15 @@ def _make_entry(candidate_id: str, proposal: Any, ev: Any, *, regime: str) -> _C
         strategy_name=proposal.strategy_name or "Debate candidate",
         thesis=proposal.thesis,
         asset_universe=list(spec.get("asset_universe", []) or []),
-        source_papers=[{"arxiv_id": a, "title": ""} for a in proposal.source_arxiv_ids],
+        source_papers=[
+            {
+                "arxiv_id": a,
+                "title": (titles.get(a) or {}).get("title", ""),
+                "mechanism": str((mech_by_id.get(a) or {}).get("mechanism", "") or ""),
```
danielscoffee, line 850: This key reaches `StrategyRecord.source_papers` and survives `_resolve_source_papers` (`entry = dict(paper)`). It does not reach the passport.

`PaperRef` already has `contribution` for exactly this sentence, and `_persist_candidate` constructs `PaperRef(arxiv_id=..., title=...)` with nothing else. Dual-write `contribution` here (or at those three constructors) when `spec_elements` is non-empty. Empty spec_elements should stay unset so the table's em-dash is the honest unattributed row, not a blank string that looks like a missing title.

danielscoffee (commented):
The mapping itself looks right. Labels, does not gate. Conflict marker is gone. `spec_elements` is checked against the validated alias set. Unattributed citations stay on the record. The six tests cover the spec they were written against, including the adversarial 5-cite / 1-mechanism case.

The hole is the surface #1739 actually named.

`GET /strategies/{id}` serves the passport from `strategy_passports.paper_refs`, not from `StrategyRecord.source_papers`. `_persist_candidate` (and the two sibling ingest sites around 1925 and 2269) still builds:

```python
PaperRef(arxiv_id=p.get("arxiv_id"), title=p.get("title", ""))
```

`PaperRef.contribution` already exists for "what this paper contributed to the synthesis," and that is the column `StrategyPassport.jsx` renders. It stays an em-dash. The table header is still `Fused from ${rows.length} papers`. So a 5-cite / 1-mechanism strategy still reads as a 5-paper fusion on the passport. The honest sentence only lands on the debate panel, as a card tagged `attribution`.

Titles flowing into `PaperRef` *is* a real win (those used to ship as `""`). Don't block on the header copy; that is a UI follow-up. The contribution dual-write is the same zero-UI trick this PR already used for `transcript_json`: pass the named mechanism into `contribution` when `spec_elements` survived, leave it unset when they did not, so the em-dash is the unattributed record rather than invented text. Keep the `mechanism` key on the JSON if you want it.

`test_persisted_source_papers_carry_title_and_mechanism` currently stops at `StrategyRecord.source_papers` + `_resolve_source_papers`. The passport path is the one a reader sees. Worth asserting `paper_refs[].contribution` (and title) after ingest.

Residual you already named (same alias on five papers still counts as 5) is fine as its own issue. Not a gate, not this PR.

danielscoffee:
One more on the user-visible half, since it changes how much the debate-card piggyback actually covers.

`GET /api/strategies/{id}/debate` is owner-only. Published is not enough. So the `attribution` sentence (`N of M cited papers name a mechanism this strategy trades`) never appears on a public passport. A visitor still gets `Fused from ${papers.length} papers` / `N fused papers` from `papers[]`, which is still one `PaperRef` per cited id.

That is why filling `PaperRef.contribution` at persist (the three constructors) is the zero-UI path that actually reaches the surface #1636 named. The heading/tag treating `papers.length` as fusion depth is the follow-up; the debate card cannot substitute for it.

backend/archimedes/agents/debate_engine.py
dbrowneup: Done in `b70da92` — you were right that the debate card cannot substitute for the passport.

One helper, `_passport_paper_refs(c)`, now feeds all three constructors (`_persist_candidate` ~1803, the returns refresh ~1925, the portfolio-backtester rebuild ~2269), so there is one place to change rather than three that drift.

```python
contribution=mechanism if (mechanism and p.get("spec_elements")) else None,
```

Empty `spec_elements` → `contribution` stays **None**, so the table renders the em-dash as the honest unattributed row (and `blankColumns` keeps the footnote accurate for a run where nothing was attributable). Titles reach the passport now too — they used to ship as `""`.

Two mutations in the PR body:

- **MUT-G** (`contribution=None`, the old constructor): `assert None == '200-day trend filter'`
- **MUT-H** (write the mechanism regardless of whether `spec_elements` survived): `assert 'carry' is None`

`test_persisted_source_papers_carry_title_and_mechanism` now asserts past `StrategyRecord.source_papers` into `strategy_passports.paper_refs`, as you asked. Header copy left alone — agreed it is a UI follow-up.

dbrowneup:
Second commit `b70da92` pushed. Three things, all from review — the first one is a defect the first commit introduced, not a polish item.

**1. BLOCKER, found in verification: unsanitized model prose reached the durable transcript.**

`_transcript_with_paper_record` scrubbed `fusion_reasoning` at the call site but wrote `debate_paper_verdicts` verbatim, and those rows carry `discard_reasons` copied by `_aggregate_paper_verdicts` straight out of the **raw** `_debate_round` turns. So the identical bear sentence that gets scrubbed on the way in as `discard[].reason` landed unscrubbed under the new `paper_verdicts` key:

```
turn discard[].reason            -> "rejected: blocked on the [redacted] [redacted] until [redacted] lands"
paper_verdicts[].discard_reasons -> ["rejected: blocked on the T3.5 cutover until Phase-4 lands"]
```

That breaks what `debate_transcript.py:26-30` states outright ("a raw read of `transcript_json` is already safe") and it is the exact class `_sanitize_claim:82-89` warns about — a shape change re-opening the leak the write-time sanitizer exists to close. Before this diff `debate_paper_verdicts` never touched the DB, so the diff introduced the path, and `GET /strategies/{id}/debate` serves it with no read-time net by design.

Fixed in the sanitizer rather than at the call site, so any future writer of that shape is covered: `sanitize_transcript` now knows `paper_verdicts` (scrubbing each row's `discard_reasons`) and `fusion_reasoning`. `title` is deliberately **not** scrubbed — corpus metadata, not model prose, and `Phase-\w*` would redact a real paper ("Phase-Locked Cycles…" → "[redacted] Cycles…"); same provenance reason validated `arxiv_ids` are left alone. New guard `test_persisted_paper_verdicts_do_not_leak_internal_jargon` asserts against the raw `transcript_json` string, with **MUT-F** (the pre-fix state) and **MUT-I** (scrub everything, titles included) both red in the PR body.

**2. `PaperRef.contribution` dual-write** — your call, replied on the thread. One `_passport_paper_refs(c)` helper feeds all three constructors; `contribution` stays `None` when `spec_elements` did not survive, so the em-dash is the unattributed record. MUT-G / MUT-H in the body.

**3. The `spec_elements` guard was not falsifiable.** `validate_strategy_spec` rebuilds `indicators` from entry/exit and ignores the list the model declares — but the fixture declared exactly `["sma_200"]`, so validating the *declared* list instead of the *derived* one passed the test anyway. The fixture now declares a `bollinger_20` it never trades; **MUT-J** (check the model's own list) is red.

All ten mutations were re-run against the final tree — every one turns a guard red. Suite: `5632 passed, 4 skipped, 2 deselected in 545.64s`, exit 0. Collect baseline 84 → 91. Anti-goals re-checked on the full branch diff: no vocabulary hits, no migration, no new `raise`/`DebateUnavailable`, the four paper constants and `_HASH_IGNORED_SPEC_KEYS` byte-identical, `paper_rag.py` / `generate_schemas.py` / `analytics-engine/` untouched.

Leaving both threads open for you to close once you have re-verified.

danielscoffee (approved):
Re-checked `b70da92`.

Passport path is fixed: `_passport_paper_refs` feeds all three constructors, `contribution` is the named mechanism only when `spec_elements` survived, and `None` keeps the em-dash as the honest unattributed row. The persistence test now asserts `strategy_passports.paper_refs`, which is what a reader actually sees.

Sanitizer fix is in the right place — `sanitize_transcript` knows `paper_verdicts[].discard_reasons` and `fusion_reasoning`, titles left alone on purpose. That restores the write-time contract the module docstring states.

`bollinger_20` decoy makes the `spec_elements` guard falsifiable against the derived indicator set. Header copy stays a UI follow-up; residual same-alias-on-five-papers is fine as its own issue.

Thread resolved. LGTM.

Issue #1746, opened 2026-09-01
[product-truth][P0] Gate endpoints contradict on the #1 leaderboard strategy


dbrowneup:
Dogfood 2026-09-01 (agent-mcp lane), reproduced 3x: `GET /api/strategies/1f9cfe96...` → `pass` / `passes_rigor_gate: true` / Sharpe 0.406, while `GET /api/strategies/passports/1f9cfe96...` → `candidate` / `false` / Sharpe null. Only the flagship pass case disagrees; fail/pending strategies agree across both. Secondary: Sharpe drifted between reads 37s apart — a live recompute presented as a persisted number. This is a claims-must-be-true violation on the exact number that gates everything downstream. Investigate which surface is authoritative, make the other read from it, and add a parity guard test that fails when the two endpoints disagree on status/passes/sharpe for the same id.

dbrowneup:
Diagnosis (mapped by a 4-lane read of the code + synthesis + 3 adversarial passes, all re-derived against `main` @ `0570f16e`). Structural, not a race.

**`1f9cfe96…` is a CURATED strategy** — re-derived by running the real id function over the corpus: it is `analytics-engine/strategies/harvey_2018_volatility_targeting.py` (`_strategy_id`, `services/strategy_provider.py` ~L135; 32-hex curated id vs 16-hex `content_hash[:16]` for generated rows). The two endpoints read two different stores:

- `GET /api/strategies/{id}` takes the **curated branch** (`strategy_provider().get_strategy()` hit → `_to_strategy_response`), which serves `passes_rigor_gate` / `rigor_gate_status` from a **live** `run_rigor_gate` run and promotes the file's `STATUS = "candidate"` to `validated` on a live pass (read-side only).
- `GET /api/strategies/passports/{id}` is a **pure read of the `strategy_passports` row** (`record.to_dict()`), and for a curated row that row is written by exactly one path — `_sync_to_unified_table` → `ingest_passport(generation_method="curated", force_update=True)` — with `passes_rigor_gate` **hardcoded `False`** (`strategy_provider.py` ~L367, the #821 fail-closed placeholder; its comment says "the served path replaces it with the live verdict", which is true of the detail route only), `status` = the file constant, `sharpe_ratio` = the #1187 fixture (harvey has no fixture row → `null`). `to_dict()` has no `rigor_gate_status` key at all.

[...]

PR #1761, opened 2026-09-01
[product-truth] Library badge and rigor numbers come from the live gate, never the generation-time verdict (#1747)

### LINKED ISSUE #1747: [product-truth][P0] Library shows Live ✓ on strategies whose passports say gate failed
Dogfood 2026-09-01 (human-explore): all 21 Generated rows show a green `Live ✓` badge; every passport opened says "Reference only — gate failed"; the leaderboard itself says 0 of 19 clear the bar. No passing example exists to calibrate against. The badge must derive from the same gate verdict the passport shows (fail-closed), and a UI guard test should pin badge==verdict. Repro: Library → Generated tab → any row → Open Passport.

dbrowneup:
Closes #1747.

## Root cause

The Library's Generated tab read a different verdict store from every other surface, and **both halves of its green badge came from the same generation-time blob**, which is why the tab's own honesty guard could never fire.

* `backend/archimedes/models/strategy_store.py:316-326` (and `:358-360` on insert) writes `status = "live" if rigor_verdict["passing"] else "rejected"` — one write, from the **fusion/synthesis** verdict. Nothing re-derives either field after a backtest.
* `backend/archimedes/api/strategies_routes.py:820-823` (`list_generated_strategies`) returned `r.to_dict()` — no live gate, no `strategy_passports` join.
* `ui/src/components/Strategies.jsx:750` `const verdict = row.rigor_verdict || null` → `:786` `passes_rigor_gate: verdict ? Boolean(verdict.passing) : null`.
* The pill demoted only on `status === 'live' && passesRigor === false` (`Strategies.jsx:120`, `:128`). On this tab both inputs were the *same fact*, so `live` + `false` was unreachable and 21 rows whose own passport says "Reference only — gate failed" rendered **Live ✓**.

Meanwhile `StrategyPassport.jsx` fetches `/api/strategies/{id}`, which for a generated id serves the passport row (`strategies_routes.py:1441`), and the leaderboard counts the same column — hence "0 of 19" on the board next to 21 green rows in the Library.

## What changed

**Backend — `backend/archimedes/api/strategies_routes.py`**

* New `_live_generated_rigor(session, ids)`: the **batched form of `selection_bias_routes._generated_strategy_rigor`** (`:676`). Same `num_trials` PROVENANCE rule (`_num_trials_for_generated_row`), same persisted per-row `library_pbo` with `pbo_library_size` deliberately unset, same look-ahead triple from `verdict_from_persisted_row`, `strictness_level` left at its default (the strictest = the badge bar). Diverging on any of those would let the Library badge and the Deploy ladder disagree about the same strategy.
* `list_generated_strategies` overlays, after the page is built and inside the same session: `rigor_gate_status` (four-state) and `passes_rigor_gate` **coupled to it** (`status == "pass"`), plus `deflated_sharpe_ratio` / `dsr_p_value` / `pbo_score` / `out_of_sample_sharpe` from **that same gate run** and `sharpe_ratio` from **the same `backtest_results` row that gate run read its returns from** (`_live_generated_rigor` returns it; `latest_backtests_by_strategy` and `get_all_daily_returns` rank with the identical `created_at DESC, id DESC` window, so it is one physical row) — all five served `None` unless the run reached `pass`/`fail`. No passport row, or fewer persisted returns than the gate accepts → `passes_rigor_gate: null`, `rigor_gate_status: "pending"`.
* **Not** the stored `strategy_passports.passes_rigor_gate`. That column is mixed-vintage: its first writer is the generation-time fusion verdict (`generation_pipeline._persist_candidate`), the live re-grade only replaces it when the post-backtest refresh reaches its write, and there is no provenance column to tell the two apart. Reading it is reading a claim, whichever value it holds. Demo 1 below is the mutation that proves this is not what the code does.
* `status` and `rigor_verdict` stay on the wire **unchanged** — the fix is a read-side overlay, not a rewrite of what the synthesis gate thought.

**UI**

* New `ui/src/libraryStatus.js` — `statusTag` / `statusLabel` / `isGatePass`, moved out of `Strategies.jsx` so a test can **run** them (`.jsx` is not importable under `node --test`; that is exactly how this rule went unpinned through #1747). Four states, four sentences: `pass` → green "Live"; `fail` → muted "Reference only — gate failed"; `pending`/absent → muted **"Not yet graded"**; `degenerate` → muted "Unevaluable — flat returns".
* Deliberately **not** the `!== true` widening. That is the smaller diff and it puts "Reference only — gate failed" on the same row as the clock icon titled "Not yet evaluated — no backtest data for the rigor gate to score" — one row asserting two contradictory things, the failing one about a verdict nothing produced (#1358's defect class). The `pending` arm gets its own label instead; only the pill *class* stops being green for a row with no affirmative pass.
* `StrategyPassport.jsx` is **not** refactored onto the new module. Its `statusTag`/`statusLabel` are near-duplicates, not duplicates — it maps `validated` to green (the Library maps it to accent), maps `rejected`/`retired` to muted, defaults unknown statuses to accent, and has no `pending_backtest` arm. Merging them would change the passport's rendering as a side effect of a Library fix. What must not diverge is the *sentence*, and the test pins that by reading the passport's source.
* `Strategies.jsx` `coerceGenerated`: `passes_rigor_gate` is taken **only** as a literal server boolean (else `null`), `rigor_gate_status` is carried through, and DSR/PBO/OOS/`dsr_p`/Sharpe now come from the server fields instead of `row.rigor_verdict` — otherwise the row would print the *other* gate's passing DSR beside a "gate failed" pill. `verdict` survives for exactly one thing: the `hasRealMetrics` probe that relabels `rejected` → `pending_backtest`.
* Both call sites take the four-state — the desktop table pill (`Strategies.jsx:282`) and the **mobile lib-card** (`:578`), which previously could render a bare green "Live" with no rigor icon beside it at all. A test asserts there are exactly two three-argument call sites and zero two-argument ones.

## Owner decisions I defaulted (all reversible, all yours)

1. **Ungraded label wording.** "Not yet graded". It is a placeholder, exported as `NOT_GRADED_LABEL` in one place. What it must not go back to being is "Live" (a claim) or "gate failed" (a different claim).
2. **Rows moving to the collapsed section.** `Strategies.jsx:1119-1121` partitions on `passes_rigor_gate === true / === false / == null`. Rows that flip to `false` drop into the collapsed **Rejected** section; rows that flip to `null` stay in the main table under "pending". That is honest, and it is also the Library visibly emptying out. Ships as-is unless you want the section renamed/expanded first.
3. **No write-back to `StrategyRecord.status`.** The alternative is writing the live re-grade back at `generation_pipeline.py:2041/:2232` so the record stops lying at rest. That destroys the record of what the synthesis gate thought and has readers outside this route (`marketplace_service.trending`). Left as a product/architecture call.
4. **Leaderboard's 2-state badge left for a follow-up.** `Leaderboard.jsx:190-198` has only pass/fail/no-backtest — a row the passport calls "unevaluable" is asserted there as "Gate failed". Same claims-must-be-true class, different surface, out of this diff.
5. **Curated Examples tab is touched too, on purpose.** A curated row with `STATUS = "live"` (5 of the 34 files) whose gate returns `pending` used to render a green "Live"; it now reads "Not yet graded". Same defect, same fix, one code path. Curated `candidate` / `validated` / `pending_backtest` rendering is unchanged — a test pins that.

## Review round 2 — commit `aecce416`

An independent verifier reproduced every claim in this PR and returned FIX-FIRST on four in-scope defects. All four are fixed in `aecce416`; the badge mechanism itself is unchanged.

**1. The display Sharpe did not come from the run the gate graded — and the comment said it did.** The gate reads its returns from the latest `backtest_results` row; the route was serving `strategy_passports.sharpe_ratio`, a denormalised snapshot with **no foreign key to `backtest_results`** (~155 backtest rows per strategy, per `scripts/archive_backtest_results.py`). Measured on a seeded row before the fix:

```
served sharpe_ratio  : 1.4      (strategy_passports column)
graded row's sharpe  : 0.031    (backtest_results — the row the gate actually read)
served DSR / OOS     : -2.606803 / -2.893566
```

A Sharpe of 1.4 printed beside a DSR of −2.61 computed from a series whose Sharpe is 0.031 — this PR's own stated invariant, broken. `_live_generated_rigor` now returns the `BacktestResultRecord` it graded as a third tuple element (it already held it at `:1445`, so this costs **no extra query**) and the route reads `sharpe_ratio` off that row.

**2. `ui/test/library-status.test.js` — the "rigor numbers" guard was VACUOUS for the DSR**, the one field that actually regresses. `strategies.indexOf(dead)` finds only the *first* occurrence, and for `verdict?.dsr` that is always the permitted `hasRealMetrics` probe at `Strategies.jsx:745`, so the loop body could never fail regardless of what appeared later in the file. Restoring `deflated_sharpe_ratio: verdict?.dsr ?? null` in `coerceGenerated` left the entire ui suite green. The loop now scans every occurrence. This also means the repo's adversarial-pass rule was **unmet** for that test in the first commit — it is met now (demo 5 below).

**3. `is_backtest_placeholder: true` put "no backtest has run" on a row whose pill says the gate ran and failed.** Listed as an open concern in round 1; it is a same-row self-contradiction, newly reachable because the "gate failed" pill was structurally unreachable on this tab before this PR. `metricDomain.absenceReason` (`metricDomain.js:298-305`) fires its pending branch on the literal, titling every empty CAGR / max-drawdown cell "PENDING — no backtest has run on this strategy yet". Now derived from the four-state: `pass`/`fail` → `false`; `pending`/`degenerate` keep `true`, where the sentence is true. Contained — on generated rows this field is read only by `absenceReason` (`Strategies.jsx:1244` scopes to `examples`; the `Leaderboard.jsx` / `StrategyPassport.jsx` reads are different payloads).

**4. `Strategies.jsx:200-202` — comment fix.** It said "coerceGenerated's own rows don't set it at all"; `coerceGenerated` now sets `rigor_gate_status` explicitly.

### Not changed, and why

* **The stored-column split is systematic, not incidental — the owner should know the size of it.** `_live_generated_rigor` routes look-ahead through `verdict_from_persisted_row` (which retires `self_attested`), while `generation_pipeline.py:2022-2029` passes `result.look_ahead_audit_passed` raw and supplies no `library_pbo`. So **every `self_attested` row will read `fail` in the Library and `pass` on its own passport page.** Measured split for one such row:

  ```
  LIBRARY /generated : passes=False status='fail'  DSR=-2.6068 PBO=0.05 p=0.0020 OOS=-2.894
  DETAIL  /{id}      : passes=True  status='pass'  DSR= 0.87   PBO=0.02 p=0.99   OOS= 1.900
  ```

  Not a reason to hold this PR: the Library is now the correct side, and it now agrees with `_generated_strategy_rigor` — the same path that already drives the Deploy ladder *on the passport page itself* (`StrategyPassport.jsx:410-419`), so the passport already contradicted itself before this change. Closing it is #1746's fix (a live overlay on the passport routes), which is a separate logical change.
* **A `StrategyRecord` with persisted returns but no `strategy_passports` row** is forced to `pending` even though the gate could grade it. Fail-closed, and `_persist_candidate` writes both rows, so it should be unreachable.
* **The route is `async def` and runs `run_rigor_gate` on the event loop.** Consistent with `list_strategies` at `:644`, authenticated, not new. A follow-up at most.

## Adversarial demos (guards-need-an-adversarial-pass)

**1. Restore the stored-column source** — `d["passes_rigor_gate"] = bool(passports[d["id"]].passes_rigor_gate) if d["id"] in passports else None`:

```
FAILED backend/tests/test_generated_badge_parity.py::test_passport_row_with_no_persisted_returns_is_pending - assert True is None
FAILED backend/tests/test_generated_badge_parity.py::test_stored_passport_true_does_not_survive_a_failing_live_gate - AssertionError: the stored True must not reach the badge
FAILED backend/tests/test_generated_badge_parity.py::test_zero_variance_series_is_degenerate_not_fail_or_pending - assert True is False
FAILED backend/tests/test_generated_badge_parity.py::test_live_gate_pass_is_served_as_a_pass - assert False is True
=================== 4 failed, 4 passed, 11 warnings in 4.27s ===================
```

**2. Restore `rigor_verdict.passing` in `coerceGenerated`** — `passes_rigor_gate: verdict ? Boolean(verdict.passing) : null`:

```
✖ coerceGenerated no longer sources the badge from rigor_verdict (1.167917ms)
  AssertionError [ERR_ASSERTION]: passes_rigor_gate is being read back out of the
  generation-time fusion verdict again — that blob is a different gate's output and
  is never re-derived after a backtest
  40825 !== -1
ℹ tests 11  ℹ pass 10  ℹ fail 1
```

**3. Restore the `=== false` guard** in `statusTag` — `if (status === "live" && passesRigor === false) return "tag-muted"`:

```
✖ pending: nothing graded this row — not green, and NOT accused of failing
  AssertionError: Expected values to be strictly equal:
    actual: 'tag-positive'   expected: 'tag-muted'
✖ a row with no verdict at all is not green
  AssertionError: Expected values to be strictly equal:
    actual: 'tag-positive'   expected: 'tag-muted'
✖ the four-state verdict outranks a disagreeing boolean
```

**4. Change the label string in one file** (em-dash → hyphen in `StrategyPassport.jsx` only):

```
✖ the demotion label is byte-identical to StrategyPassport.jsx's literal (0.513916ms)
  AssertionError [ERR_ASSERTION]: StrategyPassport.jsx no longer contains the exact
  literal "Reference only — gate failed" — the Library and the passport have started
  describing the same verdict differently
ℹ tests 11  ℹ pass 10  ℹ fail 1
```

**5. Round-2 fix (2) — re-source the DSR from the fusion verdict** in `coerceGenerated`: `deflated_sharpe_ratio: verdict?.dsr ?? null`. Green before the guard fix (924/924, EXIT=0 — the whole point); red now:

```
✖ the rigor numbers no longer come from rigor_verdict either (0.533542ms)
  AssertionError [ERR_ASSERTION]: Strategies.jsx renders verdict?.dsr — the fusion
  verdict's numbers beside a live-gate pill
ℹ tests 13  ℹ pass 12  ℹ fail 1
```

**6. Round-2 fix (1) — display Sharpe back to `strategy_passports.sharpe_ratio`**:

```
E   AssertionError: display Sharpe is the GRADED backtest row's own number, not the passport snapshot (0.4)
E   AssertionError: served sharpe_ratio=1.4 — the display Sharpe and the rigor numbers are describing two different backtest runs
FAILED backend/tests/test_generated_badge_parity.py::test_rigor_numbers_come_from_the_live_run_not_the_stored_passport
FAILED backend/tests/test_generated_badge_parity.py::test_display_sharpe_is_the_graded_backtest_rows_own_number
=================== 2 failed, 7 passed, 12 warnings in 4.02s ===================
```

**7. Round-2 fix (3) — `is_backtest_placeholder` back to the hardcoded `true`**:

```
✖ coerceGenerated derives is_backtest_placeholder from the gate verdict (0.454042ms)
  AssertionError [ERR_ASSERTION]: coerceGenerated is back to a hardcoded
  is_backtest_placeholder — a graded row's empty cells would again be titled
  'no backtest has run on this strategy yet'
ℹ tests 13  ℹ pass 12  ℹ fail 1
```

**8. Round-2 fix (3), executed half** — widen `metricDomain.absenceReason` to call a graded FAIL row pre-backtest (`row.passes_rigor_gate === false ||` added to the pending condition). Proves the executed assertion is live, not just the source-text one:

```
✖ a graded row's empty cells do not claim that no backtest has run (1.136583ms)
    actual: 'pending',   expected: 'not_measured'
ℹ tests 13  ℹ pass 12  ℹ fail 1
```

All eight mutations applied, run, and restored; working tree matches this commit (`git status` clean).

## Verification

```
$ env -i HOME=$HOME PATH=$PATH PYTHONPATH=backend python -m pytest \
    backend/tests/test_strategies_routes.py backend/tests/test_strategy_endpoints.py \
    backend/tests/test_passport_honesty.py backend/tests/test_passport_loader.py \
    backend/tests/test_api_routes.py backend/tests/test_generated_badge_parity.py \
    -q -p no:cacheprovider
155 passed, 13 warnings in 21.14s                                    EXIT=0

$ ... backend/tests/test_strategy_ownership.py test_generated_citation_truth.py \
      test_publishable_strategy_ids.py test_selection_bias_generated_gate.py \
      test_multipaper_passport.py test_brief_on_passport.py \
      test_reasoning_disclosure_gates.py test_sole_generation_route_guard.py
139 passed, 54 warnings in 13.72s                                    EXIT=0

$ cd ui && npm test
ℹ tests 926   ℹ pass 926   ℹ fail 0                                  EXIT=0

$ ruff format --check backend/archimedes/api/strategies_routes.py backend/tests/test_generated_badge_parity.py
2 files already formatted                                            EXIT=0
$ ruff check backend/tests/test_generated_badge_parity.py
All checks passed!                                                   EXIT=0
$ eslint ui/src/libraryStatus.js ui/src/components/Strategies.jsx ui/test/library-status.test.js
                                                                     EXIT=0
```

`ruff check` on `strategies_routes.py` reports one `SIM105` — pre-existing on `origin/main` in `_rollback_quietly`, verified by running the same check against `git show origin/main:...` inside the repo (same config). Not introduced here, not fixed here (one logical change per PR).

## Measured page cost

The overlay is **two SQL statements per page regardless of row count** (one `get_all_daily_returns`, one `latest_backtests_by_strategy`) plus one in-process `run_rigor_gate` per *gradeable* row, and `test_badge_overlay_is_batched_not_per_row` pins the constancy at the engine boundary (`before_cursor_execute`, same counter as `test_publishable_strategy_ids.py`) by comparing two page sizes:

| gradeable rows (300-pt series each) | wall | SQL statements on the request |
|---|---|---|
| 1 | 12 ms | 6 |
| 50 (route default limit) | 179 ms | 6 |

Measured at the engine boundary after a warm-up request (so app-lifespan DDL is not counted as page cost). Six is the whole request, auth and the citation/cost/publish reads included; the count does not move with N, which is the claim. Round 2 added no query — the graded `backtest_results` row was already loaded.

## Open concerns

* **The detail route still reads the stored column.** `GET /api/strategies/{id}` for a generated id serves `_passport_to_strategy_response`, i.e. `strategy_passports.passes_rigor_gate` (`strategies_routes.py:1441`). Badge parity holds for every row whose stored column agrees with the live gate — the pinned prod shape — but a row with a stale stored `True` and a failing live gate will now read `fail` in the Library and `pass` on its own passport page, and the four rigor numbers differ between the two routes for the same reason. **Round 2 established that this flip is systematic rather than incidental** — every `self_attested` look-ahead row is affected; see "Not changed, and why" above for the measured split and why it belongs to #1746 rather than here. `/api/leaderboard` reads the same stored column (`Leaderboard.jsx:194`/`:302`), also untouched.
* **`DeployabilityChip` is curated-only and left alone.** Its `deployMap` comes from `/api/selection-bias/gate`, whose LIST cohort is `_provider().list_strategies()` (`selection_bias_routes.py:307`), so `deployMap[generated_id]` is always undefined and the chip renders nothing on this tab. The one designed counter-signal is absent — noted, not changed.
* **The gate run is uncached.** The returns are what a cache key would have to be built from and reading them is the expensive half, so there is no free cache in front of it. At the measured ~3 ms/row this is affordable at the route's limit; a per-page memo is a real follow-up if the corpus grows.
* **The mobile lib-card still renders no rigor icon** (no check/X/clock/alert arms, unlike the desktop row). It can no longer render a bare green "Live" — it calls the same helpers — but it carries less signal than the table.

### DIFF
--- a/backend/archimedes/api/strategies_routes.py
+++ b/backend/archimedes/api/strategies_routes.py
@@ -18,6 +18,7 @@ import asyncio
 import json
 import logging
 from datetime import UTC
+from typing import TYPE_CHECKING
 
 import numpy as np
 from fastapi import APIRouter, Depends, Query, Request, Response
@@ -41,6 +42,12 @@ from archimedes.api.wallet_routes import get_linked_wallet_address
 from archimedes.models.strategy import Strategy, StrategyStatus
 from archimedes.services.live_rigor_gate import RigorGateVerdict
 from archimedes.services.rigor_evaluator import RigorGateResult
+
+if TYPE_CHECKING:
+    # Annotation only: `_live_generated_rigor` hands its caller the backtest row
+    # the gate graded. The runtime import stays inside that function, where the
+    # rest of the persisted-context reads already live.
+    from archimedes.models.backtest_store import BacktestResultRecord
 
 logger = logging.getLogger(__name__)
 
@@ -864,7 +871,7 @@ async def list_generated_strategies(
                 }
             verdicts = _live_generated_rigor(session, [i for i in page_ids if i in passports])
             for d in page:
-                verdict, rigor_result = verdicts.get(d["id"], (None, None))
+                verdict, rigor_result, graded_backtest = verdicts.get(d["id"], (None, None, None))
                 gate_status = verdict.status if verdict is not None else "pending"
                 d["rigor_gate_status"] = gate_status
                 # COUPLED to the four-state, never a separate stored boolean:
@@ -883,12 +890,17 @@ async def list_generated_strategies(
                 d["dsr_p_value"] = rigor_result.dsr_p_value if (graded and rigor_result) else None
                 d["pbo_score"] = rigor_result.pbo_score if (graded and rigor_result) else None
                 d["out_of_sample_sharpe"] = rigor_result.oos_sharpe if (graded and rigor_result) else None
-                # Display Sharpe is a descriptive backtest stat, not a gate
-                # output, so it comes from the passport row the gate re-graded —
-                # but it is still withheld unless that grading happened, so the
-                # headline number and the badge describe the same event.
-                passport = passports.get(d["id"])
-                d["sharpe_ratio"] = passport.sharpe_ratio if (graded and passport is not None) else None
+                # Display Sharpe is a descriptive backtest stat rather than a
+                # gate output, so it comes from the SAME `backtest_results` row
+                # the gate read its returns from — never from the denormalised
+                # `strategy_passports.sharpe_ratio`, which has no foreign key to
+                # backtest_results and routinely describes an older run (~155
+                # backtest rows per strategy; scripts/archive_backtest_results.py).
+                # Serving that column here printed a Sharpe of 1.4 beside a DSR
+                # computed from a series whose Sharpe was 0.031. Withheld unless
+                # the grading happened, so the headline number, the four rigor
+                # numbers and the badge all describe one event.
+                d["sharpe_ratio"] = graded_backtest.sharpe_ratio if (graded and graded_backtest is not None) else None
             # Citation truth: ``StrategyRecord.to_dict()`` returns source_papers
             # exactly as stored — arxiv_id, no title — so the Library card had no
             # real paper title to print and printed the generated strategy's own
@@ -1363,10 +1375,23 @@ def _passport_rigor_status(record, daily_returns: list[float]) -> tuple[str, boo
 
 def _live_generated_rigor(
     session, strategy_ids: list[str]
-) -> dict[str, tuple[RigorGateVerdict, RigorGateResult | None]]:
-    """LIVE four-state gate verdicts (+ the result each was reduced from) for a
-    page of GENERATED strategy ids — one ``run_rigor_gate`` per row, over that
-    row's OWN persisted returns and its OWN persisted grading context.
+) -> dict[str, tuple[RigorGateVerdict, RigorGateResult | None, BacktestResultRecord | None]]:
+    """LIVE four-state gate verdicts for a page of GENERATED strategy ids — one
+    ``run_rigor_gate`` per row, over that row's OWN persisted returns and its
+    OWN persisted grading context.
+
+    Each entry is ``(verdict, result, backtest_row)``. The third element is the
+    ``BacktestResultRecord`` the gate actually read its returns from — returned
+    rather than looked up again by the caller because the display Sharpe has to
+    describe THAT run and no other. ``latest_backtests_by_strategy`` and
+    ``get_all_daily_returns`` rank with the identical window
+    (``created_at DESC, id DESC``, backtest_repository.py), so the row handed
+    back here is the same physical row whose ``artifact_json`` produced
+    ``daily``. It is not ``strategy_passports.sharpe_ratio``: that column is a
+    denormalised snapshot with no foreign key to ``backtest_results`` (see
+    ``scripts/archive_backtest_results.py``, ~155 backtest rows per strategy),
+    so serving it beside these numbers prints a Sharpe from one run next to a
+    DSR from another.
 
     This is the batched form of ``selection_bias_routes._generated_strategy_rigor``
     (:676), and it is deliberately a re-use of that function's data path rather
@@ -1434,13 +1459,13 @@ def _live_generated_rigor(
         _rollback_quietly(session)
         return {}
 
-    graded: dict[str, tuple[RigorGateVerdict, RigorGateResult | None]] = {}
+    graded: dict[str, tuple[RigorGateVerdict, RigorGateResult | None, BacktestResultRecord | None]] = {}
     for sid in strategy_ids:
         daily = returns_by_id.get(sid) or []
         if len(daily) < _MIN_RETURNS_FOR_GATE:
             # Genuinely pre-backtest: the gate cannot run, so the badge is
             # "unknown". Never a boolean, and never the stored one.
-            graded[sid] = (RigorGateVerdict.pending(), None)
+            graded[sid] = (RigorGateVerdict.pending(), None, latest_by_id.get(sid))
             continue
         latest = latest_by_id.get(sid)
         num_trials, _num_trials_scope = _num_trials_for_generated_row(
@@ -1474,9 +1499,9 @@ def _live_generated_rigor(
             )
         except Exception as exc:  # pragma: no cover — defensive; gate-level failure
             logger.warning("generated badge: live gate failed for %s (→ pending): %s", sid, exc)
-            graded[sid] = (RigorGateVerdict.pending(), None)
+            graded[sid] = (RigorGateVerdict.pending(), None, latest)
             continue
-        graded[sid] = (RigorGateVerdict.from_result(result), result)
+        graded[sid] = (RigorGateVerdict.from_result(result), result, latest)
     return graded
 
 
--- a/backend/archimedes/api/strategies_routes.py
+++ b/backend/archimedes/api/strategies_routes.py
@@ -772,6 +772,16 @@ async def list_generated_strategies(
     canonical user; verified linked wallet handles legacy ``owner_wallet`` rows.
     Legacy ownerless rows remain invisible until purged (scripts/purge_orphan_generated.py)
     or published. Curated examples live on GET /api/strategies/ and stay public.
+
+    **The rigor badge on each row is a LIVE gate verdict (#1747)**, overlaid
+    below as ``rigor_gate_status`` (four-state) + ``passes_rigor_gate`` (the
+    fail-closed boolean coupled to it) + the four rigor numbers from that same
+    run. The row's own ``status`` and ``rigor_verdict`` are left on the wire
+    unchanged and are NOT the badge: they are the generation-time fusion
+    verdict, which is a different gate and is never re-derived after a backtest.
+    ``StrategyRecord.status`` is deliberately not written back — what the
+    synthesis gate thought is a fact worth keeping, and rewriting it has readers
+    (marketplace trending) outside this route.
     """
 
     from sqlalchemy import and_, or_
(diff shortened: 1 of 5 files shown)


danielscoffee (commented):
Library badge now comes from a live four-state gate instead of the generation-time fusion blob — that kills the #1747 Live✓ / “gate failed” split for this tab. Commenting (not approving): honesty surface, “owner vets first,” and a few leftover claim surfaces.

**What looks airtight**
- `_live_generated_rigor` overlays `rigor_gate_status` + coupled `passes_rigor_gate` (pending → `null`, never a manufactured fail) and serves DSR/PBO/OOS only when the run reached pass/fail — not `strategy_passports.passes_rigor_gate`, not `rigor_verdict`.
- Data path matches `_generated_strategy_rigor` (num_trials provenance, persisted PBO, no `pbo_library_size` relaxation, look-ahead triple, `_MIN_RETURNS_FOR_GATE`).
- `libraryStatus.js` four sentences are the right product truth: Live / Reference only — gate failed / Not yet graded / Unevaluable — flat returns. Pending is not accused of failing. GATE_FAILED string pinned to StrategyPassport.
- Mutation-style tests (stored True must not win; fusion `passing` must not feed the pill; batched SQL) are the right shape for this class of bug.

**Call-outs before merge**
1. **Rejected-section copy is now wrong.** Short/ungraded rows stay pending in the main table; `false` is fail (+ degenerate). `Strategies.jsx` still says most Rejected rows are `"return series too short"` (~1183–1188 and the empty-state blurb). Please rewrite that before this ships, or the Library trades one false claim for another.
2. **#1746 is still live and more visible.** Detail/passport still read the mixed-vintage stored column (`_passport_to_strategy_response`). Stale stored `True` + failing live gate → Library fail, passport pass. The list↔detail parity assert only holds when the stored column already agrees — worth an explicit “known diverge until #1746” note (or a test that pins the diverge).
3. **Leaderboard 2-state / no write-back / hard-coded `is_backtest_placeholder: true`** — agree these are follow-ups; flagging so they don’t get lost when the Library finally looks honest.

**CI:** Frontend + lint/analyze green when reviewed; Backend unit tests were still pending — please don’t merge on a yellow backend.

Do not merge until you vet the owner defaults (esp. “Not yet graded” wording + Rejected collapse emptying the main table). Happy to re-review a tiny copy fix on (1).

dbrowneup:
Superseded by #1792. The owner's 2026-09-01 decision ("stored": grade once at backtest time, persist with provenance, every surface reads the row) replaces this PR's premise of a live-gate overlay at read time; #1792 carries the same Library/badge fix on top of the stored verdict, plus the four-state helpers this PR introduced. Closing so there is one PR to vet.

PR #1766, opened 2026-09-01
[reliability][adr] Backtests are frozen evidence: retire the in-app refresh loop (#1760)

### LINKED ISSUE #1760: [infra][reliability] Cold-boot backtest-refresh storm: every new task re-runs the curated library at +180s, pegs the vCPU, and gets killed by its own container health check
Observed on the first deploy after #1740 (task-def 215, 2026-09-01 ~20:14–20:25Z) while watching cold boots for the #1632 lift decision. This is **not** the #1632 segfault class — no revision-215 task has exited 139 — it is a second, independent cold-boot defect.

## What happened

| time (UTC) | task 33a36e53 (rev 215) |
|---|---|
| 20:14:11 | task started; `/health` 200s at ~0.1s; ECS marks HEALTHY |
| 20:16:26–20:16:48 | a real visitor hits `/api/features`, `/api/explore/assets` (2.7s), `/api/corpus/overview`, `/api/papers`, `/api/generate/quote`, `GET /api/strategies/` |
| 20:16:50 | `compute_pbo` over the cohort (the live rigor gate for the Library list, cold) |
| **20:17:11 (+180s)** | `backtest_refresh_loop` settle delay expires → `run_backtests()` for the whole curated library runs **in the web process** (`asyncio.to_thread`): `normalize_ohlcv(CL=F)…` lines every few seconds, `REFUSING to persist backtest for …pairs…` ×4 |
| 20:17 | Container Insights: **CpuUtilized max 972 of 1024 units**, MemoryUtilized pinned at 1537 MB |
| 20:17–20:18:21 | container health check (`urlopen('http://localhost:8000/health')`, timeout **5s**, retries 3, interval 30s) fails 3× → **"(task 33a36e53…) failed container health checks" → stopped** |
| 20:19:29 | the process, still draining, logs `slow request: GET /api/strategies/ status=200 duration_ms=159933.7` |

The replacement task d78c5dc8 booted 20:17:28, hit the same storm at 20:20:33 (+185s), and survived only because `/health` came back in 1.4–2.4s (under the 5s kill line): `slow request: GET /health … 2412.7ms`. Its successor bf82be84 (for the killed one) boots into the same storm at ~20:24:30. Whether a task lives or dies is a coin-flip on whatever else is running at +3 minutes — which, on a deploy, is exactly when a visitor's first cold Library request lands.

## Why it recurs on every boot

`services/backtest_scheduler.py`:
- the missing-driven trigger fires when **any** curated strategy has no persisted row; the pairs family *never* gets a row (`run_backtests.py:456` `REFUSING to persist … realized vol/return check status='flagged'`), so `needs_refresh()` is true on every cold start;
- the backoff that is supposed to stop exactly this — `_last_unresolved_missing` — is a **process-global set**, so every new task starts with it empty and re-runs the storm once; its own comment says re-triggering "burns the box's 2 vCPUs for ~15 min right when users are generating";
- the module's single-writer note ("prod runs one backend replica") is false: `ecs_service_min_count = 2` (infra/variables.tf), and every task runs its own loop;
- the work runs in-process on a **1-vCPU task** (`cpu: 1024`); `to_thread` cannot keep the event loop responsive when pandas/numpy own the core.

## Fix (small PR, tonight — held for owner word)

1. Run the refresh as a **`nice -n 19` subprocess** (`python -m archimedes.scripts.run_backtests`, same code path) instead of a thread, so the serving process keeps the core under contention and `/health` answers.
2. Make the backoff **shared across tasks**: persist the unresolved-missing set (Redis, TTL = the refresh cadence; fall back to the process-local set when Redis is unavailable) so a permanently-missing family cannot re-trigger the storm on every boot.
3. Raise `BACKTEST_REFRESH_STARTUP_DELAY_S` default 180 → 900 so the storm never coincides with the deploy's ALB grace / first-visitor window (and pin it in the task-def env).
4. Container health check: leave the 5s timeout (it is the right kill line for a wedged server); the fix is to stop wedging it.

Follow-up (its own issue): move the refresh off the web tier entirely (scheduled Fargate task, like kb-runner) — the "single writer" premise is gone and this loop has no business in a request-serving container. Related: #1632 (the lift decision depends on clean cold boots), #1713/#1733 (cache warm-up before target registration — does not cover +180s), #1714 (rolling-update drain windows), #1594.

dbrowneup:
Closes #1760.

> **Two commits.** `ac5a9952` is the freeze itself. `c7155c14` is the verifier round: three defects, all in the *honesty of the claims* and the *reach of the guard*, not in the deletion — see [§ Commit 2](#commit-2--the-verifier-round).

## The incident

`services/backtest_scheduler.py` armed a long-lived asyncio task from the FastAPI lifespan. After a 180 s settle delay it re-ran `run_backtests()` for the **whole curated library in the serving process**, then repeated every 24 h.

On the 2026-09-01 deploy (task-def 215, ~20:14–20:27 Z) that landed at +180 s on a 1-vCPU Fargate task while a visitor's cold `GET /api/strategies/` was in flight:

| Time (UTC) | Task `33a36e53` |
|---|---|
| 20:14:11 | started; `/health` 200 at ~0.1 s; ECS marks HEALTHY |
| **20:17:11 (+180 s)** | settle delay expires → `run_backtests()` for the whole library, in the web process |
| 20:17 | Container Insights: **CpuUtilized 972 of 1024**, memory 1537 MB |
| 20:17–20:18:21 | container health check (5 s timeout) fails 3× → **task killed** |
| 20:19:29 | the draining process logs `slow request: GET /api/strategies/ status=200 duration_ms=159933.7` |

ECS replaced each dead task with a fresh one that booted into the same storm. It recurred on **every** boot because the missing-row trigger fired on the pairs family (which can never get a row — `run_backtests` refuses to persist it), the backoff that existed to stop that was a **process-global set** empty in every new process, the "single writer" premise was false (`ecs_service_min_count = 2`), and `asyncio.to_thread` cannot keep an event loop responsive when pandas owns the only core.

Mitigated that night by pinning `BACKTEST_REFRESH_ENABLED=false` into task-def **216**; both 216 tasks passed +180 s with zero refresh lines and `/health` at 0.45 s.

## The owner decision

> "Agree with removing boot-time loop and age-driven refresh from the web tier completely. We can manually re-run the curated example strategy backtests as needed, to ensure they are correct and accurate. But for a generated strategy, that should essentially never happen, because the backtest should be correct from the beginning. Let's go with the freeze."
> — Dan, 2026-09-01

Not the cheaper-loop fix (`nice -n 19` subprocess + Redis-shared backoff + 900 s settle) that #1760 originally proposed. That answers the wrong question: it keeps a clock on a thing that should not have one, and the clock had already cost more than uptime — **~155 backtest rows per strategy** (6.3 GB / 14,857 rows / 96 strategies, measured 2026-08-30), and a **moving "latest"** that is what produced #1746's Sharpe drift: same strategy, same code, a different number on a different day with no decision behind the change.

## What was deleted

- `backend/archimedes/services/backtest_scheduler.py`
- `backend/tests/test_backtest_scheduler.py`
- its lifespan wiring in `backend/archimedes/main.py`
- `BACKTEST_REFRESH_ENABLED`, `BACKTEST_REFRESH_INTERVAL_HOURS`, `BACKTEST_MAX_AGE_HOURS`, `BACKTEST_REFRESH_STARTUP_DELAY_S` — from code, from the flip-list's actionable tables, and from the docs
- the "Automated backtest refresh (no operator ritual)" section of `docs/runbooks/operations.md` and the `backtest_scheduler.py` cell in `docs/architecture.md`'s monitoring row

Removal evidence, per the flip-list's § Removing a flag:

```
$ git log --diff-filter=A -- backend/archimedes/services/backtest_scheduler.py
018b2170 2026-07-03 [backend] Automated in-app backtest refresh — retire the operator-invoked CLI ritual
```

After this commit, `grep -rn 'BACKTEST_REFRESH\|BACKTEST_MAX_AGE\|backtest_scheduler\|backtest_refresh'` returns only: the ADR, the runbook, the guard test, the flip-list's § DEAD / RETIRED row, two dated sprint docs now carrying superseded banners, and `reports/repo-scout-context-2026-07-04.md` — a dated snapshot report left as history, deliberately (same treatment `PINATA_JWT`'s removal gave historical docs).

### A latent coupling went with it

`arm_paper_advance_for_web_tier()` and the `PAPER_TRACE_PUBLISH` boot warning were nested **inside** `if refresh_enabled():`. So the 216 mitigation silently disarmed the paper-advance arming too — an unintended second effect of the kill switch. Both are now unconditional; `arm_paper_advance_for_web_tier` reads `PAPER_ADVANCE_ENABLED` itself and refuses on its own (and `PAPER_ADVANCE_ENABLED` is pinned `false` today anyway, so live behaviour is unchanged — but the coupling is gone).

## The runbook

`docs/runbooks/curated-backtests.md` — `scripts/run_backtests.py` is now the **only** way to produce a curated backtest.

- **When:** three triggers only — a curated strategy file changed, a data-quality/engine fix landed, or an explicit owner decision. **A calendar is not a trigger.** Content-hash dedup (`insert_backtest_if_missing`) already makes an unchanged strategy a no-op, so over-applying trigger 1 is safe and running "just in case" is pointless.
- **How, against prod:** a one-off `aws ecs run-task` on the dedicated single-container **`archimedes-migrate`** family (`infra/ecs_migrate.tf`) with a command override — the same mechanism `.github/workflows/deploy.yml`'s `migrate` job uses for `alembic_migrate_preflight`, with its literal cluster/family/network-configuration constants. Same image, same roles, same SSM-backed `DATABASE_URL`, no nginx sidecar, no health check, runs to completion. Includes the `ARCHIMEDES_ARTIFACT_DIR` override (the packaged artifacts dir is not writable by the nonroot user — the silent `PermissionError` that froze the board for six weeks), how to read the `backtest run summary` line, and the waiter caveat (`aws ecs wait tasks-stopped` gives up at 10 min; a full library run took ~15 min on 2 vCPU, and the migrate family is 1 vCPU).
- **Explicitly discouraged:** `aws ecs execute-command` into a serving task for a full run. That is #1760 reproduced by hand.
- **Five things to check before trusting a run** (the fifth is commit 2): exit code, the `backtest run summary` line, `ARCHIMEDES_ARTIFACT_DIR`, the data window — and **which image actually ran**. `--task-definition archimedes-migrate` resolves to that family's last *Terraform-applied* revision, tagged `latest` (`infra/ecs_migrate.tf`, `variables.tf`); CI re-registers only `archimedes-backend` but does push `:latest` every deploy, so the run executes current `main` — **not necessarily the digest the service is serving**. An ADR whose thesis is that a backtest is dated evidence should have the operator record what produced it (`describe-tasks --query 'tasks[0].containers[0].imageDigest'`).
- **What it must never do:** no clock, no boot hook, never in the serving process, never for generated strategies.

## The ADR

`docs/adr/backtests-are-frozen-evidence.md` — Accepted, Dan, 2026-09-01. Indexed in `docs/adr/README.md` (count 23 → 24), `docs/README.md`, and `mkdocs.yml`.

A backtest is a **one-time artifact with a stated data window**. Generated strategies are backtested exactly once at generation. Curated strategies are backtested when their code changes or by explicit operator action. **No periodic or boot-time refresh anywhere** — the follow-up floated in #1760 ("move it to a scheduled Fargate task like kb-runner") is explicitly *not* taken, because relocating a clock keeps the clock. Forward performance is the **paper-trading ledger's** job.

The ADR is careful about scope: it names the web tier's *other* scheduled work — the paper-advance arm (child-process isolated), the opt-in `REVENUE_SWEEP_ENABLED` sweep (off today), and one tick loop per rehydrated marketplace publisher — and says explicitly that it does **not** adjudicate those three. It removes the one that took the fleet down. (This wording is commit 2; see Defect 1.)

**What it supersedes** — the scheduler's own docstring rationale ("prod sat at *pending* for weeks", "a product whose wedge is autonomous rigor cannot depend on an operator for its evidence"). That named a real failure and answered it wrongly, and the record says so:

- the scheduler landed 2026-07-03 (`018b2170`);
- the board's newest row then sat at `backtest_end 2026-07-01` until **2026-08-18** — six weeks — *while the loop was armed and ticking daily*;
- the cause (`8e6554c5`, "Fix the frozen leaderboard refresh: artifact dir is read-only on Fargate"): every scheduled run died at the artifact write, before the DB insert.

The unattended loop did not prevent the stale board. It hid it, for longer than the operator ritual it replaced, because nobody watches a job nobody has to run. The honest answer is the manual runbook plus the passport's `pending` state.

**Consequences recorded:** the moving "latest" behind #1746's Sharpe drift disappears; ~155 rows per strategy stop accumulating (making `backtest-results-retention.md` a one-time cleanup, not a chore); the cold-boot storm is gone by construction; and, on the cost side, a stale board is now someone's decision to leave stale.

## The inert env var

`BACKTEST_REFRESH_ENABLED=false` is still on the live task definition (216+) and, after this merge, **names a flag nothing reads**. It was never in `infra/ecs.tf` or `.github/scripts/ecs_rewrite_task_def.py`, so there is nothing to remove in this repo — but **CI deploys clone the live task-def**, so it will keep riding along until an operator removes it deliberately. That is written into the flip-list's § DEAD / RETIRED row as an outstanding operator action, so the next person reading the task definition is not told a switch exists.

## Commit 2 — the verifier round

Vetting found three defects. The deletion was correct and complete; these were about what the PR *claimed*.

**Defect 1 — three "only clock left" claims were false, one of them inside an Accepted ADR.** The ADR said it twice (TL;DR and Consequences) and `paper_trading.py`'s docstring said it a third time: that the paper-advance tick is the web tier's only remaining scheduled work. It is not. The same lifespan arms **`revenue_sweep_loop()`** — `while True` + `await asyncio.sleep(REVENUE_SWEEP_INTERVAL_S)`, default 3600, `services/revenue_sweep.py` — in-process on the 1-vCPU serving task, and step 3 rehydrates marketplace publishers, each getting its own `asyncio.create_task(self._run_loop(...))` whose body is `while not self._stop.is_set()` (`marketplace/service.py`). Both are opt-in/conditional and neither is a backtest, but both are forward-looking clocks in the web process.

Why that was worth blocking: `docs/adr/README.md` forbids relitigating an Accepted ADR, so the sentence becomes what the next reader trusts — and it pointed the *next* cold-boot investigation away from `revenue_sweep_loop`, which is precisely the in-process-clock-on-a-1-vCPU-task shape #1760 was about. All three sites now name what actually ticks; the ADR adds that it does not adjudicate those three.

**Defect 2 — the guard did not enforce what the runbook said it enforced.** The runbook claimed "No boot hook. Nothing in the FastAPI lifespan, on any tier, may schedule this. Guarded by `test_backtests_are_frozen.py`." The guard banned two literal tokens. Mutant C renamed the *file* but kept the token, so the original adversarial pass never tested renaming the *symbol* — and **Mutant E (below) was green on the whole tree**. Fixed by asserting the choke point with `ast`: only `scripts/run_backtests.py` may import or call `run_backtests`. Both token scans are also lower-cased, because Mutant F — a live `os.getenv("BACKTEST_REFRESH_ENABLED")` reader — passed *both* the frozen guard (case-sensitive) and the flip-list drift guard (forward direction satisfied by the § DEAD / RETIRED row, which is the pre-existing guard's documented contract, not something this PR introduced).

**Defect 3 — the PR body miscounted its own adversarial pass**, saying "three mutations" above four. Now six, and counted.

Also folded in, from the verifier's non-blocking notes: the runbook's fifth check (which image actually ran). Left alone, as the verifier recommended: `reports/repo-scout-context-2026-07-04.md` still names the deleted module — a dated snapshot outside `docs/`, same call the `PINATA_JWT` removal made. `docs/` itself is clean, which is what the repo's deleting-PR rule requires.

## The guard, and the adversarial pass

`backend/tests/test_backtests_are_frozen.py` — a static-source guard, which is honest here because the property being guarded is *"this code path does not exist"* (same reasoning and shape as the existing `test_lifespan_no_rigor_backfill.py`). Hermetic: an import, `inspect.getsource`, and `ast.parse`. No DB, no network, no app boot.

**Five assertions, in two halves.**

*The token half* — the module does not import, the file does not exist, the lifespan source names no `backtest_refresh` / `backtest_scheduler`, and no module anywhere under `archimedes/` does either. Both scans are **case-insensitive** (see Mutant F). Comments are stripped before matching, so the tombstone comment explaining *why* there is no refresh does not fail its own guard.

*The behaviour half* — `test_only_the_cli_reaches_run_backtests`. Every token scan is pinned to the retired loop's own spelling, so a rebranded loop walks past all four while doing exactly what #1760 retired (that is Mutant E, and it was green on this branch before commit 2). The behaviour, under any name, has to reach the runner, so the choke point is what to assert: **`scripts/run_backtests.py` is the only site under `backend/archimedes/` that may import or call `run_backtests`.** Done with `ast` rather than a text scan so comments and prose stay free. Measured before it was written: exactly two matching call sites exist in the package (`scripts/run_backtests.py:307`, `:481`), both inside the owner file — so no allow-list is needed, and generated strategies are unaffected because the generation pipeline reaches the analytics engine directly and never calls `run_backtests`.

### Six mutations, each shown red before reverting

**Mutant A — re-add a stub refresh task to the lifespan** (`_app.state.backtest_refresh = asyncio.create_task(backtest_refresh_task())`):

```
FAILED ...::test_lifespan_does_not_arm_a_backtest_refresh
FAILED ...::test_no_module_schedules_a_backtest_refresh
  AssertionError: these modules reference a backtest refresh loop again: ['main.py'].
2 failed, 3 passed
```

**Mutant B — restore `services/backtest_scheduler.py` from `origin/main`, lifespan untouched:**

```
FAILED ...::test_backtest_scheduler_module_is_gone - Failed: DID NOT RAISE ModuleNotFoundError
FAILED ...::test_backtest_scheduler_file_is_gone
FAILED ...::test_no_module_schedules_a_backtest_refresh
FAILED ...::test_only_the_cli_reaches_run_backtests
4 failed, 1 passed
```

Four, up from three before commit 2: the new assertion catches the restored module's `run_backtests` import as well.

**Mutant C — a rebranded *file*, banned token kept** (`services/evidence_freshness.py` with `async def backtest_refresh_tick()`), proving the guard is not pinned to one filename:

```
FAILED ...::test_no_module_schedules_a_backtest_refresh
FAILED ...::test_only_the_cli_reaches_run_backtests
2 failed, 3 passed
```

**Mutant D — the flip-list's reverse guard**, putting a `BACKTEST_REFRESH_ENABLED` row back on the actionable § LIVE table with no reader:

```
FAILED test_feature_flag_fliplist_drift.py::test_no_actionable_row_names_a_flag_that_no_longer_exists
1 failed, 17 passed
```

**Mutant E — a rebranded *symbol*, no banned token anywhere.** `services/evidence_freshness.py::curated_evidence_tick` — `await asyncio.sleep(180.0)`, then `while True:` → `from archimedes.scripts.run_backtests import run_backtests` → `await asyncio.to_thread(run_backtests)` → `await asyncio.sleep(24*3600)`, armed from the lifespan as `_app.state.evidence_freshness_task`. That is #1760's defect byte-for-byte in behaviour, and **the entire tree was green on it** before commit 2 (the verifier reproduced it: 71 targeted tests passed). Now:

```
FAILED ...::test_only_the_cli_reaches_run_backtests
  AssertionError: these sites import or call run_backtests outside the CLI:
  ['services/evidence_freshness.py:14 (from archimedes.scripts.run_backtests import ...)'].
1 failed, 4 passed
```

**Mutant F — a live upper-case reader.** `if os.getenv("BACKTEST_REFRESH_ENABLED", "").strip().lower() == "true":` back in the lifespan. This passed **both** guards before commit 2: the flip-list's forward direction is satisfied by the flag being named anywhere in the doc (including its § DEAD / RETIRED row — the pre-existing guard's documented contract), and the frozen guard's scans were case-sensitive, so the upper-case env name missed. With both scans lower-cased:

```
FAILED ...::test_lifespan_does_not_arm_a_backtest_refresh
FAILED ...::test_no_module_schedules_a_backtest_refresh
  AssertionError: ... ['main.py'].
2 failed, 16 passed
```

All six reverted; `git status --porcelain` clean apart from the two commits, and the tree is green as pushed.

## Verification

Both commits, re-run from the worktree root at `c7155c14`:

| Command | Result |
|---|---|
| targeted: frozen guard + fliplist drift + `test_ecs_*` + lifespan/gateway wiring + paper-advance isolation & kill-switch + docs site | **110 passed** |
| `pytest -m "not integration" -q` (full unit suite) | **5636 passed, 4 skipped, 2 deselected in 530 s — `EXIT=0`** (`/tmp/1760c-suite.log`). One more than commit 1's 5635: the new assertion. |
| `make docs-check` | links 1774 scanned / **0 broken**; index 176 docs / **0 orphaned** — `EXIT=0` |
| `mkdocs build --strict` | **`EXIT=0`, zero WARNINGs** |
| `ruff format --check` + `ruff check` on all four `.py` touched | 4 files already formatted; **All checks passed!** |

## Concerns for the vet

1. **The runbook's prod command is written from the repo, not executed.** I have no AWS credentials here. Every literal in it is copied from `.github/workflows/deploy.yml`'s working `migrate` job (`archimedes-cluster`, `archimedes-migrate`, the static network-configuration) and from `infra/ecs_migrate.tf`. What I verified by reading rather than running: the image carries `/app/analytics-engine/strategies`, which `resolve_strategies_dir()` lists as a candidate, so the migrate task-def needs no `ARCHIMEDES_STRATEGIES_DIR`; `DATABASE_URL` is in its `secrets{}`; market data defaults to yfinance, which needs no token, over the private subnets' NAT egress. **The first real run should be treated as a rehearsal** — please run it once and tell me what the summary line says, and I will correct the runbook against the actual output.
2. **`docs/sprint/a6-rerun.md` and `docs/sprint/cluster-0-unblock.md`** got superseded banners rather than rewrites — they are dated execution records, and A5's "what makes the daily scheduler viable" is now describing something that does not exist. Say the word if you would rather they were left completely untouched.
3. **`reports/repo-scout-context-2026-07-04.md`** still lists `backtest_scheduler.py` in a file inventory. Left as-is: it is a dated snapshot outside `docs/`, same call the `PINATA_JWT` removal made for historical docs.
4. **The paper-advance un-nesting is a behaviour change on the merge**, though not a visible one: `PAPER_ADVANCE_ENABLED` is pinned `false` in both `infra/ecs.tf` and the CI task-def rewrite, so `arm_paper_advance_for_web_tier` still refuses and logs. It only matters the day #1632 is fixed and the tick is flipped back — at which point it would have silently stayed off, because task-def 216 turned the flag it was nested under to false.
5. **The new guard's known limits, stated rather than implied.** It catches static imports and calls plus the literal dotted module path in a string (so `importlib.import_module("archimedes.scripts.run_backtests")` trips it). It would not catch a fully dynamic reach assembled at runtime, and it scans `backend/archimedes/` only — a scheduler living outside the app package is out of its reach. It also treats a *docstring* naming the dotted path as a call site; a `#` comment is the escape hatch, which is deliberate and consistent with how `_code_lines` already frees the tombstone comment.


danielscoffee (commented):
Paper-trading UX slice from the polish lane (the latent coupling called out on #1741).

**Verified on this branch**
1. **Hoist is in.** `arm_paper_advance_for_web_tier()` now sits in its own fail-soft try/except in the lifespan, not under `refresh_enabled()`. Same shape as the revenue-sweep block. `arm_paper_advance_for_web_tier` still reads `PAPER_ADVANCE_ENABLED` itself (default `"false"`) and refuses to spawn a child, so live fleet behaviour stays off. The 216 pin no longer silently disarms the tick.
2. **`PAPER_TRACE_PUBLISH` boot warning** came along with the un-nest — also correct; that provenance claim should not have been gated on backtest refresh.
3. **Existing source guard still holds.** `test_lifespan_does_not_call_paper_advance_loop` already requires `arm_paper_advance_for_web_tier` in `main.py` (comments stripped) and forbids in-process `paper_advance_loop(`. The original "# not nested under `refresh_enabled()`" predicate is now vacuous because that `if` is gone.
4. **Child still drops INFO.** `_module_main` still has no `logging.basicConfig`, so the proving `paper advance: {...}` line cannot reach CloudWatch. That is the #1741 observation blocker (FIX-FIRST #2), not this freeze. Leave it for the lift-PR rework.

Not reviewing the reliability/ADR half. Not approving. Not merging. Not closing #1760. `PAPER_ADVANCE_ENABLED` stays false.

PR #1770, opened 2026-09-01
[ux] Generate: the model card leads with the engine, not the per-token rate

dbrowneup:
## The owner's observation

Screenshot of the Generate page, bottom — the card under **Approve & Generate**:

> **MODEL & COST**
> Running Amazon Nova Micro · $0.035 in / $0.140 out per 1M tokens

Every word of that is true. The problem is what it leads with: a billing rate,
on the one surface whose first job is to tell a new user *what they are about
to run*. And the number reads as the user's bill when it is actually the
provider's published on-demand list price out of the pricing snapshot.

## Root cause

`ui/src/components/ModelCostPanel.jsx` is, by construction, a cost-comparison
panel — the collapsed header was written as a one-line summary of the table
underneath it (`Running <model> · $x in / $y out per 1M tokens`), not as an
answer to "what engine am I running?". Nothing was wrong; the card was
answering a question the user had not asked yet.

## The change (copy only)

The collapsed header now leads with the engine and demotes the rate:

```
Strategy engine
Debate society over the research corpus, served by Amazon Nova Micro
Provider list price · $0.035 in / $0.140 out per 1M tokens
```

- **Lead line** — what the engine *is*. "Debate society over the research
  corpus" is the shipped pipeline, not a slogan: `agents/debate_engine.py` is
  the sole generation path (T1.1 Phase 3). Both halves of the phrase are the
  repo's existing wording, but the compound is new copy here:
  `Architecture.jsx:417` says *"Archimedes instead runs a **debate society** —
  the sole generation path"*, and `Architecture.jsx:224` describes that stage as
  *"proposer society drafts candidates from the research corpus and current
  market regime"*. The exact string *"debate society over the research corpus"*
  appears nowhere else in `ui/src` or `backend/archimedes` — this PR introduces
  it, compressing those two lines into one.
- **Secondary line** — the same numbers as before, now labelled **Provider
  list price** so it does not read as the viewer's bill.
- **Drill-in unchanged** — same expand toggle, same full model/cost table,
  same snapshot note, same free-tier-only selection.

Model-name honesty is preserved by construction: the served model is still read
from `/health`'s `llm_model` through the shared `fetchHealth()` cache and
matched to the snapshot by `model_id` (`priced = chosen || active`). No model
name or id appears as a literal anywhere in the component, so the card can only
name what the backend reports. When `/health` has not answered, the header
names no model and quotes no price.

Files: `ui/src/components/ModelCostPanel.jsx` (header block + comment),
`ui/test/model-cost-line.test.js` (new).

## Guard

`ui/test/model-cost-line.test.js` — the repo's source-text idiom
(`oracle-copy.test.js`, `roadmap-copy.test.js`); `.jsx` is not importable under
`node --test`. It slices the collapsed header out of the toggle button's own
subtree (so the price table below cannot satisfy it), strips comments (so a
`{/* ... list price ... */}` note cannot stand in for copy — that vacuity bit
me on the first run and is why `stripComments` exists), and checks:

1. the header names the engine, and what it does, **before** any per-token rate;
2. the rate carries the literal **Provider list price** label, ahead of the numbers;
3. no `name` or `model_id` from `modelPricing.json` is hard-coded in the component;
4. the served model still comes from `fetchHealth()` → `d?.llm_model` → `model_id` equality, and the header contains no literal dollar figure;
5. the expand toggle and the full table are still there.

Checks 1–3 are each computed by a `null`-or-reason helper (`orderingProblem`,
`priceLabelProblem`, `hardCodedModelProblem`) and each is paired with a mutation
test that feeds **the helper** an input it must reject and asserts the exact
reason — so a helper that stops checking anything fails loudly instead of
passing green. Checks 4 and 5 are flat `assert.match` regexes with no branch
that can silently stop checking, so they carry no paired mutation; the file
header says so rather than claiming blanket coverage.

## Folded in: #1777 (Daniel Reis)

The first review of this PR found that check 3's mutation test was **vacuous**:
it built `HARD_CODED` by splicing `recommended.name` into the source and then
only asserted `HARD_CODED.includes(recommended.name)` — true by construction,
and it never ran the guard. Weakening the real check left all 8 tests green.

#1777 (`cursor/hardcoded-model-mutation-68b7`, based on this branch) fixes it by
hoisting the check into `hardCodedModelProblem(src)` and pointing both the
property test and the mutation at it. Merged here as `db52bc88` with his
authorship intact (`633a06fa`, Cursor Agent). **#1777 needs no separate merge —
close it as merged-via-parent.**

Follow-up commit `fadd2864` on top: the file header claimed *"Every check below
is paired with a mutation that must turn it red"*, which was not true of checks
4 and 5 — narrowed to the three paired helpers, with the reason the other two
need none. Same commit feeds the mutation `stripComments(HARD_CODED)`, the same
comment-stripped source its property test uses, so the two cannot diverge if a
future comment in `ModelCostPanel.jsx` ever names a snapshot model.

## Adversarial demo — the guard shown red

Four mutations, each reverted; the committed files are the clean ones.

**Mutation 1 — reorder only.** Swap the two `<div>` blocks in the header so the
price leads. Not one word changes:

```
✖ collapsed header names the engine before it quotes any price
  AssertionError: ModelCostPanel.jsx's collapsed header must name the strategy
  engine (and what it does) before the per-token rate — the owner's screenshot
  showed the rate leading
  + actual   'header quotes the per-token price before it names the engine'
  - expected null
ℹ pass 7  ℹ fail 1  — exit 1
```

**Mutation 2 — claim a model we may not be running.** Write the served model in
as a literal (`, served by Amazon Nova Micro`) instead of reading `/health`:

```
✖ no model name or id from the pricing snapshot is hard-coded in the card
  AssertionError: ModelCostPanel.jsx hard-codes a model name or id from the
  pricing snapshot. The card must name only the model /health reports as served
  (or the user's pick), never a literal …
  actual: 'hard-codes model name Nova Micro'
  expected: null
ℹ pass 7  ℹ fail 1  — exit 1
```

**Mutation 3 — gut the guard, component untouched.** This is the pairing #1777
bought. Delete only the `m.name` branch from `hardCodedModelProblem` and leave
`ModelCostPanel.jsx` clean:

```
✔ no model name or id from the pricing snapshot is hard-coded in the card
✖ mutation: a hard-coded served model is rejected
  AssertionError: writing "Nova Micro" into the card is not caught by the
  hard-coded-model check — it is guarding nothing
  actual: null
  expected: 'hard-codes model name Nova Micro'
ℹ pass 7  ℹ fail 1  — exit 1
```

Before #1777, that exact weakening left **8 pass / 0 fail, exit 0** — the guard
was disabled and its own mutation test did not notice. That is the defect this
branch now closes.

**Mutation 4 — the `stripComments` probe.** Move `Provider list price` into a
JSX comment and change the rendered label to `Your cost`, so only a comment
carries the honest word:

```
✖ the per-token rate is labelled as the provider's list price
  AssertionError: ModelCostPanel.jsx's collapsed header must label the
  per-token rate as the provider's list price, not present it as the user's bill
  actual: 'the per-token rate is not labelled as a list price'
  expected: null
ℹ pass 7  ℹ fail 1  — exit 1
```

## Verification

Run on `fadd2864` (post-#1777-merge), in `.claude/worktrees` on this branch.

| check | command | result |
| --- | --- | --- |
| new guard | `cd ui && node --test test/model-cost-line.test.js` | 8 pass, 0 fail — **exit 0** |
| full ui suite | `cd ui && node --test` | 921 tests, 921 pass, 0 fail, 0 skipped — **exit 0** |
| eslint (touched) | `cd ui && npx eslint src/components/ModelCostPanel.jsx test/model-cost-line.test.js` | no output — **exit 0** |
| mutations 1–4 | see above | each 7 pass / 1 fail — **exit 1**, all reverted |
| corpus claim integrity | `pytest -p no:cacheprovider -q backend/tests/test_corpus_claim_integrity.py` | 26 passed, 1 warning — **exit 0** |
| full unit suite | `pytest -p no:cacheprovider -m "not integration" -q` | 5644 passed, 4 skipped, 2 deselected, 328 warnings in 988.02s (16:28) — **exit 0** |
| ruff | `ruff check backend` | not applicable — **0 `*.py` files in this diff** (`git diff --name-only` against the merge base = exactly the two `ui/` files). `ruff check backend` is red on this branch (37 errors) and identically red at the merge base for that reason, so CI's ruff ❌ is a pre-existing `main` condition, not this PR. |

The corpus-claim row is here because the new copy makes a corpus claim; it does
not trip `OVERCLAIM_PATTERNS` (which target embedding / semantic-retrieval / KG
claims, and this line makes none). Registering `ModelCostPanel.jsx` in that
test's `PUBLIC_SURFACES` tuple would widen the guard to cover this file — worth
doing, but it is guard-coverage hygiene rather than part of this copy fix, and
it deserves its own adversarial demo, so it is **not** in this diff.

## Known trade-off

The card's eyebrow went from "Model & cost" to "Strategy engine", so the word
"cost" no longer appears above the fold on the Generate page — costs are now one
line down under "Provider list price", and in the drill-in. That is the point of
the change, but it is a real discoverability trade and worth an owner glance at
the screenshot. "Strategy engine" is also ambiguous between the two engines
(`execution/__init__.py:15-16` — generation vs execution); this card is the
generation half. "Strategy generation engine" would disambiguate, at the cost of
a longer eyebrow. Owner's call; changing it means updating `/Strategy engine/`
in the guard too.

## Concurrency

Touches neither `dbrowneup/verdict-of-record-a`'s files (`Strategies.jsx`,
`StrategyPassport.jsx`, `strategies_routes.py`, passport writes) nor
`dbrowneup/reasoning-traces-surfaced`'s generation-stream/SSE code.
`ModelCostPanel.jsx` and `ui/test/model-cost-line.test.js` are not in either
branch's surface.

### DIFF
--- a/ui/src/components/ModelCostPanel.jsx
+++ b/ui/src/components/ModelCostPanel.jsx
@@ -15,6 +15,16 @@ import pricing from '../data/modelPricing.json'
 //
 // `selectedModel` / `onSelectModel` lift the choice into the Generate flow. The
 // data is a snapshot (ui/src/data/modelPricing.json); the active row is live.
+//
+// COLLAPSED HEADER copy: it leads with what the engine IS (a debate society
+// over the research corpus — agents/debate_engine.py, the sole generation
+// pipeline) and only then quotes the per-token rate, labelled as the
+// provider's list price. The old header led with "Running <model> · $x in /
+// $y out per 1M tokens", which is true but reads as a bill on a page whose
+// first job is to tell you what you are about to run. The served model is
+// never a literal in this file — it comes from /health's `llm_model` and is
+// matched against the snapshot, so the card cannot name a model the backend
+// is not running. Guarded by ui/test/model-cost-line.test.js.
 
 const blended = (m) => m.input * 0.75 + m.output * 0.25
 const fmt = (x) => (x == null ? '—' : `$${x.toFixed(x < 1 ? 3 : 2)}`)
@@ -39,6 +49,11 @@ export default function ModelCostPanel({ selectedModel = null, onSelectModel }) 
   const active = rows.find((m) => m.model_id && m.model_id === activeModel)
   // The chosen model row (for the collapsed-header summary), if any.
   const chosen = rows.find((m) => m.model_id && m.model_id === selectedModel)
+  // The row whose published rate the collapsed header quotes: the user's pick
+  // if they made one, else the model /health reports as being served. Null
+  // when neither is known (or the served id is not in the snapshot) — the
+  // header then quotes no price at all rather than guessing one.
+  const priced = chosen || active
 
   // Free-tier-only selection. Premium rows are never selectable here.
   const selectable = (m) => Boolean(m.works_now && m.model_id && onSelectModel)
@@ -61,18 +76,28 @@ export default function ModelCostPanel({ selectedModel = null, onSelectModel }) 
         }}
       >
         <div>
-          <div className="label" style={{ marginBottom: 2 }}>Model &amp; cost</div>
+          <div className="label" style={{ marginBottom: 2 }}>Strategy engine</div>
+          {/* Lead line — WHAT you are running. The served model is read from
+              /health (activeModel) or the user's own pick; it is never a
+              literal here, so this line cannot name a model we are not
+              running. */}
           <div className="caption" style={{ color: 'var(--text-3)' }}>
+            Debate society over the research corpus
             {chosen ? (
-              <>Will run <strong style={{ color: 'var(--text-1)' }}>{chosen.provider} {chosen.name}</strong>
-                {' · '}{fmt(chosen.input)} in / {fmt(chosen.output)} out per 1M tokens</>
+              <>, next run served by <strong style={{ color: 'var(--text-1)' }}>{chosen.provider} {chosen.name}</strong></>
             ) : active ? (
-              <>Running <strong style={{ color: 'var(--text-1)' }}>{active.provider} {active.name}</strong>
-                {' · '}{fmt(active.input)} in / {fmt(active.output)} out per 1M tokens</>
+              <>, served by <strong style={{ color: 'var(--text-1)' }}>{active.provider} {active.name}</strong></>
             ) : activeModel ? (
-              <>Running <strong style={{ color: 'var(--text-1)' }}>{activeModel}</strong></>
+              <>, served by <strong style={{ color: 'var(--text-1)' }}>{activeModel}</strong></>
+            ) : null}
+          </div>
+          {/* Secondary line — the rate, explicitly labelled as the provider's
+              published list price so it does not read as your bill. */}
+          <div className="caption" style={{ color: 'var(--text-3)', marginTop: 2 }}>
+            {priced ? (
+              <>Provider list price{' · '}{fmt(priced.input)} in / {fmt(priced.output)} out per 1M tokens</>
             ) : (
-              <>Compare model costs across {rows.length} options on Bedrock</>
+              <>Compare provider list prices across {rows.length} models on Bedrock</>
             )}
           </div>
         </div>
--- a/ui/test/model-cost-line.test.js
+++ b/ui/test/model-cost-line.test.js
@@ -26,9 +26,14 @@
 //
 // Idiom is the repo's source-text one (oracle-copy.test.js,
 // roadmap-copy.test.js): `.jsx` is not importable under `node --test`, so the
-// component is read as text. Every check below is paired with a mutation that
-// must turn it red, so a check that stops checking anything fails loudly
-// instead of passing vacuously.
+// component is read as text. The three properties above are each computed by a
+// `null`-or-reason helper (`orderingProblem`, `priceLabelProblem`,
+// `hardCodedModelProblem`) and each is paired with a mutation test that feeds
+// the helper an input it must reject and asserts the exact reason — so a helper
+// that stops checking anything fails loudly instead of passing vacuously. The
+// two remaining tests (the /health wiring and the drill-in) are flat
+// `assert.match` regexes with no branch that can silently stop checking, so
+// they carry no paired mutation.
 
 import { readFileSync } from "node:fs";
 import assert from "node:assert/strict";
@@ -196,7 +201,7 @@ test("mutation: a hard-coded served model is rejected", () => {
 		`Debate society over the research corpus, served by ${recommended.provider} ${recommended.name}`,
 	);
 	assert.equal(
-		hardCodedModelProblem(HARD_CODED),
+		hardCodedModelProblem(stripComments(HARD_CODED)),
 		`hard-codes model name ${recommended.name}`,
 		`writing ${JSON.stringify(recommended.name)} into the card is not caught by the hard-coded-model check — it is guarding nothing`,
 	);
--- a/ui/test/model-cost-line.test.js
+++ b/ui/test/model-cost-line.test.js
@@ -98,6 +98,18 @@ function priceLabelProblem(text) {
 	return null;
 }
 
+// `null` = no snapshot model name or id is a literal in the source; a string
+// = the problem. Shared by the property test and the mutation that must turn
+// it red — a check inlined only on the clean source cannot be shown to reject
+// a hard-coded model.
+function hardCodedModelProblem(src) {
+	for (const m of pricing.models) {
+		if (src.includes(m.name)) return `hard-codes model name ${m.name}`;
+		if (src.includes(m.model_id)) return `hard-codes model id ${m.model_id}`;
+	}
+	return null;
+}
+
 // ── 1. The header leads with the engine ───────────────────────────────────
 
 test("collapsed header names the engine before it quotes any price", () => {
@@ -169,16 +181,11 @@ test("no model name or id from the pricing snapshot is hard-coded in the card", 
 		pricing.models.length > 1,
 		"pricing snapshot is empty — this check would be vacuous",
 	);
-	for (const m of pricing.models) {
-		assert.ok(
-			!panel.includes(m.name),
-			`ModelCostPanel.jsx hard-codes the model name ${JSON.stringify(m.name)}. The card must name only the model /health reports as served (or the user's pick), never a literal — a literal keeps claiming that model after the backend moves off it.`,
-		);
-		assert.ok(
-			!panel.includes(m.model_id),
-			`ModelCostPanel.jsx hard-codes the model id ${JSON.stringify(m.model_id)} — same problem as a hard-coded name.`,
-		);
-	}
+	assert.equal(
+		hardCodedModelProblem(stripComments(panel)),
+		null,
+		"ModelCostPanel.jsx hard-codes a model name or id from the pricing snapshot. The card must name only the model /health reports as served (or the user's pick), never a literal — a literal keeps claiming that model after the backend moves off it.",
+	);
 });
 
 test("mutation: a hard-coded served model is rejected", () => {
@@ -188,8 +195,9 @@ test("mutation: a hard-coded served model is rejected", () => {
 		"Debate society over the research corpus",
 		`Debate society over the research corpus, served by ${recommended.provider} ${recommended.name}`,
 	);
-	assert.ok(
-		HARD_CODED.includes(recommended.name),
+	assert.equal(
+		hardCodedModelProblem(HARD_CODED),
+		`hard-codes model name ${recommended.name}`,
 		`writing ${JSON.stringify(recommended.name)} into the card is not caught by the hard-coded-model check — it is guarding nothing`,
 	);
 });
--- a/ui/test/model-cost-line.test.js
+++ b/ui/test/model-cost-line.test.js
@@ -1,0 +1,248 @@
+// Claim + orientation guard for the Generate page's engine card
+// (ui/src/components/ModelCostPanel.jsx).
+//
+// Owner observation (screenshot, Generate page, under "Approve & Generate"):
+// the card read
+//
+//     MODEL & COST
+//     Running Amazon Nova Micro · $0.035 in / $0.140 out per 1M tokens
+//
+// Every word of that is true, but it leads with a billing rate on the one
+// surface whose first job is to tell a new user WHAT they are about to run.
+// The reworked header leads with the engine ("Strategy engine — debate
+// society over the research corpus, served by <model>") and keeps the rate as
+// a secondary line, labelled as the provider's published list price rather
+// than as the user's bill.
+//
+// This file pins three properties of that header:
+//
+//   1. it names the engine BEFORE it quotes any per-token price;
+//   2. the price is labelled as the provider's list price;
+//   3. it can never claim a model the backend is not running — no model name
+//      or model id from the pricing snapshot appears as a literal in the
+//      component, and the served model is still read from /health's
+//      `llm_model` through the shared fetchHealth cache, matched against the
+//      snapshot by `model_id` (the same source the card used before).
+//
+// Idiom is the repo's source-text one (oracle-copy.test.js,
+// roadmap-copy.test.js): `.jsx` is not importable under `node --test`, so the
+// component is read as text. Every check below is paired with a mutation that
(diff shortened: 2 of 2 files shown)


danielscoffee (approved):
Looks good — in the Generate honesty lane.

The old collapsed line was true and still wrong for the job: it led with a rate that reads like the viewer's bill on the surface whose first ask is "what am I about to run?". Leading with the engine and labelling the secondary line **Provider list price** fixes both. Keeping the model off literals (`fetchHealth` → `llm_model` → `model_id`) is the right honesty constraint, and the source-text guard with the reorder / hard-code mutations is the right shape for this repo.

One small accuracy note on the PR claim: Architecture does not use the exact phrase "Debate society over the research corpus". It does say **debate society** (sole generation path) and talks about the research corpus in the Debate stage — so the lead line is consistent in spirit, just not a verbatim lift. Fine as Generate copy; worth not overselling the match.

Drill-in / free-tier selection / snapshot note look untouched. Leaving merge to the owner as noted.

dbrowneup:

**Verdict: FIX-FIRST** — the copy change itself is honest and well-scoped, but one of the two shipped mutation tests is vacuous, and one PR-body claim is not true against the repo.

### Defects

**1. `ui/test/model-cost-line.test.js:196-207` — "mutation: a hard-coded served model is rejected" does not exercise the check it claims to.**
It builds `HARD_CODED` by string-splicing `recommended.name` into the source, then asserts:
```js
assert.ok(HARD_CODED.includes(recommended.name), ...)
```
That is true by construction — it asserts that a string you just concatenated contains the thing you concatenated. It never runs the hard-coded-model property against `HARD_CODED`. Delete or weaken the real check above it (e.g. drop the `m.name` loop, keep only `m.model_id`) and this "mutation" test still passes green. The PR body's *Mutation 2* red output came from a manual edit of the component, not from this file — so the repo rule (a guard is shown red on the input it rejects) is met in the description but not in the committed guard.
Exact edit — hoist the check into a `null`-or-reason function, the same idiom this file already uses for `orderingProblem` / `priceLabelProblem`:
```js
function hardCodedModelProblem(src) {
	for (const m of pricing.models) {
		if (src.includes(m.name)) return `hard-codes model name ${m.name}`;
		if (src.includes(m.model_id)) return `hard-codes model id ${m.model_id}`;
	}
	return null;
}
```
then `assert.equal(hardCodedModelProblem(panel), null, ...)` in the property test, and in the mutation test `assert.equal(hardCodedModelProblem(HARD_CODED), \`hard-codes model name ${recommended.name}\`, ...)`.

**2. PR body — "`Architecture.jsx` already describes it in exactly those words" is false.**
`grep -rni "over the research corpus" ui/src backend/archimedes` → no matches. `Architecture.jsx:417` says `<strong>debate society</strong> — the sole generation path`; the phrase *"over the research corpus"* is new copy introduced by this PR. The substance is fine (the debate society does retrieve from the corpus), but the body claims prior-art wording that does not exist. Edit the body to: "`Architecture.jsx:417` already calls it a *debate society*; *over the research corpus* is new copy here."

**3. PR body verification table — the `pytest -m "not integration" -q` row is still "still running at PR-open time".**
No Python files are touched, so this row should either read `— no Python files touched` (as the `ruff` row does) or carry a real exit code. Leaving an unresolved row in a verification table is the shape of a claim that outran its evidence.

**Nit (not blocking):** the hard-coded-name check scans the *whole* `panel` including comments, and `modelPricing.json` contains generic names (`Mistral Small`, `Nova Pro`). A future prose comment mentioning one of those words would fail the guard for no real reason. Scanning `stripComments(panel)` would be tighter.

### What holds up

- **Model-name honesty is genuinely by construction.** No model name or id is a literal in the component; `activeModel` comes from `/health`'s `llm_model` (`backend/archimedes/main.py:1624`) and is matched by `model_id` equality against the snapshot. Verified the served id is real: `backend/archimedes/services/llm_backend.py:34` `DEFAULT_CONVERSE_MODEL = "amazon.nova-micro-v1:0"` (overridable by `LLM_BEDROCK_MODEL`), and `ui/src/data/modelPricing.json` carries that exact id as `Nova Micro`. So "Running Amazon Nova Micro" was accurate and the reworked header stays accurate without hard-coding it.
- **Fail-closed on the unknown case.** With no `chosen`/`active`/`activeModel`, the header names no model and (`priced` null) quotes no price — it falls back to "Compare provider list prices across N models on Bedrock". Correct direction.
- **"Provider list price" relabel is the right fix** for the actual complaint: the number is the provider's published on-demand rate, not the viewer's bill, and it now says so.
- **One logical change** — copy + its guard, two files, no behaviour change, no new claims beyond the engine description.
- **No analogy branding**; paper trading is not mentioned.

### What I ran (exit codes)

- `gh pr view 1770 --json title,body,files,mergeable,mergeStateStatus` → 0 (`MERGEABLE` / `CLEAN`)
- `gh pr diff 1770` → 0 (read in full; 32+/7- in the component, 248 new test lines)
- `grep -rn "llm_model" --include="*.py" backend` → 0
- `grep -rn "BEDROCK_MODEL\|nova-micro\|LLM_MODEL" backend/archimedes` → 0
- `grep -rni "debate society" ui/src` → 0 (5 hits, none with the corpus phrase)
- `grep -rni "over the research corpus" ui/src backend/archimedes` → no matches
- `python3` read of `ui/src/data/modelPricing.json` → 0

### What I did NOT check (time box)

- **`ui` tests were not run.** `ui/node_modules` does not exist in this checkout, so `cd ui && node --test` was skipped per instructions — the "8 pass" and "921 pass, 0 fail" rows in the body are unverified here, including whether defect 1's vacuous test currently passes (by inspection it must).
- Did not render the Generate page or take a screenshot; the guard is a source-text test and I did not confirm the *rendered* header ordering in a browser.
- Did not inspect `/api/generate/quote`'s payload (`backend/archimedes/api/generate_routes.py:350`) — this PR does not touch pricing served to the client, only the snapshot-driven display, so it was out of the blast radius, but I did not confirm the snapshot's numbers still match what `/quote` charges.
- Did not run eslint, pytest, or any Python suite.

dbrowneup:
Fix pass on the FIX-FIRST review above. Head is now `fadd2864` (was `a31bf26f`). Three commits: `633a06fa` (Daniel Reis, via #1777), the merge `db52bc88`, and `fadd2864`.

---

### Defect 1 — the hard-coded-model "mutation" test proved nothing — **fixed**

`test("mutation: a hard-coded served model is rejected")` asserted `HARD_CODED.includes(recommended.name)` — true by construction, since `HARD_CODED` was built by splicing that same name in two lines earlier. It never ran the guard.

Fixed by **folding in #1777** (Daniel Reis, branch `cursor/hardcoded-model-mutation-68b7`, based on this branch), which hoists the check into a `null`-or-reason helper — the idiom this file already uses for `orderingProblem` / `priceLabelProblem` — and points both the property test and the mutation at it:

```js
function hardCodedModelProblem(src) {
	for (const m of pricing.models) {
		if (src.includes(m.name)) return `hard-codes model name ${m.name}`;
		if (src.includes(m.model_id)) return `hard-codes model id ${m.model_id}`;
	}
	return null;
}
```

Merged with `git merge --no-ff`, no conflicts, his authorship intact. **#1777 needs no separate merge — it can be closed as merged-via-parent** (commented there; I have not touched its state).

**The pairing shown real, both directions.** Component left clean, only the helper weakened (`m.name` branch deleted):

| state | result |
| --- | --- |
| before #1777 | **8 pass / 0 fail — exit 0** (guard disabled, mutation test did not notice) |
| after #1777 | **7 pass / 1 fail — exit 1** |

```
✔ no model name or id from the pricing snapshot is hard-coded in the card
✖ mutation: a hard-coded served model is rejected
  AssertionError: writing "Nova Micro" into the card is not caught by the
  hard-coded-model check — it is guarding nothing
  actual: null
  expected: 'hard-codes model name Nova Micro'
```

Two follow-ups on top, in `fadd2864`:

- **The vacuity claim itself was too broad.** The file header said *"Every check below is paired with a mutation that must turn it red"*. Not true of the two flat `assert.match` tests (the `/health` wiring and the drill-in), which carry no mutation. Narrowed to the three `null`-or-reason helpers, and it now says why the other two need none (no branch that can silently stop checking). Same narrowing applied to the PR body's Guard section.
- **Your nit, carried into the mutation too.** #1777 made the *property* test scan `stripComments(panel)`; the mutation still fed raw `panel`. It now feeds `stripComments(HARD_CODED)` — same input shape, so the two cannot diverge if a future comment in `ModelCostPanel.jsx` ever names a snapshot model.

### Defect 2 — the `Architecture.jsx` prior-art claim was false — **fixed in the body**

You were right: `grep -rni "debate society over the research corpus" ui/src backend/archimedes docs README.md` → **no match outside `ModelCostPanel.jsx`**. The body now says what is actually there, and that the compound is new:

> `Architecture.jsx:417` says *"Archimedes instead runs a **debate society** — the sole generation path"*, and `Architecture.jsx:224` describes that stage as *"proposer society drafts candidates from the research corpus and current market regime"*. The exact string *"debate society over the research corpus"* appears nowhere else in `ui/src` or `backend/archimedes` — this PR introduces it, compressing those two lines into one.

### Defect 3 — the unresolved pytest row — **fixed, run with exit code**

Run on `fadd2864`, in this worktree:

```
$ /opt/homebrew/.../envs/archimedes/bin/pytest -p no:cacheprovider -m "not integration" -q
=== 5644 passed, 4 skipped, 2 deselected, 328 warnings in 988.02s (0:16:28) ====
$ echo $?
0
```

Plus the one that is actually load-bearing for this diff, since the new copy makes a corpus claim:

```
$ pytest -p no:cacheprovider -q backend/tests/test_corpus_claim_integrity.py
26 passed, 1 warning in 1.66s
$ echo $?
0
```

The body's verification table now carries both rows with their exit codes; no row says "still running".

`ruff` stays `— no Python files touched`; `git diff --name-only` against the merge base is exactly the two `ui/` files.

### Nit — scan `stripComments(panel)` — **taken** (#1777, plus the mutation, above).

---

### Everything re-run on `fadd2864`

| check | result |
| --- | --- |
| `cd ui && node --test test/model-cost-line.test.js` | 8 pass, 0 fail — **exit 0** |
| `cd ui && node --test` | 921 tests, 921 pass, 0 fail, 0 skipped — **exit 0** |
| `npx eslint src/components/ModelCostPanel.jsx test/model-cost-line.test.js` | no output — **exit 0** |
| mutation 1 — reorder the two header `<div>`s | 7 pass / 1 fail — **exit 1** |
| mutation 2 — hard-code `, served by Amazon Nova Micro` | 7 pass / 1 fail — **exit 1** |
| mutation 3 — gut `hardCodedModelProblem`, component clean | 7 pass / 1 fail — **exit 1** |
| mutation 4 — `Provider list price` into a comment, render `Your cost` | 7 pass / 1 fail — **exit 1** |
| `pytest -q backend/tests/test_corpus_claim_integrity.py` | 26 passed — **exit 0** |
| `pytest -m "not integration" -q` | 5644 passed, 4 skipped, 2 deselected in 16:28 — **exit 0** |

All four mutations reverted; the committed files are the clean ones (`git status` clean at commit time).

### Not done, on purpose

- **`ModelCostPanel.jsx` is still not in `test_corpus_claim_integrity.py`'s `PUBLIC_SURFACES`** (your non-blocking note 3). It is one line, and the file is clean against `OVERCLAIM_PATTERNS` today — but registering a new claim surface deserves its own adversarial demo (an over-claim written into the card, shown red), and it is guard-coverage hygiene rather than part of this copy fix. Worth its own PR; flagged in the body.
- **"Strategy engine" vs "Strategy generation engine"** (your note 1) — left as the owner's call, since the string was prescribed. Now written up in the body under *Known trade-off*, together with your note 2 (the word "cost" no longer appears above the fold), so the owner sees both before vetting the screenshot.

Body updated so every claim matches the diff. Still **do not merge — owner vets first.**

danielscoffee (approved):
Fix-pass on `fadd2864` closes the FIX-FIRST defects: `hardCodedModelProblem` is real (property + mutation share it; mutation uses `stripComments`), Architecture prior-art claim corrected, verification table carries exit codes. Copy + guards look solid. Remaining notes (PUBLIC_SURFACES; "Strategy generation engine") are owner/follow-up as written. Still owner-vet before merge.

Issue #1768, opened 2026-09-01
[infra][cloudfront] Session-dependent paths (/app/*, /sign-in) must use CachingDisabled with cookies forwarded — belt to #1767's nginx no-store


dbrowneup:
Follow-up to PR #1767 (the 2026-09-01 sign-in ping-pong). #1767 fixes it at the origin: nginx now sends `Cache-Control: private, no-store` on the gated `/app` locations and the `@sign_in` redirect, and CloudFront honours it because the `html` cache policy has `min_ttl = 0`.

That leaves the structural hazard in place: the default behavior in `infra/cloudfront.tf` applies the `html` cache policy (`default_ttl = 60`, `cookie_behavior = "none"`) to every path that is not `/api/*`, `/health*`, `/events/*`, `/assets/*`, `/static/*` — including `/app/*` and `/sign-in`, whose responses depend on the session cookie. Any future nginx edit that drops the header, or any new gated path, reintroduces "an anonymous visitor's 302 is replayed to the next signed-in user".

**Change:** add `ordered_cache_behavior` blocks for `/app/*` (and `/app`) and `/sign-in*` using `data.aws_cloudfront_cache_policy.caching_disabled` + the existing `all_viewer` origin request policy (cookies + headers forwarded), before the default behavior. Keep `/assets/*` and `/static/*` on `static_assets`. `terraform plan` must show only those behaviors added; apply is an owner step.

**Guard:** extend `backend/tests/test_cloudfront_health_uncached.py`'s pattern to assert the new behaviors exist and point at `caching_disabled` (mutation: delete one block → red).

Evidence for the class: PR #1767's body (x-cache: Hit on the 302 with a session cookie; 165 get-session calls in 45 min for one user).

dbrowneup:
Closed by #1772 (merged d545fa41) + `terraform apply` run from the owner's session at 01:33Z under the owner's standing authorization.

[...]

Issue #1520, opened 2026-08-30
[infra] /health is served from the CloudFront cache — including its 504s


onder-akkaya:
## Summary

`/health` is served from the CloudFront cache. It matches no `ordered_cache_behavior`, so it falls through to `default_cache_behavior`, which uses the `html` cache policy (`infra/cloudfront.tf:138-143` — `default_ttl = 60`, `max_ttl = 60`). The origin sends no `Cache-Control` header on `/health`, so CloudFront applies `default_ttl` and caches it for 60 seconds.

A cached health check is not a health check. `CLAUDE.md` tells every reader that current status comes from the live system — "`/health`, `GET /api/config/contracts`, and the deploy history" — and for up to a minute `/health` answers from a CDN edge instead.

## Measured

Four probes just now, same POP:

```
probe 1  HTTP/2 200  x-cache: Miss from cloudfront  total=1.11s
probe 2  HTTP/2 200  x-cache: Hit from cloudfront   total=0.13s
probe 3  HTTP/2 200  x-cache: Hit from cloudfront   age: 1  total=0.06s
probe 4  HTTP/2 200  x-cache: Hit from cloudfront   age: 1  total=0.10s
```

A later probe showed `age: 17` with no `Cache-Control` on the response at all.

Earlier in the same session, the error side of this:

```
health #1  code=504  total=90.40s   x-response-time-ms: 3203.0
health #2  code=504  total=0.12s
health #3  code=504  total=0.07s
```

Note the first one: `X-Response-Time-Ms: 3203.0` (the header #1505 added) says the **application answered in 3.2 seconds**, while the client saw a 504 at 90.4s. The two that followed came back in 0.12s and 0.07s — far too fast to have reached the origin. Those are cached 504s. There is no `custom_error_response` block anywhere in `infra/cloudfront.tf`, so CloudFront's default error-caching applies.

## Why it matters

1. **The signal can be wrong in the dangerous direction.** A cached `200 {"status":"ok"}` keeps reporting healthy for up to 60s after the origin starts failing. That is the fail-soft shape `docs/architectural-principles.md` warns about: the degraded state has to be a loud, visible absence, and here it is a plausible substitute served from an edge.

2. **It amplifies a blip into a window.** One slow origin response becomes an instant 504 for every viewer hitting that POP until the error TTL expires. That is what the three probes above show, and it makes a brief origin hiccup look like a sustained outage.

[...]

PR #1772, opened 2026-09-01
[infra][cloudfront] /app, /app/* and /sign-in* bypass the edge cache — the structural half of #1767

dbrowneup:
**Structural follow-up to #1767 (the 2026-09-01 sign-in ping-pong), issue #1768.**

#1767 fixed the outage at the origin — `Cache-Control: private, no-store` on the gated `/app` locations and the `@sign_in` redirect, honoured because the `html` policy has `min_ttl = 0`. That is live and it is what stopped the thrash. **It is necessary but not sufficient.**

The hazard it left in place: `infra/cloudfront.tf`'s `default_cache_behavior` applies the `html` policy (`default_ttl = 60`, `cookie_behavior = "none"`) to every path that is not `/api/*`, `/health*`, `/events/*`, `/assets/*` or `/static/*` — `/app/*` and `/sign-in` included. The gated pages under `/app` answer differently depending on the session cookie; the edge keys without it. The origin header is opt-in per response, in a different file, and its guard enumerates the locations that gate on `auth_request /_auth_session`; it catches a deleted header but cannot see a gated response produced anywhere else. One new gated path and the edge is back in the business of replaying one viewer's redirect to another:

```
GET /app/generate                            -> 302  x-cache: Miss from cloudfront
GET /app/generate                            -> 302  x-cache: Hit from cloudfront
GET /app/generate  (Cookie: __Secure-...=…)  -> 302  x-cache: Hit from cloudfront   # the bug
```

**Change.** Three `ordered_cache_behavior` blocks — `/app/*`, `/app`, `/sign-in*` — on `data.aws_cloudfront_cache_policy.caching_disabled`, with the existing `all_viewer` origin request policy and the same response headers policy the default behaviour uses. Opt-out per path: with CachingDisabled the edge has nothing to replay even when the origin forgets to say so. The `html` policy is untouched — these paths simply stop matching it, the same shape as the `/health` behaviours (#1520).

### The public `/app` carve-outs are de-cached deliberately

Not every path these patterns cover is session-dependent, and the PR used to claim otherwise. `nginx/nginx.conf:293-326` is the anonymous-browse carve-out block (#1194 revision d, the owner's product call), and nginx picks the **longest** matching prefix, so it wins over the gated `^~ /app`. Bare `/app` (the SPA's alias for Explore), `/app/explore`, `/app/leaderboard`, `/app/corpus` and `/app/strategy/*` are **public**: no `auth_request`, no `error_page 401 = @sign_in`, the same shell for every viewer. `/app/*` and `/app` pull them off the 60s `html` policy along with the gated pages.

**That is the owner's ruling (2026-09-01), not an oversight**, and the code comment, the runbook and the guard now all say so:

- **Cost.** Every anonymous visitor to those pages reaches the origin instead of being answered at the edge for up to 60s. What they get is the ~4 KB `ui/index.html` shell, read straight off nginx's disk (`root /usr/share/nginx/html; try_files $uri $uri/ /index.html`) with **no backend call**; the SPA's data then comes over `/api/*`, which was never cached at the edge anyway.
- **Benefit.** Promoting one of those carve-outs to gated later cannot reintroduce #1767's cached anonymous 302. And a per-carve-out list here would make `infra/cloudfront.tf` a **fourth** copy of the anon-page list that `nginx/nginx.conf` and `ANON_APP_PAGES` in `ui/src/routes.js` already have to keep in lockstep — a fourth copy needing its own lockstep guard, to save origin hits on a static file.

The rest of the decisions, all argued in the comment above the blocks:

- **`all_viewer` is load-bearing, not decoration.** CachingDisabled alone forwards no cookies, so `auth_request` would see an anonymous request from every viewer and `/app` would be permanently signed out. Cookies, headers and query strings forwarded means `?next=` still reaches the SPA.
- **`allowed_methods` and the response headers policy mirror the default behaviour.** What changes for these paths is caching and edge compression — not which verbs the edge forwards, and not whether HSTS/frame-options reach the gated half of the site.
- **`compress = false` is a description, not a free lunch.** The AWS-managed CachingDisabled policy sets `EnableAcceptEncodingGzip = false` (Brotli likewise), and CloudFront compresses only when the behaviour's `compress` flag **and** the cache policy's accept-encoding settings are on — so `compress = true` on these blocks would be inert. The real cost lands at the origin: nginx sets `gzip off` on every one of these locations, so the shell reaches the browser uncompressed, **~4.0 KB where gzip would have sent ~1.4 KB** (measured on `ui/index.html`: 4042 bytes, 1385 gzipped).
- **Ordered ahead of `*.js` / `*.css`, after `/static/*`.** Those patterns match any path with that suffix, `/app/x.js` included, and bind it to `static_assets`: 1h at the edge, keyed without cookies — the same defect with a longer TTL. Nothing is served under `/app/*.js` today, which is why the ordering has to be right before someone puts one there.
- **`/sign-in*` is prophylactic and the comment says so.** Today `/sign-in` is the identical anonymous SPA shell for every viewer, so caching it harms nothing; it is the other end of the redirect loop and one nginx edit away from being session-dependent.

**Named residual, not papered over:** nginx's `^~ /app` is a prefix match, so `/appfoo` is gated too and still resolves to the default behaviour. A single `/app*` pattern would cover it at the cost of swallowing any future top-level path starting with "app" (`/apply`, `/app-store`). What is left uncovered is an anonymous 302 for a path the router does not have.

**Guard.** `backend/tests/test_cloudfront_session_paths_uncached.py` reuses the HCL reader and the CloudFront first-match resolver from the #1520 sibling rather than growing a second copy, and pins the behaviours, their cache/origin-request/response-headers policies, their **position** in the list, the paths' resolution (`/app/nonsense` included — the gated catch-all answers unknown paths identically), and the comment's claims. `TestThePublicCarveOutsAreDeCachedOnPurpose` pins the **decision**: the four carve-outs and bare `/app` resolve to CachingDisabled, and the comment must name them as public and state the cost — so reversing the ruling is a red test, not a quiet cache "optimisation". Hermetic: one file in the repo, no AWS, no terraform binary.

Mutations run (each restored afterwards):

| mutation | result |
|---|---|
| delete the `/sign-in*` block | 16 failed — `no ordered_cache_behavior for '/sign-in*'`, and `/sign-in resolves to '<default_cache_behavior>'` |
| rebind `/app/*` to the `html` policy | **9 failed** — `test_the_behaviour_binds_caching_disabled['/app/*']`, the four gated `/app/` resolutions (`generate`, `insights`, `library`, `nonsense`), and all four public carve-outs (`/app/explore`, `/app/leaderboard`, `/app/corpus`, `/app/strategy/abc123`). Bare `/app` stays green — it has its own behaviour, which is exactly why there are two patterns and not one. |
| move `/app/*` behind `*.js` / `*.css` | 5 failed — `'*.js' is ordered in front of '/app/*'` |
| strip `/app`'s origin-request + response-headers policies, narrow its methods | 3 failed — `'/app' forwards None, not aws_cloudfront_origin_request_policy.all_viewer.id` |
| delete the file header's new bullet | 1 failed in the #1520 sibling — `the file header lists the never-cached paths but omits [...]` |

Restored: `102 passed` (this suite + both siblings, exit 0). `terraform fmt -check -recursive` clean, `terraform validate` → *Success! The configuration is valid.* `ruff check` + `ruff format --check` clean. `docs_links.py` 0 broken, `docs_index.py` 0 orphaned.

**Docs.** `docs/runbooks/cloudfront-cache-behaviour-apply.md` (new, indexed in both `docs/README.md` and `docs/runbooks/README.md`): merging this changes nothing on the live stack — CI runs `fmt` + `validate` only, never `plan`. After merge the owner runs `infra/apply.sh --apply`, and the expected plan is **exactly the three new behaviours on `aws_cloudfront_distribution.main` and no other change** (`Plan: 0 to add, 1 to change, 0 to destroy`). It carries the mid-list-insert rendering caveat **above** the pseudo-plan it warns about (terraform may render the index shift as changes to the trailing behaviours — check that every rendered change is a re-ordering only), the ruling on the public carve-outs, the post-apply invalidation (`/app*` wildcarded, because the `html` policy forwards query strings), the `x-cache` verification — including `curl -sI https://archimedes-arc.com/app/explore` **twice, Miss both times, expected and deliberate** — and the rollback for the dangerous failure mode: `/app` signed out for everyone if the origin request policy is wrong.

**Not in this PR:** the apply itself, any change to the `html` policy's TTLs, any change to `/api/*`, `/events/*` or the static-asset behaviours, and any per-carve-out cache behaviour (ruled out above).


dbrowneup:

**FIX-FIRST.** The precedence, the policy bindings and the "no apply ran" framing all check out. One claim in the PR body and the code comment is false against `nginx/nginx.conf`, and it has a real (small) consequence the guard is blind to.

**1. `infra/cloudfront.tf` (the `/app` and `/app/*` blocks, and PR body ¶2/¶3) — "session-dependent" is untrue for five of these paths, and the change silently de-caches the public half of `/app`.** `nginx/nginx.conf:288-320` is the anonymous-browse carve-out block (#1194 rev d, owner's product call), and nginx picks the LONGEST matching prefix, so it wins over the gated `^~ /app` at :372:

- `location = /app` (:296) — comment says verbatim "Bare /app is the SPA's alias for Explore — **public** with it". No `auth_request`.
- `^~ /app/explore` (:302), `^~ /app/leaderboard` (:307), `^~ /app/corpus` (:312), `^~ /app/strategy` (:317) — all ungated, no `auth_request`, no `error_page 401 = @sign_in`.

So the new `/app` behaviour covers a page that is public by design, and `/app/*` pulls Explore/Leaderboard/Corpus/strategy-detail off the 60s `html` policy onto CachingDisabled. Those are the highest-traffic anonymous pages on the site; every visitor now goes to origin. The PR body's "their responses depend on the session cookie" and the block comment's "nginx's `auth_request` gate on `^~ /app` answers an anonymous visitor" are true for `/app/generate` and `/app/library`, false for the five above.

Exact edit, preferred (keeps the fix, restores the status quo for the public pages) — insert AHEAD of the new `/app/*` block, and DELETE the bare `/app` block entirely:

```hcl
  # Anonymous-browse carve-outs — mirror of nginx.conf:288-320 (#1194 rev d).
  # Ungated, identical shell for every viewer: they stay on the html policy.
  # Ordered ahead of /app/* below, which is the gated catch-all.
  ordered_cache_behavior {
    path_pattern               = "/app/explore*"   # + /app/leaderboard*, /app/corpus*, /app/strategy*
    ...                                            # copy the default behaviour's html-policy block verbatim
  }
```

and drop `path_pattern = "/app"` (exact match, ungated, public).

If you'd rather keep all of `/app` uncached (defensible: a carve-out could be gated later), that is fine — but then the prose has to stop saying these paths are session-dependent and say instead "the public `/app` carve-outs are de-cached deliberately; the cost is N origin hits, the benefit is that promoting a carve-out to gated cannot reintroduce #1767." Either way the current text is a claim the diff does not support.

**2. `backend/tests/test_cloudfront_session_paths_uncached.py` — `STILL_CACHED` is the anti-goal test and it omits exactly the paths that regressed.** It pins `/`, `/explore`, `/index.html`, two `/assets/` bundles. Add the carve-outs so this is a guard rather than a restatement:
```python
    "/app/explore": HTML_POLICY,
    "/app/leaderboard": HTML_POLICY,
    "/app/corpus": HTML_POLICY,
    "/app/strategy/abc123": HTML_POLICY,
```
Run it against the current head first and confirm it goes RED (it will — they resolve to `/app/*` → CachingDisabled today). That is the input that should fail the check, and nothing in the suite currently rejects it. Related: `nginx.conf:292` says the carve-out list "must stay in lockstep with `ANON_APP_PAGES` in `ui/src/routes.js`" — this PR makes `infra/cloudfront.tf` a third copy of that list with no guard binding it to the other two.

**3. `infra/cloudfront.tf`, the `compress = false` justification — "gives up nothing" is not right.** All these nginx locations do set `gzip off` (:298/:303/:308/:313/:318/:375), which is why the default behaviour's `compress = true` was the only compressor for the SPA shell. `compress = false` + `gzip off` = the HTML now reaches the browser uncompressed. (CloudFront largely won't compress an uncached response anyway, so I'd keep `false` — but change the comment from "gives up nothing" to "gives up the edge gzip on the shell; nginx sends it uncompressed, and CloudFront does not compress what it does not cache.") The PR body's "the ONLY thing that changes for these paths is caching" is likewise not exactly true, and `compress` is not mentioned in the body at all.

**Checked and clean:** ordering — the three blocks land after `/static/*` and before `*.js` / `*.css`, and none of `/health`, `/health/*`, `/api/*`, `/events/*`, `/assets/*`, `/static/*` can capture an `/app` or `/sign-in` path (first-match verified by hand against the file); `cache_policy_id` on all three = `data.aws_cloudfront_cache_policy.caching_disabled.id`, same source the `/health` behaviours use; `all_viewer` really does forward what the comment says (`cookie_behavior = "all"`, `header_behavior = "allViewerAndWhitelistCloudFront"`, `query_string_behavior = "all"`), so `auth_request` and `?next=` survive; `response_headers_policy_id` and `allowed_methods` match `default_cache_behavior` exactly; the `html` policy itself is untouched and `/`, `/explore`, `/index.html` still resolve to it; `/app/*` vs bare `/app` as two patterns is correct CloudFront semantics; the body does NOT claim `terraform apply` ran and explicitly reserves it for the owner — accurate.

**What I ran:** `gh pr view 1772 --json ...` (exit 0), `gh pr diff 1772` (exit 0), `grep`/`sed` over `infra/cloudfront.tf` + `nginx/nginx.conf` on a local main checkout (exit 0).

**NOT checked, for lack of time:** `terraform validate` — `infra/.terraform` does not exist in my checkout and I was instructed not to `init`, so the body's `fmt`/`validate` results are unverified; the five mutation rows in the body (not re-run — I did not restore-and-mutate the file); the new runbook `docs/runbooks/cloudfront-cache-behaviour-apply.md` beyond its existence; the docs index/link checks; whether `ui/src/routes.js`'s `ANON_APP_PAGES` matches the nginx carve-out list today; the `TestTheCommentTellsTheTruth` class body past the first test.

dbrowneup:
Fixed at `467c77bc`. The owner's ruling on this review: **keep the behaviour exactly as built, make the prose true.**

All of `/app` — bare `/app`, `/app/*`, plus `/sign-in*` — stays on CachingDisabled. The public carve-outs (`/app`, `/app/explore`, `/app/leaderboard`, `/app/corpus`, `/app/strategy/*`) are de-cached **deliberately**, for two reasons: a carve-out promoted to gated later must not be able to reintroduce #1767's cached anonymous 302 (the product has already moved pages across that line), and a per-carve-out list in terraform would be a fourth copy of the anon-page list that `nginx/nginx.conf` and `ANON_APP_PAGES` in `ui/src/routes.js` already have to keep in lockstep — a fourth copy needing its own lockstep guard, bought to save origin hits on a static file.

**1 — "session-dependent" is untrue for five of these paths.** Correct, and the prose was the defect, not the diff. `infra/cloudfront.tf`'s file header and the block comment no longer call these paths session-dependent. The comment now says plainly that `nginx/nginx.conf:293-326` is the anonymous-browse carve-out block (#1194 rev d, owner's product call), that nginx's longest-prefix match puts it ahead of the gated `^~ /app`, and that bare `/app` and the four carve-outs are **PUBLIC** — no `auth_request`, no `error_page 401 = @sign_in`. Both sides of the trade are written down rather than implied:

> **Cost.** Every anonymous visitor to those pages reaches the origin instead of being answered at the edge for up to 60s. What they fetch is the ~4 KB `ui/index.html` shell, read straight off nginx's disk (`root /usr/share/nginx/html; try_files $uri $uri/ /index.html`) with no backend call; the SPA's data then comes over `/api/*`, which was never cached at the edge anyway.
>
> **Benefit.** Promoting one of those carve-outs to gated later cannot reintroduce the cached anonymous 302 of #1767/#1768.

The PR body's ¶2/¶3 are rewritten to match, with the ruling as its own section rather than a footnote. The runbook records it too.

**2 — the anti-goal test omits exactly the paths that regressed.** Taken, inverted to match the ruling. Rather than adding the carve-outs to `STILL_CACHED` (which would now assert the opposite of the decision), there is a new `TestThePublicCarveOutsAreDeCachedOnPurpose` that **pins the decision**: `/app/explore`, `/app/leaderboard`, `/app/corpus`, `/app/strategy/abc123` and bare `/app` must resolve to CachingDisabled, and two prose tests require the comment to name them as public and to state the origin cost — so a later "optimisation" that quietly puts them back on the `html` policy is a red test, not a surprise in production. The class docstring names the ruling, the date, and #1194 rev d, so the next reader finds the reasoning instead of re-deriving it.

Mutation, run and restored: **rebind `/app/*` to `aws_cloudfront_cache_policy.html.id` → 9 failed** — `test_the_behaviour_binds_caching_disabled[/app/*]`, the four gated `/app/` resolutions (`generate`, `insights`, `library`, `nonsense`), and all four public carve-outs. Bare `/app` stays green, because it has its own behaviour — which is exactly why there are two patterns and not one. Restored: 102 passed, exit 0.

**3 — `compress = false` "gives up nothing" is not right.** Correct, and it was wrong in a second way the review was generous about: the managed CachingDisabled policy sets `EnableAcceptEncodingGzip = false` (Brotli likewise), and CloudFront compresses only when the behaviour's `compress` flag **and** the cache policy's accept-encoding settings are on — so `compress = true` here would be inert, not a choice foregone. The comment now says that, and names the real cost, which lands at the origin: nginx sets `gzip off` on every one of these locations, so the shell reaches the browser uncompressed — **~4.0 KB where gzip would have sent ~1.4 KB** (measured: `ui/index.html` is 4042 bytes, 1385 gzipped). "The ONLY thing that changes is caching" is now "what changes is caching and edge compression", in both the comment and the body, and `compress` is in the body.

**Also, from the runbook half of the review:** the mid-list-insert rendering caveat now sits **above** the fenced pseudo-plan it warns about, so it is read before the plan is read rather than after. Post-apply verification gains `curl -sI https://archimedes-arc.com/app/explore` twice → `x-cache: Miss` both times, labelled as expected and deliberate, pointing at the pin test as the thing that would have caught a reversal earlier.

**Two of your "NOT checked" items, now checked:** `terraform validate` → *Success! The configuration is valid.* (`infra/.terraform` exists in this worktree), and `terraform fmt -check -recursive` is clean. `ruff check` + `ruff format --check` clean, `docs_links.py` 0 broken, `docs_index.py` 0 orphaned. The other four mutation rows in the body are carried forward from the original run and were **not** re-run this pass — only the `/app/*` rebind was.

**One deliberate deviation from your text:** you cited `nginx/nginx.conf:288-320` for the carve-out block; the block is actually at **293-326** on this branch (288-291 is the tail of `@sign_in`). Since the whole point of this pass is prose that is true, the comment, the test and the runbook all cite 293-326.

Still not merged — this is the owner's call, and `terraform apply` is a separate owner step after it.

Reply via the owner's session

Issue #1436, opened 2026-08-20
[infra][perf] API latency tail: p95 6.2s / p99 16.8s while p50 is 46ms — and the p95>2s alarm sits OK through it


dbrowneup:
## Summary

Median latency is excellent and the tail is not. Real user-facing numbers from the ALB, six-hour window, 2026-08-20:

```
p50  0.046 s
p95  6.222 s
p99 16.789 s
```

30-minute bins reached maxima of **17.3 s** and **30.0 s**. Separately, a cold `GET /api/strategies/` measured **78.9 s** to first byte (warm: 0.59 s and 0.75 s on the two immediately following calls), and `GET /api/selection-bias/gate` takes **13.2 s even warm**.

A p50 of 46 ms with a p99 of 17 s means most requests are static assets and the API calls that actually render a page are living in the tail.

## Where it lands for a user

`/app/library` issues three calls on mount (`Strategies.jsx`): `/api/strategies/`, `/api/strategies/generated`, `/api/selection-bias/gate`. Until they resolve the page paints `Generated (0)` / `Examples (0)` and the "NO GENERATED STRATEGIES YET" body as statements of fact — see the reproduction comment on #1356. That empty-state rendering is a separate (fixed) concern; **this** issue is that the window it is exposed for can be tens of seconds rather than milliseconds.

## ~~The existing alarm cannot be relied on to catch it~~ — CORRECTED: it fires constantly

**The original text of this section was wrong and is retracted.** I claimed the alarm was tuned so it could not catch this, based on observing it in `OK` at one moment. That was a between-flaps read and I drew a false inference from a single sample.

`describe-alarm-history` for `archimedes-alb-target-latency-high`, last 24 h:

```
27 OK -> ALARM transitions
current state: ALARM since 2026-08-20T16:27:48-05:00
```

Sample of the flapping:

```
16:27:48  OK -> ALARM
16:02:48  ALARM -> OK
15:57:48  OK -> ALARM
10:58:09  ALARM -> OK
10:56:09  OK -> ALARM      <- 2 minutes
 8:59:09  ALARM -> OK
 8:54:09  OK -> ALARM      <- 5 minutes
 2:51:09  OK -> ALARM
 2:49:09  ALARM -> OK      <- 2 minutes
```

[...]

Issue #1324, opened 2026-08-20
[frontend] Library still ships a live "Published" tab wired to the hidden Marketplace API — dead door missed by #1266


dbrowneup:
## Summary

PR #1266 hid Marketplace, Publish, Subscriptions, Portfolio/vaults and Learnings behind `ROADMAP_SURFACES_ENABLED` (default off), and its dead-door audit concluded "every remaining call site targeting a hidden page is either inside an already-hidden component or backstopped by the routing layer."

The **Library page is a hole in that audit.** It still ships a live, clickable **"Published (N)"** tab that fetches from the hidden Marketplace API on every load.

## Evidence

`ui/src/components/Strategies.jsx`:

- **L617-621** — the tab's data is fetched unconditionally on mount, alongside the real tabs:
  ```js
  const [seedRes, genRes, gateRes, publishedRes] = await Promise.allSettled([
    ...
    apiGet('/api/marketplace/my-published'),
  ])
  ```
- **L680-683** — the tab is rendered with a live count:
  ```js
  className={`tag ${activeTab === 'published' ? 'tag-accent' : 'tag-muted'}`}
  onClick={() => setActiveTab('published')}
  ...
  Published ({published.length})
  ```
- **L806** — `{activeTab === 'published' && ( ... )}` renders its panel.

None of this is behind `featureEnabled()` or `ROADMAP_SURFACES_ENABLED`.

`grep -n "ROADMAP_SURFACES_ENABLED\|featureEnabled" ui/src/components/Strategies.jsx` → **no match.**

## User impact

A signed-in user on `/app/library` sees a third tab advertising a Marketplace capability that PR #1266 deliberately removed from the product. Clicking it opens a panel for a feature with no publish flow, no marketplace page and no subscription surface to reach it from — the exact "dead door" #1266 set out to eliminate. It also means every Library load makes a request to a roadmap subsystem.

## Acceptance criteria (machine-checkable)

1. With the flag off (the default), the tab is absent and the request is not made:
   ```
   grep -n "my-published" ui/src/components/Strategies.jsx
   ```
   → the call site must be gated by `ROADMAP_SURFACES_ENABLED`, not unconditional.
[...]

Issue #1187, opened 2026-07-28
[claim-integrity] Leaderboard numeric fields still fall back to migrated fixture data — NOT fixed by #1172


dbrowneup:
## Summary

Six of the twenty graded rows on the public leaderboard are served with **numeric values that trace back to migrated test-fixture data**, not live computation. This was assumed to be fixed by PR #1172. **It is not** — verified by diffing that PR.

This is a live claim-integrity defect on the product's most important public surface, and it needs its own tracking issue so it doesn't get lost behind the assumption that #1172 covered it.

## Evidence that #1172 does not fix it

`gh pr diff 1172` touches `backend/archimedes/api/strategies_routes.py` at **exactly one hunk — `@@ -483,8 +483,22 @@`**. The fallback logic lives at roughly lines **154–179** and is untouched:

- `deflated_sharpe_ratio` — ~154-160
- `dsr_p_value` — ~161-165
- `pbo_score` — ~166-169
- `out_of_sample_sharpe` — ~171-179

Each falls back to `s.<field>` / `bt.<field>` whenever `_live_rigor_results_for_strategies` cannot produce a live result (fewer than 10 persisted daily returns, or a cohort-compute failure). Those DB columns were populated by a migration from `backend/tests/fixtures/backtest_fixtures_snapshot.json` (PR #863).

#1172 fixes a **different and also real** problem — cohort agreement between the list route, the passport, and the deploy gate. Both matter; neither substitutes for the other.

## Independent confirmation from the live API

Querying `GET /api/leaderboard` and bucketing by `pbo_score` shows the contamination directly — the values cluster onto fixture constants rather than varying per strategy:

```
pbo=0.112821  n=14  passes=3   <- live-computed
pbo=0.299456  n=3   passes=0   <- fixture constant
pbo=0.065113  n=2   passes=0   <- fixture constant
pbo=0.083838  n=1   passes=0   <- fixture constant
```

**Mitigating fact:** all three passing rows are live-computed. Every stale row is a *failing* pairs-trading variant. So the contamination currently inflates the denominator without corrupting which strategies pass — but that is luck, not design, and it will not stay true.

## Why the fixture is not a safe fallback source

The fixture is a fossil. It cannot be reproduced by any single code version:

- `num_trials_in_selection` takes ten distinct values across 23 rows because the generator is add-only and stamps library-size-at-time-of-add — it records *when a strategy was added*, not any coherent trial count.
[...]

PR #1779, opened 2026-09-02
[ux] Examples tab: stop presenting curated references as failures

dbrowneup:
## What the owner saw

Library → **Examples (34)**: every hand-curated card carried the pill
`Reference only — gate failed` **and** a red `blocked` pill, next to metrics.
Owner: *"lots of things don't look good."*

Under the new grading model — the passport is the verdict of record, and a
curated row is not graded until its backtest is actually run — the Examples tab
was presenting the reference library as 34 failures.

This PR takes the **copy** and the **`blocked` pill**. The
`Reference only — gate failed` pill is `statusTag` / `statusLabel`, which
PR #1792 (`dbrowneup/verdict-of-record-a`) owns; nothing here touches those
helpers, or `coerceGenerated`. See **Merge order** below — that is a hard
ordering constraint, not a nice-to-have.

## Root cause

`blocked` is not a research verdict — it is a **deployability** verdict.
`DeployabilityChip` (ui/src/components/Strategies.jsx) grades a row off
`/api/selection-bias/gate`, and `blocked_by_floor` is the same field the vault
deploy gate reads (`backend/archimedes/api/vaults_routes.py::_deployable_levels`).
Its own title says so: *"Fails an always-on correctness floor — cannot deploy at
any level."*

Vaults are out of the MVP cut (#1266, `ROADMAP_SURFACES_ENABLED`, off unless
`VITE_ROADMAP_SURFACES=true`). So the shipped build was grading hand-curated
reference implementations against a capability it does not have, in red, on the
one tab whose whole job is to teach the card format.

The metrics beside them have the same shape of problem, and the API already
knew: `GET /api/strategies/` serves `display_metrics_source` per row
(`strategies_routes.py::_display_metrics_source`), whose own comment says
`strategy_record` is **not** "measured" — for the curated library it traces to
the backtest-fixture snapshot (#1187). The UI dropped the field, so a fixture
value and a real persisted backtest rendered as the same black number.

## The change

**1. The deployability chip does not render unless `ROADMAP_SURFACES_ENABLED`.**
The *whole* chip, not just the red branch. Suppressing only `blocked` would
leave the green `deployable` claim standing on a surface where nothing can be
deployed — keeping the flattering half of a vault verdict and hiding the rest.
Every branch (`blocked` included) is intact and comes back with
`VITE_ROADMAP_SURFACES=true`; this is a gate, not a deletion, and there is a
test pinning that ordering.

**2. The gate fetch is gated on the same flag.** `/api/selection-bias/gate`
recomputes the whole cohort gate (the slowest request on the page, ~8-10s, and
504-ing against prod on 2026-08-31). Paying for it to annotate something that no
longer renders is the exact defect #1324 was filed against — same one-line
ternary as `publishedPromise`, and the empty payload leaves `deployMap`,
`gateLoading` and the `gateError` banner behaving exactly as before.

**3. The Examples intro is rewritten to the new identity** — reference
implementations to learn the card format; *not* fusion output; **not a
scoreboard**; graded only once a backtest has been run for it, so an absent
verdict is not a failure; a strategy's passport is its verdict of record; and
where the numbers come from (`fixture`, or `placeholder` where only the
strategy module's declared constants exist).

**4. `ui/src/metricsSource.js` (new)** turns `display_metrics_source` into a
one-word mark under the row's Sharpe (table + card): `fixture` for
`strategy_record`, `placeholder` for `stub_placeholder`.

Deliberately an **allow-list**, and the omissions are the honest part:
`persisted_backtest` is a real run and stays unmarked — marking it "fixture"
would be a fresh false claim, worse than the silence being fixed — and an
absent/unknown value (a generated row reaches the table through
`coerceGenerated`, which does not carry the field) makes no claim in either
direction. That is also why the intro says *"where a row's metrics trace to a
stored fixture snapshot rather than to a backtest run here, the row is marked
fixture"* rather than declaring every number on the tab a fixture: the paragraph
never generalises about rows it cannot see.

## Merge order — depends on PR #1792 (PR-A)

**Land this after or with #1792 (`dbrowneup/verdict-of-record-a`).**

This PR's intro copy — *"until then it carries no verdict, and an absent verdict
is not a failure"* — is only true once #1792 lands. `statusLabel(...)`
(Strategies.jsx:161, correctly untouched here) still returns
`'Reference only — gate failed'` for
`status === 'live' && passes_rigor_gate === false`, and `passes_rigor_gate` is
the **fail-closed** boolean: false for a `pending` verdict too. #1792 replaces
that with the four-state `rigor_gate_status` and
`NOT_GRADED_LABEL = "Not yet graded"` (`ui/src/libraryStatus.js`), which is what
stops an ungraded curated row from rendering a failure pill. **Merged alone,
this paragraph sits directly above 34 rows asserting the verdict it denies.**

This is an **ordering** constraint, not a conflict one:

```
git merge-tree --write-tree --name-only \
  origin/dbrowneup/ux-curated-examples-copy \
  origin/dbrowneup/verdict-of-record-a
-> EXIT=0, tree 656059e6, no conflicted paths   (217f8b8f vs b6b8b790)
```

`git merge-tree` against `origin/dbrowneup/reasoning-traces-surfaced` is clean
too (EXIT=0, 0 conflict markers). The dependency is
also recorded in the source, as a JSX comment immediately above the dependent
paragraph (Strategies.jsx, `MERGE ORDER — depends on PR #1792`), so it cannot be
lost when this body scrolls out of view.

The shared-file overlap with #1792 is textual proximity only: it owns
`statusTag` / `statusLabel` (~lines 118-131 on main); the hunks here are the
chip body (~line 28), a new component after `STATUS_ORDER`, the Sharpe cells,
the `gatePromise` line in `load()`, and the Examples intro.

## Adversarial demo — every guard shown red on the input it rejects

Guard file: `ui/test/curated-examples-copy.test.js` (10 tests, green on this
branch). Each pattern also carries in-file anti-vacuity: it is asserted to
*reject* its own pre-fix input, so a pattern that stops matching anything fails
loudly instead of guarding nothing. Beyond that, every defect was re-introduced
into the source and the guards run. **8 mutations, all red (exit 1), tree
restored after each (`git status --porcelain` clean).**

| # | mutation | suite | result |
| --- | --- | --- | --- |
| 1 | delete `if (!ROADMAP_SURFACES_ENABLED) return null` from `DeployabilityChip` | curated-examples-copy | pass 8 / **fail 2**, EXIT=1 |
| 2 | restore the pre-fix Examples intro copy | curated-examples-copy | pass 8 / **fail 2**, EXIT=1 |
| 3 | un-gate the `/api/selection-bias/gate` fetch | curated-examples-copy | pass 9 / **fail 1**, EXIT=1 |
| 4 | mark a real `persisted_backtest` as "fixture" | curated-examples-copy | pass 9 / **fail 1**, EXIT=1 |
| 5 | half-gate: move the flag check *below* the `!deploy` / "checking…" branch | curated-examples-copy | pass 9 / **fail 1**, EXIT=1 |
| 6 | drop the card's `MetricsSourceTag`, keep the table row's | curated-examples-copy | pass 9 / **fail 1**, EXIT=1 |
| 7 | rename `strategy_record` → `record_metrics` in the route **and** in `backend/tests/test_metrics_provenance.py` | curated-examples-copy | pass 9 / **fail 1**, EXIT=1 |
| 8 | un-gate the fetch (again), against the *tightened* neighbour guard | library-loading | pass 7 / **fail 1**, EXIT=1 |

Failure text for the two that pin the honesty signal itself:

```
# 7 — the rename lands "normally" everywhere else:
pytest backend/tests/test_metrics_provenance.py  ->  9 passed, EXIT=0   (green)

✖ every marked source is a value the backend actually returns
AssertionError [ERR_ASSERTION]: _display_metrics_source no longer returns "strategy_record"

# 4 — a real run must never be labelled a fixture:
✖ only non-measured sources are marked
+ actual - expected
+ { label: 'fixture', title: 'x' }
- null
```

## Behavioural proof (the built bundle, not just the source text)

| build | `always-on correctness floor` in `dist/assets/Strategies-*.js` | `selection-bias/gate` |
| --- | --- | --- |
| default flags | **0** | **0** |
| `VITE_ROADMAP_SURFACES=true` | **1** | **1** |

The whole chip *and* the gate request are dead-code-eliminated in the shipped
build: the red pill cannot render and the page's slowest request is genuinely
not made. With the flag on, the `blocked` branch is back — gated, not deleted.

## Verification

| check | command | result |
| --- | --- | --- |
| new guard | `node --test test/curated-examples-copy.test.js` | tests 10 / pass 10 / fail 0, `EXIT=0` |
| full ui suite | `cd ui && node --test` | `tests 923 / pass 923 / fail 0`, `EXIT=0` |
| eslint (touched files) | `npx eslint src/components/Strategies.jsx src/metricsSource.js test/curated-examples-copy.test.js test/library-loading.test.js` | `EXIT=0` |
| ruff | `ruff check ui/` | "No Python files found", All checks passed, `EXIT=0` |
| routes/sitemap | `npm run check:routes` | orphans OK, sitemap OK, `EXIT=0` |
| vite build | `npm run build` (both flag states) | `EXIT=0` |
| backend behaviour this PR reads | `pytest -p no:cacheprovider -q backend/tests/test_metrics_provenance.py` | `9 passed`, `EXIT=0` |
| python unit suite | `pytest -m "not integration" -q -p no:cacheprovider` | `5644 passed, 4 skipped, 2 deselected in 951.01s`, `EXIT=0` — run at 8f05f165; **this PR touches zero Python files** (`git diff --name-only origin/main...HEAD` = 4 UI files), so it cannot be affected |

Pre-existing suites that pin this file still pass unchanged:
`rigor-tristate.test.js` (the pending / degenerate / blocked branch ordering
inside the chip), `fail-soft.test.js` (per-feed error state), `routes.test.js`
(#1324). `library-loading.test.js` still pins the chip signature and the
`gatePending` plumbing; its gate assertion is **tightened** by this PR (below).

## Review round — what changed after verification

Verdict was FIX-FIRST (2 required, 3 optional). All five applied; no defect
left open.

- **Required 1 — `ui/test/curated-examples-copy.test.js`.** `METRICS_SOURCE_NOTES`'
  keys are bare string literals of `_display_metrics_source`'s return values, and
  nothing pinned them. Renaming one in the route *and* in its own backend test —
  the normal way such a rename lands — left that file at 9 passed and the whole
  UI suite green while every `fixture` mark silently stopped rendering **and** the
  intro's promise that "the row is marked fixture" became false: fail-open on the
  exact honesty signal this PR adds. New guard reads the backend source from the
  UI suite (same idiom as `security-claims.test.js`, `account-deletion.test.js`),
  with an anti-vacuity check on all four backend states. Mutation 7 above.
- **Required 2 — merge-order dependency on #1792.** Now recorded in this body
  (**Merge order** section) and as a JSX comment at the dependent paragraph in
  `Strategies.jsx`.
- **Optional 3 — `ui/test/library-loading.test.js`.** Its comment ("the gate
  still runs, still lands") became false, and
  `assert.match(strategies, /apiGet\('\/api\/selection-bias\/gate'\)/)` was
  satisfied by the substring *inside* the new ternary — it would pass on a
  default build that never makes the call. Comment corrected and the assertion
  tightened to require the `ROADMAP_SURFACES_ENABLED ?` prefix. Mutation 8 above.
- **Optional 4 — `Strategies.jsx` intro.** It promised `fixture` while a row can
  render `placeholder` (`stub_placeholder`). One clause added.
- **Optional 5 — `Strategies.jsx` gateError banner comment.** It still claimed to
  be "the visible signal that replaces that silence"; with the flag off the gate
  is never fetched, so it cannot fire and there are no chips left for it to
  describe. Scope note added. Comment-only.

Part of #1436

### DIFF
--- a/ui/src/components/Strategies.jsx
+++ b/ui/src/components/Strategies.jsx
@@ -21,11 +21,31 @@ import {
   UNKNOWN_RIGOR_TITLE,
 } from '../rigorGateStatus.js'
 import { signClass } from '../signClass.js'
+import { metricsSourceNote } from '../metricsSource.js'
 import { strategies as ROADMAP_COPY } from '../roadmapCopyApp.js'
 
 // A compact "deployable at your level" chip for a library row, driven by the
 // strategy's min_passing_level (from the live gate) and the user's strictness.
 function DeployabilityChip({ deploy, level, gatePending }) {
+  // Vaults are out of the MVP cut (#1266) — and EVERY branch below answers a
+  // VAULT question. "deployable" / "needs level N" / "blocked" all grade a
+  // strategy against `/api/selection-bias/gate`, the same verdict the vault
+  // deploy gate reads (`api/vaults_routes.py::_deployable_levels`). With the
+  // roadmap flag off there is no vault to deploy into, so every one of those
+  // chips grades a strategy against a capability the shipped build does not
+  // have. On the Examples tab that is what put a red `blocked` pill on
+  // hand-curated reference implementations.
+  //
+  // Suppressing only the red branch was the other option, and it is worse: it
+  // would leave the green "deployable" claim standing on a surface where
+  // nothing can be deployed — the flattering half of a verdict kept and the
+  // rest hidden. The whole chip belongs to the flag, and comes back intact
+  // (blocked branch included) under VITE_ROADMAP_SURFACES=true.
+  //
+  // The gate FETCH is gated on the same flag in load() below: a hidden
+  // annotation that still costs the page its slowest request is the #1324
+  // defect, not a fix.
+  if (!ROADMAP_SURFACES_ENABLED) return null
   // #1645: the Library now renders rows as soon as the strategy list resolves,
   // without waiting for the (much slower) /api/selection-bias/gate call. During
   // that window there is no entry for this row yet — and a SILENTLY ABSENT chip
@@ -79,6 +99,20 @@ function DeployabilityChip({ deploy, level, gatePending }) {
 }
 
 const STATUS_ORDER = ['live', 'validated', 'candidate', 'retired']
+
+// A one-word provenance mark that rides directly under a row's headline
+// number. `metricsSourceNote` allow-lists only the sources that are NOT a run
+// made here (see ../metricsSource.js), so a real persisted backtest renders
+// unmarked and an unknown/absent value makes no claim in either direction.
+function MetricsSourceTag({ source }) {
+  const note = metricsSourceNote(source)
+  if (!note) return null
+  return (
+    <div style={{ fontSize: '0.68rem', color: 'var(--text-4)' }} title={note.title}>
+      {note.label}
+    </div>
+  )
+}
 
 function downloadStrategy(strategy, format) {
   let content, filename, type
@@ -329,6 +363,11 @@ function StrategyRow({ s, isHighlighted, onOpenRigorExplainer, onOpenPassport, d
               (DSR conf=<MetricValue metric="dsr_p_value" value={s.dsr_p_value} row={s} surface="Library table" />)
             </div>
           )}
+          {/* Sharpe is the representative field of the whole display block —
+              the backend derives `display_metrics_source` from it because one
+              link of the fallback chain populates all of them — so one mark in
+              this cell describes the row's numbers, not four repeats. */}
+          <MetricsSourceTag source={s.display_metrics_source} />
         </td>
         <td className={`mono ${signClass(s.cagr)}`} style={{ textAlign: 'right' }}>
           <MetricValue metric="cagr" value={s.cagr} row={s} surface="Library table" />
@@ -593,7 +632,7 @@ function StrategyCard({ s, isHighlighted, onOpenRigorExplainer, onOpenPassport, 
         <DeployabilityChip deploy={deploy} level={level} gatePending={gatePending} />
       </div>
       <div className="lib-card-stats">
-        <div><div className="caption">Sharpe</div><div className="mono"><MetricValue metric="sharpe_ratio" value={s.sharpe_ratio} row={s} surface="Library card" /></div></div>
+        <div><div className="caption">Sharpe</div><div className="mono"><MetricValue metric="sharpe_ratio" value={s.sharpe_ratio} row={s} surface="Library card" /></div><MetricsSourceTag source={s.display_metrics_source} /></div>
         <div><div className="caption">CAGR</div><div className={`mono ${signClass(s.cagr)}`}><MetricValue metric="cagr" value={s.cagr} row={s} surface="Library card" /></div></div>
         <div><div className="caption">Max DD</div><div className="mono negative"><MetricValue metric="max_drawdown" value={s.max_drawdown} row={s} surface="Library card" /></div></div>
         <div title={genCost.title}><div className="caption">Gen tokens</div><div className={genCost.measured ? 'mono' : 'caption'}>{genCost.label}</div></div>
@@ -895,7 +934,16 @@ export default function Strategies({ highlightStrategyId, defaultTab, onNavigate
     // the 2026-08-30 external product review ("0 of 20 examples pass").
     const seedPromise = apiGet('/api/strategies/?limit=100')
     const genPromise = apiGet('/api/strategies/generated')
-    const gatePromise = apiGet('/api/selection-bias/gate')
+    // Deployability is a VAULT question and vaults are out of the MVP cut
+    // (#1266): DeployabilityChip returns null with the flag off, so an
+    // unconditional call here would spend the slowest request on the page
+    // (this route recomputes the whole cohort gate — ~8-10s, and 504-ing
+    // against prod on 2026-08-31) on an annotation nothing renders. Same
+    // ternary shape, and the same reason, as publishedPromise below: gating
+    // the render while still hitting the hidden API every load is the exact
+    // defect #1324 was filed against. The empty payload keeps every consumer
+    // downstream — deployMap, gateLoading, the gateError banner — unchanged.
+    const gatePromise = ROADMAP_SURFACES_ENABLED ? apiGet('/api/selection-bias/gate') : Promise.resolve({ strategies: [] })
     // Kept on one line deliberately: ui/test/routes.test.js's #1324 guard
     // anchors on this exact expression, and reformatting it would silently
     // disarm someone else's regression test rather than break it loudly.
@@ -1203,12 +1251,44 @@ export default function Strategies({ highlightStrategyId, defaultTab, onNavigate
 
       {activeTab === 'examples' && (
         <>
+          {/* This tab is REFERENCE MATERIAL, and the intro has to say so before
+              the reader starts reading the rows as scores. The previous copy
+              ("see what a rigor-gate verdict looks like") invited exactly the
+              opposite reading, and the rows obliged: 34 hand-curated
+              implementations of published papers, each one carrying a red
+              failure pill next to numbers nobody re-ran here.
+
+              Three claims, each of which has to stay true:
+                1. what they are — reference implementations, to learn the card
+                   format; not fusion-engine output;
+                2. what a missing verdict means — a curated example is graded
+                   only once a backtest has actually been run for it (the live
+                   gate needs persisted returns), so "no verdict" is a state,
+                   not a failure, and the passport is where the verdict lives;
+                3. where the numbers come from — the mark under each row's
+                   Sharpe, driven by the API's own `display_metrics_source`
+                   (see ../metricsSource.js), so this paragraph never has to
+                   generalise about rows it cannot see. */}
           <div className="caption mb-3 text-[var(--text-3)] leading-relaxed">
-            <strong>Example strategies</strong> — hand-curated single-paper implementations
-            from published research. <em>Not</em> outputs of the fusion engine. Included
-            so you can read a strategy card, understand the metrics, and see what a
-            rigor-gate verdict looks like. They're also the candidate pool the curated-library
-            path of Generate picks and weights from.
+            <p style={{ marginBottom: 6 }}>
+              <strong>Example strategies</strong> — hand-curated reference implementations
+              of single published papers. They are here to be read: open one to learn the
+              card format — what each field means, which paper it came from, and where a
+              verdict appears once there is one. They are <em>not</em> outputs of the
+              fusion engine.
+            </p>
+            <p style={{ marginBottom: 6 }}>
+              They are <strong>not a scoreboard</strong>. A curated example is graded only
+              once a backtest has been run for it here; until then it carries no verdict,
+              and an absent verdict is not a failure. The numbers beside them are not all
+              measurements either: where a row's metrics trace to a stored fixture snapshot
+              rather than to a backtest run here, the row is marked <strong>fixture</strong>.
+              A strategy's passport is its verdict of record.
+            </p>
+            <p>
+              Generate's curated-library path picks and weights its candidates from this
+              same set.
+            </p>
           </div>
           {loading && <StrategyListSkeleton />}
           {/* Gated on !loadError, matching the Published branch below (#1356
--- a/ui/src/components/Strategies.jsx
+++ b/ui/src/components/Strategies.jsx
@@ -1296,11 +1296,11 @@ export default function Strategies({ highlightStrategyId, defaultTab, onNavigate
                 failure pill. Merged alone, this paragraph contradicts the pill two
                 inches below it, so land this after or with #1792.
 
-                This is an ORDERING constraint, not a conflict one: a 3-way merge
-                against that branch is clean (base=origin/main, git merge-file
-                EXIT=0, zero conflict markers). Do not "fix" it by softening the
-                copy — the copy is the correct end state; #1792 is what makes the
-                rows agree with it. */}
+                This is an ORDERING constraint, not a conflict one:
+                `git merge-tree --write-tree --name-only` of this branch against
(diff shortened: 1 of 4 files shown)


danielscoffee (approved):
In-lane UX honesty fix. Looks good.

**What works**
- Gating the *whole* `DeployabilityChip` (not just `blocked`) is the right call — keeping green `deployable` while hiding red would be a worse lie on a surface with no vaults.
- Same-flag gate on the `/api/selection-bias/gate` fetch matches the #1324 pattern (no pay-for-hidden-annotation).
- `metricsSourceNote` allow-list is load-bearing: marking `persisted_backtest` as fixture would be a fresh false claim. Prototype-safe lookup (`Object.hasOwn`) is appreciated.
- Examples intro retracts the "see what a rigor-gate verdict looks like" invitation that made the red pills look intentional.
- Guard file is strong: each pattern rejects its own pre-fix input, plus the adversarial re-intro demos.

**Scope note (not a blocker)**
This PR correctly leaves `statusTag` / `statusLabel` (`Reference only — gate failed`) to `verdict-of-record-a`. Until that lands, Examples still carries one of the three pills the owner reacted to — worth sequencing so the full "34 failures" reading actually clears.

**Concurrency**
Proximity conflict with `verdict-of-record-a` on `Strategies.jsx` is expected; hunks look non-overlapping semantically.

Not merging — owner vets first, as the PR asks.

dbrowneup:
## Review fixes applied — 2 required + 3 optional, all five landed

Commits: `217f8b8f` (the five fixes) and `06644d53` (merge-evidence citation).
Files now in the PR: `ui/src/components/Strategies.jsx`, `ui/src/metricsSource.js`,
`ui/test/curated-examples-copy.test.js`, `ui/test/library-loading.test.js` (new to the PR).

### REQUIRED 1 — the mark's vocabulary was the backend's, and nothing pinned it
`ui/test/curated-examples-copy.test.js` (+50, and `METRICS_SOURCE_NOTES` added to the import)

`METRICS_SOURCE_NOTES`' keys are bare string literals of what
`_display_metrics_source` returns. I ran the rename the way it would actually land —
`"strategy_record"` → `"record_metrics"` in `backend/archimedes/api/strategies_routes.py`
**and** in its own `backend/tests/test_metrics_provenance.py`:

```
pytest -p no:cacheprovider -q backend/tests/test_metrics_provenance.py -> 9 passed, EXIT=0   (green)
```

…while every `fixture` mark silently stops rendering and the intro's promise that
"the row is marked **fixture**" becomes false above 34 unmarked rows. Fail-open on the
exact honesty signal this PR exists to add.

New guard `every marked source is a value the backend actually returns` reads
`strategies_routes.py` from the UI suite (same idiom as `security-claims.test.js` and
`account-deletion.test.js`), extracts `_display_metrics_source`'s `return "..."` literals,
and checks every `METRICS_SOURCE_NOTES` key against them. Anti-vacuity: all four backend
states (`strategy_record`, `persisted_backtest`, `stub_placeholder`, `unavailable`) must be
found, so it cannot pass by comparing against an empty set.

```
green on HEAD:   node --test test/curated-examples-copy.test.js -> tests 10 / pass 10 / fail 0, EXIT=0
red under the rename:                                             pass 9 / fail 1, EXIT=1
  ✖ every marked source is a value the backend actually returns
  AssertionError [ERR_ASSERTION]: _display_metrics_source no longer returns "strategy_record"
```

### REQUIRED 2 — the merge-order dependency on #1792 was unrecorded
`ui/src/components/Strategies.jsx` (JSX comment above the dependent paragraph) + PR body

Recorded in **both** places asked for:

- **PR body**: new "## Merge order — depends on PR #1792 (PR-A)" section, and the summary
  paragraph up top now points at it.
- **Source**: a `{/* MERGE ORDER — depends on PR #1792 (dbrowneup/verdict-of-record-a) ... */}`
  comment sitting immediately above the `<p>` that makes the dependent claim, so it cannot be
  lost when the body scrolls out of view. It names the mechanism (`statusLabel(...)` at :161
  still returns `'Reference only — gate failed'` for
  `status === 'live' && passes_rigor_gate === false`; `passes_rigor_gate` is the fail-closed
  boolean, false for a `pending` verdict too), names what fixes it in #1792
  (`NOT_GRADED_LABEL = "Not yet graded"` on the four-state `rigor_gate_status`, in
  `ui/src/libraryStatus.js`), and says explicitly: do not "fix" it by softening the copy.

The dependency was **not** removed and no attempt was made to remove it. Verified it is an
ordering constraint only, not a conflict one — #1792's branch is on origin now, so this is a
real branch-to-branch 3-way rather than the earlier `git merge-file` against a sibling worktree:

```
git merge-tree --write-tree --name-only \
  origin/dbrowneup/ux-curated-examples-copy origin/dbrowneup/verdict-of-record-a
-> EXIT=0, tree 656059e6, no conflicted paths   (217f8b8f vs b6b8b790)
```

### OPTIONAL 3 — a neighbouring guard's comment was false and its assertion too loose
`ui/test/library-loading.test.js:100-102`

It read *"The gate still runs, still lands, and still reports its own errors"*, and
`assert.match(strategies, /apiGet\('\/api\/selection-bias\/gate'\)/)` was satisfied by the
substring **inside** the new ternary — it would pass on a default build that never makes the
call (0 hits in the built chunk). Comment corrected to say "under `VITE_ROADMAP_SURFACES=true`"
and to explain why the prefix is in the pattern; assertion tightened to
`/ROADMAP_SURFACES_ENABLED \? apiGet\('\/api\/selection-bias\/gate'\)/`.

Shown adversarially — un-gate the fetch, and the tightened guard now catches it:

```
node --test test/library-loading.test.js -> tests 8 / pass 7 / fail 1, EXIT=1
  ✖ the skeleton comes down without waiting for the deployability gate
restored -> tests 8 / pass 8 / fail 0, EXIT=0
```

### OPTIONAL 4 — the intro promised `fixture` but a row can render `placeholder`
`ui/src/components/Strategies.jsx` (intro copy)

One clause: "…the row is marked **fixture** — or **placeholder**, where only the strategy
module's declared constants exist." (`stub_placeholder`, previously explained only in a
`title` tooltip.)

### OPTIONAL 5 — stale comment on now-unreachable UI
`ui/src/components/Strategies.jsx` (gateError banner)

The banner's comment still claimed to be "the visible signal that replaces that silence,
near the chips it describes". With the flag off `load()` never fetches the gate, so
`gateError` stays null, the banner cannot fire, and there are no chips left for it to
describe. Scope note added; banner and chips come back together under
`VITE_ROADMAP_SURFACES=true`. Comment-only, no behaviour change.

---

## Verification after the fixes

| check | result |
| --- | --- |
| `node --test test/curated-examples-copy.test.js` | tests 10 / pass 10 / fail 0, `EXIT=0` |
| `node --test test/library-loading.test.js` | tests 8 / pass 8 / fail 0, `EXIT=0` |
| `cd ui && node --test` (full) | `tests 923 / pass 923 / fail 0`, `EXIT=0` |
| `npx eslint` on the 4 touched files | `EXIT=0` |
| `ruff check ui/` | "No Python files found", All checks passed, `EXIT=0` |
| `npm run check:routes` | orphans OK, sitemap OK, `EXIT=0` |
| `npm run build` (default) | `EXIT=0`; `always-on correctness floor` → **0**, `selection-bias/gate` → **0** in `dist/assets/Strategies-*.js` |
| `VITE_ROADMAP_SURFACES=true npm run build` | `EXIT=0`; both → **1** (gated, not deleted) |
| `pytest -p no:cacheprovider -q backend/tests/test_metrics_provenance.py` | `9 passed`, `EXIT=0` |

Python suite not re-run: this PR touches **zero** Python files
(`git diff --name-only origin/main...HEAD` = 4 UI files), so the 8f05f165 run
(`5644 passed, 4 skipped, 2 deselected`, `EXIT=0`) still stands.

**Mutations — 8, all red (exit 1), working tree restored after each
(`git status --porcelain` clean before commit):**

| # | mutation | suite | result |
| --- | --- | --- | --- |
| 1 | delete `if (!ROADMAP_SURFACES_ENABLED) return null` | curated-examples-copy | pass 8 / **fail 2** |
| 2 | restore the pre-fix Examples intro | curated-examples-copy | pass 8 / **fail 2** |
| 3 | un-gate the gate fetch | curated-examples-copy | pass 9 / **fail 1** |
| 4 | mark `persisted_backtest` as "fixture" | curated-examples-copy | pass 9 / **fail 1** |
| 5 | half-gate (flag check moved below `!deploy`) | curated-examples-copy | pass 9 / **fail 1** |
| 6 | drop the card's `MetricsSourceTag` | curated-examples-copy | pass 9 / **fail 1** |
| 7 | rename `strategy_record` in the route **and** its backend test | curated-examples-copy | pass 9 / **fail 1** |
| 8 | un-gate the fetch, vs the tightened neighbour guard | library-loading | pass 7 / **fail 1** |

Nothing left open from the verdict. PR body rewritten so every claim matches the diff.

danielscoffee:
Checked the follow-up on `06644d53` — all five look landed.

1. Backend vocabulary pin in `curated-examples-copy.test.js` (keys vs `_display_metrics_source` return literals) ✓
2. `#1792` merge-order recorded in Strategies.jsx + PR body ✓
3. `library-loading.test.js` gate assert tightened with `ROADMAP_SURFACES_ENABLED ?` prefix ✓
4. Intro now mentions **placeholder** alongside **fixture** ✓
5. `gateError` banner scope note ✓

Still: do not merge — owner vets first. Land after / with #1792 so the intro's "absent verdict ≠ failure" line isn't sitting above `Reference only — gate failed` pills.

Re-approving on HEAD.

danielscoffee (approved):
Follow-up looks solid on `06644d53`. The five review fixes (vocab pin, #1792 merge-order comment + body, tightened library-loading gate assert, placeholder in intro, gateError scope note) are in the tree. Frontend CI green; PR still touches no Python.

Same sequencing caveat as before: after / with #1792. Not merging — owner vets first.

Issue #1748, opened 2026-09-01
[funnel][P0] Free-tier activation: verification delivery unproven for real inboxes; resend is fire-and-forget; audit the dual-unlock


dbrowneup:
Dogfood 2026-09-01 (human-signup + agent-cli) + SES reality check. Facts: SES IS out of sandbox (ProductionAccessEnabled=true, 50k/day) but SentLast24Hours=4 and delivery to a REAL inbox has never been verified (docs/account-authentication.md's own admission). Every new account gets `free_generations_remaining: 3` locked behind `email_unverified`; resend returns `200 {status:true}` forever with no delivery feedback; dogfood fake addresses bounced into the SES suppression list (expected). Scope: (1) OPERATOR STEP: one real-inbox signup test end to end. (2) In-product delivery feedback: surface bounce/suppression state instead of eternal silent 200. (3) AUDIT: agent-cli's generate probe returned exit 6 with a DUAL unlock (email verify + wallet) — determine from code whether the $2 paid path is also email-gated; if so a fresh account has NO path to generation at all. (4) Suppression-list hygiene for retried addresses.

dbrowneup:
Item (3), answered from code on `main` @ `adda6d5e` — **the paid path is NOT email-gated; there is no dual lock.**

`backend/archimedes/api/generate_routes.py` (payment-gate block, ~L462–541):

1. `free_grant_id = free_generations.claim(user.id, email_verified=user.email_verified)` — the ONLY place `email_verified` is consulted. `services/free_generations.py:91-103` (`locked_reason`) and `:112-131` (`claim`) return `None`/no slot when unverified.
2. If no free slot **and no linked wallet** → `409 {reason: "wallet_link_required", free_generations_locked_reason: "email_unverified" | null}`. The message names *both* unlocks ("verify your email … **or** link a wallet … and pay per run") — that is what the agent-cli probe saw as a "dual unlock". They are two **alternative** exits, not two locks in series.
3. If no free slot **and a linked wallet** → `payment, credit_id = await _paywall_with_credit(request, linked_wallet, user.id)` — the x402 402 flow. No `email_verified` read anywhere on this branch; `enforce_model_entitlement` after it checks the wallet, not the email.

[...]

PR #1780, opened 2026-09-02
SES suppression-list hygiene: operator script + runbook (#1748 item 4)

dbrowneup:
Closes part of #1748 — **item 4 (suppression-list hygiene)**. Item 2 (delivery feedback) is a separate PR; items 1 and 3 are already answered on the issue.

## The observation

From the issue: *"dogfood fake addresses bounced into the SES suppression list (expected) … Suppression-list hygiene for retried addresses."*

Every hard bounce or spam complaint puts the recipient on the **account-level** SES suppression list automatically. From that moment:

- `SendEmail` to that address **succeeds** — it returns a `MessageId`.
- The message is then **dropped inside AWS** and never sent.
- Nothing errors. Nothing in CloudWatch shows a failure. The recipient sees silence, and so do we.

**The failure mode is a success response.** That is what makes this list different from an outage.

Our list was populated correctly — AWS did exactly what it should when fake dogfood addresses bounced. The residue is the problem: a **real** address that was once typo'd, temporarily undeliverable, or used as a probe stays blocked forever, and the human retrying it gets nothing. Before this PR there was no way to see the list, and no way to clear one entry.

## Design

Three commands in `backend/archimedes/scripts/ses_suppression.py`, read-first:

| Command | Mutates? | What it does |
|---|---|---|
| `list [--reason BOUNCE\|COMPLAINT] [--limit N] [--json]` | never | Every suppressed destination with its reason and the date AWS added it. Paginates. |
| `check <address>` | never | One address. Exit `0` suppressed, `3` not on the list, `2` the AWS call failed. |
| `remove <address> [--apply]` | only with `--apply` | **Dry run by default**: looks the address up, prints exactly what it would delete, changes nothing. |

**There is deliberately no bulk clear, and that absence is the design.** Clearing the list re-sends to every address that ever hard-bounced; AWS reads that as a sender who does not honour bounces, our bounce rate climbs, and delivery degrades **for every real user** — ending, at a high enough rate, in a sending pause on the whole account. So `remove` takes exactly one address, there is no `--all`, and `test_remove_takes_exactly_one_address_and_no_bulk_flag_exists` pins it: a later "convenience" flag has to delete that test on purpose.

**A failed AWS call is never rendered as "not suppressed"** (exit `2`, error name on stderr). Silence looking like success is the entire failure this tool addresses, so the tool must not reproduce it. Same rule the product side follows.

`docs/runbooks/ses-suppression.md` carries the operator judgement the script cannot encode — the three conditions that must **all** hold before an address comes off:

1. A human **owns it and has confirmed it is real and reachable** — in writing, in the ticket. Not "looks real".
2. **The cause is understood and is over.** If you cannot say why it bounced, you cannot say it will not bounce again.
3. **It is one address.**

Plus: never remove a `COMPLAINT` entry unless its owner has asked to receive our mail again (a complaint is a person pressing "this is spam"); never remove dogfood/probe addresses at all (they bounced because they are fake — leaving them costs nothing and protects the bounce rate).

Credentials come from the ambient operator profile/role, like every other script in that directory. The read path needs `ses:ListSuppressedDestinations` + `ses:GetSuppressedDestination`; `--apply` needs `ses:DeleteSuppressedDestination`. The app's own ECS task role has neither delete nor list.

## Adversarial demos — the guards, shown red

Each mutation was applied by hand to the source, the suite re-run, and the source restored. Pasted verbatim.

**M1 — make `remove` delete regardless of `--apply`** (delete the `if not apply:` early return):

```
E   AssertionError: a dry run issued a DeleteSuppressedDestination
E   assert ['real@example.com'] == []
E   assert (True is True and True is False)
FAILED backend/tests/scripts/test_ses_suppression.py::test_remove_without_apply_deletes_nothing_and_says_what_it_would_do - AssertionError: a dry run issued a DeleteSuppressedDestination
FAILED backend/tests/scripts/test_ses_suppression.py::test_json_output_is_machine_readable_for_every_command - assert (True is True and True is False)
2 failed, 12 passed, 1 warning in 3.29s
```

**M2 — add a bulk mode** (`removal.add_argument("address", nargs="*")` + `--all`):

```
backend/tests/scripts/test_ses_suppression.py::test_remove_takes_exactly_one_address_and_no_bulk_flag_exists FAILED [100%]
E   Failed: DID NOT RAISE <class 'SystemExit'>
FAILED backend/tests/scripts/test_ses_suppression.py::test_remove_takes_exactly_one_address_and_no_bulk_flag_exists - Failed: DID NOT RAISE <class 'SystemExit'>
1 failed, 1 warning in 3.61s
```

**M3 — report a failed AWS lookup as "not suppressed"** (`check_suppressed` swallows every exception and returns `None`):

```
FAILED backend/tests/scripts/test_ses_suppression.py::test_a_failed_aws_call_exits_nonzero_and_never_reports_the_address_as_clean - assert 3 == 2
FAILED backend/tests/scripts/test_ses_suppression.py::test_a_failed_lookup_during_remove_aborts_before_any_delete - AssertionError: assert 3 == 2
2 failed, 12 passed, 1 warning in 2.49s
```

Source restored after each; clean run: `14 passed, 1 warning in 4.84s`.

## Verification

```
$ pytest -p no:cacheprovider backend/tests/scripts/test_ses_suppression.py -q
14 passed, 1 warning in 4.45s                                   EXIT 0

$ pytest -p no:cacheprovider backend/tests/scripts/test_ses_suppression.py backend/tests/test_docs_site.py -q
32 passed, 1 warning in 3.01s                                   EXIT 0

$ ruff check backend/archimedes/scripts/ses_suppression.py backend/tests/scripts/test_ses_suppression.py
All checks passed!                                              EXIT 0

$ ruff format --check backend/archimedes/scripts/ses_suppression.py backend/tests/scripts/test_ses_suppression.py
2 files already formatted                                       EXIT 0

$ mkdocs build --strict
(no warnings, no errors)                                        EXIT 0

$ pytest -p no:cacheprovider -m "not integration" -q backend/tests
5658 passed, 4 skipped, 2 deselected, 328 warnings in 983.55s   EXIT 0
```

Tests are hermetic: a hand-written `StubSes` with the three SESv2 methods the script calls, injected through `main(client=...)`. No boto3, no credentials, no network, no AWS API ever reached. The stub's not-found exception is a plain class named `NotFoundException` because `_not_found()` classifies by class name — so the stub travels the same code path production does without importing botocore's exception factory.

**Not run against live AWS.** Nothing in this PR has touched the real suppression list; the first `--apply` is an owner action, gated by the runbook's § 3.

## Concerns / what a reviewer should push on

- **The runbook is the actual control.** The script cannot verify that an address's owner confirmed it is real — that is a human judgement, and the tool's dry-run default exists to force the operator through it. If you think a `--yes-i-read-the-runbook`-style prompt would help more than the dry run, say so.
- **`_not_found()` matches on the exception's class name**, not on a botocore import. That keeps the module and its tests botocore-free, but it would silently stop matching if AWS renamed the error. The consequence of a miss is exit `2` ("we could not look"), never a false "clean" — fail-closed, but worth knowing.
- **Item 2's `GET /api/auth/verification-status`** is the in-product half of the same fact and ships in the sibling PR. This runbook references it by name but does not link a section, so the two PRs can land in either order.


dbrowneup:
Both doc nits addressed in `5c1a529`.

**(a) `docs/README.md:143` — status vocabulary.** The new row carried `current`. Every *procedure* runbook in that section is `runbook` (`arc-testnet-e2e`, `spec-1-walkthrough`, `t3.2-contract-redeploy`, `docs-site-setup`, `backtest-results-retention`, `erc8004-identity-registration`, `runner-ec2-wedge`, `email-verification-validation`); `current` is used only for the section's two standing docs (`runbooks/README.md`, `operations.md`). Row → `runbook`, and the doc's own front-matter with it.

**(b) `docs/runbooks/ses-suppression.md` § 2 — the "cannot delete" clause.** The old sentence ("the app's own ECS task role has the GET only and **cannot** delete") is not true today: the only SES statement on the task role right now is `ses:SendEmail` / `ses:SendRawEmail` (`infra/ecs.tf:314`) — it has no GET either. § 2 now names both states:

- **Today** — send only; it can neither read the suppression list nor delete from it.
- **After the `terraform apply`** that adds `ses:GetSuppressedDestination` for `GET /api/auth/verification-status` (#1748 item 2) — one-address lookup, still no list, still no delete.
- `ses:DeleteSuppressedDestination` is on the task role in **neither** state, by design; removal is an operator action (§ 4).

**One thing I found while there** (flagging because it is the same truth class, and § 2's new wording exposed it): the runbook asserted `GET /api/auth/verification-status` as a live surface in three places, and § 7 claimed `api/auth-and-accounts.md` documents it. That route exists nowhere in the tree — not in `backend/`, not in the sidecar, not in that API doc. It is #1748 **item 2**, unshipped; only item 4 (this PR) is landing. Those four references now say "not shipped yet" and describe the post-ship state, and § 5 step 3 is gated on item 2 shipping. Revert that part if you would rather the runbook stay written for the end state.

Guards, all green in the worktree:

```
docs_index.py : 175 docs under docs/; 175 indexed, 0 orphaned          rc=0
docs_links.py : scanned 1743 links in 180 markdown files; broken 0     rc=0
pytest -q backend/tests/test_docs_site.py                              18 passed
mkdocs build --strict                                                  built in 5.97s (INFO-level anchor notes only, all pre-existing)
```

Docs-only commit; no Python touched.

PR #1781, opened 2026-09-02
[frontend] Landing hero is visible in every animation state: the entrance no longer starts at opacity 0 (#1750)

### LINKED ISSUE #1750: [frontend] Landing hero renders as a solid black void
Dogfood 2026-09-01 (human-explore): headline, lede, CTAs, screenshot all present in DOM and fetched over network, but nothing paints in the hero section; no console errors; reproduced on fresh load, full reload, post-scroll. First pixel every visitor sees. Suspect stacking/theme-token/animation regression. Repro: https://archimedes-arc.com/ under the header. Fix + a visual guard (the public-visuals test family) pinning computed visibility of the hero title.

dbrowneup:
Closes #1750.

**Reworked.** The first attempt on this branch changed `animation-fill-mode` from `both` to `backwards` and added an explicit `to` frame. An **independent verifier** pulled it apart and found it inert: it changes what a spec-compliant browser paints in *zero* of the four states it claimed to fix, and leaves the one state that produces the reported black void untouched. The owner accepted that analysis. This is the rework.

## Why the fill-mode story was wrong

Nothing in the hero declares an `opacity` of its own — I grepped the whole hero block, and the only `opacity` in it was the keyframe's `from` frame. So the element's *underlying* computed opacity is 1, and CSS Animations L1 §3 says a missing `to` frame is synthesised from exactly that underlying value. Which means the old `both` forwards-fill was already holding **opacity 1** at the end.

The verifier's phase × fill-mode table, before vs. after that first attempt:

| state | before (`both`, implicit `to`) | after (`backwards`, explicit `to`) |
| --- | --- | --- |
| before phase / play-pending (hold time 0) | **0** (the backwards half of `both`) | **0** — unchanged |
| `prefers-reduced-motion: reduce` | 1 (the rules live inside `no-preference`) | 1 — unchanged |
| interrupted: paused/frozen mid-active | partial (→ 0 at the head) | unchanged; cancelled → idle = 1 either way |
| hidden tab → shown | finished before hide → 1; suspended at t=0 → **0** | identical |
| finished | forwards fill of the implicit frame = **1** | no fill = **1** |

Active-phase interpolation was byte-identical too — the implicit `to` resolves to `opacity 1 / transform none`, which is precisely the explicit `to` that was added. Same computed animation, before and after.

The only state that rests at 0 is the before/pending phase, and `backwards` is *defined* as filling it.

## The actual bug, and the ruling

**`from { opacity: 0 }` is the whole bug**, and no fill mode reaches it:

- `backwards` and `both` apply the first frame throughout the before phase, by definition.
- `none` and `forwards` do not help either: three of the five animated elements have **no delay**, so their play-pending hold time lands on active time 0 — which is the `from` frame again.

So any state where the animation is applied but its start time never resolves — a background-tab load with a suspended WebKit document timeline, a bfcache/tab-snapshot restore before the first style commit, a compositor stall — paints the hero at opacity 0 over `.public-hero__stage`'s near-black `--public-stage` (`#0c0c11`). Indefinitely. That is the reported symptom exactly: DOM present, assets fetched, no console errors, a solid black rectangle.

`@starting-style` is not the escape it looks like — a CSS transition is spec'd with `fill: backwards` as well, so a play-pending transition holds the same opacity 0.

**Owner ruling: the hero is fail-visible by construction.** Opacity comes out of the entrance entirely. The `translateY(10px)` → `none` slide stays; **the fade is dropped, and that cost is accepted.**

```css
@keyframes public-arrive {
	from { transform: translateY(10px); }
	to   { transform: none; }
}
```

The hero's opacity is now 1 in **every** state there is: pending, before phase, active, finished, idle, cancelled, reduced motion, a stale bundle, a stylesheet that failed to load. An entrance that never starts leaves the hero sitting 10px low and completely readable.

Built CSS confirms it: `public-arrive{0%{transform:translateY(10px)}to{transform:none}`.

The two rules keep `backwards` (now immaterial — the first frame is a transform, and holding it is the intended entrance offset). The `@media (prefers-reduced-motion: reduce)` block stays, relabelled plainly as what it is: **a motion-preference courtesy, not a visibility guard.** It changes nothing today (the animation rules already sit inside `no-preference`), and since the entrance no longer touches opacity, deleting it could not reintroduce #1750.

I also confirmed no rule anywhere in `App.css` matching `.public-hero`, `.public-hero__stage`, `.public-hero__copy`, `.public-product-frame` or `#public-hero-title` declares `opacity: 0`, `visibility: hidden`, `display: none`, or any `content-visibility`. (`.public-product-frame__bar span:last-child { display: none }` at :8692 is the frame's chrome on narrow screens, not the frame.)

## Guard, rewritten — `ui/test/public-hero-visible-at-rest.test.js`

The verifier found the old guard green on the reported symptom itself: adding `opacity: 0` to `.public-hero__copy` passed 4/4, as did a 999999ms delay and a fill-mode longhand injected in a separate rule. It only ever inspected the rules that *named* the animation.

The rewrite pins the structural invariant instead of a fill mode. Three checks, all over `ui/src/App.css` source text (comments blanked out first, so prose about `opacity` can never read as a declaration):

1. `@keyframes public-arrive` contains **no `opacity` token at all**, and still animates something — so an emptied keyframe cannot pass it.
2. **No rule targeting the hero hides it** — opacity below 1, `visibility: hidden`/`collapse`, `display: none`, `content-visibility`. Checked in every context, **media queries included**, because a rule that hides the hero at one viewport width is #1750 at that width. A full leaf-rule walker replaces the old "rules that name the animation" scan.
3. **Every keyframe named by an animation on a hero rule is opacity-free** — not just `public-arrive`.

On the verifier's vacuity points:

- Each check asserts it **actually looked at something**: the keyframe exists, each required hero selector still matches a rule, both known `public-arrive` users are still present. A rename or a deletion fails loudly rather than passing over an empty set. (This caught a real thing while I wrote it — `#public-hero-title` has no rule of its own, it is styled through `.public-hero h1`, so it is scanned but not required.)
- **No passengers.** The old check 2 could only fail after check 1 had already failed (`endsVisible` was module-level and check 1 already asserted it) — it is gone. So is the reduced-motion assertion, for the same reason: with opacity out of the entrance, that block is no longer load-bearing for visibility, and asserting on it here would be a second passenger. Check 3 is *not* a passenger of check 1: mutation D below reddens it alone.

### Mutations — every check red on its own, none vacuous

Each applied to the fixed CSS one at a time, then restored (pristine md5 re-verified after the run).

```
A  restore `from { opacity: 0 }` in @keyframes public-arrive
✖ @keyframes public-arrive animates transform only — no opacity anywhere in it
✔ no rule targeting the hero hides it
✖ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 1   ℹ fail 2   exit code: 1
  AssertionError: src/App.css:6357: @keyframes public-arrive mentions `opacity`. The hero
  has no opacity of its own, so any opacity in this entrance is the only opacity the
  stylesheet states for the headline, lede, CTAs and screenshot — and a pending or stalled
  animation holds its first frame forever, under every fill mode. Animate `transform` only,
  so a broken entrance leaves the hero 10px low and readable rather than invisible. (#1750)
  AssertionError: src/App.css:6369: `.public-hero__copy > *` runs @keyframes public-arrive
  (defined at line 6357), which animates `opacity`. …

B  `opacity: 0` added to `.public-hero__copy > *`   ← the mutation the old guard passed
✔ @keyframes public-arrive animates transform only — no opacity anywhere in it
✖ no rule targeting the hero hides it
✔ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 2   ℹ fail 1   exit code: 1
  AssertionError: src/App.css:6368: `.public-hero__copy > *` (inside @media
  (prefers-reduced-motion: no-preference)) declares `opacity: 0` — the hero must never be
  less than fully opaque — that is #1750. The hero sits on a near-black stage; anything
  that stops it painting is the black void the issue reports. (#1750)

C  @keyframes public-arrive deleted entirely   ← must fail loudly, not pass vacuously
✖ @keyframes public-arrive animates transform only — no opacity anywhere in it
✔ no rule targeting the hero hides it
✖ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 1   ℹ fail 2   exit code: 1
  AssertionError: src/App.css: @keyframes public-arrive is gone. If the hero's entrance was
  retired, delete this guard in the same change; do not leave it passing over an animation
  that no longer exists. (#1750)   [both failures carry this message]

D  a DIFFERENT opacity-fading keyframe applied to the hero copy   ← check 3 alone
✔ @keyframes public-arrive animates transform only — no opacity anywhere in it
✔ no rule targeting the hero hides it
✖ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 2   ℹ fail 1   exit code: 1
  AssertionError: src/App.css:6374: `.public-hero__copy > *` runs @keyframes
  public-soft-fade (defined at line 6367), which animates `opacity`. …

E  `opacity: 0` on `.public-product-frame` inside @media (max-width: 980px)
✔ @keyframes public-arrive animates transform only — no opacity anywhere in it
✖ no rule targeting the hero hides it
✔ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 2   ℹ fail 1   exit code: 1
  AssertionError: src/App.css:8463: `.public-product-frame` (inside @media
  (max-width: 980px)) declares `opacity: 0` — the hero must never be less than fully
  opaque — that is #1750. …

F  `.public-hero__copy` renamed away   ← the vacuity trip
✔ @keyframes public-arrive animates transform only — no opacity anywhere in it
✖ no rule targeting the hero hides it
✖ every animation the hero runs comes from an opacity-free keyframe
ℹ pass 1   ℹ fail 2   exit code: 1
  AssertionError: src/App.css: no rule matches `.public-hero__copy` any more. This guard
  scans the hero by selector; a rename leaves it scanning nothing. Update HERO_SELECTORS in
  the same change. (#1750)
  AssertionError: src/App.css: no hero-copy rule runs public-arrive any more; this guard
  would have nothing to check. (#1750)
```

## Tests

```
ui suite, node --test:  ℹ tests 916   ℹ pass 916   ℹ fail 0    exit code: 0
npx eslint test/public-hero-visible-at-rest.test.js            exit code: 0
npm run lint                                                   exit code: 0
npm run build                                                  exit code: 0
```

(916, not the previous 917: the old guard had four checks and one was the passenger described above; the rewrite has three.)

## Residual — stated, not hidden

- **The trigger on the owner's iPhone is unconfirmed.** This removes the *dependency* — the hero can no longer rest, stall, or start invisible — but it does not prove *why* the entrance failed to complete on that particular load. I have no way to render the page; the confirming screenshot is the owner's to take.
- **A stale-bundle check is still worth doing before closing #1750.** Compare the CSS hash actually served by CloudFront against the one in the current build output. If the black void was a stale cached bundle, this fix does not reach it until that bundle expires — and that would be a different issue with a different remedy.
- Also unrejected by this change: correcting the record on the first commit here, whose message claimed the hero was "invisible in every other state." That was false. Finished, idle and reduced-motion all resolved to opacity 1 under the old CSS. The verifier caught it; the new commit message says so plainly.

### DIFF
--- a/ui/src/App.css
+++ b/ui/src/App.css
@@ -6323,16 +6323,48 @@ body:has(.public-site) {
 	color: var(--accent-text, var(--accent));
 }
 
+/* The hero's resting appearance must come from the cascade, never from a
+   retained animation value — issue #1750.
+
+   What broke: this keyframe declared only a `from` frame and both rules below
+   used `animation-fill-mode: both`. `both` adds the backwards fill (the first
+   frame, opacity 0, applied before the animation starts) *and* the forwards
+   fill (the last frame, applied after it ends) — and with no `to` frame the
+   last frame is one the UA synthesises from the element's underlying style.
+   Nothing in the hero declares an opacity of its own (the only `opacity` in
+   this whole block was the `from` frame), so opacity 0 was the only value the
+   stylesheet ever stated, and the hero's visibility depended entirely on the
+   animation running to completion *and* on that implicit frame resolving to 1.
+   When it didn't, the hero's children painted at opacity 0 over
+   `.public-hero__stage`'s own `--public-stage` (#0c0c11) — the solid black
+   void in #1750.
+
+   The two halves of the fix, and what each one buys:
+   • an explicit `to` frame: the end state is stated, so a forwards fill can
+     never resolve to an implicit frame the UA has to synthesise.
+   • `backwards` instead of `both`: no forwards fill at all, so once the
+     animation is over the element is drawn from its own cascade — where
+     opacity is 1 — rather than from a value the animation left behind.
+   Together: with the animation absent, idle, or finished, the hero is visible.
+   opacity 0 remains only *while* the animation is running (the ≤120ms delay
+   window and the fade itself), which is the animation doing its job.
+
+   Guard: ui/test/public-hero-visible-at-rest.test.js. */
 @keyframes public-arrive {
 	from {
 		opacity: 0;
 		transform: translateY(10px);
 	}
+
+	to {
+		opacity: 1;
+		transform: none;
+	}
 }
 
 @media (prefers-reduced-motion: no-preference) {
 	.public-hero__copy > * {
-		animation: public-arrive 360ms both;
+		animation: public-arrive 360ms backwards;
 	}
 
 	.public-hero__copy > :nth-child(2) {
@@ -6345,7 +6377,20 @@ body:has(.public-site) {
 
 	.public-hero__copy > :nth-child(4),
 	.public-product-frame {
-		animation: public-arrive 400ms 120ms both;
+		animation: public-arrive 400ms 120ms backwards;
+	}
+}
+
+/* Belt and braces. The rules above already sit inside a
+   `prefers-reduced-motion: no-preference` block, so a reduce preference never
+   applies them today — this block states the intent anyway, so that hoisting
+   the animation out of that media query (or adding another `public-arrive`
+   user outside it) cannot silently hand reduced-motion visitors an entrance
+   animation, and an opacity-0 entry state, that they asked not to have. */
+@media (prefers-reduced-motion: reduce) {
+	.public-hero__copy > *,
+	.public-product-frame {
+		animation: none;
 	}
 }
 
--- a/ui/src/App.css
+++ b/ui/src/App.css
@@ -6323,41 +6323,43 @@ body:has(.public-site) {
 	color: var(--accent-text, var(--accent));
 }
 
-/* The hero's resting appearance must come from the cascade, never from a
-   retained animation value — issue #1750.
+/* The hero must be visible in every animation state, by construction — issue
+   #1750.
 
-   What broke: this keyframe declared only a `from` frame and both rules below
-   used `animation-fill-mode: both`. `both` adds the backwards fill (the first
-   frame, opacity 0, applied before the animation starts) *and* the forwards
-   fill (the last frame, applied after it ends) — and with no `to` frame the
-   last frame is one the UA synthesises from the element's underlying style.
-   Nothing in the hero declares an opacity of its own (the only `opacity` in
-   this whole block was the `from` frame), so opacity 0 was the only value the
-   stylesheet ever stated, and the hero's visibility depended entirely on the
-   animation running to completion *and* on that implicit frame resolving to 1.
-   When it didn't, the hero's children painted at opacity 0 over
-   `.public-hero__stage`'s own `--public-stage` (#0c0c11) — the solid black
-   void in #1750.
+   What broke: this keyframe entered from `opacity: 0`, and nothing else in the
+   hero declares an opacity at all — the only `opacity` anywhere in the hero
+   block was that `from` frame. So opacity 0 was the only opacity the
+   stylesheet ever stated for the eyebrow, `#public-hero-title`, the lede, the
+   CTA row and the product screenshot, and the hero was visible only as a
+   consequence of the animation actually running. Every state that holds the
+   first frame paints the hero at opacity 0 over `.public-hero__stage`'s own
+   near-black `--public-stage` (#0c0c11): the before phase, and — the one that
+   never ends — a play-pending animation whose start time never resolves,
+   which is what a suspended document timeline in a background tab, a bfcache
+   restore before the first style commit, or a compositor stall produce. That
+   is the solid black void in #1750.
 
-   The two halves of the fix, and what each one buys:
-   • an explicit `to` frame: the end state is stated, so a forwards fill can
-     never resolve to an implicit frame the UA has to synthesise.
-   • `backwards` instead of `both`: no forwards fill at all, so once the
-     animation is over the element is drawn from its own cascade — where
-     opacity is 1 — rather than from a value the animation left behind.
-   Together: with the animation absent, idle, or finished, the hero is visible.
-   opacity 0 remains only *while* the animation is running (the ≤120ms delay
-   window and the fade itself), which is the animation doing its job.
+   No fill mode fixes that. `backwards` and `both` are *defined* as applying
+   the first frame throughout the before phase, and `none`/`forwards` do not
+   help either: three of the five animated elements have no delay, so their
+   pending hold time lands on active time 0 — which is the `from` frame again.
+   The fill mode only ever changed the far end, where the underlying value is
+   opacity 1 already.
+
+   The fix is to take opacity out of the entrance entirely. The hero animates
+   `transform` only, so its opacity is 1 in every state there is: pending,
+   before phase, active, finished, idle, cancelled, reduced motion, a stale
+   bundle, a stylesheet that failed to load. An entrance that never starts now
+   leaves the hero sitting 10px low and completely readable. The fade is gone
+   — that is the accepted cost; the slide stays.
 
    Guard: ui/test/public-hero-visible-at-rest.test.js. */
 @keyframes public-arrive {
 	from {
-		opacity: 0;
 		transform: translateY(10px);
 	}
 
 	to {
-		opacity: 1;
 		transform: none;
 	}
 }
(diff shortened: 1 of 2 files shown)


danielscoffee (approved):
Looks right for #1750.

The resting state was hanging off an implicit end frame under `both`; writing `to { opacity: 1; transform: none; }` and dropping to `backwards` puts visibility back on the cascade when the animation is absent, idle, or finished. The source-text guard (especially the forwards-fill check and the deliberate green on `both` + explicit `to`) is the right shape for this class of bug.

Residual noted and fine to leave open: opacity 0 during the delay/active window is the animation doing its job, not a rest-state failure. The reduce `animation: none` block is redundant today given the `no-preference` wrap, but documenting that intent is cheap insurance.

Approve. Worth a quick visual on the live landing once Frontend CI is green — author could not render it.

dbrowneup:
**Reworked to the owner's ruling — force of the change is now structural, not fill-mode.**

The independent verifier's verdict on my first attempt was: *the diff is inert.* It changes what a spec-compliant browser paints in zero of the four states it claimed to fix, and leaves the one state that produces the black void untouched. The owner accepted that. I have rewritten both files and rewritten the PR body around the verifier's **phase × fill-mode table** — see "Why the fill-mode story was wrong" above; it is the load-bearing part of this PR now, and it shows the `both` → `backwards` swap was a no-op in every row.

What changed in this push (`f238a310`):

- **`ui/src/App.css`** — `@keyframes public-arrive` animates `transform` only: `translateY(10px)` → `none`, **no `opacity` token anywhere in it**. The hero's opacity is 1 in every state there is — pending, before phase, active, finished, idle, cancelled, reduced motion, a stale bundle, a stylesheet that failed to load. A stalled entrance now leaves the hero 10px low and completely readable instead of invisible. **The fade is dropped; the slide is kept.** That is the accepted cost of fail-visible-by-construction. The `prefers-reduced-motion: reduce` block stays, relabelled plainly as a motion-preference courtesy rather than a visibility guard.

- **`ui/test/public-hero-visible-at-rest.test.js`** — rewritten to pin the invariant instead of a fill mode. The verifier showed the old guard passed 4/4 on `opacity: 0` added to `.public-hero__copy` — the reported symptom itself — because it only inspected rules that *named* the animation. The new one walks every leaf rule in the sheet, media queries included, and each check asserts it actually looked at something so a rename or deletion fails loudly rather than passing over an empty set. The old check 2 was a passenger of check 1 (`endsVisible` was module-level, already asserted) and is gone; so is the reduced-motion assertion, which would be a second passenger now that opacity is out of the entrance.

Six mutations, full output in the body: restoring `from { opacity: 0 }`; `opacity: 0` on `.public-hero__copy > *` (**the one the old guard passed**); deleting the keyframe (loud, not vacuous); a *different* opacity-fading keyframe on the hero copy, which reddens check 3 alone and proves it is not a passenger; `opacity: 0` inside a width media query; and renaming `.public-hero__copy` away. Every check has a mutation that reddens it independently.

ui suite **916 pass / 0 fail**, `npm run lint` and `npm run build` clean, exit code 0 throughout. Built CSS reads `public-arrive{0%{transform:translateY(10px)}to{transform:none}`.

Two residuals I am not claiming to have closed, both in the body: **the trigger on the iPhone is still unconfirmed** — this removes the dependency, not the mystery — and **a stale-bundle check (served CloudFront CSS hash vs. the build output) is still worth doing before #1750 closes**, since a stale cached bundle would not be reached by this fix at all.

I also corrected the previous commit message's claim that the hero was "invisible in every other state." False: finished, idle and reduced-motion all resolved to opacity 1, because the implicit `to` frame takes the element's underlying computed value.

dbrowneup:
Owner-session check of the stale-bundle residual: **not stale.** At 00:45Z `https://archimedes-arc.com/` serves `/assets/index-ULAEj82R.css` (222,080 B) whose keyframe reads `public-arrive{0%{opacity:0;transform:translateY(10px)}}` — byte-for-byte the from-only keyframe on main at `ui/src/App.css:6326-6331`, i.e. the current build. So the black void on the owner's iPhone was produced by the current CSS, not an old bundle, which is consistent with the pending/stalled-animation mechanism this PR removes. Residual closed; the iPhone screenshot after deploy remains the owner's to take.

danielscoffee (approved):
**Improvements (polish lane) — approve on the reworked head (`f238a310`).**

Lane: in. User-visible landing polish for #1750 (black-void hero), not a crash/latency fire.

### Prior approval superseded
My earlier approve (`a2daecf0`) praised the fill-mode + explicit-`to` attempt. The independent verifier was right that that diff was inert: with no underlying opacity declaration on the hero, the synthesised end frame was already opacity 1, and `backwards`/`both` both hold the first frame through before/pending — so swapping fill mode never touched the black-void state. Treat that earlier review as withdrawn on the merits.

### Why this rewrite is sound
`from { opacity: 0 }` was the whole bug. Taking opacity out of `@keyframes public-arrive` leaves only `translateY(10px) → none`, so opacity stays at the cascade default (1) in every state that previously could hold the first frame forever:

| state | before (opacity in `from`) | after (transform only) |
| --- | --- | --- |
| before / play-pending (incl. stalled timeline) | **0** → black void | **1** (may sit 10px low) |
| active | interpolating toward 1 | **1**, slide only |
| finished / idle / cancelled | 1 | **1** |
| `prefers-reduced-motion: reduce` | 1 (rules already under `no-preference`) | **1**; courtesy `animation: none` is documentation, not load-bearing |

That matches the owner fail-visible-by-construction ruling. Dropping the fade is the accepted cost; the slide stays. `backwards` is now just holding the entrance offset, which is fine.

### Guard
The new source-text guard pins the structural invariant: no opacity token in the keyframe; no hiding declarations on hero selectors across media contexts; every keyframe named by a hero animation is opacity-free. Vacuity assertions make rename/delete fail loud. The six mutations in the PR body (including opacity 0 on the hero copy, which the old guard missed) are the right shape. Frontend CI green on this head.

### Residuals (ok to leave open; do not block merge on them)
- iPhone stall trigger still unproven — this removes the dependency, which is enough for the user-visible symptom.
- Stale CDN CSS vs build hash is still worth a post-deploy check before closing the issue.

No further code change needed from this lane. Owner merge gate noted; technical sign-off from polish is approve.

Issue #1557, opened 2026-08-30
[SECURITY] is_strategy_visible grants reasoning disclosure on is_published — debate/returns/detail leak to anon


dbrowneup:
**SECURITY — latent, one flag away from live. Found by the 2026-08-30 red-team.**

## The hole

`backend/archimedes/services/strategy_visibility.py:49` — `is_strategy_visible` returns True on `is_example OR is_published`. So "visible" means "readable by ANYONE," including anonymous.

**Consumers that surface private reasoning through this predicate:**
- `strategies_routes.py:2292` — `GET /{id}/debate` (the full bull/bear LLM transcript, shipped tonight in #1542)
- `strategies_routes.py:2217` — `/{id}/returns`
- `strategies_routes.py:2338` — `/{id}`
- `selection_bias_routes.py:678`, `paper_routes.py:42`

## The attack

The moment anything flips `is_published=True` on a user's strategy, an **anonymous** `GET /api/strategies/{id}/debate` returns that strategy's full generation debate. The route's own docstring claims "404 unless the caller owns the row" — **that is false**: the predicate it calls returns True for any published row, to anyone. (This is the same class as the brief leak caught and fixed tonight in #1547 — the brief gate had to be re-asked in *ownership-only* form; the trace/debate/returns surfaces still use the public-visibility form.)

## Fix

Reasoning-disclosure surfaces (debate transcript, full returns series, trace detail) must gate on **ownership**, not on `is_strategy_visible`. The `_owned_generated_strategy_responses` ownership shape used for the brief fix (#1547) is the pattern — apply it to the debate/returns/detail reasoning fields. Public-visibility stays correct for the *card* (name, methodology, headline metrics); it is wrong for the *reasoning*.

## Acceptance
- Anonymous `GET /{id}/debate` on a published strategy → 404/403, never the transcript.
- A different authenticated user → same.
- Guard demonstrated (non-tautological: fails against the current predicate).

Also fix the false docstring on the debate route.

PR #1782, opened 2026-09-02
[generation][passport] Surface the strategy engine's reasoning: live debate turns in the generation stream, a Reasoning section on the passport, human event copy

dbrowneup:
## The owner's words

Dan, 2026-09-01, dogfooding a live generation on his phone after producing **"Volatility-Relief Swing"**:

> You still can't see the actual reasoning traces anywhere as best I can tell. Seems like it should be somewhere in the generation stream and also in the strategy passport, or at least linked well in the strategy passport… the Reasoning page is supposed to be for reasoning from the execution engine rather than the strategy engine — so we shouldn't have strategy generation reasoning on the reasoning page.

What his stream showed him, verbatim:

```
#17 Tool result — synthesize → leader=Volatility-Relief Swing dsr=0.082005 of 1 entries
#18 Candidate drafted — … grounded in 5 papers
#19 Candidate evaluated — cand_neutral — DSR 0.082005 · OOS 0.198029
#20 Best selected — Picked cand_neutral from 1
#21 Trace hashed — 6ae4b039607e14…
#22 Persisted — /app/library?highlight=…
#24 Done
```

Nothing of the debate — proposer turns, bear/bull critiques, per-paper verdicts, why papers were discarded — appears during or after the run.

---

## The map: this was all already being produced

Every byte this PR puts on a screen was **already computed and already sanitized-on-write**. No new LLM call, no migration, no new endpoint, no gate change.

| Stage | Where | What it holds |
|---|---|---|
| Produced | `debate_engine._debate_round` | 4 paid LLM turns, fixed `[bull-r1, bear-r1, bull-r2, bear-r2]`. Each = `{role, round, verdict, claims[], discard[]}`; `claims[].arxiv_ids` id-checked against the evidence whitelist. |
| Per-paper tally | `debate_engine._aggregate_paper_verdicts` | 0-token deterministic roll-up: per paper, who cited it, who discarded it and why, with `unused` kept on purpose. |
| Attribution entry (#1739) | `generation_pipeline._transcript_with_paper_record` | One trailing `role:"attribution"` list entry carrying `paper_verdicts` + `fusion_reasoning`. |
| Persisted | `_persist_debate_transcripts` → `record_debate_transcript` | Sanitized at WRITE time — the module contract is that a raw read of `transcript_json` is already safe. |
| Read back | `GET /api/strategies/{id}/debate` | **Owner-only by design (#1557)**, 404 not 403. |

**Why none of it reached a screen.** On the wire, the debate existed only as four `tool_called` markers whose `args_summary` was `cards[:120]` — a truncated evidence card, never the turn. On the passport, the transcript panel *did* already exist and *was* already mounted — but it sits at the bottom of a ~1450-line page under a heading that does not say which engine it belongs to, the run ends by pointing at `/app/library?highlight=…` rather than at the passport, and the renderer walked the `role:"attribution"` entry as if it were a turn, so the per-paper verdicts and the discard reasons rendered as an almost-blank card.

---

### (a) The generation stream — the debate, live, sanitized

Two new SSE events, chosen so the stream is **1:1 with what gets stored** — what you watched is what was written.

* **`debate_turn`** ×4, emitted from `_debate_round` immediately after each `transcript.append(...)`: `{candidate_id, role, round, verdict, claims[], discard[], headline}`. `headline` is server-written copy over the *sanitized* turn — counts, not adjectives: *"Bear researcher, rebuttal — verdict: decline. 3 claims, 2 grounded in a named paper; 1 paper set aside."* Every count in it is re-derivable from the card rendered underneath.
* **`debate_attribution`** ×1, after the leader is chosen: the summary sentence, the per-paper verdict rows, `fusion_reasoning`, and the `papers_offered` / `distinct_mechanism_papers` pair.

**The load-bearing detail — one summary sentence, one writer.** `_paper_attribution_entry` is extracted out of `_transcript_with_paper_record`, and the persist path and the SSE path both call it. Two copies of that sentence would be two plausible, disagreeing claims about how many papers a single run engaged with.

**Sanitization is mandatory on this path and is why `_emit_debate_turn` exists.** `sanitize_transcript` normally runs inside `record_debate_transcript`, on the way into the DB. The SSE path never touches the DB, so without an explicit call the raw model text reaches a browser through a transport the write-time scrubber never sees — the exact leak `_sanitize_claim` / `_sanitize_paper_verdict` were written to close, one transport to the left. Both emitters scrub before emitting; a scrub failure drops the **event**, never the turn, and never falls back to the raw dict.

**Honest absence.** No LLM backend ⇒ `_debate_round` returns `[]` ⇒ **zero** `debate_turn` and **zero** `debate_attribution`. This is the same reasoning that already gates `debate_paper_verdicts` on `transcript` rather than on the evidence map: a table of all-`unused` rows reads as "the researchers looked at 30 papers and engaged with none" rather than as the absence it is. The gate narrows the *event*; the persisted row keeps its own #1739 rule unchanged.

### (b) The passport — a Reasoning section that names its engine

`StrategyReasoning.jsx` only; **`StrategyPassport.jsx` is untouched** (concurrent `dbrowneup/verdict-of-record-a` owns it), and no verdict/badge logic is anywhere near this diff.

1. The transcript is split into turns and the `role:"attribution"` entry, which now renders properly: summary sentence, the proposer's per-paper mechanism prose, and a **per-paper verdict table** — arXiv id · title · verdict chip (`cited`/`discarded`/`contested`/`unused`) · who cited or threw it out · the discard reasons. That last column is Dan's *"why papers were discarded"*, and it has been in the DB since #1739.
2. The two sections are now headed **"Strategy engine — generation debate"** and **"Execution engine — trading decisions"**, with a one-line pointer that `/app/reasoning` is the execution engine's page. Those two engines are the product; until now the panel was one heading called "Reasoning" over two unrelated kinds of it.
3. **The empty state stopped lying.** It said *"No debate transcript for this strategy"* — but the 404 has two causes, and for a non-owner reading a published strategy the other one is "not yours". It now states both and says it cannot tell you which.

### (c) Human copy, machine fields kept

New plain module `ui/src/generation-copy.js` — same pattern as `trace-binding.js`, and for the same reason: the copy is runnable in tests, and a component cannot quietly re-derive its own version of it.

`eventHeadline(name, data)` is the human line; `eventDetail(name, data)` is the backend's own `args_summary` / `result_summary` / candidate ids / redirect urls, **moved under a "Show machine details" toggle, not deleted**. One toggle rather than one per row, because this screen is watched on a phone.

Nothing verifiable is lost by default — the numbers a reader could independently check stay in the headline:

| Before | After |
|---|---|
| `Tool result — synthesize → leader=… dsr=0.082005 of 1 entries` | `Finished — Ranked the survivors and picked a leader` |
| `Tool called — propose_pool(steers=18, …)` | `Started — Searching the paper corpus and drafting candidate strategies` |
| `Candidate evaluated — cand_neutral — DSR 0.082005 · OOS 0.198029` | `Candidate graded — Rigor gate ran on the draft — DSR 0.082005 · PBO 0.12 · OOS 0.198029` |
| `Trace hashed — 6ae4b039607e14…` | `Provenance — Stamped this generation with a content hash — 6ae4b039607e14…` |
| `Persisted — /app/library?highlight=…` | `Saved — Now in your Library` |

`agent_iteration` stays deliberately conservative (`Progress — Stage 3 of 4`): the debate emits stages 1-3 of a declared max of 4, so an invented stage-name map would be a false label. An unknown tool is **named**, not described — copy written before a tool existed must not narrate it.

Also: `EVENT_LABELS` is the `addEventListener` subscription list, so an event absent from it is never received by the browser at all. That is why `backtest_running` / `backtest_done` / `backtest_failed` are emitted by the backend and invisible to this day. Keeping the label map and the copy map in one module lets a guard pin them against each other.

### (d) Reachability

`GenerationStream` gains one CTA on the done state — **"See the full reasoning →"** → `onNavigate('strategy', { strategyId })`. `persisted.redirect_url` is deliberately left alone (see Known findings).

### (e) The Reasoning page is untouched

Not a change — a **pinned absence**. `ui/src/components/Reasoning.jsx` has no diff, and a guard now asserts it fetches no `/debate`, handles no `debate_turn` / `debate_attribution`, and imports none of these renderers.

---

## Adversarial demos — every guard pasted red

Each mutation was applied to the real source, run, and reverted.

### Backend — `backend/tests/test_debate_stream_events.py` (8 tests)

**B1 — emit the raw turn instead of the sanitized one** (`payload = safe[0]` → `payload = turn`):
```
E   AssertionError: 'T3.5' reached the wire unscrubbed
FAILED tests/test_debate_stream_events.py::test_streamed_debate_turns_are_sanitized
```

**B2 — emit only round 1** (drop `_emit_debate_turn` from the round-2 loop):
```
E   AssertionError: assert 2 == 4
E   AssertionError: assert [('bull', 1), ('bear', 1)] == [('bull', 1),..., ('bear', 2)]
FAILED tests/test_debate_stream_events.py::test_streamed_debate_turns_are_sanitized
FAILED tests/test_debate_stream_events.py::test_every_debate_turn_is_streamed_exactly_once_and_matches_the_stored_turn
```

**B3 — reword the summary sentence on the stream copy only** (the drift the extraction exists to prevent):
```
E   AssertionError: assert {'fusion_reas...d': None, ...} == {'fusion_reas...d': None, ...}
E     Differing items:
E     {'verdict': 'Papers accounted for.'} != {'verdict': 'Paper attribution: 0 of 6 retrieved paper(s) were cited or discarded by name in this debate; 0 of 2 cited paper(s) name a mechanism this strategy trades.'}
FAILED tests/test_debate_stream_events.py::test_streamed_attribution_is_the_entry_that_gets_persisted
```

**B4 — emit the attribution unconditionally** (`if transcript:` → `if True:`):
```
E   AssertionError: assert [{'candidate_...ts': [], ...}] == []
E     Left contains one more item: {'candidate_id': 'cand_1', 'distinct_mechanism_papers': 0, 'fusion_reasoning': 'reasoning', 'paper_verdicts': [], ...}
FAILED tests/test_debate_stream_events.py::test_leaderboard_with_no_transcript_emits_no_attribution
```

### UI — `ui/test/generation-reasoning.test.js` (16 tests)

**U1 — add a label with no copy** (`backtest_done: "Backtest done"`) — the class of bug that already hides `backtest_*`:
```
✖ every subscribed event has a headline, and every headline is subscribed
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
  + actual - expected
  +   'backtest_done',
✖ every fixture is exercised — a new event cannot slip past the copy guards
```

**U2 — delete `debate_turn` from the subscription list** (the browser stops hearing the event):
```
✖ every subscribed event has a headline, and every headline is subscribed
  + actual - expected
  -   'debate_turn',
✖ the debate's own events are on the subscription list
✖ every fixture is exercised — a new event cannot slip past the copy guards
```

**U3 — restore the old terse copy** for `tool_result` and `candidate_evaluated`:
```
✖ no headline contains engine jargon
  AssertionError [ERR_ASSERTION]: tool_result headline still says dsr=: synthesize → leader=Volatility-Relief Swing dsr=0.082005 of 1 entries
✖ the numbers a reader can check stay in the headline
✖ tool copy distinguishes starting from finishing, and never narrates an unknown tool
```

**U4 — `eventDetail` drops the backend's own summary** (the opposite failure: human copy that swallows the machine fields):
```
✖ the machine strings survive verbatim under the details toggle
```

**U5 — import the debate panel into the execution Reasoning page** — Dan's line, as an enforceable one:
```
✖ the execution Reasoning page carries none of the generation debate
  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
```

**U6 — drop the "Strategy engine" heading:**
```
✖ the passport names both engines
  AssertionError [ERR_ASSERTION]: The input did not match the regular expression /Strategy engine — generation debate/
```

**U7 — render only the summary sentence** (drop the per-paper table — the state this PR fixes):
```
✖ the passport panel reads the per-paper record, not just the summary sentence
  AssertionError [ERR_ASSERTION]: The input did not match the regular expression /discard_reasons/
```

**U8 — `splitTranscript` walks the attribution entry as a turn** (the near-blank card):
```
✖ the attribution entry is split out of the turns rather than walked as one
  AssertionError [ERR_ASSERTION]: The input did not match the regular expression /e\.role !== "attribution"/
```

---

## Verification

| Run | Result |
|---|---|
| `pytest` — **full backend unit suite** | **5652 passed, 6 skipped, EXIT=0** — re-run after the fix-first commit (16m26s) |
| Targeted: `test_debate_stream_events`, `test_debate_engine`, `test_reasoning_disclosure_gates`, `test_strategies_routes` | **112 passed** |
| `ui` — `node --test` (full) | **934 passed, 0 failed** (913 before this PR + 21 new) |
| `npx eslint .` (ui, repo-wide) | clean, EXIT=0 |
| `npm run check:routes` (orphans + sitemap) | OK — 133 src files, 0 orphans; 6 public routes all in sitemap |
| `ruff format --check` on touched files | clean |
| `ruff check` on touched files | clean on `debate_engine.py`; **EXIT=1 on `generation_pipeline.py`** — 2 × `SIM905`, both confirmed pre-existing on `origin/main` (Known findings #4) |

**The owner-only gate is unchanged and pinned.** `test_reasoning_disclosure_gates.py` already carries the guard the design asked for — `test_published_debate_transcript_never_reaches_a_non_owner` runs a **published, non-owned** row against `/debate` and asserts a 404 with an owner-side positive control in the same test body. It passes on this branch, so this PR demonstrably did not widen `is_strategy_reasoning_visible`. Adding a duplicate to `test_strategies_routes.py` would have been a second copy of an existing guard, so I ran the existing one instead. The SSE stream has its own independent owner gate (`generate_routes._require_job_access`), so putting the debate on the wire needed no new gate either.

**Two existing source-text pins were updated, not weakened**, because the code they pin moved:
- `ui/test/anchor-state.test.js` — the claim-shape rules (`claimText`, both shapes, "not attributed to a listed paper") now read `DebatePaperVerdicts.jsx`, plus a **new** assertion that `StrategyReasoning.jsx` reaches them through that shared renderer rather than growing a second copy. The empty-state assertion follows the new, honest copy and gains two assertions about the second reason.
- `ui/test/app-visuals.test.js` — the "papers count comes only from real per-candidate citations" pin now reads `generation-copy.js`, and still asserts it against `GenerationStream.jsx` too.

---

## Known findings, filed separately rather than folded in

1. **`/app/reasoning` still advertises strategy-engine work as its own.** Its blurb says traces include *"a strategy construction from the Generate page"*, and `construction` is a live filter chip. That IS generation reasoning on the execution page — the same class of thing Dan flagged. Its empty state (`Reasoning.jsx:184-190`) carries a second, pre-existing false claim on top: *"each construction emits a trace"* — the generation pipeline emits none (`grep 'save_trace\|ReasoningTrace' generation_pipeline.py` is empty). It is a copy + filter change entangled with `trace-binding.js`, `Portfolio.jsx` and `ui/test/anchor-state.test.js`, so it wants its own PR rather than riding this one.
2. **`persisted.redirect_url` disagrees with its spec.** `docs/specs/generation-streaming-spec.md` says `/strategy/:id`; the code emits `/app/library?highlight=…`. Left alone deliberately — the "See the full reasoning →" CTA solves the reachability problem without changing a payload other consumers may read.
3. **`backtest_running` / `backtest_done` / `backtest_failed` are still invisible.** They are emitted and not subscribed. The new label/copy guard is what would catch this class going forward, but wiring the three events up is its own change.
4. **Two pre-existing `ruff` SIM905 findings in `generation_pipeline.py`** (`_COMMON_WORDS` / `_FINANCE_WORDS` built with `str.split()`) are untouched. `ruff check --fix` "fixed" them mid-work and I reverted that — unrelated churn in a one-logical-change PR.

---

## Fix-first pass (second commit)

The vet came back **FIX-FIRST**: the engineering held and every guard bit, but three pieces of new user-facing copy asserted things the code does not do — on the screen this PR exists to make trustworthy. Two stale guards and four nits came with it. All fixed below; every fix has its own mutation, pasted red.

### D1 — "so it can't be edited after the fact" was false twice over

`ui/src/generation-copy.js`. The line I shipped:

```js
`Hashed the reasoning so it can't be edited after the fact — ${String(d?.trace_hash || "").slice(0, 14)}…`
```

**(a) The hash covers no reasoning.** `generation_pipeline._persist_candidate` keccaks the canonical `{brief, candidate_id, strategy_name, weights, rigor_verdict}` tuple. The debate transcript and `fusion_reasoning` are nowhere in the preimage.

**(b) Nothing makes it tamper-evident.** It lands as `strategy_store.provenance_hash`. `grep -n 'save_trace\|ReasoningTrace' backend/archimedes/agents/generation_pipeline.py` returns **nothing** — the generation pipeline writes no `ReasoningTrace` and anchors nothing on chain (the docstring's "mirrored on-chain" is v1.5). Meanwhile `PATCH /api/strategies/{id}` renames the row afterwards and *deliberately does not recompute the hash*. The claim also borrows the affordance `ui/src/trace-binding.js` invented `anchored_only` precisely to deny.

Now: **`Provenance — Stamped this generation with a content hash — 6ae4b039607e14…`**. The label changed too: "Reasoning hashed" made the same false claim in two words.

### D2 — the researchers were not shown those papers

`ui/src/components/DebatePaperVerdicts.jsx`. The table rows come from `evidence_by_id`, which `debate_engine._propose_pool` documents as *"every paper that entered ANY **proposer** prompt on this run"*. What the bull/bear turns are actually handed is `_candidate_cards` — only the papers cited by the top `_DEBATE_CARD_MAX = 5` pool candidates. Most `unused` rows are papers the debaters never saw. The backend is careful here (`_aggregate_paper_verdicts` says "entered a proposer prompt"; the summary sentence says "retrieved paper(s)"); my copy escalated it.

| Was | Is |
|---|---|
| `Every paper the researchers were shown — including the ones neither of them named.` | `Every paper this run retrieved and put in front of the proposers — including the ones the debate never named.` |
| `unused: "Retrieved and shown to the researchers, but neither one named it."` | `unused: "Retrieved for this run, but named by neither researcher."` |

### D3 — `contested` does not imply two researchers

`_aggregate_paper_verdicts` sets `contested` on `cited_by AND discarded_by` and never compares the two lists: the bull citing a paper in round 1 and dropping it in round 2 lands there with one role on both sides.

`contested: "One researcher cited this paper; another threw it out."` → **`contested: "Cited in one turn and thrown out in another."`**

### D4 — a pre-existing guard had gone vacuous

`ui/test/anchor-state.test.js:315-316` asserted `panel.includes("Generation debate")` / `("Trading decisions")`. The engine-named headings are lowercase after the em dash, so those strings survived only in the `// ── Generation debate ──` section-divider comments. Demonstrated: blanking **both** rendered headings left that file at 31 pass / 0 fail. Retargeted to the rendered strings; the mutation is now red there too (M8 below).

### D5 — a sentence that claimed an out-of-scope fix had landed

`ui/src/components/StrategyReasoning.jsx` said *"The Reasoning page carries this engine's traces … nothing from the generation debate above appears there."* The second half is true; the first implies `/app/reasoning` is already execution-only, and it is not — its own blurb advertises *"a strategy construction from the Generate page"* and `construction` is a live filter chip. That is Known finding #1, filed and not fixed here. Now:

> The Reasoning page carries the execution engine's traces across every strategy and vault; the generation debate above is not among them, because an argument is not a decision that moved money.

### Nits taken

- **N6 — label/headline stutter.** The row renders `<label> — <headline>`, and four rows said their label twice: `Queued — Queued your brief`, `Stage — Stage 3 of 4`, `Saved — Saved to your Library`, `Done — Done — the strategy is ready`. A new guard walks every event name and fails on any headline that opens with its own label. It also caught one I had not spotted: `No candidate — No candidate beat the passive null`, where the *backend's* message collided with my label (now `Candidate dropped`).
- **N7 — unreachable tool copy.** Kept (a future emit site should get real copy, not `Running <tool>`), but the module comment and the test comment now say honestly which phase the backend actually emits — `tool_called` for six tools, `tool_result` for five — instead of reading as if all twenty sentences reach a screen.
- **N8 — kwarg collision.** `emit.emit("debate_attribution", candidate_id=…, **sanitize_transcript([attribution])[0])` would raise `TypeError: got multiple values for keyword argument 'candidate_id'` the day `_paper_attribution_entry` grows that key. Built as one merged dict instead.
- **N9 — hover-only legend.** `VERDICT_TITLE` shipped only as a `title=` tooltip, invisible on the phone this PR was written for. The meanings now render as a visible key **outside** the table's horizontal scroller (inside it, a legend scrolls away from the columns it explains), listing only the verdicts this run actually produced.
- **N10 — PR body accuracy.** The Verification table said `ruff check` was clean; it exits 1 on two pre-existing `SIM905` findings. Corrected below.

### Mutations for the fix-first pass — each applied to real source, run, reverted

| # | Mutation | Red |
|---|---|---|
| M1 | restore `Hashed the reasoning so it can't be edited after the fact` | `✖ the hash line claims a stamp, never immutability` — *did not match `/Stamped this generation with a content hash/`* |
| M2 | restore the `trace_hashed: "Reasoning hashed"` label | `✖ the hash line claims a stamp, never immutability` — *expected to not match `/Reasoning/`* |
| M3 | restore `unused: "Retrieved and shown to the researchers…"` | `✖ the passport panel reads the per-paper record…` + `✖ the per-paper copy claims only what the backend actually recorded` — *expected to not match `/shown to the researchers/`* |
| M4 | restore `contested: "One researcher cited this paper; another threw it out."` | `✖ the per-paper copy claims only what the backend actually recorded` — *expected to not match `/another threw it out/`* |
| M5 | restore the `Every paper the researchers were shown` caption | `✖ the per-paper copy claims only what the backend actually recorded` |
| M6 | delete the visible verdict legend, leaving the hover-only `title=` | `✖ the verdict meanings are visible, not hover-only` — *did not match `/\{VERDICT_TITLE\[v\]\}/`* |
| M7 | restore `carries this engine's traces … nothing … appears there` | `✖ the passport names both engines` — *did not match `/The Reasoning page carries the execution engine's traces/`* |
| M8 | blank the `Strategy engine — generation debate` heading | `✖ StrategyReasoning keeps the debate and the trading decisions separate` (**anchor-state.test.js — this is the assertion that was vacuous**) + `✖ the passport names both engines` |
| M9 | restore `persisted: () => "Saved to your Library"` | `✖ no log row says its label twice` — *`persisted renders "Saved — Saved to your Library"`* |
| M10 | add `candidate_id` to `_paper_attribution_entry`'s return | pre-fix: `TypeError: emit() got multiple values for keyword argument 'candidate_id'`; post-fix: green |

### On the vet's read of my three concerns

1. **Not duplicating the `/debate` 404 guard** — the vet agreed and re-verified the existing one bites at **8 failures** when `is_published` is added to `is_strategy_reasoning_visible`. Left as is.
2. **`debate_attribution` gated on a non-empty transcript while the persisted entry is not** — accepted as deliberate and documented. Unchanged.
3. **Log height 320 → 460px** — accepted ("with dual-regime runs the log now carries 8 turn cards plus 2 verdict tables"). Unchanged.
4. **Accessibility** — the vet agreed the `aria-live` announcement of a whole card per turn is the right call, still untested against a real AT.

---

## Copy nit (third commit) — a bear turn declines; it does not abstain

Daniel, on review: `TOOL_COPY` said *"Bear researcher is arguing for abstention"* / *"…argued for abstention"*, but a debate **turn's** verdict on the wire is `act` | `decline` — `debate_engine`'s critique prompt asks for exactly those two words.

**Abstain is a different, first-class outcome.** `_abstain_result` returns `generation_method="debate_abstain"` and `"Debate — abstain (hold current weights)"`, and it is reached by the CRISIS regime gate or by no candidate beating the passive null by ≥ 5 bps — **never** by a bear turn. So the old copy both announced an outcome the run had not reached and spent the word the real path needs.

| Was | Is |
|---|---|
| `Bear researcher is arguing for abstention` | `Bear researcher is arguing to decline` |
| `Bear researcher argued for abstention` | `Bear researcher argued to decline` |

`debate_bear_r2` already spoke about rebutting rather than about a verdict, and the bull keeps **"stating its case"**: both turns carry the same verdict field, so mirroring an "arguing to act" onto the bull would invent a claim the copy cannot see.

**Guard.** A new test pins the two bear rows to "to decline" and forbids `/abstain|abstention/` across **every** `TOOL_COPY` sentence, so the word cannot drift back onto a turn.

**MUTATION (verified red)** — restore `Bear researcher is arguing for abstention`:

```
✖ debate turn copy uses the turn contract's words, not the abstain path's
  AssertionError [ERR_ASSERTION]: debate_bear_r1.started names the abstain path
  for a turn whose verdict is act|decline
ℹ pass 20  ℹ fail 1
```

No existing test pinned the old strings — the stutter guard walks `EVENT_LABELS`, not `TOOL_COPY` — so nothing needed retargeting.

**The optional phone polish is not taken.** The generation-debate empty state already leads with its own short sentence — `"No generation debate is shown here."`, pinned as a whole sentence by `ui/test/anchor-state.test.js` — before the two causes. What is left would be splitting those causes onto their own block: a layout change, not the one-line copy edit that was offered, so it stays out of this PR.

`ui`: `node --test` **934 passed, 0 failed** (933 before). `npx eslint` clean on both touched files.

---

## Scope

- One logical change: carry reasoning that already exists to the two screens that should show it.
- `StrategyPassport.jsx` and `Strategies.jsx` are **not** in this diff (`dbrowneup/verdict-of-record-a` owns them), and neither is any verdict or badge logic.
- No API contract change beyond two additive SSE event names, both documented in the streaming spec.

Part of #1800


danielscoffee (commented):
**Improvements lane review** (honesty / public copy / passport polish) — comment only, not an approve. Owner vets before merge, as the PR says.

This is squarely in lane: live generation copy, passport reasoning headings, honest empty states, and provenance-shaped reachability ("See the full reasoning →"). The structural honesty choices look right — engines named separately, machine fields kept under one toggle, paper counts only from real citations, unknown tools named not narrated, SSE scrubbed before emit, owner-only `/debate` gate left alone.

### One copy honesty nit

`ui/src/generation-copy.js` `TOOL_COPY` for the bear:

- `debate_bear_r1`: "arguing for abstention" / "argued for abstention"
- (bull stays the softer "stating its case")

Debate turn verdicts on the wire are `act` | `decline` (see `_turn_headline` / the JSON schema in `debate_engine.py`). **Abstain** is a different first-class outcome — C-null / regime force / no candidate clears the cost bar (`debate_abstain`, "hold current weights"). Framing every bear opening as "arguing for abstention" collapses those two claims. Prefer something that matches the turn contract, e.g. "Bear researcher is arguing to decline" / "…argued to decline", and keep "abstention" for the actual abstain path if/when that surfaces in stream copy.

### Affirm the deferred finding

Known finding (1) — `/app/reasoning` still advertising strategy-construction as its own — is the same class of engine-mix Dan flagged. Right to keep it out of this PR; it is a natural follow-up in this lane once `trace-binding` / Portfolio / the filter chip can move together.

### Small optional polish (non-blocking)

The passport empty state for generation debate is correctly dual-cause and refuses to pick one. On a phone (the dogfood surface that started this) it is a long block; if it reads heavy in the live UI, a shorter first sentence plus the two causes as a quiet second line would keep the honesty without burying the panel.

No merge recommendation from me — leave for owner vet. Happy to take a tiny follow-up PR for the bear abstention→decline wording if you want it split out.

dbrowneup:
@danielscoffee — taken, in `7e5fa8d4`.

You were right about the contract: the bear's **turn** verdict is `act` | `decline` (that is literally what `debate_engine`'s critique prompt asks for), while abstain is the separate outcome `_abstain_result` produces — `generation_method="debate_abstain"`, *"Debate — abstain (hold current weights)"* — reached by the CRISIS regime gate or by no candidate beating the passive null by ≥ 5 bps, never by a bear turn. The old copy both announced an outcome the run had not reached and spent the word the real path needs.

- `debate_bear_r1` → **"Bear researcher is arguing to decline"** / **"…argued to decline"**.
- `debate_bear_r2` already talked about rebutting rather than about a verdict, so it is unchanged, and the bull keeps **"stating its case"** — both turns carry the same verdict field, so mirroring an "arguing to act" onto the bull would have invented a claim the copy cannot see.
- New guard in `ui/test/generation-reasoning.test.js`: the two bear rows are pinned to "to decline", and `/abstain|abstention/` is forbidden across **every** `TOOL_COPY` sentence, so the word cannot drift back onto a turn. Mutation verified red — restoring "arguing for abstention" gives 20 pass / 1 fail in that file.
- Nothing needed retargeting: no existing test pinned those strings, since the stutter guard walks `EVENT_LABELS` rather than `TOOL_COPY`.

**The phone polish I did not take**, because it is not the one-line change you offered: the generation-debate empty state already leads with a short standalone sentence — *"No generation debate is shown here."* — before the two causes, and `ui/test/anchor-state.test.js` pins that sentence whole. The remaining improvement is splitting the two causes onto their own block, which is a layout change; happy to do it, but not folded in here.

`ui`: `node --test` 934 passed / 0 failed (933 before). `npx eslint` clean on both touched files. PR body's copy section updated.

Still **do not merge — owner vets first.**

— via the owner's session

Issue #1411, opened 2026-08-20
APIN - Generation - Spike: offload pipeline execution to Lambda + measured per-generation cost in the quote


dbrowneup:
## Why

Generation currently executes inside the single shared Fargate task (~65% of its vCPU for ~48s per run, debate_backtest-dominated). Admission control (#1408) makes this safe but caps throughput at GENERATION_MAX_CONCURRENT. Scaling by raising task CPU scales the whole web tier for a bursty batch workload — a per-invocation compute service fits better and prices per run.

Measured unit economics (2026-08-20): ~15 nova-micro calls ≈ $0.03 LLM per generation; Lambda at ~1.8GB × 48s ≈ 85 GB-s ≈ $0.0015 compute. A measured cost model can back the quote instead of the current flat env price — `generation_payment.quote()` already documents this exact seam (#1217's measured budget).

## What (spike → ADR, not a direct build)

1. Prototype the pipeline worker as a Lambda container image reusing the backend package (entry: `run_generation` for one job; events still pushed to the job store so SSE is unchanged). Identify blockers: package size, cold start vs 48s runtime, VPC access to Redis/Aurora, Bedrock IAM, 15-min limit headroom.
2. Cost-model module: per-generation cost = measured LLM spend (cost_meter already tracks calls) + compute (duration × memory price) + fixed overhead; expose components in `GET /api/generate/quote` so the price can be cost + margin instead of a flat number. Flat env price remains the fallback and the default until flipped.
3. Deliverable: short ADR (offload architecture + quote model) with the prototype's measured numbers. Infra cost commitments need Dan's ack before any terraform.

## Acceptance (spike)

- ADR in docs/adr/ with measured cold-start, per-run duration, per-run cost from a real prototype invocation (or a documented blocker list if infeasible).
- Quote seam design references `generation_payment.quote()` and keeps the public-quote == 402-quote invariant.

## Anti-goals

- No production terraform in this issue.
- Do not change the flat-price default or any payment flag.
- Pricing/margin strategy itself lives in the private docs repo, not here.

dbrowneup:
[...]

PR #1785, opened 2026-09-02
[ux] A generation that delivers nothing gives the free slot back

dbrowneup:
## What the owner saw

Two generations attempted on a verified, wallet-less account. The second was a
brief about **BKLN/FLOT floating-rate carry**, and it died inside the pipeline
with:

> Generation is unavailable right now: the corpus yielded <2 papers for this
> steer — the society cannot fuse.

The Generate page then read **"Your account has 1 of 3 free generation left"**
(`ui/src/freeGenerations.js:129`, singular at `shown === 1` via the `plural`
ternary on :95). Nothing was delivered, and the free allowance had
gone 3 → 1 anyway.

## Root cause

`POST /api/generate/start` claims a free slot **before** the gate it opens, then
releases it if the request itself fails — `except BaseException:
free_generations.release(...); raise` around the entitlement gate and the
enqueue. That window closes at the enqueue.

The corpus failure lives on the other side of it. In
`agents/generation_pipeline.py` the insufficient-corpus branch emits
`error`/`GENERATION_UNAVAILABLE`, writes `update_status(job_id, "error", …)` and
**returns** — it does not raise. So `_run_with_cleanup` sees an ordinary
completion, and the only trace of the failure is the status left in the job
store. `_release_credit_if_undelivered` reads exactly that status for the PAID
path (#1441); the free tier had no equivalent, so every post-enqueue failure —
thin corpus, pipeline crash, cancel, timeout — kept the slot spent on a
generation that produced nothing.

## The change

`_run_with_cleanup`'s `finally` now calls `_release_free_slot_if_undelivered`
beside the credit one. It keeps the slot spent in exactly three cases and hands
it back otherwise:

| Case | Slot | Why |
|---|---|---|
| status `done` | stays spent | the run delivered |
| a `persisted` event carrying a `strategy_id` | stays spent | the run put a library row in the account and died afterwards — the cancel poll (`backtest_persist`) and the backtest fan-out both run AFTER `_persist_candidate`. Releasing would hand out the strategy **and** the free generation. |
| the job record or event log cannot be read | stays spent | silence is not evidence of non-delivery; the ledger row keeps its `job_id`, so it is still releasable by hand |
| anything else terminal, no strategy | **released** | the owner's case |

Deliberately **stricter than the paid path**, which restores on every non-`done`
status: money erring toward the payer is the right direction for an accounting
bug on a paywall, but re-granting an allowance that DID buy a library row is
just an over-grant.

Idempotent by construction — `release_grant_for_job` only moves a `used` row
stamped with that `job_id`, so a retried cleanup cannot mint a second free
generation, and a paid or flag-off run matches no row at all. No schema change:
`free_generation_grants` already carries `job_id`, `status` and `released_at`.

**Paid path: verified, not changed.** `_release_credit_if_undelivered` already
restores a consumed credit on every terminal state except `done`, so a $2 credit
already survived this exact failure. Two new tests pin that end to end through
`_run_with_cleanup` (`restore_for_job` called once for the corpus failure, never
for a delivered run) so the free-tier change cannot regress it. Left alone
otherwise — one logical change per PR.

Three comments that this change falsifies, and one log line that made a claim
that is no longer complete, were corrected alongside it: two in `/start` that
were true only up to the enqueue; `free_generation_grant.py`'s `job_id` column
comment, which said a `released` row is "NULL forever" when a row released by
`release_grant_for_job` deliberately KEEPS its job id; and `stamp_job`'s "only
its provenance is missing" — the stamp is now what the release finds the row
by.

## Adversarial pass — every new guard shown red

`pytest backend/tests/test_free_generation_gate.py` (43 tests, green on the
branch) run against six mutations of the shipped code:

**1. The pre-fix code** — drop `_release_free_slot_if_undelivered` from
`_run_with_cleanup`'s `finally`:

```
E   AssertionError: assert ['used'] == ['released']
E   AssertionError: assert ['used', 'used'] == ['released', 'released']
E   AssertionError: assert 2 == 3
FAILED TestAPipelineFailureReturnsTheFreeSlot::test_the_owners_sequence_a_corpus_failure_leaves_the_allowance_whole - AssertionError: assert ['used'] == ['released']
FAILED TestAPipelineFailureReturnsTheFreeSlot::test_two_attempts_that_both_fail_leave_all_three_slots - AssertionError: assert ['used', 'used'] == ['released', 'released']
FAILED TestTheReleaseIsIdempotentAndScoped::test_releasing_the_same_job_twice_returns_exactly_one_slot - AssertionError: assert 2 == 3
3 failed, 40 passed
```

`assert 2 == 3` is the owner's counter, reproduced.

**2. The over-grant** — hard-code `persisted = False` (i.e. the paid path's
"non-`done` ⇒ hand it back"):

```
FAILED TestTheReleaseGuardRejects::test_a_crash_AFTER_the_strategy_persisted_keeps_the_slot_spent - AssertionError: assert ['released'] == ['used']
FAILED TestTheReleaseGuardRejects::test_a_cancel_AFTER_the_strategy_persisted_keeps_the_slot_spent - AssertionError: assert ['released'] == ['used']
FAILED TestTheReleaseGuardRejects::test_an_unreadable_event_log_keeps_the_slot_spent_rather_than_guessing - AssertionError: assert ['released'] == ['used']
3 failed, 40 passed
```

**3. Guessing from missing data** — replace the `if job is None` branch with the
credit path's `status = (job or {}).get("status")`:

```
FAILED TestTheReleaseGuardRejects::test_a_vanished_job_record_keeps_the_slot_spent_rather_than_guessing - AssertionError: assert ['released'] == ['used']
1 failed, 42 passed
```

**4. Non-idempotent release** — drop `status=GRANT_USED` from
`release_grant_for_job`'s filter:

```
FAILED TestTheReleaseIsIdempotentAndScoped::test_releasing_the_same_job_twice_returns_exactly_one_slot - AssertionError: assert True is False
1 failed, 42 passed
```

**5. Guessing when the log cannot be read** — collapse
`_job_persisted_a_strategy`'s two `return None` exits (no `list_events` surface;
the read raised) onto `False`, i.e. "unreadable ⇒ nothing was delivered":

```
FAILED TestTheReleaseGuardRejects::test_an_unreadable_event_log_keeps_the_slot_spent_rather_than_guessing - AssertionError: assert ['released'] == ['used']
1 failed, 42 passed
```

**6. An unscoped release** — drop `job_id` from `release_grant_for_job`'s filter,
so the cleanup releases *any* `used` row:

```
FAILED TestTheReleaseIsIdempotentAndScoped::test_a_job_that_never_spent_a_free_slot_is_a_no_op - AssertionError: assert True is False
FAILED TestTheReleaseIsIdempotentAndScoped::test_one_users_failure_never_releases_anothers_slot - AssertionError: assert False is True
FAILED TestTheReleaseIsIdempotentAndScoped::test_an_unstamped_slot_is_reported_as_not_released_rather_than_swept - AssertionError: assert True is False
3 failed, 40 passed
```

One user's failed generation cannot hand another user's slot back, and that is
pinned rather than assumed.

(A further attempt — `if False:` in place of the `job is None` branch — did NOT go
red, because it crashes into the outer `except` and the slot stays spent for the
wrong reason. Recorded because a mutation that passes for an accidental reason
proves nothing; it was replaced by mutation 3 above.)

The stand-in for the pipeline is not a bare `raise`: `_corpus_failure` performs
the same two store writes the real branch performs and then RETURNS, which is
the shape that made this bug invisible.
`test_the_corpus_failure_message_still_exists_in_the_pipeline` pins the literal
reason string and `code="GENERATION_UNAVAILABLE"` against
`generation_pipeline`'s source, so the reproduction cannot drift away from the
branch it reproduces.

## Verification

```
pytest backend/tests/test_free_generation_gate.py -q            → 43 passed        (exit 0)
pytest test_free_generation_gate + test_generation_credits
     + test_generation_timeout + test_generation_concurrency
     + test_generate_credits_route -q                           → 89 passed        (exit 0)
ruff check <4 touched python files>                             → All checks passed (exit 0)
ruff format --check <4 touched python files>                    → 4 files already formatted (exit 0)
pytest -m "not integration" -q  (6 chunks, see note)            → 5659 passed, 4 skipped
                                                                  (exit 0 on every chunk)
```

No `ui/` files touched, so no npm run.

**About the full-suite run.** One `pytest -m "not integration"` process on this
machine is killed by the 10-minute tool bound, so it was run as six file-chunks
covering all 336 test files. Chunk 3 died with **exit 139 (SIGSEGV)** — twice,
at two DIFFERENT files (`test_strategy_fusion.py`, then `test_paper_rag.py`),
i.e. not attributable to any one test. Control: the same chunk on a pristine
`origin/main` worktree with none of this branch's changes **also exits 139**, at
`test_strategy_fusion.py`. It is a pre-existing native crash in a long-lived
process in this conda env, not something this PR introduces — nothing here
touches fusion, RAG or any native library. Split in two, the same 825 tests pass
green (615 + 210, exit 0 each). Totals above are the sum across chunks; every
chunk exited 0.

## Concerns for the owner

- **The banner does not refresh itself.** `FreeGenerationBanner.jsx` fetches
  `GET /api/account/usage` once on mount (`useEffect(..., [])`). The number is
  now correct on the server, but a user who watches a generation fail on the
  Generate page still sees the pre-failure count until the component remounts.
  Left out of this PR: it is a UI change on the Generate page, which
  `dbrowneup/reasoning-traces-surfaced` is editing.
- **`scripts/run_generation_job.py` bypasses this cleanup entirely.** It calls
  `run_generation` directly rather than through `_run_with_cleanup`, so on that
  path neither the free slot nor the paid credit is released. Pre-existing
  parity gap — the credit had it too — and nothing in the serving path imports
  that module (it is the #1411 offload spike), but its docstring's "the cost
  meter and the credit ledger all run on their normal code paths" is now wrong
  about one more thing. Worth a line in the ADR.
- **A run whose process dies never runs its `finally`.** A container roll mid-run
  leaves the job `running` until the stalled-heartbeat read reports it, and no
  cleanup executes — so the slot stays spent. The paid credit has the identical
  gap. A sweeper over stalled jobs would close both; not attempted here.
- **A slot whose `stamp_job` write failed cannot be released automatically** —
  the release finds the row by `job_id`. `stamp_job`'s error log now says so
  instead of reassuring.
- **The delivery oracle is the `persisted` event.** It is the honest signal
  available at cleanup time (the job record carries no strategy id on a crash
  path), but the event log has a shorter TTL than the job record. That is why an
  unreadable log keeps the slot spent rather than releasing it. Reading it does
  not touch `generation_pipeline`.

## Collision check

No code overlap with the concurrent branches, and the one shared file is a doc.
`dbrowneup/verdict-of-record-a`
edits `Strategies.jsx`, `StrategyPassport.jsx`, `strategies_routes.py`,
`passport_loader` and `generation_pipeline`'s passport writes;
`dbrowneup/reasoning-traces-surfaced` edits the generation stream components and
the SSE emit code. This PR touches **no** `ui/` file, and **no** file under
`agents/` — `generation_pipeline.py` is only read (by a test asserting its error
string still exists). In `generate_routes.py` the new code sits in the
background-cleanup helpers, ~200 lines below `/start`'s handler body.

Two shared files, both checked rather than assumed:

- `docs/agent-quickstart.md` **is** also edited by `dbrowneup/verdict-of-record-a`
  (~:567 and ~:712). This PR's edit is at ~:170; the hunks are far apart and the
  file merges clean.
- `dbrowneup/ux-too-few-papers` restructures the *exact* insufficient-corpus
  branch that this PR's source-pinning test asserts against — the highest-risk
  neighbour. Union tested in review: `origin/main` + this branch + `ux-too-few-papers`
  merged in a scratch worktree with **no conflicts**, and
  `test_free_generation_gate.py` plus that branch's two new files ran **61
  passed**. Both pinned forms survive there — the reason string ("the corpus
  yielded <2 papers for this steer — the society cannot fuse", `:1332` on this
  branch) and the keyword form `code="GENERATION_UNAVAILABLE",` (`:1338`), which
  the merge relocates but does not alter.

### DIFF
--- a/backend/archimedes/api/generate_routes.py
+++ b/backend/archimedes/api/generate_routes.py
@@ -539,6 +539,11 @@ async def start_generation(
     # gate can raise 402 and the enqueue can error, and either would leave the
     # allowance spent on a generation that never ran. Same shape, and the same
     # reason, as _paywall_with_credit's `except BaseException: void; raise`.
+    # This covers only what THIS request can see fail. Everything that goes
+    # wrong after the enqueue — the corpus yielding too few papers to fuse, a
+    # pipeline crash, a cancel — is released by
+    # `_release_free_slot_if_undelivered` in the run's own `finally`, keyed on
+    # the job id stamped below.
     try:
         # Paid-tier gating (T1.8): a premium (Anthropic) model requires a
         # wallet-connected entitlement. Enforced BEFORE the job is enqueued so a
@@ -584,7 +589,9 @@ async def start_generation(
 
     # The free slot is bound to its generation only now, once the job is queued
     # — the same "spend only what was delivered" point at which a paid credit is
-    # consumed below.
+    # consumed below. The stamp is also what the terminal-failure release finds
+    # the row by, so an unstamped slot cannot be handed back later (see
+    # `free_generations.stamp_job`, which says so where it fails).
     if free_grant_id is not None:
         free_generations.stamp_job(free_grant_id, job_id=job_id)
 
@@ -713,6 +720,136 @@ async def _release_credit_if_undelivered(job_id: str, store) -> None:
         logger.exception("could not evaluate credit release for job %s", sanitize_log_value(job_id))
 
 
+async def _job_persisted_a_strategy(job_id: str, store) -> bool | None:
+    """Did this run put a strategy in the caller's library? ``None`` = cannot tell.
+
+    ``status == "done"`` is NOT the same question. A run can persist the
+    winning strategy (``generation_pipeline`` § "K=1 persistence") and only
+    then die — the backtest fan-out crashing, the user cancelling at the
+    ``backtest_persist`` stage boundary, the process rolling — which leaves a
+    terminal status of ``error``/``cancelled`` beside a strategy that is
+    genuinely in the library. Handing a free generation back for that run would
+    give the account the strategy AND the slot.
+
+    The oracle is the ``persisted`` event the pipeline pushes immediately after
+    ``_persist_candidate`` returns a real ``strategy_id``. Read-only: this is a
+    consumer of the event log, and nothing here writes to it.
+
+    Three-valued on purpose. The event log has a shorter TTL than the job
+    record, the read can fail, and a store double may not expose the surface at
+    all (the same "a store that predates this surface is skipped" allowance
+    ``_abort_if_cancel_requested`` makes). Collapsing any of those onto
+    ``False`` would report "nothing was delivered" on no evidence, and the
+    caller would hand back a slot that may have bought a library row.
+    ``None`` says only what is true — we could not look — and the caller keeps
+    the slot spent, which is recoverable: the ledger row is on disk with its
+    job id.
+    """
+    lister = getattr(store, "list_events", None)
+    if lister is None:
+        logger.debug("free-slot release: store exposes no event log for job %s", sanitize_log_value(job_id))
+        return None
+    try:
+        events = await lister(job_id)
+    except Exception:
+        logger.warning(
+            "free-slot release: could not read the event log for job %s",
+            sanitize_log_value(job_id),
+            exc_info=True,
+        )
+        return None
+    for ev in events:
+        if ev.get("event") != "persisted":
+            continue
+        data = ev.get("data")
+        if isinstance(data, dict) and data.get("strategy_id"):
+            return True
+    return False
+
+
+async def _release_free_slot_if_undelivered(job_id: str, store) -> None:
+    """Give a free generation back unless the job actually delivered (#1643).
+
+    The free-tier sibling of :func:`_release_credit_if_undelivered`, and the
+    fix for what the owner saw on 2026-09-01: a brief whose corpus yielded
+    fewer than two papers failed INSIDE the pipeline with "the society cannot
+    fuse", and the allowance still went 3 → 2. ``/start`` only ever released a
+    slot for failures it could see itself (the entitlement gate, the enqueue —
+    its ``except BaseException: release; raise``); every failure after the
+    enqueue kept the slot spent on a generation that produced nothing.
+
+    Two states keep the slot spent, and only two:
+
+    ``done``
+        The run delivered. Checked first and cheaply.
+    a persisted strategy
+        The run put a strategy in the library and died afterwards. Giving the
+        slot back here would hand out the strategy and the free generation.
+        Deliberately stricter than the paid path, which restores on every
+        non-``done`` status (see :func:`_release_credit_if_undelivered`) —
+        money erring toward the payer is the right direction for an accounting
+        bug on a paywall, but re-granting an allowance that DID buy a library
+        row is just an over-grant.
+
+    …and one more that keeps it spent without deciding anything: a job record
+    or event log this cannot read. Silence is not evidence of non-delivery.
+
+    Idempotent through ``release_grant_for_job``: only a ``used`` row with this
+    ``job_id`` moves, so a retried cleanup cannot mint a second free slot out of
+    one claim. A paid or flag-off run matches no row and this is a no-op.
+
+    Fail-safe: this runs in a background task's ``finally`` with nobody to
+    report to, so it logs and returns. A read failure leaves the slot spent —
+    the honest direction when we cannot tell whether the run delivered, and
+    recoverable, since the ledger row is still on disk with its job id.
+    """
+    try:
+        job = await store.get(job_id)
+        if job is None:
+            # Deliberately NOT the credit path's "reads as undelivered". This
+            # runs in the run's own `finally`, so a job record that is already
+            # gone means the store lost it, not that time passed — and with no
+            # record there is no event log either, so "no strategy persisted"
+            # would be an inference from missing data rather than a reading of
+            # it. The slot stays spent and says so; the ledger row keeps its
+            # job id, so it can still be released by hand.
+            logger.warning(
+                "job %s has no job record at cleanup — the free generation stays spent (undecidable)",
+                sanitize_log_value(job_id),
+            )
+            return
+        status = job.get("status")
+        if status == "done":
+            return
+        persisted = await _job_persisted_a_strategy(job_id, store)
+        if persisted is None:
+            logger.warning(
+                "job %s ended as %s and its event log could not be read — the free generation "
+                "stays spent rather than being handed back on no evidence",
+                sanitize_log_value(job_id),
+                sanitize_log_value(str(status)),
+            )
+            return
+        if persisted:
+            logger.info(
+                "job %s ended as %s but persisted a strategy — the free generation stays spent",
+                sanitize_log_value(job_id),
+                sanitize_log_value(str(status)),
+            )
+            return
+        if free_generations.release_for_job(job_id):
+            logger.warning(
+                "job %s ended as %s with no strategy — free generation returned to the account",
+                sanitize_log_value(job_id),
+                sanitize_log_value(str(status)),
+            )
+    except Exception:
+        logger.exception(
+            "could not evaluate the free-generation release for job %s — the slot stays spent",
+            sanitize_log_value(job_id),
+        )
+
+
 async def _run_with_cleanup(
     job_id: str,
     brief: GenerateBrief,
@@ -811,6 +948,10 @@ async def _run_with_cleanup(
         with contextlib.suppress(asyncio.CancelledError):
             await heartbeat_task
         await _release_credit_if_undelivered(job_id, store)
+        # …and the free-tier equivalent. Both run: a run is funded by a credit
+        # or by a free slot, never both, and each helper is a no-op for the
+        # funding it does not own.
+        await _release_free_slot_if_undelivered(job_id, store)
 
 
 @generate_router.get("/stream/{job_id}")
--- a/backend/archimedes/models/free_generation_grant.py
+++ b/backend/archimedes/models/free_generation_grant.py
@@ -26,10 +26,13 @@ the thing it authorises):
     A free slot was spent on a generation. ``job_id`` is stamped once the job
     reaches the queue.
 ``released``
-    The slot was claimed but the request never reached the queue (the premium
-    -model entitlement gate refused it, the enqueue errored). Nothing was
-    delivered, so the allowance is handed back. Kept as a row rather than
-    deleted so the ledger shows the attempt.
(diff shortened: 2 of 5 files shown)


danielscoffee (approved):
In lane: free-tier honesty when a generation delivers nothing. The owner's corpus-failure case (slot spent, nothing in the library) is the right bug, and the guard is correctly stricter than the paid path — a `persisted` strategy keeps the slot even on a later error/cancel, so we don't over-grant.

What looks good
- `_release_free_slot_if_undelivered` in `_run_with_cleanup`'s `finally`, keyed on stamped `job_id`
- Three-valued event-log read (`True` / `False` / `None`) so silence keeps the slot spent
- Idempotent `release_grant_for_job` (only `used` rows move)
- Adversarial mutations on the four failure modes, plus paid-path pins so this PR can't regress #1441
- Docs line in `agent-quickstart.md` matches the behavior (re-read remaining after a failure)

Noted, correctly deferred
- Banner still mounts once — server count is right, Generate page can look stale until remount (tracked against `reasoning-traces-surfaced`)
- Process death / no `finally` is the same gap the paid credit already has; a stalled-job sweeper would close both
- Unstamped grant can't auto-release; the new `stamp_job` log is honest about that

Owner-vet gate stands ("Do not merge — owner vets first"). No follow-up PR from me unless the banner refresh lands orphaned after that other branch merges.

dbrowneup:
Review verdict was **FIX-FIRST — 2 defects, both prose-truth**. Both applied at `7b803ca3`; no behaviour, signature or schema change.

**Defect 1 — `backend/archimedes/models/free_generation_grant.py:83-88`, a column comment this PR itself falsified.** It read "NULL until the job reaches the queue, and NULL forever on a `released` row — there was no job", while `release_grant_for_job` deliberately leaves `job_id` in place (its own docstring at :206-207 says so, and `test_the_owners_sequence_a_corpus_failure_leaves_the_allowance_whole` asserts `rows[0].job_id == "job-carry"` at `test_free_generation_gate.py:855`). The module docstring's `released` paragraph had been widened and this comment missed, so the file contradicted itself two screens apart. Replaced with the verifier's exact text: NULL only on the `release_grant` path (request never queued); a row released by `release_grant_for_job` KEEPS its job id, which is both how the ledger shows which run burned the slot and the key the terminal-failure release finds the row by.

**Defect 2 — the owner's screenshot was misquoted as plural, in three places.** `ui/src/freeGenerations.js:95` is `const plural = shown === 1 ? "generation" : "generations";` and :129 renders `Your account has ${shown} of ${allowance} free ${plural} left.`, so at one remaining the product says **"1 of 3 free generation left"** — singular, exactly as the screenshot. Fixed in:
- `backend/tests/test_free_generation_gate.py:820` docstring — `generations left` → `generation left`;
- the PR body's "What the owner saw", which additionally attributed the plural form to `ui/src/freeGenerations.js`, a file that cannot emit it — now quotes the singular and cites `:129` / the `:95` ternary;
- the commit body. History was not rewritten (the reviewed commit `62efac4c` is left intact for the owner's vet); `7b803ca3`'s message carries the correction of record and says so explicitly.

**Also folded into the PR body**, all true against the diff and the review:
- Adversarial section goes from four mutations to **six** — the two the verifier ran that the body did not list are now written up with their real output: collapsing `_job_persisted_a_strategy`'s two `return None` exits onto `False` (**1 failed, 42 passed**) and dropping `job_id` from `release_grant_for_job`'s filter so the cleanup releases *any* `used` row (**3 failed, 40 passed** — one user's failure cannot hand back another user's slot).
- Collision section no longer opens "Nothing here overlaps". `docs/agent-quickstart.md` **is** shared with `dbrowneup/verdict-of-record-a` (~:567 / ~:712 there, :170 here — merges clean), and the union with `dbrowneup/ux-too-few-papers`, which restructures the very insufficient-corpus branch the source-pinning test asserts against, is recorded: merged with no conflicts, 61 passed, both pinned forms surviving (the reason string at `generation_pipeline.py:1332` and `code="GENERATION_UNAVAILABLE",` at :1338 on this branch — the review's :1357/:1364 were the merged tree's offsets, so the body cites this branch's and says the merge only relocates them).
- New owner concern: `backend/archimedes/scripts/run_generation_job.py:196` calls `run_generation` directly, bypassing `_run_with_cleanup`, so on that path neither the free slot nor the paid credit is released. Pre-existing parity gap on the #1411 offload spike, nothing in the serving path imports it — but its docstring's "the cost meter and the credit ledger all run on their normal code paths" is now wrong about one more thing.
- The prose-correction paragraph now says three comments were corrected, not two.

**Re-verification after the edit** (`-p no:cacheprovider`, from the branch worktree):

```
pytest backend/tests/test_free_generation_gate.py -q            → 43 passed        (exit 0)
pytest free_generation_gate + generation_credits
     + generation_timeout + generation_concurrency
     + generate_credits_route -q                                → 89 passed        (exit 0)
ruff check <4 touched python files>                             → All checks passed (exit 0)
ruff format --check <4 touched python files>                    → 4 files already formatted (exit 0)
```

All six mutations re-run independently, each reproducing the counts and assert texts claimed:

| # | Mutation | Result |
|---|---|---|
| 1 | drop `_release_free_slot_if_undelivered` from the `finally` | **3 failed, 40 passed** — `assert ['used'] == ['released']`, `assert ['used','used'] == ['released','released']`, `assert 2 == 3` |
| 2 | hard-code `persisted = False` | **3 failed, 40 passed** — `assert ['released'] == ['used']` ×3 |
| 3 | `status = (job or {}).get("status")` in place of the `job is None` branch | **1 failed, 42 passed** |
| 4 | drop `status=GRANT_USED` from the release filter | **1 failed, 42 passed** — `assert True is False` |
| 5 | collapse `_job_persisted_a_strategy`'s `return None` → `False` | **1 failed, 42 passed** |
| 6 | drop `job_id` from the release filter | **3 failed, 40 passed** |

No `ui/` file touched, so no npm/node run. Full-suite claim and its SIGSEGV caveat are unchanged from the previous commit and still disclosed in the body.

danielscoffee (approved):
Re-checked HEAD `7b803ca3` against the two FIX-FIRST prose defects. Both now match the code.

- `job_id` column comment: NULL only on `release_grant` (request never queued). `release_grant_for_job` keeps the id — it only flips status/`released_at`, and the test still asserts `rows[0].job_id == "job-carry"`.
- Owner screenshot quote is singular: **1 of 3 free generation left**, matching `freeGenerations.js:95` (`shown === 1 ? "generation"`) and `:129`.

No behaviour, signature, or schema change. Approve stands on HEAD. Owner-vet still ("Do not merge").

PR #1786, opened 2026-09-02
[infra][waf] /api/rigor/verify may carry a year of returns: scoped SizeRestrictions_BODY override + explicit row cap (#1749)

### LINKED ISSUE #1749: [infra] WAF 403s verify payloads over ~8KB — one year of daily returns cannot be verified
Dogfood 2026-09-01 (agent-cli), deterministic: 160 rows (8,076B) → 200; 165 rows (8,326B) → 403 HTML from `awselb/2.0`; 252 rows (one trading year) → 403 every time. The product punishes exactly the input that gives DSR/OOS statistical power, and the CLI folds it into a generic `exit 2 http_error`. Fix: raise/scope the WAF body-size rule for the verify route (security tradeoff: scope to authenticated POST /api/... verify path, keep the global limit elsewhere) + make the CLI surface a WAF/413-class error distinctly. Guard: an integration-marked test with a 300-row payload against the local stack, and a CLI unit test for the error mapping.

dbrowneup:
Closes #1749 (WAF half + the application cap; the CLI error-mapping half is left open — see "Not in this PR").

## The boundary, measured

`AWSManagedRulesCommonRuleSet` is attached to the ALB in full BLOCK mode with **no `rule_action_override` and no `scope_down_statement`** (`infra/waf.tf:46-66` before this change). Its `SizeRestrictions_BODY` rule blocks any body over the WAF body-inspection limit for a REGIONAL web ACL — 8,192 bytes. `POST /api/rigor/verify` serialises one `{date, daily_return}` object per bar and had **no `max_length`** (`rigor_verify_routes.py:120-126`).

Dogfood 2026-09-01 (agent-cli), deterministic:

| rows | bytes | result | |
|---|---|---|---|
| 160 | 8,076 | `200` | measured |
| 165 | 8,326 | `403` | measured — HTML from `awselb/2.0`, not a FastAPI error body |
| 252 | — | `403` **every time** | measured — **one trading year** |

Those two measured points give **50.5 B/row** including the envelope (`8,076 / 160 = 50.5`; `(8,326 − 8,076) / 5 = 50.0`), so 8,192 is crossed at **~163 rows** and a 252-bar year is ~12.7 KB (extrapolated, not measured). `8,076 < 8,192 < 8,326`.

The endpoint whose entire purpose is DSR/OOS statistical power rejected exactly the sample size that gives it power.

## Design choice: scoped, not a blanket override

A bare `rule_action_override → count` on `SizeRestrictions_BODY` fixes the 403 and **drops the 8 KB body ceiling for the whole site** — a real DoS/parser-abuse control. So the override is paired with a replacement:

1. **`rule_action_override { name = "SizeRestrictions_BODY"; action_to_use { count {} } }`** on the CRS group (priority 10). `count` rather than deletion keeps the managed rule's CloudWatch metric and `awswaf:managed:aws:core-rule-set:SizeRestrictions_Body` label, so oversize traffic stays observable. **Only this one rule** is softened — XSS, LFI, RFI, the other `SizeRestrictions_*`, `NoUserAgent_HEADER` etc. still block on every path including `/api/rigor/verify`, and `override_action { none {} }` on the group itself is untouched.
2. **New custom rule `oversize-body-except-rigor-verify`, priority 11** — immediately after the group it replaces, before `aws-known-bad-inputs` (20) / `aws-ip-reputation` (30) / `aws-sqli` (40). `and_statement` of a `size_constraint_statement` on `body` (`GT 8192`, `oversize_handling = "MATCH"`) AND a `not_statement` around a `byte_match_statement` on `uri_path`, `search_string = "/api/rigor/verify"`, `positional_constraint = "EXACTLY"`. Every other route keeps the 8 KB ceiling.

   Mechanism, per AWS's "Oversize handling for request components": on a REGIONAL ACL protecting an ALB only the first 8,192 bytes reach WAF, so the `GT 8192` comparison can never be satisfied by the *inspected* portion alone — `oversize_handling = "MATCH"` is what actually fires on a >8 KB body. The `GT` comparison is kept as a backstop in case the limit is ever raised. **Not verified without an apply:** which of the two conditions fires in the live ACL. Either way the effective condition is "body > 8 KB".

   `EXACTLY`, not a prefix, is load-bearing: `/api/rigor/verifyX` and `/api/rigor/verify/2` keep the ceiling. The narrowing is defensible because the route is auth-gated (`require_current_user`) and its schema is closed.

3. **Application-side cap** (`rigor_verify_routes.py`): `returns` is capped at `_MAX_RETURN_ROWS = 2600` — ten years of daily bars (10 × 252 = 2,520 plus calendar headroom), **~122 KB measured** at 6-decimal returns, ~131 KB extrapolated at the production 50.5 B/row. `max_length` carries the declarative contract into the OpenAPI schema; a `mode="before"` field validator runs first and produces the actual message, because pydantic's own `"List should have at most 2600 items after validation"` tells the caller nothing:

   > `returns has 2601 rows; the maximum is 2600 (~10 years of daily bars). This is a payload cap, not a statistical one: split the series or aggregate to a coarser frequency.`

   **What that cap is and is not** (corrected in review — the first draft of this PR claimed a decade was "the honest outer edge of what a daily-bar rigor verdict can claim" and that ~122 KB sits "comfortably inside FastAPI/uvicorn's defaults"; both were wrong):

   - It is a **row** cap, not a byte ceiling, and it is applied **post-parse**. FastAPI's request handler does `await request.body()` / `await request.json()` *before* `solve_dependencies`, so the whole body is buffered and `json.loads`'d before `require_current_user` and before this validator runs. An over-cap — or unauthenticated — request is still parsed in full first.
   - FastAPI and uvicorn have **no** request-body size default; there was nothing to be "inside". With the WAF exception in place the only **byte** ceiling left on this path is nginx's implicit 1 MB `client_max_body_size` (`nginx/nginx.conf` sets none; nginx is the ALB target, `infra/ecs.tf:949-952`). That residual — ~1 MB against nginx's own `limit_req api_read 60r/m burst=20`, under the 1000/5min WAF rate rule — is bounded and accepted here; an explicit `client_max_body_size` is the right hardening but is a second logical change (see "Not in this PR").
   - The cap is not statistical: `compute_dsr_hac_and_iid` / `compute_oos_sharpe` are fine over 3,000 bars and nothing in the module derives a 10-year validity horizon. 2,600 is a size choice, and the 422 now says so.

### `association_config` — verified, then deliberately omitted

The scope IS `REGIONAL`, but the associated resource type must also support raising the limit. **Verified locally** against the provider this stack resolves to (`terraform init -backend=false` under `~> 5.0` from `infra/main.tf` → `hashicorp/aws v5.100.0`), by writing each candidate block into `aws_wafv2_web_acl.association_config.request_body` and running `terraform validate`:

```
api_gateway                 Success! The configuration is valid.
app_runner_service          Success! The configuration is valid.
cloudfront                  Success! The configuration is valid.
cognito_user_pool           Success! The configuration is valid.
verified_access_instance    Success! The configuration is valid.

application_load_balancer   Error: Unsupported block type
                            Blocks of type "application_load_balancer" are not expected here.
alb / load_balancer         (same error)
```

There is no ALB form of `default_size_inspection_limit` to write, matching the AWS docs. **Independently reproduced in review** on a scratch copy against the same 5.100.0 — `application_load_balancer` / `alb` / `load_balancer` all `Unsupported block type`, the other five valid — so this is settled, not an inference. **Consequence, stated plainly:** the ALB body-inspection limit stays at 8,192 bytes. A 13 KB verify payload now *reaches* the application (that is the fix), but WAF's signature rules will have inspected only its first 8 KB — bytes past 8,192 on that one path arrive un-inspected. On an authenticated route with a closed schema and a 2,600-row cap that residual is bounded and accepted. Full-body inspection there would mean moving enforcement onto CloudFront, a different change with its own WCU cost.

## Guards, shown red on the input they reject

`backend/tests/test_waf_verify_body.py` — 15 tests, hermetic (parses `infra/waf.tf` as HCL and exercises the model + router over in-process ASGI; no AWS, no terraform binary, no network, no DB). `terraform validate` catches none of this: a deleted override, an override on the wrong rule name, a prefix-widened exception, an **inverted NOT/AND nesting** and a rule ordered ahead of the group are all valid HCL.

The rule's statement **tree is parsed and asserted node by node**, not substring-matched. Review caught the first version of these guards passing the one mutation that is a site-wide outage (V1 below): the only nesting assertion was `"and_statement" in _collapse(rule)`, and that string survives inversion. The guard now asserts `rule.statement` **is** the `and_statement`; that its two branches are a `size_constraint_statement` and a `not_statement` **as siblings**; that no `size_constraint_statement` appears anywhere beneath the `not_statement`; and that the whole rule holds exactly one and/not/size/byte_match and zero `or_statement`. Attribute reads (`GT`, `8192`, `oversize_handling`, `EXACTLY`, `uri_path`, `text_transformation NONE`) bind to the block that owns them.

**V1 — invert the nesting: `not_statement` WRAPS the `and_statement`** (the outage case; `terraform validate` **Success**, `terraform fmt -check` clean, and it passed all 14 of the previous guards)

```
E   AssertionError: rule.statement must contain exactly one block, `and_statement {...}`; got ['not_statement']
FAILED test_custom_rule_blocks_oversize_bodies
FAILED test_the_exception_is_exactly_the_verify_path_and_is_negated
FAILED test_the_negation_is_a_sibling_of_the_size_test_not_a_wrapper
3 failed, 12 passed          (pytest exit 1)
```

What it would have meant applied: `NOT(body > 8 KB AND path != /api/rigor/verify)` — block every request site-wide **except** an oversize POST to `/api/rigor/verify`.

**V1b — negate the SIZE test instead of the path** (`terraform validate` Success)

```
E   AssertionError: the and_statement's two branches must be a size_constraint_statement and a not_statement, as SIBLINGS; got ['byte_match_statement', 'not_statement']
FAILED test_custom_rule_blocks_oversize_bodies
FAILED test_the_exception_is_exactly_the_verify_path_and_is_negated
FAILED test_the_negation_is_a_sibling_of_the_size_test_not_a_wrapper
3 failed, 12 passed          (pytest exit 1)
```

**M1 — delete the `rule_action_override`**
```
E   AssertionError: no 'rule_action_override' block found — the scoped WAF fix for #1749 is gone
E   AssertionError: exactly one managed rule may be overridden (SizeRestrictions_BODY); found []
E   assert [] == ['SizeRestrictions_BODY']
FAILED test_size_restrictions_body_is_overridden_to_count_on_the_crs_group
FAILED test_no_other_managed_rule_is_overridden_anywhere_in_the_web_acl
2 failed, 13 passed          (pytest exit 1)
```

**M2 — widen the exception to `/api/` `STARTS_WITH`**
```
E   AssertionError: the exception must be the literal path /api/rigor/verify; got: search_string = "/api/" positional_constraint = "STARTS_WITH" field_to_match { } text_transformation { }
E   AssertionError: exception widened to "/api/" — every other route loses the 8 KB ceiling
FAILED test_the_exception_is_exactly_the_verify_path_and_is_negated
FAILED test_no_wildcard_or_prefix_exception_anywhere_in_the_custom_rule
2 failed, 13 passed          (pytest exit 1)
```

**M3 — order the custom rule ahead of the group (priority 5)**
```
E   AssertionError: WAF evaluates by ascending priority; the replacement must run AFTER the group whose rule it replaces
E   assert 5 > 10
FAILED test_custom_rule_is_evaluated_after_the_managed_group_and_before_the_rest
1 failed, 14 passed          (pytest exit 1)
```

**M4 — sneak a second managed-rule override in (`NoUserAgent_HEADER`)**
```
E   AssertionError: the CRS override must name SizeRestrictions_BODY; got: name = "NoUserAgent_HEADER" action_to_use { count {} }
E   AssertionError: exactly one managed rule may be overridden (SizeRestrictions_BODY); found ['NoUserAgent_HEADER', 'SizeRestrictions_BODY']
FAILED test_size_restrictions_body_is_overridden_to_count_on_the_crs_group
FAILED test_no_other_managed_rule_is_overridden_anywhere_in_the_web_acl
2 failed, 13 passed          (pytest exit 1)
```

**M5 — drop the application row cap (`max_length` + validator)**
```
E   Failed: DID NOT RAISE <class 'pydantic_core._pydantic_core.ValidationError'>
E   KeyError: 'maxItems'
E   assert 200 == 422
FAILED test_schema_rejects_one_row_over_the_cap_with_a_message_that_names_the_limit
FAILED test_max_length_is_declared_on_the_field_as_the_backstop
FAILED test_over_cap_payload_is_a_422_over_http
3 failed, 12 passed          (pytest exit 1)
```

M2's `got:` line now prints the byte_match block's own attributes rather than a flattened blob, because the read is scoped to that block. All seven mutations reverted; tree is the committed state.

## Runs (green)

```
pytest backend/tests/test_waf_verify_body.py backend/tests/test_rigor_verify_routes.py
    30 passed          (exit 0)

ruff check   backend/tests/test_waf_verify_body.py backend/archimedes/api/rigor_verify_routes.py   All checks passed!
ruff format --check (same two files)                                                               2 files already formatted
ruff check backend/                          37 errors — byte-identical to origin/main, none in these files

terraform fmt -check -recursive infra/       (clean)
terraform validate                           Success! The configuration is valid.
```

## Owner apply step

The terraform half is **not live until applied**:

```
infra/apply.sh              # plan first
infra/apply.sh --apply
```

Expected plan: **one resource change only** — `aws_wafv2_web_acl.main` updated in place (the `rule_action_override` added to the `aws-core-rules` group, plus the new `oversize-body-except-rigor-verify` rule at priority 11). No ALB, ECS, Aurora or association changes. `terraform init` is needed first if the working copy has not been initialised since the `archive` provider was added (`infra/main.tf`). Anything else in the plan means drift unrelated to this PR — stop and read it.

**Post-apply confirmation the guard cannot give** (nobody has seen the fixed path go green at the edge yet — no apply was run from this branch): re-run the 252-row verify against prod and confirm `200`, then confirm a 9 KB body to any *other* `/api/*` route still `403`s. Worth also reading the ACL's sampled requests for `terminatingRuleId`.

**The "firing rule was inferred, not read from logs" caveat is now retired.** The earlier draft called the override name a guess. Reviewed by elimination instead: `infra/cloudfront.tf:75-114` shows the CLOUDFRONT-scope ACL holds **only** a rate-based rule, and the distribution has no `origin_path` — so a byte-deterministic 403 served by `awselb/2.0` can only come from the regional ALB ACL, where `SizeRestrictions_BODY` is the **only** rule keyed on a byte threshold (the rate-based, KnownBadInputs, IpReputation and SQLi rules have none). The override names the right rule. The sampled-requests check is still worth doing; it is no longer a coin flip.

## Not in this PR

- **CLI error mapping.** #1749 also asks that an `awselb/2.0` 403 surface as a distinct payload-too-large error instead of a generic `exit 2 http_error`. Separate file, separate logical change; the issue stays open for it.
- **The inspection limit** — see above; there is nothing to write for an ALB.
- **An explicit `client_max_body_size` in `nginx/nginx.conf`.** The 1 MB implicit default is the only byte ceiling left on this path; pinning it is the right hardening, but it is a second logical change.
- **`PBORequest.returns_matrix`** (`selection_bias_routes.py:257-260`) carries a *trial matrix* — structurally many times larger than one series, with no cap on rows, columns or `s_partitions` — and is **not** in this WAF exception, so it is almost certainly 403ing at the same 8,192-byte boundary today. That is the endpoint this route's own PBO `not_evaluable` reason points callers at. Needs its own issue.

### Considered and rejected: keying the exception on the CRS label

`label_match_statement` on `awswaf:managed:aws:core-rule-set:SizeRestrictions_Body` would be tidier, but it fails **open**: if the label name is wrong the rule never fires and the 8 KB ceiling is silently gone site-wide. The self-contained `size_constraint_statement` here fails **closed** — if anything about it is wrong, verify keeps 403ing and nothing widens. Owner ruling: keep the self-contained shape.

### DIFF
--- a/backend/archimedes/api/rigor_verify_routes.py
+++ b/backend/archimedes/api/rigor_verify_routes.py
@@ -72,7 +72,7 @@ import math
 from typing import Literal
 
 from fastapi import APIRouter, Depends, Request, Response
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from archimedes.api.account_auth import CurrentUser, require_current_user
 from archimedes.api.limiter import limiter
@@ -121,9 +121,52 @@ class ReturnPoint(BaseModel):
     daily_return: float
 
 
+# #1749: the ceiling on a verify payload belongs to the APPLICATION, at a number
+# we chose, not to whatever byte count the edge happens to enforce.
+#
+# Until this cap existed, `returns` had `min_length=1` and no upper bound, so the
+# only ceiling anywhere in the path was AWS WAF's `SizeRestrictions_BODY` —
+# 8,192 bytes on a REGIONAL web ACL fronting an ALB. Measured (dogfood
+# 2026-09-01, agent-cli): 160 rows = 8,076 B -> 200; 165 rows = 8,326 B -> 403.
+# That is 50.5 bytes per serialised row including the envelope, so the limit is
+# crossed at ~163 rows and one trading year (252 bars, ~12.7 KB extrapolated)
+# 403'd at the edge with an `awselb/2.0` HTML body the CLI could only report as
+# a generic http_error. See infra/waf.tf for the scoped edge fix.
+#
+# 2,600 rows is a decade of daily bars (10 x 252 = 2,520, plus headroom for
+# leap-year/exchange-calendar variation). A full-cap payload measures ~122 KB of
+# JSON with 6-decimal returns and extrapolates to ~131 KB at the 50.5 B/row seen
+# in production — comfortably inside FastAPI/uvicorn's defaults, and small
+# enough that the DSR/OOS math stays sub-second. A decade is the honest outer
+# edge of what a daily-bar rigor verdict can claim anyway; beyond it the caller
+# wants a different endpoint, not a bigger POST.
+#
+# Fail-closed: over the cap is a 422 with a message that names the limit, the
+# count received and the reason — never a truncation, never a silent accept.
+_MAX_RETURN_ROWS = 2600
+
+
 class RigorVerifyRequest(BaseModel):
-    returns: list[ReturnPoint] = Field(min_length=1)
+    # `max_length` is the declarative contract (it lands in the OpenAPI schema
+    # and is the backstop if the validator below is ever removed); the
+    # mode="before" validator runs first and is what produces the explicit
+    # message, because pydantic's own max_length error ("List should have at
+    # most 2600 items after validation") does not tell the caller what to do.
+    returns: list[ReturnPoint] = Field(min_length=1, max_length=_MAX_RETURN_ROWS)
     trials: int = Field(default=1, ge=1, description="Self-attested trial count for the DSR deflation.")
+
+    @field_validator("returns", mode="before")
+    @classmethod
+    def _cap_returns(cls, value: object) -> object:
+        """Reject over-long series with a message that names the limit (#1749)."""
+        if isinstance(value, list) and len(value) > _MAX_RETURN_ROWS:
+            raise ValueError(
+                f"returns has {len(value)} rows; the maximum is {_MAX_RETURN_ROWS} "
+                f"(~10 years of daily bars). Split the series or aggregate to a "
+                f"coarser frequency — a rigor verdict over a longer daily window "
+                f"is not something this endpoint can honestly compute."
+            )
+        return value
 
 
 class RigorCheck(BaseModel):
--- a/backend/archimedes/api/rigor_verify_routes.py
+++ b/backend/archimedes/api/rigor_verify_routes.py
@@ -121,8 +121,9 @@ class ReturnPoint(BaseModel):
     daily_return: float
 
 
-# #1749: the ceiling on a verify payload belongs to the APPLICATION, at a number
-# we chose, not to whatever byte count the edge happens to enforce.
+# #1749: the size ceiling on a verify payload belongs to the APPLICATION, at a
+# row count we chose, instead of being set implicitly by whatever byte count the
+# edge happens to enforce.
 #
 # Until this cap existed, `returns` had `min_length=1` and no upper bound, so the
 # only ceiling anywhere in the path was AWS WAF's `SizeRestrictions_BODY` —
@@ -136,10 +137,13 @@ class ReturnPoint(BaseModel):
 # 2,600 rows is a decade of daily bars (10 x 252 = 2,520, plus headroom for
 # leap-year/exchange-calendar variation). A full-cap payload measures ~122 KB of
 # JSON with 6-decimal returns and extrapolates to ~131 KB at the 50.5 B/row seen
-# in production — comfortably inside FastAPI/uvicorn's defaults, and small
-# enough that the DSR/OOS math stays sub-second. A decade is the honest outer
-# edge of what a daily-bar rigor verdict can claim anyway; beyond it the caller
-# wants a different endpoint, not a bigger POST.
+# in production — well under nginx's implicit 1 MB `client_max_body_size`
+# default (nginx/nginx.conf sets none; nginx is the ALB target, infra/ecs.tf),
+# which with the WAF exception in place is now the only BYTE ceiling on this
+# path. Note this is a ROW cap: FastAPI buffers and json-parses the entire body
+# before dependencies or field validation run, so an over-cap — or
+# unauthenticated — request is still parsed in full first. It is also small
+# enough that the DSR/OOS math stays sub-second.
 #
 # Fail-closed: over the cap is a 422 with a message that names the limit, the
 # count received and the reason — never a truncation, never a silent accept.
@@ -162,9 +166,8 @@ class RigorVerifyRequest(BaseModel):
         if isinstance(value, list) and len(value) > _MAX_RETURN_ROWS:
             raise ValueError(
                 f"returns has {len(value)} rows; the maximum is {_MAX_RETURN_ROWS} "
-                f"(~10 years of daily bars). Split the series or aggregate to a "
-                f"coarser frequency — a rigor verdict over a longer daily window "
-                f"is not something this endpoint can honestly compute."
+                f"(~10 years of daily bars). This is a payload cap, not a statistical "
+                f"one: split the series or aggregate to a coarser frequency."
             )
         return value
 
--- a/backend/tests/test_waf_verify_body.py
+++ b/backend/tests/test_waf_verify_body.py
@@ -28,13 +28,20 @@ The fix is deliberately SCOPED, and these tests exist to keep it scoped:
     "EXACTLY"` matters: widening it to a `/api/*` prefix would hand every API
     route an unlimited body, which is the DoS/parser-abuse control the managed
     rule was there to provide;
+  * the rule's statement NESTING is asserted as a parsed tree, not as substrings.
+    `not_statement { and_statement { size, byte_match } }` — the negation wrapped
+    AROUND the pair rather than ANDed beside the size test — is valid HCL, passes
+    `terraform validate`, and contains every string a substring guard looks for,
+    while meaning "block everything EXCEPT an oversize POST to /api/rigor/verify":
+    a site-wide outage the moment it is applied;
   * the application, not the edge, owns the real ceiling — `RigorVerifyRequest`
     caps `returns` at 2,600 rows (a decade of daily bars) and fails closed with
     a 422 that names the limit.
 
 `terraform validate` cannot catch any of this. A deleted override, an override
-on the wrong rule name, a prefix-widened exception and a custom rule ordered
-ahead of the managed group are all syntactically valid HCL.
+on the wrong rule name, a prefix-widened exception, an inverted NOT/AND nesting
+and a custom rule ordered ahead of the managed group are all syntactically valid
+HCL.
 
 Hermetic: reads one file from the repo and exercises the pydantic model plus the
 router over an in-process ASGI transport. No AWS, no terraform binary, no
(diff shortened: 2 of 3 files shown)


dbrowneup:
Review verdict was FIX-FIRST on the guards and two claims, not on the change itself. Pushed `4b6a3f38` — `infra/waf.tf` is untouched by it; the rule shape stays as-is per the owner ruling (self-contained size constraint, fails closed; **not** a `label_match`, which fails open if the label name is wrong).

**Defect 1 — the guard suite could not see the NOT/AND inversion.** `not_statement { statement { and_statement { size, byte_match } } }` is valid HCL, `terraform validate` Success, `terraform fmt -check` clean — and means "block everything EXCEPT an oversize POST to `/api/rigor/verify`", a site-wide outage on apply. It passed all 14 guards, because the only nesting assertion was `"and_statement" in _collapse(rule)` and that string survives inversion.

The guard is now **structural**: `backend/tests/test_waf_verify_body.py` parses the custom rule's statement tree with a brace/string-aware child-block scanner and asserts the nesting node by node —

- `rule.statement` **is** the `and_statement` (nothing may wrap it),
- its two branches are a `size_constraint_statement` and a `not_statement` **as siblings**,
- no `size_constraint_statement` appears anywhere beneath the `not_statement`,
- the `not_statement`'s inner statement is the `byte_match_statement` (`uri_path`, `EXACTLY`, `/api/rigor/verify`, `text_transformation NONE`),
- the size branch reads `GT` / `8192` / `field_to_match → body → oversize_handling = "MATCH"` / `NONE`,
- and the whole rule holds exactly one and/not/size/byte_match and zero `or_statement`.

Attribute reads now bind to the block that owns them (`_own_attrs` blanks child blocks), so a value cannot be satisfied by a substring somewhere else in the rule.

Shown red on the inversion, and on a second one I wrote to check the guard is not just pattern-matching that one shape:

```
V1  not_statement WRAPS the and_statement       terraform validate: Success
E   AssertionError: rule.statement must contain exactly one block, `and_statement {...}`; got ['not_statement']
3 failed, 12 passed          pytest exit 1

V1b negate the SIZE test instead of the path    terraform validate: Success
E   AssertionError: the and_statement's two branches must be a size_constraint_statement and a not_statement, as SIBLINGS; got ['byte_match_statement', 'not_statement']
3 failed, 12 passed          pytest exit 1
```

M1–M5 re-run against the new suite, all still red, exit 1 each: **M1** 2 failed/13 passed · **M2** 2/13 · **M3** 1/14 · **M4** 2/13 · **M5** 3/12. Green on the committed tree: **30 passed, exit 0** (15 + 15). `ruff check`/`format --check` on both touched files exit 0; `ruff check backend/` still 37, byte-identical to `origin/main`; `terraform fmt -check -recursive` and `terraform validate` exit 0.

**Defect 2 — "comfortably inside FastAPI/uvicorn's defaults" was false.** Neither has a request-body size default; there was nothing to be inside. Two corrections in `rigor_verify_routes.py` and in the PR body:

- With the WAF exception in place the only **byte** ceiling left on this path is nginx's **implicit 1 MB `client_max_body_size`** — `nginx/nginx.conf` sets none, and nginx is the ALB target (`infra/ecs.tf:949-952`, `container_name = "nginx"`, port 8080).
- 2,600 is a **row** cap applied **post-parse**. FastAPI's request handler does `await request.body()` / `await request.json()` before `solve_dependencies`, so the whole body is buffered and `json.loads`'d before `require_current_user` and before this validator — an over-cap, or unauthenticated, request is parsed in full first.

The residual (~1 MB against nginx's own `limit_req api_read 60r/m burst=20`, under the 1000/5min WAF rate rule) is bounded and accepted; an explicit `client_max_body_size` is the right hardening and is listed in "Not in this PR" as its own change.

**Defect 3 — the 422 dressed an operational cap as a statistical one.** "a rigor verdict over a longer daily window is not something this endpoint can honestly compute" is not supported by the code: `compute_dsr_hac_and_iid` / `compute_oos_sharpe` are fine over 3,000 bars and nothing derives a 10-year validity horizon. The reason is payload size. Message is now `"...This is a payload cap, not a statistical one: split the series or aggregate to a coarser frequency."`, and the "honest outer edge" line is gone from the comment.

**Defect 4 — stale number in a failure message.** `~135 KB` → `~122 KB` (measured 121,935 B), and the band binding that claim tightened from `100_000 < payload < 200_000` to `118_000 < payload < 126_000` so the comment cannot drift to 190 KB and still pass.

**PR body** also updated with the two concerns the review retired: (#2) the firing rule is no longer inferred — `infra/cloudfront.tf:75-114` shows the CLOUDFRONT-scope ACL holds only a rate-based rule and the distribution has no `origin_path`, so a byte-deterministic 403 from `awselb/2.0` can only be the regional ALB ACL, where `SizeRestrictions_BODY` is the only rule keyed on a byte threshold; (#4) `association_config` genuinely has no ALB form in `hashicorp/aws` 5.100.0, reproduced independently. Also noted as a follow-up: `PBORequest.returns_matrix` (`selection_bias_routes.py:257-260`) is uncapped, not in this exception, and is almost certainly 403ing at the same 8,192-byte boundary today — needs its own issue.

Terraform half is still not live until applied. **Do not merge — owner vets first.**

PR #1788, opened 2026-09-02
[ux] rejected strategy cards state their own rigor-gate reason

dbrowneup:
## The owner's observation

Screenshot of Library → Generated → **"Rejected (1) — did not pass the rigor gate"**:

- the rejected card shows `—` for Sharpe, CAGR and Max DD, and one measured number, `Gen tokens 65,826`;
- the only text about *why* is a paragraph shared by every rejected candidate:

  > These candidates were generated but failed at least one rigor check (DSR, PBO, chronological OOS, or look-ahead audit). **Most rejections at this stage are "return series too short"** — the agent generated a strategy but there isn't enough backtest history yet to compute DSR / PBO with statistical confidence. **A longer backtest window typically unlocks them.**

Nothing had measured "most", and the sentence is simply false for a candidate rejected on the numbers (a DSR below the bar, a PBO above the ceiling, a negative OOS Sharpe). The one fact a reader wants — which check said no — was the one fact absent.

## Root cause

The reason was on the row the whole time, unexposed.

`StrategyRecord.rigor_verdict` (JSON) carries `dsr_p_value` / `pbo` / `oos_sharpe` / `in_sample_sharpe` / `lookahead_audit_passed` (+ `look_ahead_status` on the fusion path), and a verbatim `reason` string for the branches where the gate never graded the candidate at all. On today's live write paths that blob comes from `debate_engine._rigor_verdict_dict` (the graded fusion/debate verdict — the only writer of the four-state `look_ahead_status`), from `debate_engine._abstain_result` and `generation_pipeline`'s fixture branch (both non-graded, both carrying a `reason`), and is then patched by `generation_pipeline._patch_pbo` / `_patch_dsr_with_pool_correlation`. `generation_pipeline._rigor_verdict_for` — the writer of the `"return series too short for rigor evaluation"` branch — has **no production callers** (repo-wide it appears only at its own `def` and in tests), so that shape is reachable only on **stored** rows.

`GET /api/strategies/generated` returned the raw blob via `to_dict()`, but nothing turned it into per-check statements, and `coerceGenerated` mapped only the four numbers onto `deflated_sharpe_ratio` / `pbo_score` / `out_of_sample_sharpe` / `dsr_p_value` — all of which the *rejected card layout* does not render. So the card had numbers it never showed and a paragraph it could not justify.

Deriving the reasons in JS was the wrong fix: it would put the gate's thresholds in a second place, where they can drift from `rigor_profiles`.

## The change

**Backend — `backend/archimedes/services/rigor_reasons.py` (new).** A pure function over the verdict dict:

```python
rigor_reasons_for_verdict(verdict) -> {
  "bar": "Archimedes Verified", "bar_level": 1, "passing": bool,
  "recorded_reason": str | None,          # the strategy's own reason, verbatim
  "reason_code": "short_return_series" | "fixture_mode" | None,
  "min_returns_for_gate": 10,
  "checks": [{key, label, status, detail, value, threshold}, ...],
  "unattributed": bool,
}
```

- Thresholds come from `rigor_profiles.get_profile(STRICTEST_LEVEL)` (`dsr_p_min` 0.90, `pbo_max` 0.50, `oos_is_ratio_min` 0.50) plus `OOS_ABS_FLOOR` — **read, never re-declared**, so the card quotes the bar the gate enforces.
- Three statuses, not two. `fail` is reserved for a check that ran and did not clear its bar; a check with nothing on record is `not_computed` — fail-closed (it still blocks) but never *printed* as a failure. Same FAIL-vs-NOT_RUN distinction `RigorGateResult.gate_details` already keeps.
- **A default is not a measurement.** A verdict carrying its own `reason` was never graded, so no numeric leg on it is reported as a graded result. This is load-bearing, not defensive: `_patch_pbo` (`generation_pipeline.py:660`) stamps `pbo = 0.0  # PBO undefined for N<2` onto exactly those candidates (every one is `has_real_rigor=False`, so every one is in its `agent_cands`), and 0.0 sits below the 0.50 ceiling. Without this rule the card renders **"Passed: PBO (overfitting probability) — 0.00 < 0.50 required"** on a candidate the gate never scored.
- **A look-ahead claim keys off `look_ahead_status` only.** Neither value of the fail-closed `lookahead_audit_passed` boolean is an audit result on its own: `_lookahead_for_candidate` is documented as *"vacuously True when none expose auditable source"*, and `dsl_lookahead_audit.verdict_from_persisted_row` explicitly retired this rendering (*"the alternative is a live surface showing 'look_ahead: PASS' on the strength of a sentence an LLM wrote about itself"*). A bare `True` with no four-state status is therefore `not_computed`, and a `False` on a row carrying a `reason` is `not_computed` rather than an invented leak finding. Conclusive rows are unaffected — the fusion writer derives the boolean *from* `look_ahead_status == "pass"` (`fusion_evaluator.py:143`).
- `unattributed` covers the one honest gap: the agent path's own `passing` uses a stricter DSR bar (0.95, hardcoded in `_rigor_verdict_for`) than the badge's 0.90, so a p-value in [0.90, 0.95) lands a rejected row with every badge-bar check clear. The card then says it cannot attribute the rejection rather than naming a culprit. With `_rigor_verdict_for` currently uncalled in production, that shape reaches the surface via stored rows rather than new writes — the state is still rendered, because a row the surface cannot attribute must say so either way.

**Backend — `strategies_routes.list_generated_strategies`.** One added line inside the existing per-page loop:

```python
d["rigor_reasons"] = rigor_reasons_for_verdict(d.get("rigor_verdict"))
```

Additive, batched by construction: the verdict dict is already decoded by `to_dict()` above, so this is a pure transform and adds **zero queries** (guarded — see verification).

**Frontend — `ui/src/rejectionReasons.js` (new)** holds the reading/selection logic as a plain importable module (the `.jsx` is not importable by `node --test`, so this is what makes the render behaviour testable rather than regex-pinned). Fail-closed: a row without the field yields empty lists and a null summary.

**Frontend — `ui/src/components/Strategies.jsx`.** New `RejectionReasons` component, rendered un-collapsed on the mobile card (right under the four `—` tiles, the surface in the screenshot) and in the shared detail panel for the desktop table row. Scoped to `passes_rigor_gate === false`, so a row still awaiting a verdict is never handed one. Example output for a graded rejection (a fusion row, which carries `look_ahead_status: "pass"`):

> **Failed:** DSR confidence — 0.08 < 0.90 required · PBO (overfitting probability) — 0.62 ≥ 0.50 — at or above the overfitting ceiling
> **Passed:** Out-of-sample Sharpe — 0.41 > 0.00 required · Look-ahead audit — no forward-looking data access found
> Thresholds are the Archimedes Verified bar.

and for one the gate never scored:

> **Reason on record:** society could not agree on a candidate
> **Not computed:** DSR confidence — not scored — see the reason on record · PBO (overfitting probability) — not scored — see the reason on record · … — a check with nothing on record does not count as passed.

The population-wide paragraph is deleted. In its place: one sentence that is unconditionally true ("Each row names the checks it failed, read from that strategy's own recorded verdict"), plus `rejectedSectionSummary(rejected)` — computed from the rows on screen, counting them, quoting `min_returns_for_gate` from the payload, and returning `null` when there is nothing true to say. The same correction is applied to the empty-state card above it, which carried the identical unmeasured parenthetical.

Untouched, as agreed with the concurrent branches: `statusTag` / `statusLabel`, `coerceGenerated`'s verdict lines (the new key is appended after `generation_cost`), `StrategyPassport.jsx`, `passport_loader`, `generation_pipeline`, and the generation-stream/SSE components. The `strategies_routes.py` hunk is a comment block plus a local import (line 839) and one assignment inside `list_generated_strategies`'s existing page loop (line 844), far from the `_passport_to_strategy_response` verdict fields `dbrowneup/verdict-of-record-a` edits.

## Adversarial demo — every guard shown red on the input it rejects

**1. Drop the payload field** (`d["rigor_reasons"] = …` removed from the route):

```
FAILED tests/test_rejected_reason_payload.py::test_generated_route_carries_each_rows_own_reasons - KeyError: 'rigor_reasons'
FAILED tests/test_rejected_reason_payload.py::test_generated_route_reasons_add_no_queries - KeyError: 'rigor_reasons'
FAILED tests/test_rejected_reason_payload.py::test_a_row_with_no_verdict_carries_a_null_reasons_field - AssertionError: assert 'rigor_reasons' in {...}
================== 3 failed, 18 passed, 4 warnings in 15.04s ===================   # exit 1
```

**2. Drop the field on the way to the card** (`rigor_reasons: row.rigor_reasons ?? null` removed from `coerceGenerated`):

```
✖ coerceGenerated carries the per-row reasons through to the card (1.777375ms)
ℹ pass 11
ℹ fail 1
```

**3. Render the generic sentence for strategies whose reason is different** (`rejectedSectionSummary` stops filtering on `reason_code`, so every rejected row is described as a short-series rejection):

```
✖ the short-series sentence appears only for rows that actually carry that reason (2.791459ms)
✖ the short-series sentence counts the rows it is true of (1.950333ms)
ℹ tests 12  ℹ pass 10  ℹ fail 2                                        # exit 1
```

**4. Bring the deleted paragraph back** (the two sentences restored verbatim in the JSX):

```
✖ the population-wide rejection prose is gone from Strategies.jsx (4.297083ms)
ℹ pass 10
ℹ fail 1
```

**5. Report a check that never ran as a failure** (the `has_recorded_reason` branch removed; a missing DSR p-value returned as `FAIL`):

```
FAILED tests/test_rejected_reason_payload.py::test_ungraded_checks_are_not_computed_never_failed - AssertionError: dsr
FAILED tests/test_rejected_reason_payload.py::test_fixture_mode_does_not_claim_a_look_ahead_failure - AssertionError: assert 'fail' == 'not_computed'
FAILED tests/test_rejected_reason_payload.py::test_nan_is_not_a_measurement - AssertionError: assert 'fail' == 'not_computed'
================== 3 failed, 13 passed, 4 warnings in 10.52s ===================
```

**6. Show the block on a row with no verdict yet** (`showsRejectionReasons` stops checking `passes_rigor_gate`):

```
✖ a row still awaiting a verdict is never handed one (0.883875ms)
ℹ pass 11
ℹ fail 1
```

**7. Report a default number on an ungraded verdict as a passed check** (the `if recorded_reason:` downgrade removed, so `_patch_pbo`'s `pbo = 0.0` sentinel renders as "PBO — 0.00 < 0.50 required" under **Passed**):

```
FAILED tests/test_rejected_reason_payload.py::test_a_default_number_on_an_ungraded_verdict_is_not_a_passed_check - AssertionError: assert 'pass' == 'not_computed'
================== 1 failed, 20 passed, 4 warnings in 14.05s ===================   # exit 1
```

The fixture for that guard is not hand-written: it is the blob `debate_engine._abstain_result` actually produces after the real `_patch_pbo` runs on it, so it cannot drift from the pipeline.

**8. Let a bare `lookahead_audit_passed: True` short-circuit the four-state handling** (`if status == PASS or bool(verdict.get("lookahead_audit_passed")):` restored, the `is True` branch removed):

```
FAILED tests/test_rejected_reason_payload.py::test_a_bare_lookahead_true_is_not_a_passed_audit - AssertionError: assert 'pass' == 'not_computed'
FAILED tests/test_rejected_reason_payload.py::test_an_inconclusive_status_beats_a_stored_lookahead_true - AssertionError: assert 'pass' == 'not_computed'
================== 2 failed, 19 passed, 4 warnings in 12.04s ===================   # exit 1
```

The second one is the sharper case: a verdict that *explicitly says the audit reached no verdict* (`look_ahead_status: "pending"`) still rendered as a pass because the boolean happened to read `True`.

**9. Misattribute the not-computed ratio** (the single `"no positive in-sample Sharpe to compare against"` string restored, so a row with `in_sample_sharpe = 1.2` and no OOS Sharpe is told it has no positive in-sample Sharpe):

```
FAILED tests/test_rejected_reason_payload.py::test_the_ratio_check_names_the_side_that_is_actually_missing - AssertionError: assert 'no positive ...mpare against' == 'no out-of-sa...pe...'
=================== 1 failed, 20 passed, 4 warnings in 7.66s ===================   # exit 1
```

## Verification

New guards, green:

```
$ pytest -p no:cacheprovider backend/tests/test_rejected_reason_payload.py -q
21 passed, 4 warnings in 13.98s                                        # exit 0

$ cd ui && node --test test/rejected-reason.test.js
ℹ tests 12  ℹ pass 12  ℹ fail 0                                        # exit 0
```

Neighbouring backend suites on the same route/model (guards included):

```
$ pytest -p no:cacheprovider backend/tests/test_rejected_reason_payload.py \
    backend/tests/test_strategies_routes.py backend/tests/test_strategy_store.py \
    backend/tests/test_generated_citation_truth.py backend/tests/test_unify_curated_generated.py -q
130 passed, 12 warnings in 30.54s                                      # exit 0
```

The write paths the new honesty rules are pinned against (`_patch_pbo`, `_abstain_result`, the passport):

```
$ pytest -p no:cacheprovider backend/tests/test_generation_pipeline.py \
    backend/tests/services/test_generation_pipeline.py backend/tests/test_brief_on_passport.py \
    backend/tests/test_publishable_strategy_ids.py backend/tests/test_strategy_ownership.py -q
108 passed, 33 warnings in 32.64s                                      # exit 0
```

Full UI suite:

```
$ cd ui && node --test
ℹ tests 925  ℹ pass 925  ℹ fail 0                                      # exit 0
```

Lint:

```
$ ruff check backend/archimedes/services/rigor_reasons.py backend/tests/test_rejected_reason_payload.py
All checks passed!                                                     # exit 0
$ ruff format --check <same two files>
2 files already formatted                                              # exit 0
$ npx eslint src/rejectionReasons.js src/components/Strategies.jsx test/rejected-reason.test.js
(no output)                                                            # exit 0
```

`ruff check` on `backend/archimedes/api/strategies_routes.py` reports one `SIM105` at line 1263 (`try/except/pass` around `session.rollback()`). **Pre-existing** — reproduced identically with my changes stashed, and present in `origin/main`'s copy of the file; not in my hunk, not touched.

Backend unit suite, `-m "not integration"`, run on the branch's first commit. The single full-suite invocation cannot complete in this environment — it is killed (`EXIT=143`, SIGTERM) partway through, and one file segfaults the interpreter on shutdown (`EXIT=139`). Both reproduce on unmodified `origin/main`. It was therefore run in chunks covering every collected file; log at `/tmp/rejected-reason-suite.log`:

| chunk | exit | result |
| --- | --- | --- |
| `backend/tests/api` | 0 | 53 passed |
| `backend/tests/chain` | 0 | 336 passed, 2 skipped |
| `backend/tests/marketplace` | 0 | 202 passed |
| `backend/tests/models` | 0 | 16 passed |
| `backend/tests/scripts` | 0 | 62 passed, 1 deselected |
| `backend/tests/services` | 1 | 2 failed, 1374 passed — **pre-existing** |
| `backend/tests/test_*.py` chunks 1–3 | 0 | all passed |
| `backend/tests/test_*.py` chunk 4a | 1 | 5 failed, 102 passed — **pre-existing** |
| `backend/tests/test_*.py` chunk 4b | 0/139 | all tests passed; SIGSEGV at interpreter shutdown — **pre-existing, flaky** |
| `backend/tests/test_*.py` chunk 5 | 0 | 751 passed, 1 skipped |

Every failure baselined against a clean tree (`git stash -u`, same command):

- `services/test_generation_pipeline.py::test_pipeline_{writes,does_not_ledger}_…` — `sqlite3.OperationalError: no such table`, a cross-file DB-isolation problem. Clean main: **identical** `2 failed, 1374 passed`.
- `test_metrics_routes.py` × 5 — same `no such table` shape. Clean main: **identical** `5 failed, 102 passed`.
- `test_paper_rag.py` — segfault (139) at interpreter shutdown after all 37 tests pass. Ran 3× on each tree: this branch 139/0/0, clean main 139/139/0. Flaky and **pre-existing**.

That chunked run predates the second commit. The second commit touches only `rigor_reasons.py` (a new file nothing else imports except the route line) and its own test file, so the suites re-run above are the ones that can reach it; the chunked table has not been repeated.

## What the second commit fixed

The first commit's `rigor_reasons.py` stated the right honesty rules in its docstring and then broke two of them on inputs today's write paths actually produce. Review caught both:

1. **`pbo = 0.0` rendered as a passed overfitting check** on every ungraded candidate — mutation 7 above. Fixed by refusing to grade any numeric leg of a verdict that carries its own `reason`.
2. **A bare `lookahead_audit_passed: True` rendered as a passed audit**, short-circuiting the module's own four-state handling — including on a row whose `look_ahead_status` explicitly said `pending`. Mutation 8 above.

Plus two smaller corrections: the not-computed ratio detail now names the side that is actually missing (mutation 9), and the module/test docstrings no longer attribute live writes to `_rigor_verdict_for`, which has no production callers.

## Concerns for the vet

1. **The 0.90-vs-0.95 divergence is real and is only *reported*, not fixed here.** `generation_pipeline._rigor_verdict_for` hardcodes a 0.95 DSR bar; the badge profile is 0.90. This PR grades the report against the badge bar and adds `unattributed` for the gap, rather than silently quoting a threshold that did not produce the stored `passing`. Reconciling the two bars is a separate change (and Önder's call).
2. **`_rigor_verdict_for` has no production callers at all.** Two knock-ons: `SHORT_SERIES_REASON` / `MIN_RETURNS_FOR_GATE = 10` match **legacy stored rows only**, so `rejectedSectionSummary`'s short-series sentence is a legacy-only path (kept, with its pinning test, because those rows are still in the library); and the `unattributed` state is likewise reached via stored rows. Whether that function is dead code worth deleting is a separate question this PR does not answer.
3. **The report grades the STORED verdict, not a live gate run.** That is deliberate — the rejected card's `passes_rigor_gate` comes from `coerceGenerated(row.rigor_verdict.passing)`, so the reasons and the badge now come from the same blob and cannot disagree. It does mean a stale stored verdict yields stale reasons; the live gate stays the passport's job.
4. **Desktop rows show the block only when expanded.** The mobile card (the screenshot's surface) shows it above the fold; the desktop table has no free row to put it in without a new `<tr>`. Both layouts render "Gen tokens", so the screenshot could be either — worth confirming which surface you were on. A `colSpan` summary row is the fix if it was the table.
5. `_pbo_below_power_floor` (#819's CSCV N floor) is **not** modelled — the stored verdict does not carry `pbo_library_size`, so a PBO that the live gate would treat as non-load-bearing is still reported as a fail here when it is ≥ 0.50. Fixing it needs a new field at write time. (Distinct from the N<2 sentinel fixed above, which is caught because those verdicts carry a `reason`.)
6. **Shared-file contact:** `strategies_routes.py` and `Strategies.jsx` are both touched by concurrent branches. `dbrowneup/reasoning-traces-surfaced` touches neither file. `dbrowneup/verdict-of-record-a` touches `list_generated_strategies` and, in `Strategies.jsx`, `statusTag` / `statusLabel` and `coerceGenerated`'s verdict lines — all of which this branch deliberately leaves alone; a hunk-range diff of both branches against this one came back disjoint. The one spot to eyeball on merge is `strategies_routes.py`'s import block, where the two branches' inserts land a couple of lines apart.

### DIFF
--- a/backend/archimedes/api/strategies_routes.py
+++ b/backend/archimedes/api/strategies_routes.py
@@ -831,9 +831,17 @@ async def list_generated_strategies(
                 [p.get("arxiv_id") for d in page for p in (d.get("source_papers") or []) if isinstance(p, dict)],
                 session,
             )
+            # Each row's OWN rejection reasons, derived from the rigor_verdict
+            # blob `to_dict()` already decoded above — a pure function, zero
+            # extra queries, so the page cost is unchanged (the Library's
+            # "Rejected — did not pass the rigor gate" cards used to share one
+            # paragraph of guessed prose because this field did not exist).
+            from archimedes.services.rigor_reasons import rigor_reasons_for_verdict
+
             rows = []
             for d in page:
                 d["source_papers"] = _resolve_source_papers(d.get("source_papers"), corpus_meta)
+                d["rigor_reasons"] = rigor_reasons_for_verdict(d.get("rigor_verdict"))
                 rows.append(_redact_owner_wallet(d, caller))
     except Exception as exc:
         # Full exception detail is logged server-side only — never echoed to
--- a/backend/archimedes/services/rigor_reasons.py
+++ b/backend/archimedes/services/rigor_reasons.py
@@ -9,9 +9,15 @@ for a DIFFERENT reason the surface stated something false about that strategy.
 
 This module turns a strategy's OWN stored ``rigor_verdict`` into a per-check
 report the card can render. That blob is the JSON in
-``StrategyRecord.rigor_verdict``, written by
-``generation_pipeline._rigor_verdict_for`` on the buy-and-hold/agent path and by
-``debate_engine._rigor_verdict_dict`` on the fusion/debate path.
+``StrategyRecord.rigor_verdict``. On today's live write paths it comes from
+``debate_engine._rigor_verdict_dict`` (the graded fusion/debate verdict, the only
+writer of the four-state ``look_ahead_status``), from
+``debate_engine._abstain_result`` and ``generation_pipeline``'s fixture branch
+(both non-graded, both carrying a verbatim ``reason``), and is then patched by
+``generation_pipeline._patch_pbo`` / ``_patch_dsr_with_pool_correlation``.
+``generation_pipeline._rigor_verdict_for`` — the writer of the too-short-series
+branch — currently has NO production callers (repo-wide, only its own ``def``
+and tests), so the shapes it produces are reachable only on STORED rows.
 
 It is a PURE function over a dict the caller already holds — no DB read, no gate
 re-run — so a list route calls it once per row and adds ZERO queries to the page
@@ -34,13 +40,20 @@ Honesty rules this module keeps
 * ``recorded_reason`` is the verdict's own ``reason`` string, verbatim. The
   pipeline writes it exactly for the branches where the gate did NOT grade the
   candidate at all — return series too short, fixture mode, debate abstain. When
-  it is present it IS this strategy's reason and no derived sentence replaces it.
+  it is present it IS this strategy's reason and no derived sentence replaces it,
+  AND no numeric leg on that blob is reported as a graded result: the numbers
+  sitting on an ungraded verdict are defaults (``_patch_pbo`` stamps ``pbo=0.0``
+  as "undefined for N<2" on exactly these candidates), and printing a default as
+  a passed check is the same overclaim this module exists to remove.
 * ``unattributed`` is True when a row did not pass yet no recorded check fails
   the bar above. The surface then says exactly that instead of naming a culprit
-  it cannot support. This is reachable, not theoretical: the agent path's own
-  ``passing`` uses a STRICTER DSR bar (0.95, hardcoded in ``_rigor_verdict_for``)
-  than the badge profile's ``dsr_p_min`` (0.90), so a p-value in [0.90, 0.95)
-  lands a rejected row with every badge-bar check clear.
+  it cannot support. The known source of that gap is a DSR-bar divergence: the
+  agent path's ``passing`` uses a STRICTER 0.95 (hardcoded in
+  ``_rigor_verdict_for``) than the badge profile's ``dsr_p_min`` (0.90), so a
+  p-value in [0.90, 0.95) lands a rejected row with every badge-bar check clear.
+  With ``_rigor_verdict_for`` currently uncalled in production that shape reaches
+  the surface via stored rows rather than new writes — the state is still
+  rendered, because a row the surface cannot attribute must say so either way.
 """
 
 from __future__ import annotations
@@ -76,6 +89,11 @@ MIN_RETURNS_FOR_GATE = 10
 # branches. Matching them yields a stable ``reason_code`` the frontend can
 # classify on without string-matching prose in JS. An unrecognised reason gets
 # ``reason_code = None`` and is still surfaced verbatim — never dropped.
+#
+# ``SHORT_SERIES_REASON``'s only writer is ``_rigor_verdict_for``, which has no
+# production callers today, so it matches STORED rows only — the short-series
+# summary sentence on the frontend is a legacy path, not one new rows enter.
+# ``FIXTURE_REASON`` is written inline by the fixture branch and is live.
 SHORT_SERIES_REASON = "return series too short for rigor evaluation"
 FIXTURE_REASON = "fixture mode — no LLM call, rigor gate not run"
 
@@ -173,18 +191,22 @@ def _oos_cliff_check(verdict: dict[str, Any], profile) -> dict[str, Any]:
     """The in-/out-of-sample cliff leg.
 
     The gate only applies this ratio when there is a POSITIVE in-sample Sharpe to
-    divide by (``_rigor_verdict_for`` / ``fusion_evaluator``'s
-    ``in_sample_sharpe > 0`` guard). With no such reference the check did not run
-    — reported as not-computed, not as a pass it never earned.
+    divide by (the ``in_sample_sharpe > 0`` guard in ``_rigor_verdict_for`` and in
+    the fusion evaluator's cliff leg). With no such reference — or with no
+    out-of-sample Sharpe to divide — the check did not run, and is reported as
+    not-computed naming the side that is missing, never as a pass it never earned.
     """
     label = "Out-of-sample / in-sample ratio"
     bar = profile.oos_is_ratio_min
     oos = _finite(verdict.get("oos_sharpe"))
     in_sample = _finite(verdict.get("in_sample_sharpe"))
     if oos is None or in_sample is None or in_sample <= 0:
-        return _check(
-            "oos_is_ratio", label, NOT_COMPUTED, "no positive in-sample Sharpe to compare against", threshold=bar
+        # Name the side that is actually missing. "No positive in-sample Sharpe"
+        # is false on a row that HAS one (say 1.20) and is missing the OOS leg.
+        detail = (
+            "no out-of-sample Sharpe to compare" if oos is None else "no positive in-sample Sharpe to compare against"
         )
+        return _check("oos_is_ratio", label, NOT_COMPUTED, detail, threshold=bar)
     ratio = oos / in_sample
     if ratio >= bar:
         return _check("oos_is_ratio", label, PASS, f"{ratio:.2f} ≥ {bar:.2f} required", value=ratio, threshold=bar)
@@ -201,23 +223,45 @@ def _oos_cliff_check(verdict: dict[str, Any], profile) -> dict[str, Any]:
 def _look_ahead_check(verdict: dict[str, Any], *, has_recorded_reason: bool) -> dict[str, Any]:
     """The look-ahead leg, four-state where the writer supplied four states.
 
-    ``look_ahead_status`` (fusion/debate path) is the honest four-state verdict;
-    ``lookahead_audit_passed`` is the fail-closed boolean every path writes. A
-    bare ``False`` is a real audit finding ONLY when the gate actually graded
-    this candidate — on a row whose verdict carries its own ``reason`` the gate
-    never ran, and reporting "the audit found a leak" there would invent a
-    finding out of a default.
+    ``look_ahead_status`` (fusion/debate path) is the honest four-state verdict
+    and is the ONLY field a rendered claim may key off. ``lookahead_audit_passed``
+    is the fail-closed boolean every path writes, and neither of its values is an
+    audit result on its own:
+
+    * a bare ``False`` is a real audit finding ONLY when the gate actually graded
+      this candidate — on a row whose verdict carries its own ``reason`` the gate
+      never ran, and reporting "the audit found a leak" there would invent a
+      finding out of a default;
+    * a bare ``True`` is never a finding at all — it is what the agent path
+      writes when nothing auditable was found to audit.
     """
     label = "Look-ahead audit"
     status = verdict.get("look_ahead_status")
     reason = verdict.get("look_ahead_reason")
-    if status == PASS or bool(verdict.get("lookahead_audit_passed")):
+    if status == PASS:
         return _check("look_ahead", label, PASS, "no forward-looking data access found")
     if status in _LA_INCONCLUSIVE:
         detail = reason.strip() if isinstance(reason, str) and reason.strip() else "the audit reached no verdict"
         return _check("look_ahead", label, NOT_COMPUTED, f"{detail} — blocks admission (fail-closed)")
     if has_recorded_reason:
         return _check("look_ahead", label, NOT_COMPUTED, "not run — see the reason on record")
+    if verdict.get("lookahead_audit_passed") is True:
+        # A bare ``True`` with no four-state status is a DEFAULT, not an audit
+        # result. ``_lookahead_for_candidate`` is documented as "vacuously True
+        # when none expose auditable source" (generation_pipeline.py:622), and
+        # the repo has explicitly retired this rendering:
+        # ``dsl_lookahead_audit.verdict_from_persisted_row`` refuses to honour a
+        # stored True because "the alternative is a live surface showing
+        # 'look_ahead: PASS' on the strength of a sentence an LLM wrote about
+        # itself". Any look-ahead CLAIM keys off ``look_ahead_status``
+        # (debate_engine.py:770-773) — which the fusion writer always sets, so no
+        # conclusive row reaches this branch.
+        return _check(
+            "look_ahead",
+            label,
+            NOT_COMPUTED,
+            "no four-state audit verdict on record — the stored boolean is not an audit result",
+        )
     if verdict.get("lookahead_audit_passed") is False:
         return _check("look_ahead", label, FAIL, "the audit found a forward-looking pattern")
     return _check("look_ahead", label, NOT_COMPUTED, "no look-ahead audit result on record")
@@ -266,6 +310,24 @@ def rigor_reasons_for_verdict(verdict: Any) -> dict[str, Any] | None:
         _oos_cliff_check(verdict, profile),
         _look_ahead_check(verdict, has_recorded_reason=bool(recorded_reason)),
     ]
+    # A verdict that carries its own ``reason`` was NEVER GRADED — the pipeline
+    # writes that string exactly for the branches where the gate did not run
+    # (too-short series, fixture mode, debate abstain). Numbers left on such a
+    # blob are defaults, not measurements: ``_patch_pbo``
+    # (agents/generation_pipeline.py:660) stamps ``pbo = 0.0  # PBO undefined
+    # for N<2`` onto every one of those candidates (they are all
+    # ``has_real_rigor=False``, so they are all in its ``agent_cands``), and
+    # patches ``0.0`` in for any agent candidate missing from the CSCV map even
+    # when N≥2. Reporting that sentinel as "PBO 0.00 < 0.50 required — passed"
+    # claims an overfitting evaluation that never happened, which is the exact
+    # class of overclaim this module exists to remove. The look-ahead leg is
+    # excluded because ``_look_ahead_check`` already handles the recorded-reason
+    # case with its own wording.
+    if recorded_reason:
+        for c in checks:
+            if c["key"] != "look_ahead" and c["status"] != NOT_COMPUTED:
+                c.update(status=NOT_COMPUTED, detail="not scored — see the reason on record", value=None)
+
     passing = bool(verdict.get("passing"))
     any_failed = any(c["status"] == FAIL for c in checks)
 
--- a/backend/archimedes/services/rigor_reasons.py
+++ b/backend/archimedes/services/rigor_reasons.py
@@ -1,0 +1,281 @@
+"""Per-strategy rigor-gate reasons — which checks failed, which passed, against what bar.
+
+The Library's "Rejected (N) — did not pass the rigor gate" section used to
+explain every rejected candidate with ONE shared paragraph of prose ("Most
+rejections at this stage are 'return series too short' … A longer backtest
+window typically unlocks them"). That sentence was written from a guess about
+the population, not read from any strategy's record, so for a candidate rejected
+for a DIFFERENT reason the surface stated something false about that strategy.
+
+This module turns a strategy's OWN stored ``rigor_verdict`` into a per-check
+report the card can render. That blob is the JSON in
+``StrategyRecord.rigor_verdict``, written by
+``generation_pipeline._rigor_verdict_for`` on the buy-and-hold/agent path and by
+``debate_engine._rigor_verdict_dict`` on the fusion/debate path.
+
+It is a PURE function over a dict the caller already holds — no DB read, no gate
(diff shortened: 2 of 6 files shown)


danielscoffee (approved):
**Lane:** honesty / Library UX — rejected cards no longer claim a population-wide reason nobody measured. In scope for Improvements.

**What lands**
- Per-row `rigor_reasons` derived from the stored `rigor_verdict` against `rigor_profiles` (thresholds read, not re-declared).
- `fail` vs `not_computed` kept distinct — same FAIL / NOT_RUN honesty the gate already has.
- Population paragraph deleted; section summary only when it is true of the rows on screen.
- Guards look right: drop the field, restore the old prose, or print an ungraded check as a fail and the new tests go red.

**Non-blocking notes (author already flagged these)**
1. **0.90 vs 0.95 DSR** — reporting `unattributed` instead of inventing a culprit is the right call for this PR. Reconciling the agent-path bar with the badge bar is still open and should stay separate.
2. **`_pbo_below_power_floor`** — still unmodelled without `pbo_library_size` on the stored verdict. Correct call not to fake it; worth a write-time field later so the card does not over-claim a PBO fail the live gate would treat as non-load-bearing.
3. **Desktop** — reasons only when the row is expanded is fine for v1; mobile above-the-fold matches the screenshot. A dedicated desktop row later is optional polish, not a merge blocker.
4. **CI** — Frontend + critical Ruff are green; Backend unit still pending as of this review. Quality Gate’s broader ruff “49 errors” looks like the non-critical / repo-wide bucket (critical Ruff job passed).

LGTM from the polish lane. Leaving merge to owner vet as the PR asks.

dbrowneup:
**Review fixes applied** — `a39488c0`. Two blocking defects, two minor, one fixture correction. Files: `backend/archimedes/services/rigor_reasons.py`, `backend/tests/test_rejected_reason_payload.py` (+178 / −24).

---

### Blocking 1 — `pbo = 0.0` rendered as a **passed** overfitting check on an ungraded candidate

Confirmed on the real blob: `debate_engine._abstain_result` → `generation_pipeline._patch_pbo` (line 660) stamps `pbo = 0.0  # PBO undefined for N<2` onto every candidate carrying a recorded `reason` (all `has_real_rigor=False`, so all in `agent_cands`), and 0.0 sits below the 0.50 ceiling. Probe before the fix:

```
pass  PBO (overfitting probability) | 0.00 < 0.50 required     <-- on a candidate the gate never scored
```

Fixed generally rather than by special-casing PBO: a verdict carrying its own `reason` was never graded, so **no** numeric leg on it is reported as a graded result.

```python
if recorded_reason:
    for c in checks:
        if c["key"] != "look_ahead" and c["status"] != NOT_COMPUTED:
            c.update(status=NOT_COMPUTED, detail="not scored — see the reason on record", value=None)
```

Guard `test_a_default_number_on_an_ungraded_verdict_is_not_a_passed_check`. Its fixture is not hand-written — it is `_abstain_result(...)` run through the real `_patch_pbo`, with an assertion pinning the `0.0` sentinel, so it cannot drift from the pipeline.

### Blocking 2 — a bare `lookahead_audit_passed: True` printed as a passed audit

`if status == PASS or bool(verdict.get("lookahead_audit_passed")):` fired before the `_LA_INCONCLUSIVE` branch. Probes before the fix:

```
look_ahead_status='pending', look_ahead_reason='spec had no auditable source', lookahead_audit_passed=True
  -> pass | no forward-looking data access found
no status, lookahead_audit_passed=True
  -> pass | no forward-looking data access found
```

The first is the sharper case: a verdict that *explicitly says the audit reached no verdict* still rendered as a pass. Fixed to `if status == PASS:` plus an explicit `is True` → `not_computed` branch ("no four-state audit verdict on record — the stored boolean is not an audit result"), citing `_lookahead_for_candidate`'s "vacuously True when none expose auditable source" and `dsl_lookahead_audit.verdict_from_persisted_row`, which retired this exact rendering. No live path regresses: the fusion writer derives the boolean *from* `look_ahead_status == "pass"` (`fusion_evaluator.py:143`), and agent/fixture/abstain rows write `False`.

Guards `test_a_bare_lookahead_true_is_not_a_passed_audit`, `test_an_inconclusive_status_beats_a_stored_lookahead_true`, plus `test_a_conclusive_look_ahead_pass_still_passes` pinning the no-regression direction.

### Minor 3 — the not-computed ratio detail misattributed the cause

`"no positive in-sample Sharpe to compare against"` was emitted unconditionally, and is false on a row with `in_sample_sharpe = 1.2` and no OOS Sharpe. Split: `"no out-of-sample Sharpe to compare"` when `oos is None`, the original string otherwise. Guard `test_the_ratio_check_names_the_side_that_is_actually_missing`.

### Minor 4 — `_rigor_verdict_for` has no production callers

Confirmed by repo-wide grep and `graft callers`: only its own `def` and tests. Docstring corrections in `rigor_reasons.py` (module docstring, the `unattributed` note, `SHORT_SERIES_REASON`, `_oos_cliff_check`) and in the test module docstring; the same claim is corrected twice in the PR body. Knock-ons written down rather than coded around: `SHORT_SERIES_REASON` / `MIN_RETURNS_FOR_GATE` match **stored rows only**, so the frontend's short-series summary sentence is a legacy path, and `unattributed` is likewise reached via stored rows. No code change — the pinning test stays.

### Fixture correction (found while fixing #2)

`_GRADED_FAIL` carried `lookahead_audit_passed: True` with **no** `look_ahead_status` — a shape no live path produces, since `_rigor_verdict_dict` (the only writer of a graded verdict) always emits the four-state status and derives the boolean from it. Added `"look_ahead_status": "pass"` so the "graded rejection" fixture is actually a graded row. `test_a_passing_verdict_reports_no_failures` gained the same field plus an assertion that all five checks read `pass`, and `test_unattributed_...` likewise.

---

### Mutations — every new guard red on the bug it exists for

| mutation | result |
| --- | --- |
| drop the `if recorded_reason:` downgrade | `1 failed, 20 passed` — `assert 'pass' == 'not_computed'` (exit 1) |
| restore the `or bool(lookahead_audit_passed)` short-circuit | `2 failed, 19 passed` — both look-ahead guards (exit 1) |
| restore the single ratio detail string | `1 failed, 20 passed` — `assert 'no positive ...' == 'no out-of-sa...'` (exit 1) |
| drop `d["rigor_reasons"] = …` from the route (verifier's) | `3 failed, 18 passed` — `KeyError: 'rigor_reasons'` ×2 + missing-field assert (exit 1) |
| force `rejectedSectionSummary` unconditional (verifier's) | `ℹ tests 12  ℹ pass 10  ℹ fail 2` (exit 1) |

### Green

```
pytest -p no:cacheprovider backend/tests/test_rejected_reason_payload.py -q
  21 passed, 4 warnings in 13.98s                                    # exit 0   (was 16)

pytest -p no:cacheprovider .../test_rejected_reason_payload.py .../test_strategies_routes.py \
       .../test_strategy_store.py .../test_generated_citation_truth.py .../test_unify_curated_generated.py -q
  130 passed, 12 warnings in 30.54s                                  # exit 0

pytest -p no:cacheprovider .../test_generation_pipeline.py .../services/test_generation_pipeline.py \
       .../test_brief_on_passport.py .../test_publishable_strategy_ids.py .../test_strategy_ownership.py -q
  108 passed, 33 warnings in 32.64s                                  # exit 0   (the write paths the new rules pin against)

cd ui && node --test test/rejected-reason.test.js
  ℹ tests 12  ℹ pass 12  ℹ fail 0                                    # exit 0
cd ui && node --test
  ℹ tests 925  ℹ pass 925  ℹ fail 0                                  # exit 0

ruff check <both touched files>          All checks passed!          # exit 0
ruff format --check <both touched files> 2 files already formatted   # exit 0
npx eslint src/rejectionReasons.js src/components/Strategies.jsx test/rejected-reason.test.js
  (no output)                                                        # exit 0
```

The `SIM105` on `strategies_routes.py:1263` remains pre-existing and untouched.

### Not done

The desktop-table coverage note is left as **concern 4** in the body rather than fixed — it needs the owner's call on which surface the screenshot was taken from, and a `colSpan` row is a design change, not a defect fix. `_pbo_below_power_floor` (#819) still isn't modelled; that needs a new field at write time. The chunked full-backend run in the body predates this commit and was not repeated — this commit touches only `rigor_reasons.py` (which nothing imports but the one route line) and its own test file, so the suites above are the ones that can reach it.

danielscoffee (approved):
**Re-review on `a39488c0`** (prior approve was on the first commit).

The two blocking honesty fixes land correctly:
1. **Default ≠ measurement** — a verdict with its own `reason` downgrades every numeric leg to `not_computed`, so `_patch_pbo`'s `pbo = 0.0` sentinel can no longer render as a passed overfitting check on an ungraded candidate.
2. **Look-ahead keys off `look_ahead_status` only** — bare `lookahead_audit_passed: True` (and an inconclusive status that still carried a True boolean) no longer short-circuits to PASS. Matches the retired `verdict_from_persisted_row` stance.

Minor 3 (ratio detail names the missing side) and the fixture/`_rigor_verdict_for` docstring corrections are fine. Guards that go red on those mutations are the right shape.

Still leaving merge to owner vet. Prior non-blocking notes (0.90 vs 0.95, `_pbo_below_power_floor`, desktop expand-only) unchanged and still not merge blockers.

dbrowneup:
On your two standing notes: the 0.90-vs-0.95 DSR bar is now its own issue, #1794 (one source, Önder's call on the number, guard against a literal threshold outside `rigor_profiles`); the `_pbo_below_power_floor` write-time field goes with PR-B of the stored-verdict program, where the passport writer gains fields anyway. Desktop expand-only stays v1. — via the owner's session

PR #1790, opened 2026-09-02
Verification-mail delivery feedback: replace the eternal 200 with sent / suppressed / rate-limited (#1748 item 2)

dbrowneup:
Closes part of #1748 — **item 2 (in-product delivery feedback)**. Item 4 (suppression-list hygiene) is [#1780](https://github.com/aprin-labs/archimedes/pull/1780); items 1 and 3 are already answered on the issue.

## The observation

`POST /api/auth/send-verification-email` returns `200 {status:true}` **forever**:

- for an address Amazon SES has already dropped onto the account suppression list — where `SendEmail` **succeeds**, returns a `MessageId`, and the message is then binned inside AWS;
- for an address whose last send threw (`AccessDeniedException`, `MessageRejected`, …);
- for an address that was never sent to at all.

The only trace a send left anywhere was a `console.error` on the failure path (`auth/auth.js`'s fire-and-forget `.catch`). So nothing in the product could tell "it is in your spam folder" from "SES is silently dropping this", and nothing could tell the account owner. The UI's most honest possible answer was `VERIFICATION_REQUESTED_MESSAGE` — *"requested — delivery isn't confirmed"* — which is true forever and therefore never becomes useful.

The free tier hangs on this. `free_generations.claim(..., email_verified=...)` is the only consumer of `emailVerified`, so a first-time human's activation path runs through a button whose feedback is a constant.

## Design

**Every send now leaves a receipt, and a session-required endpoint reads it back.**

```
sendVerificationEmail ─┐
sendResetPassword      ├─► auth/mailer.js ──► auth_email_deliveries   (one row per send:
sendChangeEmail…       │   (ses | console)                             MessageId, or the
notifyAccountChange   ─┘                                               AWS error NAME)
                                                                              │
GET /api/auth/verification-status  ◄── auth/verification-status.js  ◄─────────┤
  (session-required, self-only)          (pure resolver)                      │
                                   ◄── auth/suppression.js  ◄── SESv2 GetSuppressedDestination
```

| Piece | What it is |
|---|---|
| `backend/migrations/versions/d4b1f7c8e206_…` + `models/account.py` | `auth_email_deliveries`. Node writes the rows, alembic owns the DDL — the same split the four existing `auth_*` tables already live under. The write order is a DB-assigned `seq` (`BIGINT GENERATED BY DEFAULT AS IDENTITY`, `UNIQUE`), not a timestamp — see the concern-6 section. |
| `auth/delivery-log.js` | Insert/read against the service's existing `pg` pool. Never throws; a failed record cannot turn into a failed send. |
| `auth/mailer.js` | Records `sent` + `MessageId`, or `failed` + the error **name**, then rethrows so `auth.js`'s own `.catch` still makes it fail-soft. |
| `auth/suppression.js` | SESv2 `GetSuppressedDestination`, lazily imported, same task-role credentials as the sender. |
| `auth/verification-status.js` | Pure. Owns the state precedence and the resend window. |
| `auth/server.js` | The route. Session check, then the resolver. |
| `ui/src/verificationDelivery.js` + `VerificationDeliveryStatus.jsx` | One derivation, one component, both mount points. |

### Why a sibling GET rather than a richer response on the resend POST

`POST /api/auth/send-verification-email` is reachable with **no session** (`better-auth/dist/api/routes/email-verification.mjs:95-117`), and Better Auth defends it with a 500 ms constant-time floor so that "unknown or already-verified" and "known and unverified" are indistinguishable to an anonymous caller. That floor is exactly why `auth.js`'s `sendVerificationEmail` is fire-and-forget — an awaited SES round trip walked straight through it (measured 504 ms vs 922 ms) and turned the endpoint into an account-existence oracle.

Teaching **that** response to say `suppressed` vs `sent` would hand any anonymous caller a per-address oracle and undo the fix. This endpoint **requires a live session and reports on `session.user.email` only** — it reads no address from the request at all, so it can be as specific as it likes: the caller already owns the account it is asking about. A test drives `?email=victim@example.com` at it and asserts the response is still about the session's own address.

### State precedence, and the distinctions that carry the honesty

`suppressed` → `failed` → `rate_limited` → `sent` → `unknown` (`verified` short-circuits first and spends no SES call).

- **`suppressed` first** because it is the only state where pressing the button again *cannot* work. It is asserted only when the lookup actually ran (`checked && suppressed`); a lookup that threw falls through rather than guessing either way.
- **A `MessageId` is acceptance, not delivery.** SES returns one for a suppressed address too. No copy anywhere in this PR says "delivered", "sent successfully", or "check your inbox".
- **`sends: null` ≠ `sends: 0`.** `null` means the delivery log could not be read; `0` means nothing was ever sent. `delivery-log.js`'s `recent()` returns `null` rather than `[]` for exactly this, and the UI renders two different sentences.
- **`suppression.checked: false` is not "clean."** Throttling, a missing IAM grant, or the console mailer all produce it, and it renders as "we could not look" — in the API *and*, since `586ae4ac`, in the copy a human reads (follow-up 1 below).
- **`retryAfterSeconds` rides on every state**, computed from the `RESEND_WINDOW_MAX`-th newest send (the sliding-window answer — using the oldest row would clear the wait too early).

`auth.js`'s `rateLimit.customRules['/send-verification-email']` is now **built from** `verification-status.js`'s constants, because the endpoint quotes that window back to a human ("try again in 42 seconds"). Same 60/3 as before — `auth/test/email-flows.test.js` still pins the literal `{ window: 60, max: 3 }` on the built options, which is what stops the constants drifting.

### Infra

`infra/ecs.tf` gains `ses:GetSuppressedDestination` on `"*"` — the suppression list is an **account-level** SESv2 resource with no per-identity ARN, unlike the existing send statement. Read-only, one action. If the grant is missing the lookup fails `AccessDeniedException`, which `suppression.js` reports as "we could not look" — an honest unknown, **never** a false all-clear. The same fail-closed shape covers a missing migration: `record()` logs `EMAIL_DELIVERY_RECORD_FAILED`, `recent()` returns `null`, and the endpoint answers `unknown`.

### PII

The rows carry an email address, so `user_id` is `ON DELETE CASCADE` — account deletion takes them, matching migration `85ca5310b7a1`'s erasure policy — and `ui/src/account-deletion.js` gains a row saying so (the existing `ui/test/account-deletion.test.js` guard caught the omission before I did; that is the guard working). Nothing else about the message is stored: no subject, no body, no verification URL (a one-time bearer sign-in credential), and no error *message* — only the AWS SDK error's **name**, a fixed vocabulary that cannot carry an address or a token. A test asserts no row's JSON matches `token=|https?://`.

## Review follow-ups — `586ae4ac`

A review pass found one real defect and one stale comment; CI found a third thing neither of us had — the ordering that this feature's central claim rests on was never guaranteed, only lucky.

### 1. `suppression.checked === false` was computed, shipped, and then dropped by the UI

The bullet above claims a failed suppression lookup "renders as *we could not look*". The API said it. **The UI did not.** `ui/src/verificationDelivery.js` read `status.suppression` only for `.reason`, inside the `suppressed` branch — `grep -n checked ui/src/verificationDelivery.js ui/src/components/VerificationDeliveryStatus.jsx` returned nothing, and no UI test ever passed `checked: false`.

Reproduced against the real modules (`auth/verification-status.js` → `ui/src/verificationDelivery.js`), simulating day one in production — `terraform apply` has not run, so every `GetSuppressedDestination` is `AccessDeniedException`, **and the address really is suppressed**:

```
API state       : sent
API suppression : {"checked":false,"suppressed":false,"reason":null,"since":null,"detail":"suppression lookup failed"}
UI message      : Our mail provider accepted the last verification email for this address.
                  That is acceptance, not proof it reached you — it can take a few minutes.
UI canResend    : true
```

That is this PR's own failure mode with better prose: an optimistic constant in front of a permanently blocked address. And because concern 2 says the IAM grant is an unapplied owner step, `checked: false` is the **first** state production will actually serve. The `sent` and `unknown` copy now appends:

> We could not check whether our mail provider is blocking this address, so a permanent block cannot be ruled out.

Side effect, deliberately kept: under the console mailer (local dev) every `sent` carries the caveat — locally there genuinely is no suppression state to read, and dev is where you want to see that.

### 2. CI red — a tied millisecond decided which send was "the newest"

`not ok 54 - a send is recorded per address with its SES MessageId, and read back most-recent-first`, `expected 'failed', actual 'sent'` on ubuntu, green on macOS. Not ordering dependence between test files, not shared pool/DB state, not a stub mailer default: **timestamp resolution.**

`delivery-log.record()` stamped `created_at` with `new Date()`. Two records written back to back — a send and the record of the failure it threw; a signup send and a resend fired straight after — land on the **same millisecond** as a matter of routine. Measured on this exact code path:

```
$ node probe.mjs          # 200 back-to-back record() pairs, timestamps compared
ties (same ms) out of 200: 197
```

`ORDER BY created_at DESC` is then a tie. Postgres may return either row first; the test fake's `.sort()` is stable, so it returned **insertion order — oldest first**, the exact inverse of the contract. The test only ever passed because the *first* `record()` in a cold process pays for `randomUUID()`'s entropy-pool initialisation and crosses a millisecond boundary. A warm CI runner does not.

Reproduced deterministically by warming the process the way CI already had:

```
$ cat warm.mjs
import { randomUUID } from 'node:crypto'
for (let i = 0; i < 5000; i++) { randomUUID(); new Date() }

$ node --import warm.mjs --test --test-reporter=tap test/delivery-feedback.test.js
not ok 1 - a send is recorded per address with its SES MessageId, and read back most-recent-first
# pass 16
# fail 1
```

**Why this is not a test bug.** `verification-status.js` reads `rows[0]` as THE latest attempt. That one row sets `lastError` and decides `failed` vs `sent` — whether the account owner is told "the last attempt was refused" or "our provider accepted it". A tie at the newest millisecond flips that coin, and the wrong side of it is an optimistic claim about someone's mail, which is the thing this entire PR exists to end. The assertion was right. The ordering key was not.

**The fix, three parts:**

- `delivery-log.js` gains `nextCreatedAt()` — the row's ordering key, **strictly increasing within a process**: never behind the wall clock, never equal to the row before it. The drift that introduces is bounded by how many rows one process writes inside a single millisecond (a handful, at a few mails per second).
- The test fake now tie-breaks in the **worst legal direction** (older row first) rather than letting a stable sort quietly hand back the answer the test wants. Postgres promises no order for a tie, so the fake models the hostile case, and a tie reaching it now fails loudly instead of passing by luck.
- A new test **freezes `Date.now()` outright** and asserts both records still come back in order, that the key yields three distinct values from one input millisecond, and that it never lags the wall clock (a lagging stamp would age rows out of the 24 h window early and understate the resend count).

**Residual, stated rather than hidden:** this is per *process*. `ecs_service_desired_count` is 1 today but the service autoscales, and two tasks writing for one address inside the same millisecond can still tie. No key available in this file can recover an order those two events do not have. → **Closed in `5ae33a28`**: the order is now a DB-assigned `seq`, and `nextCreatedAt()` is demoted to the window/countdown timestamp it always also was. See the next section.

### 3. A comment pointed at a file that does not exist

`auth/verification-status.js`'s header said its rules are "executable in `auth/test/verification-status.test.js`". They are executable in `auth/test/delivery-feedback.test.js` — which is what the file's own sibling reference and `docs/api/auth-and-accounts.md:181` already said.

## Owner ruling on concern 6 — the write order belongs to the database — `5ae33a28`

Concern 6 said the ordering key was monotonic per *process*, that two tasks writing for one address inside a millisecond could still tie, and that the only real fix was a database-assigned sequence — "a migration change I did not make in a follow-up commit". **Owner ruling: `d4b1f7c8e206` has not shipped, so it gets the column now rather than a second migration later.**

### What changed

`auth_email_deliveries` gains `seq`. The Postgres DDL alembic actually emits, rendered offline:

```sql
CREATE TABLE auth_email_deliveries (
    id VARCHAR(32) NOT NULL,
    seq BIGINT GENERATED BY DEFAULT AS IDENTITY,   -- the write order
    ...
    PRIMARY KEY (id),
    UNIQUE (seq),
    FOREIGN KEY(user_id) REFERENCES auth_users (id) ON DELETE CASCADE
);
```

- `delivery-log.js` `record()` leaves `seq` **out of its `INSERT` column list entirely**. The database assigns it, which is the whole point: the sequence is the one thing two tasks share and neither of them can compute.
- `recent()` is now `ORDER BY seq DESC, created_at DESC`. `seq` alone decides — it is `UNIQUE`, so the secondary key is unreachable; it is named only so the statement reads the way the window predicate above it does.
- `verification-status.js` still reads `rows[0]`. **Nothing about the endpoint, the response body, or the UI changes** — only which row is genuinely first.
- `created_at` stays, and stays what the 24 h window and `retryAfterSeconds` are measured from. `nextCreatedAt()` stays with it: it no longer decides which row is newest, but a stamp that repeated or went backwards would still misreport how long a caller has to wait.
- `id` remains the primary key — Node mints it and reads it back; `seq` is an ordering column, not an identifier.

### Portability

`BY DEFAULT`, not `ALWAYS`, and the type is `sa.BigInteger().with_variant(sa.Integer(), "sqlite")` — the same shape `e41c7a9b2d63` uses for `paper_marks.id`. SQLite has neither `IDENTITY` nor sequences, and every `alembic upgrade head` test in this repo runs against a throwaway SQLite file, so those tests supply `seq` themselves; `BY DEFAULT` is what keeps an explicit value legal on both dialects. Nothing that writes this table for real ever runs on SQLite — `delivery-log.js` holds a node-postgres pool and nothing else.

### Residual, stated rather than hidden

The sequence value is taken when the `INSERT` **executes**, not when it commits, so a row inside a long transaction could commit after a row with a higher `seq`. Every write here is its own autocommit statement from `delivery-log.js`, so for this table that ordering *is* the write order. Two sends that are genuinely simultaneous across two tasks are ordered by whichever reached the sequence first — an answer the database has and neither process does.

### Tests

- **The same-millisecond test no longer stops the clock and hopes.** It records two rows and then **flattens `created_at` to one identical value across both**, so no ordering a clock could have produced survives in the data at all. The only thing left that can produce the answer is `seq`. (Strictly harsher than production, where `nextCreatedAt` still separates two rows written by one process.)
- **A new cross-process test** hands two `createDeliveryLog`s two `fakeDb`s over **one shared store** — two writers, one database, one sequence — interleaves four sends ending in a `failed` one, flattens every `created_at`, and asserts *both* readers return the same strictly-descending `seq` order and that `resolveVerificationStatus` answers `failed` with `lastError: AccessDeniedException`, not `sent`.
- **The fake now obeys the statement instead of assuming it.** It reads the `ORDER BY` out of the SQL text and sorts accordingly, breaks ties in the worst legal direction, and projects only the columns the `SELECT` names. Reverting `delivery-log.js` to `ORDER BY created_at DESC`, or dropping `seq` from the `SELECT`, makes the fake follow it down — which is the only reason these tests mean anything (M14 / M16 below).
- **Python.** The alembic round-trip now proves the `UNIQUE` *fires* (a duplicate `seq` raises `IntegrityError`) and that `NOT NULL` does too. A separate test renders the migration's **Postgres** DDL offline (`alembic upgrade <down_rev>:<rev> --sql`, no connection, no server) and asserts `seq` is `BIGINT GENERATED BY DEFAULT AS IDENTITY` — a property SQLite cannot express, so without it the identity could disappear with every SQLite test in the file still green (M15).

### Honest note on M12

The mutation "the ordering key goes back to `new Date()`" is now **green** — `# pass 19  # fail 0`. That is correct rather than a regression: with the order carried by `seq`, `created_at` colliding on a millisecond costs sub-millisecond precision on `retryAfterSeconds` and nothing else, so no test is owed. The job M12 used to do is now M14's. M13 — `nextCreatedAt` losing monotonicity — is still red, because the countdown claim is still real.

### Also in this commit

`backend/migrations/env.py` gains `AuthEmailDelivery` in its side-effect import list and `__all__`. That file's own docstring says the list "mirrors `archimedes.db.init_db()`'s import list exactly ... so the two never drift apart", and `586ae4ac`'s predecessor added the model to `db.py` only. The table registered on `Base.metadata` either way (the module import does it), so this restores a stated invariant rather than changing behaviour.

## Review push — one click, one answer — `57d577be`

Both reviews (00:22, 01:53) asked for the same thing and `586ae4ac` / `5ae33a28` did not touch the mount points. On a successful click each surface rendered `VERIFICATION_REQUESTED_MESSAGE` — *"requested — delivery isn't confirmed and may take a few minutes"* — **above** `<VerificationDeliveryStatus />`. Correct while the eternal 200 was the only fact available; two competing truths once the status GET answers. The caption could sit over a panel saying **suppressed** / **failed** / **rate_limited**, or repeat a second, softer "requested" over an honest **sent**.

### The fix

`shouldShowRequestedFallback(status)` in `ui/src/verificationDelivery.js` — **one predicate, both surfaces**, next to the derivation it wraps:

```js
export function shouldShowRequestedFallback(status) {
	return deriveVerificationDeliveryView(status) === null;
}
```

- `AccountSettings.jsx` — `{verifyStatus === 'sent' && shouldShowRequestedFallback(verifyDelivery) && (…)}`
- `ResendVerificationControl.jsx` — `{status === "sent" && shouldShowRequestedFallback(delivery) && (…)}`

The caption is now the **pre-status fallback only**: the GET is still in flight, it failed closed to `null` (401 / 503 / offline), or the body named a state this build has never heard of. In each of those the panel renders nothing and the click still needs acknowledging, and "a send was requested" is the one thing always true. A client-side `429` (`RATE_LIMITED_BY_CLIENT`) counts as a specific answer, so it suppresses the caption too.

No new string, no second copy of the rule: the predicate lives in the pure module and both components are pinned against re-deriving `deriveVerificationDeliveryView(x) === null` inline.

### Tests

`ui/test/verification-delivery.test.js` gains three (18 → 21):

- **absent when a view exists** — executable, for each of `sent` / `suppressed` / `failed` / `rate_limited` (plus `RATE_LIMITED_BY_CLIENT`): the state renders a view *and* `shouldShowRequestedFallback` is `false`.
- **present while there is nothing** — `null`, `undefined`, and a state from a hypothetical future build all return `true`.
- **both mount points, one predicate** — source-pinned on the exact gated JSX in each file, on the import, and against an inline re-derivation; plus the predicate being defined once, in the module.

Mutations M18 / M19 below.

## Adversarial demos — every guard shown red

Each mutation is applied by hand to the source, the suite re-run, and the source restored. **Every JS transcript below was re-run against the tree as committed at `%s`**; ordinals shifted because the cross-process test was inserted at position 3. Verbatim, `node --test --test-reporter=tap`.

`auth/test/delivery-feedback.test.js` — 19 tests:

**M1 — `suppressed` collapses into `sent`** (the suppression branch never fires):
```
### M1  [auth]  EXIT 1
   not ok 9 - SUPPRESSED wins over everything — resending cannot work, so nothing else is worth saying
   # pass 18
   # fail 1
```

**M2 — a failed suppression lookup reported as `checked: true, suppressed: false`** (i.e. "not suppressed"):
```
### M2  [auth]  EXIT 1
   not ok 7 - the suppression lookup reports BOUNCE, absence, and its own failure as three different things
   not ok 8 - the console mailer has no suppression state to report, and says so rather than "clean"
   # pass 17
   # fail 2
```

**M3 — drop the session requirement on the status route** (anonymous callers get a default user):
```
### M3  [auth]  EXIT 1
   not ok 18 - no session is 401, and an unwired status route is 503 — never a cheerful default
   # pass 18
   # fail 1
```

**M4 — an unreadable delivery log returns `[]` instead of `null`** (the fail-soft-into-a-plausible-substitute failure):
```
### M4  [auth]  EXIT 1
   not ok 4 - an unreadable log answers null, never an empty history — they are different claims
   # pass 18
   # fail 1
```

**M5 — the mailer stops recording the failure** (pre-#1748 behaviour: rethrow without a receipt):
```
### M5  [auth]  EXIT 1
   not ok 6 - the SES mailer records the MessageId on success and the error NAME on failure
   # pass 18
   # fail 1
```

**M9 — a failed send counted as one that "went out"** (`sends = rows.length` instead of the accepted-only filter). The spam hint quotes this number at a human, and the rate window deliberately counts the other thing — every attempt, because a refused send still consumed a limiter slot:
```
### M9  [auth]  EXIT 1
   not ok 3 - two processes writing for one address on one millisecond still agree which attempt was last
   not ok 14 - the spam hint counts ACCEPTED sends; the rate window counts every attempt
   # pass 17
   # fail 2
```

**M12 — the timestamp goes back to `new Date()`** — now **GREEN**, and that is the correct answer, not a hole. See "Honest note on M12" above: `created_at` no longer carries the order, and sub-millisecond drift on `retryAfterSeconds` is not a claim anyone reads:
```
### M12  [auth]  EXIT 0
   # pass 19
   # fail 0
```

**M13 — `nextCreatedAt` keeps the clock but loses monotonicity** (`lastIssuedAt = now`). Still red: the 24 h window and the resend countdown are measured from `created_at`, so a repeating stamp still misreports how long a caller must wait:
```
### M13  [auth]  EXIT 1
   not ok 2 - two records on one frozen timestamp still come back in order — the database assigns it
   # pass 18
   # fail 1
```

**M14 — `recent()` goes back to `ORDER BY created_at DESC`** (the ordering claim handed back to the clock; the mutation M12 used to be). The fake reads the `ORDER BY` out of the statement, so it follows the source down and breaks the tie in the worst legal direction:
```
### M14  [auth]  EXIT 1
   not ok 2 - two records on one frozen timestamp still come back in order — the database assigns it
   not ok 3 - two processes writing for one address on one millisecond still agree which attempt was last
   # pass 17
   # fail 2
```

**M16 — `seq` dropped from the `SELECT` column list** (ordered by a column never read back, so nothing downstream can check it):
```
### M16  [auth]  EXIT 1
   not ok 2 - two records on one frozen timestamp still come back in order — the database assigns it
   not ok 3 - two processes writing for one address on one millisecond still agree which attempt was last
   # pass 17
   # fail 2
```

`ui/test/verification-delivery.test.js` — 21 tests. M6 / M7 / M8 / M11 are unchanged from the `586ae4ac` run (no `ui/` file is touched by `5ae33a28`); their ordinals shift by the three tests `57d577be` appends. M18 / M19 are new and were run on `57d577be`:

**M6 — `suppressed` still offers the resend button** (`canResend: true`):
```
### M6  [ui]  EXIT 1
   not ok 2 - suppressed is the only state that takes the button away, and it says why
   # pass 17
   # fail 1
```

**M7 — `sent` claims delivery** ("The verification email has been sent and delivered to this address."):
```
### M7  [ui]  EXIT 1
   not ok 4 - sent claims ACCEPTANCE, never delivery — SES accepts mail it then drops
   not ok 9 - no state anywhere in this module claims the mail was delivered
   # pass 16
   # fail 2
```

**M8 — an unrecognised state silently mapped to "on its way"** instead of rendering nothing:
```
### M8  [ui]  EXIT 1
   not ok 8 - nothing is rendered for a verified account, an absent response, or a state this build does not know
   # pass 17
   # fail 1
```

**M11 — the `checked: false` caveat removed from the `sent` copy** (the defect this follow-up fixes, re-armed):
```
### M11  [ui]  EXIT 1
   not ok 18 - a suppression lookup that never ran is never hidden behind an optimistic 'accepted'
   # pass 17
   # fail 1
```

**M18 — either unconditional mount restored** (the defect these reviews flagged, re-armed — here both at once):
```
### M18  [ui]  EXIT 1
   not ok 21 - both mount points gate the caption on the SAME predicate, and neither keeps a copy
   # pass 20
   # fail 1
```

**M19 — the predicate neutered to `return true`** (the caption comes back everywhere, panel or no panel):
```
### M19  [ui]  EXIT 1
   not ok 19 - once the panel has something real to say, the eternal 'requested' caption stands down
   not ok 20 - the caption is still there while the status GET has nothing
   # pass 19
   # fail 2
```

Python-side, `pytest -q backend/tests/test_alembic_migrations.py -k "auth_email_deliveries or delivery_log_sql"` (3 tests), re-run at `5ae33a28`:

**M10 — the migration renames a column the Node sidecar `INSERT`s** (`message_id` → `ses_message_id`). This is the seam nothing type-checks: `auth/delivery-log.js`'s SQL is a string literal, so a rename here would break delivery feedback in production with every other Python test green:
```
### M10  [backend]  EXIT 1
   E   AssertionError: delivery-log.js INSERTs columns that do not exist: {'message_id'}
   FAILED backend/tests/test_alembic_migrations.py::test_auth_delivery_log_sql_names_the_columns_the_migration_creates
   2 failed, 1 passed, 28 deselected, 1 warning in 8.81s
```

**M15 — `sa.Identity()` dropped from the `seq` column** (the column survives, but Node would now have to fill it in itself and the order goes back to being per-process). SQLite renders the same `INTEGER NOT NULL` either way, so every other test in the file stays green — this is exactly why the Postgres DDL is asserted separately:
```
### M15  [backend]  EXIT 1
   E   AssertionError: seq must be assigned by the database — a Node-computed key only orders one process's writes
   FAILED backend/tests/test_alembic_migrations.py::test_alembic_auth_email_deliveries_seq_is_database_assigned
   1 failed, 2 passed, 28 deselected, 1 warning in 11.58s
```

**M17 — `unique=True` dropped from `seq`** (a total order assumed rather than enforced):
```
### M17  [backend]  EXIT 1
   E   AssertionError: seq is the write order; a repeated value makes 'newest' a coin flip
   FAILED backend/tests/test_alembic_migrations.py::test_alembic_auth_email_deliveries_table_added_and_removed
   2 failed, 1 passed, 28 deselected, 1 warning in 7.46s
```

Source restored after each; `git status --porcelain` empty afterwards. Clean runs: `auth` 94 pass / 0 fail; `ui` 931 pass / 0 fail; those 3 backend tests 3 pass / 0 fail.

## Verification

Re-run in full at `5ae33a28`:

```
$ cd auth && npm test
ℹ tests 94   ℹ pass 94   ℹ fail 0                              EXIT 0

$ cd ui && node --test          (re-run at `57d577be`)
ℹ tests 934  ℹ pass 934  ℹ fail 0                              EXIT 0

$ cd ui && npx eslint src/verificationDelivery.js \
      src/components/AccountSettings.jsx \
      src/components/ResendVerificationControl.jsx \
      test/verification-delivery.test.js    (no output)        EXIT 0

$ cd auth && npm test           (unchanged by `57d577be`, re-run anyway)
ℹ tests 94   ℹ pass 94   ℹ fail 0                              EXIT 0

$ cd ui && npm run lint                     (no output)        EXIT 0

$ pytest -p no:cacheprovider -q backend/tests/test_alembic_migrations.py
31 passed, 1 warning in 130.24s (0:02:10)                      EXIT 0

$ pytest -p no:cacheprovider -q \
      backend/tests/test_alembic_migrations.py backend/tests/test_account_schema.py \
      backend/tests/test_account_deletion_cascade.py backend/tests/test_identity_schema.py \
      backend/tests/test_db.py backend/tests/test_api_docs_drift.py
71 passed, 1 skipped, 1 warning in 120.20s (0:02:00)           EXIT 0

$ pytest -p no:cacheprovider -m "not integration" -q backend/tests
5647 passed, 4 skipped, 2 deselected, 328 warnings in 570.25s (0:09:30)   EXIT 0
(measured on `5ae33a28`; `9b1f3bc5` is comment-only. 5646 → 5647 is the one new backend test.)

$ ruff format --check <touched .py>          4 files already formatted   EXIT 0
$ ruff check --select E9,F63,F7,F40,F82 <touched .py>
                                             All checks passed!          EXIT 0
```

(The single skip is `test_identity_schema.py`'s Postgres-only case — no `DATABASE_URL` DSN configured locally.)

`5ae33a28` (+ `9b1f3bc5`, a stale-comment fix on the ORM model's index — `d4b1f7c8e206`'s twin comment was updated and the model's was not) touches no terraform, and the only docs change is one paragraph in `docs/api/auth-and-accounts.md` saying where "the last attempt" comes from. Rather than re-assert those locally, they are CI results on this exact commit: **Build (mkdocs build --strict)** ✅, **Docs — links + index** ✅, **Infra — terraform fmt + validate (infra)** ✅, **Ruff — format + critical lint rules** ✅, **Migrations — chain fork guard** ✅.

`ruff check` (the full, informational rule set — CI's blocking gate is `ruff format --check` plus `--select E9,F63,F7,F40,F82`) reports one **pre-existing** `RUF022` (`__all__` not sorted), now on `backend/migrations/env.py` as well as `backend/archimedes/db.py`. Confirmed present on those files before my additions — `git stash push backend/migrations/env.py && ruff check backend/migrations/env.py` still reports it — and both additions preserve the existing ordering. Not fixed here: re-sorting a 38-entry `__all__` in a delivery-feedback PR is unrelated churn.

Everything is hermetic. The auth tests use a hand-written `{query}` fake for the delivery log and a `loadClient` seam for SESv2 — the AWS SDK is never imported, no credentials, no network. The migration tests run real `alembic upgrade head` / `downgrade` subprocesses against throwaway SQLite files with a whitelist-only env.

**Not verified against live AWS.** No message was sent, no suppression list was queried, and the IAM change has not been applied — `terraform apply` is an owner step. The endpoint's first production answer will be `unknown` until the migration runs and a send is recorded, which is the correct honest first answer.

## Concerns / what a reviewer should push on

1. **The migration must land before the auth image that writes to it.** If it does not, `record()` logs `EMAIL_DELIVERY_RECORD_FAILED` per send and the endpoint answers `unknown` — fail-closed and honest, but noisy. Worth sequencing deliberately.
2. **`ses:GetSuppressedDestination` on `"*"`.** I could not scope it: SESv2's suppression list is account-level and has no per-identity ARN. If you would rather not grant it at all, the feature degrades to "we could not look" rather than lying — but then `suppressed` is never reported, and item 4's operator script is the only way to see the list.
3. **`retryAfterSeconds` is computed from our own rows, and Better Auth's limiter keys on client IP, not address.** So the two can disagree: a shared CloudFront edge can be refused while our log shows room. The client compensates by rendering a real `429` as rate-limited directly (`RATE_LIMITED_BY_CLIENT`), but the *number* only exists on the server-side path. Both are true statements about different things; if you would rather the endpoint say nothing about timing, that is a one-line removal.
4. **`createPool` was extracted from `createAuth`** so `server.js` can hand one pool to both Better Auth and the delivery log. Body and comment are unchanged, `createAuth`'s default path is unchanged, and all 94 auth tests still pass — but it is the one refactor in this PR and worth a look.
5. **`notifyAccountChange` now awaits one extra INSERT** on the `/link-social` and `/unlink-account` paths (it is the only caller that awaits `mailer.send`). Those carry no timing claim, unlike the three fire-and-forget senders, so this does not touch the anti-enumeration argument — but it is a real added round trip on a user-facing request.
6. ~~**The delivery log's ordering key is monotonic per PROCESS, not per account.**~~ **RESOLVED in `5ae33a28`** on the owner's ruling — `d4b1f7c8e206` has not shipped, so the table gets a DB-assigned `seq` now instead of a second migration later. See *Owner ruling on concern 6* above. What is still worth pushing on: the sequence value is taken at `INSERT` time rather than commit time (immaterial for the single-statement autocommit writes this table sees, stated anyway), and `seq` costs one extra `UNIQUE` index on a table that takes a handful of writes per account per day.
7. **No UI test renders React.** `ui/` has no jsdom/testing-library by design (CLAUDE.md anti-goal), so the component wiring is source-text-pinned while every honesty rule is a real unit test of the pure module. If you want the rendering itself executed, that is a bigger change to `ui/`'s test posture than this PR should make. `57d577be`'s mount-point gate is pinned the same way — the *decision* (`shouldShowRequestedFallback`) is a real unit test; that the JSX actually calls it is a regex.

Part of #1804


danielscoffee (commented):
**Lane:** honesty / verification UX polish (#1748 item 2). In scope for Improvements; not rubber-stamping merge (migration + `ses:GetSuppressedDestination` + author “owner vets first”).

### What looks solid
- Acceptance ≠ delivery is held end-to-end: `sent` copy, tests (M7), and docs all refuse “delivered.”
- Session-only status GET (no address in the request) is the right fix for the anonymous POST oracle.
- Fail-closed unknowns (`sends: null` vs `0`, `suppression.checked: false`, unknown states → render nothing) match the honesty posture.
- Deletion inventory adds `auth_email_deliveries` with the address called out — good privacy copy.
- Shared `deriveVerificationDeliveryView` + one `VerificationDeliveryStatus` for Account Settings and Generate avoids two surfaces inventing different answers.

### Push: dual status after resend
On a successful click, both surfaces still show `VERIFICATION_REQUESTED_MESSAGE` *and* `<VerificationDeliveryStatus />`:

- `AccountSettings.jsx` — `verifyStatus === 'sent'` block still mounts the eternal “requested — delivery isn't confirmed…” line above the new panel.
- `ResendVerificationControl.jsx` — same pattern with the shared free-tier string.

That was correct when the only answer was the eternal 200. After a refresh it can contradict the specific state:

1. Panel says **suppressed** / **failed** / **rate_limited**, while the caption still says delivery “isn't confirmed and may take a few minutes.”
2. Panel says **sent** (acceptance, not delivery) next to a second, softer “requested” sentence — two competing truths for one click.

**Ask:** once `verifyDelivery` / `delivery` is non-null and `deriveVerificationDeliveryView` returns a view, drop (or never mount) `VERIFICATION_REQUESTED_MESSAGE` for that click. Keep the eternal string only as the brief pre-status fallback (status endpoint still loading / failed closed to `null`), so the panel is the single source of truth when it has something real to say.

### Out of Improvements merge call
Owner sequencing for migration-before-auth-image and the account-level IAM grant stay owner decisions. Happy to re-check copy once the dual-message bit is sorted; not approving.

dbrowneup:
## Follow-up pushed — `586ae4ac`

Two real defects and one stale comment. CI is green on all 19 checks.

### 1. `suppression.checked === false` was computed, shipped, and dropped by the UI

`ui/src/verificationDelivery.js` read `status.suppression` only for `.reason`, inside the `suppressed` branch. So a lookup that never ran — a missing IAM grant, throttling, the console mailer — rendered as an ordinary optimistic "accepted". Day one in production is exactly that state, because the `ses:GetSuppressedDestination` grant is an unapplied owner step:

```
API state       : sent
API suppression : {"checked":false,"suppressed":false,...,"detail":"suppression lookup failed"}
UI message      : Our mail provider accepted the last verification email for this address. …
UI canResend    : true
```

The `sent` and `unknown` copy now carries "We could not check whether our mail provider is blocking this address, so a permanent block cannot be ruled out." One new UI guard, red without the fix.

### 2. The CI red: a tied millisecond decided which send was "the newest"

`not ok 54 … expected 'failed', actual 'sent'` on ubuntu, green on macOS. **Not** ordering dependence between files, shared pool state, or a stub default — **timestamp resolution.**

`delivery-log.record()` stamped `created_at` with `new Date()`. Two records written back to back share a millisecond routinely — measured on this exact code path, **197 pairs in 200**. `ORDER BY created_at DESC` is then a tie, the fake's `.sort()` is stable, so it returned insertion order: oldest first, the inverse of the contract. It only ever passed because the *first* `record()` in a cold process pays for `randomUUID()`'s entropy-pool init and crosses a millisecond boundary. A warm CI runner does not. Reproduced deterministically here by preloading `for (…5000…) { randomUUID(); new Date() }` before the file.

This is not a test bug. `verification-status.js` reads `rows[0]` as THE latest attempt — the row that sets `lastError` and decides `failed` vs `sent`, i.e. whether the account owner is told "the last attempt was refused" or "our provider accepted it". A tie flips that coin, and the wrong side is an optimistic claim about someone's mail: the thing this PR exists to end.

Fixed at the root, three parts:

- `nextCreatedAt()` — a **strictly increasing per-process** ordering key. Never behind the wall clock, never equal to the row before it.
- The test fake now tie-breaks in the **worst legal direction** (older row first) instead of letting a stable sort return the answer the test wants, so a tie can never again look like a correct answer.
- A new test **freezes `Date.now()`** and pins the ordering, the key's distinctness, and that it never lags the clock.

Residual, in the PR body as concern 6: this is per *process*. Two ECS tasks writing for one address in the same millisecond can still tie; only a DB-assigned sequence fixes that, which is a migration change I did not make here.

### 3. `auth/verification-status.js`'s header named `auth/test/verification-status.test.js`, a file that does not exist

Now `auth/test/delivery-feedback.test.js`, which is what the docs already said.

### CI — `Frontend + Better Auth`: **pass** (49s)

<https://github.com/aprin-labs/archimedes/actions/runs/33578644726/job/100088015151> — the previously red assertion, on the ubuntu runner:

```
ok 54 - a send is recorded per address with its SES MessageId, and read back most-recent-first
ok 55 - two records in the same millisecond still come back in order — the key is monotonic, not the clock
# tests 93   # pass 93   # fail 0

ok 907 - a suppression lookup that never ran is never hidden behind an optimistic 'accepted'
# tests 931  # pass 931  # fail 0
```

All 19 checks pass, 2 skipped, 0 failing.

### Local verification

```
$ cd auth && npm test                 93 pass / 0 fail        EXIT 0
$ cd ui && node --test               931 pass / 0 fail        EXIT 0
$ cd ui && npx eslint src/ test/     (no output)              EXIT 0
$ pytest -q backend/tests/test_alembic_migrations.py \
          backend/tests/test_api_docs_drift.py
                                      33 passed in 138.22s    EXIT 0
```

Twelve mutations re-applied by hand against this tree — every one red, source restored, `git status --porcelain` empty. Transcripts refreshed in the PR body; the previous ones were pasted from a 16-test revision and the ordinals no longer matched. Concern 4's "91 auth tests" is now 93.

danielscoffee (commented):
**Re: `586ae4ac`** — honesty lane still, still not an approve / merge call.

### Follow-up that landed well
- **`checked: false` caveat** on `sent` / `unknown` is the right day-one fix. Without the IAM grant applied, production would have served optimistic “accepted” forever; the new copy + the UI guard (`could not check|cannot be ruled out`) close that hole.
- **`nextCreatedAt()`** is the honesty fix, not a test tweak. `rows[0]` deciding failed vs sent cannot rest on a lucky millisecond. Hostile tie-break in the fake + frozen-clock test are solid. Residual cross-process ties (concern 6) stay owner-side when desired count > 1.
- Stale header path cleanup: fine.

CI green on the tip (19 pass / 2 skip) matches what you reported.

### Still open from the earlier review: dual message after resend
`586ae4ac` did not touch the mount points. Both surfaces still show the eternal string **and** the new panel on a successful click:

- `AccountSettings.jsx` — `verifyStatus === 'sent'` still mounts `VERIFICATION_REQUESTED_MESSAGE` above `<VerificationDeliveryStatus />`
- `ResendVerificationControl.jsx` — same pattern

So after status loads you can still get “delivery isn’t confirmed…” next to **suppressed** / **failed** / **rate_limited**, or a second softer “requested” next to honest **sent**.

**Ask (unchanged):** once `verifyDelivery` / `delivery` is non-null and `deriveVerificationDeliveryView` returns a view, drop (or never mount) `VERIFICATION_REQUESTED_MESSAGE` for that click. Keep it only as the brief pre-status fallback while the GET is loading / failed closed to `null`.

Owner bits stay owner: migration-before-auth-image, `ses:GetSuppressedDestination` on `*`, DB sequence if you ever scale past one auth task. Happy to re-check copy when the dual-message bit lands; not approving.

dbrowneup:
**`5ae33a28` — concern 6 closed: the write order is the database's now.**

Owner ruling taken: `d4b1f7c8e206` has not shipped, so the column goes into it rather than into a second migration later.

**What landed**

`auth_email_deliveries` gains `seq` — `BIGINT GENERATED BY DEFAULT AS IDENTITY`, `UNIQUE`. `record()` leaves it out of the `INSERT` column list entirely, because the database assigning it is the point: the sequence is the one thing two ECS tasks share and neither can compute. `recent()` is now `ORDER BY seq DESC, created_at DESC` — `seq` alone decides, `created_at` is named second only for readability and, being behind a unique key, can never be reached.

`verification-status.js` still reads `rows[0]`. The endpoint, the response body and the UI are byte-identical; only *which row is genuinely first* changed.

`created_at` and `nextCreatedAt()` both stay. They no longer decide which row is newest, but the 24 h window and `retryAfterSeconds` are measured from them, so a stamp that repeated or went backwards would still misreport how long a caller has to wait.

**Portability.** `BY DEFAULT` (not `ALWAYS`) plus `sa.BigInteger().with_variant(sa.Integer(), "sqlite")` — the shape `e41c7a9b2d63` already uses for `paper_marks.id`. SQLite has no `IDENTITY` and no sequences, so the alembic round-trip tests supply `seq` themselves; `BY DEFAULT` is what keeps that legal on both dialects. Nothing that writes this table for real ever runs on SQLite.

**Residual, stated not hidden.** The value is taken when the `INSERT` executes, not when it commits, so a row inside a long transaction could commit behind a higher `seq`. Every write here is its own autocommit statement, so for this table that ordering *is* the write order.

**Tests**

- The same-millisecond test now **flattens `created_at` to one identical value across both rows** — no ordering a clock could have produced survives in the data, so only `seq` can produce the answer.
- New cross-process test: two `createDeliveryLog`s over two `fakeDb`s sharing **one store** — two writers, one sequence — four interleaved sends ending in a `failed` one, every `created_at` flattened; both readers return the same strictly-descending `seq` order and `resolveVerificationStatus` answers `failed`, not `sent`.
- The fake now **obeys the statement rather than assuming it**: it reads the `ORDER BY` out of the SQL and projects only the `SELECT`ed columns. Verified red both ways — reverting to `ORDER BY created_at DESC` (M14) and dropping `seq` from the `SELECT` (M16) each fail two tests.
- Python: the alembic round-trip proves `UNIQUE` and `NOT NULL` actually *fire* (duplicate → `IntegrityError`), and a new test renders the migration's **Postgres** DDL offline (`alembic upgrade <down_rev>:<rev> --sql`) to pin `GENERATED BY DEFAULT AS IDENTITY` — a property SQLite cannot express, so without it the identity could vanish with every SQLite test still green. Verified red by removing it (M15) and by dropping the `UNIQUE` (M17).

**Honest note:** the old M12 mutation ("the ordering key goes back to `new Date()`") is now **green**, and that is correct rather than a hole — `created_at` no longer carries the order, so a millisecond collision costs sub-millisecond precision on `retryAfterSeconds` and nothing a human reads. M13 (`nextCreatedAt` losing monotonicity) is still red, because the countdown claim is still real. M14 now does the job M12 used to.

Also in the commit: `backend/migrations/env.py` gains `AuthEmailDelivery`, which its own docstring requires ("mirrors `init_db()`'s import list exactly ... so the two never drift apart") and the previous commit added to `db.py` only. No behaviour change — the module import already registered the table on `Base.metadata`.

**Verification (local, at `5ae33a28`)**

```
cd auth && npm test                    ℹ tests 94  ℹ pass 94  ℹ fail 0     EXIT 0
cd ui && node --test                   ℹ tests 931 ℹ pass 931 ℹ fail 0     EXIT 0
cd ui && npm run lint                  (no output)                         EXIT 0
pytest -q backend/tests/test_alembic_migrations.py
                                       31 passed in 130.24s                EXIT 0
pytest -q test_alembic_migrations + test_account_schema + test_account_deletion_cascade
       + test_identity_schema + test_db + test_api_docs_drift
                                       71 passed, 1 skipped in 120.20s     EXIT 0
pytest -m "not integration" -q backend/tests
                                       5647 passed, 4 skipped, 2 deselected in 570.25s   EXIT 0
ruff format --check <touched .py>      4 files already formatted           EXIT 0
ruff check --select E9,F63,F7,F40,F82 <touched .py>
                                       All checks passed!                  EXIT 0
```

Determinism, since this file is the one that went red on CI once before:

```
30 × node --test test/delivery-feedback.test.js                       0 failures
15 × node --import warm.mjs --test test/delivery-feedback.test.js     0 failures
```

(`warm.mjs` is the 5 000-`randomUUID()` pre-warm from the original CI-red reproduction — the thing that used to make every `record()` pair share a millisecond.)

CI, both shas, polled via `gh api .../check-runs`:

| Check | `5ae33a28` | `9b1f3bc5` (head) |
|---|---|---|
| **Frontend + Better Auth** | ✅ success | ✅ success |
| **Backend — unit tests** | ✅ success | ✅ success |

Every other check on the head commit is `success` or `skipped` — nothing failed, nothing cancelled.

dbrowneup:
**Re: the dual message after resend — `57d577be`.** Landed exactly as asked, in both surfaces, through one predicate.

`ui/src/verificationDelivery.js` gains the rule next to the derivation it wraps:

```js
export function shouldShowRequestedFallback(status) {
	return deriveVerificationDeliveryView(status) === null;
}
```

- `AccountSettings.jsx` — `{verifyStatus === 'sent' && shouldShowRequestedFallback(verifyDelivery) && (…)}`
- `ResendVerificationControl.jsx` — `{status === "sent" && shouldShowRequestedFallback(delivery) && (…)}`

So `VERIFICATION_REQUESTED_MESSAGE` survives only as the pre-status fallback: GET still in flight, GET failed closed to `null` (401 / 503 / offline), or a state this build has never heard of. Once the panel has a recognised view for the click it owns the answer alone — no "isn't confirmed…" over **suppressed** / **failed** / **rate_limited**, no second softer "requested" over an honest **sent**. A client-side 429 (`RATE_LIMITED_BY_CLIENT`) counts as a specific answer too, so it also suppresses the caption. No second copy of the rule: both components are pinned against re-deriving `deriveVerificationDeliveryView(x) === null` inline.

Guards — `ui/test/verification-delivery.test.js`, 18 → 21 tests:

- **absent when a view exists**, executable, for each of `sent` / `suppressed` / `failed` / `rate_limited` (each asserted to render a view at all, then `shouldShowRequestedFallback === false`).
- **present while there is nothing** — `null`, `undefined`, unknown future state.
- **both mount points, one predicate** — source-pinned on the exact gated JSX in each file, plus the import, plus the no-inline-re-derivation check.

Mutations, verbatim:

```
### M18  [ui]  the unconditional mounts restored (both)   EXIT 1
   not ok 21 - both mount points gate the caption on the SAME predicate, and neither keeps a copy
   # pass 20
   # fail 1

### M19  [ui]  the predicate neutered to `return true`    EXIT 1
   not ok 19 - once the panel has something real to say, the eternal 'requested' caption stands down
   not ok 20 - the caption is still there while the status GET has nothing
   # pass 19
   # fail 2
```

```
$ cd ui && node --test        ℹ tests 934  ℹ pass 934  ℹ fail 0        EXIT 0
$ cd ui && npx eslint <4 touched files>   (no output)                  EXIT 0
$ cd auth && npm test         ℹ tests 94   ℹ pass 94   ℹ fail 0        EXIT 0   (untouched, re-run anyway)
```

CI on `57d577be`: **Frontend + Better Auth** ✅ (plus Build, Docs, Infra ×3, Ruff, Migrations, CodeQL, import-guard all green; backend/python still running at the time of writing — no backend file is touched by this commit).

PR body updated with a *Review push — one click, one answer* section and the two new mutation transcripts. Owner bits stay yours: migration-before-auth-image, `ses:GetSuppressedDestination` on `*`, and the DB sequence past one auth task. Still not a merge — owner vets first.

— via the owner's session

PR #1774, opened 2026-09-01
[cleanup] Remove six dead modules and their 73 tests (audit P1-1)

dbrowneup:
## What this is

**Audit P1-1**, the first (zero-risk) item of Phase 1 in the 2026-09-01 test-suite + dead-code decision package:

> | P1-1 | Delete 6 dead modules + empty `_deprecated/` + their 73 tests: `services/sharpe_statistics.py` (377 LOC — orphaned duplicate of the live `_rigor_helpers.py` math, `services/rigor_evaluator.py:30-49`), `services/dynamodb_paper_index.py` (141), `services/s3_artifact_store.py` (103, + doc edit `infra/iam/README.md:168`), `services/regime_weight_schedule.py` (140, #164 wiring never happened), `agents/base.py` (36, zero in-edges per graft), `benchmarks/__init__.py` stub | −1,313 LOC, −73 tests | None found; **re-run graft in-edge check at PR time** |

The audit's own caveat — *re-run the in-edge check at PR time* — was the first thing done here. **All seven are still dead on today's `main` (`03bff3dc`). Nothing has gained a caller since the audit; nothing was left behind.**

### 1. Import-graph reachability (the authoritative check)

AST-walked every `import` / `from … import` in `backend/archimedes` (242 modules), resolved each target to its longest known module prefix, then BFS'd from **22 entry points**: `archimedes.main`, `chain.oracle_runner`, `chain.agent_runner`, `services.kb_runner`, and every `archimedes.scripts.*` CLI.

**Before deletion** — 222 of 242 modules reachable:

```
archimedes.services.sharpe_statistics       in_index=True  reachable=False in_edges=[]
archimedes.services.dynamodb_paper_index    in_index=True  reachable=False in_edges=[]
archimedes.services.s3_artifact_store       in_index=True  reachable=False in_edges=[]
archimedes.services.regime_weight_schedule  in_index=True  reachable=False in_edges=[]
archimedes.agents.base                      in_index=True  reachable=False in_edges=[]
archimedes.benchmarks                       in_index=True  reachable=False in_edges=[]
archimedes.services._deprecated             in_index=True  reachable=False in_edges=[]
```

`in_edges=[]` is the strong claim: not merely unreachable from the entry points, but **referenced by no module in the package at all**, dead or alive.

**After deletion** — 235 modules, **still 222 reachable**. The reachable set is byte-identical before and after. Deleting these seven removed nothing anything can reach.

### 2. graft in-edges (the audit's own cross-check, re-run)

```
[...]

PR #1792, opened 2026-09-02
[product-truth][adr] The passport is the rigor verdict of record (PR-A: generated strategies) — closes the #1746/#1747 split

### LINKED ISSUE #1747: [product-truth][P0] Library shows Live ✓ on strategies whose passports say gate failed

dbrowneup:
## The owner decision this implements

> **Dan, 2026-09-01:** generation, backtesting and grading are one-time events. The rigor
> verdict is graded ONCE at backtest time and **stored** on the strategy passport with its
> inputs; every surface reads the stored verdict; a re-grade is an explicit, versioned
> event, never a silent overwrite and never a recompute on read. #821's principle survives
> tightened: the stored verdict must be written by the real gate, never a fixture, and
> provenance proves it. Board-level FDR stays a live relational signal off the passport
> (#1654 option 1). Backtests are frozen evidence (#1760 / PR #1766).

ADR: [`docs/adr/rigor-verdict-of-record.md`](../blob/dbrowneup/verdict-of-record-a/docs/adr/rigor-verdict-of-record.md) — **Accepted, pending quant sign-off**, Dan, 2026-09-01.

**Scope of this PR: generated strategies.** Curated grading is PR-B; the explicit re-grade
of existing rows is PR-C. Plan at the bottom.

---

## Since the verifier's FIX-FIRST (3 new commits)

A review pass found one **new** mislabel this PR had introduced, plus a cluster of claims in
the decision record and on the agent-facing surfaces that were false — two of them denied by
this PR's own body. All eleven are fixed.

| # | Defect | Fix |
|---|---|---|
| 1 | A generated strategy the real gate graded **`fail`** rendered as amber **"Pending Backtest"** on the Library — a badge asserting no gate had run, on a row a gate ran and failed | Retired the client-side `pending_backtest` invention entirely; the store status travels unchanged and `rigor_gate_status` is the only answer to "has a gate graded this?" |
| 2 | The pill and its own tooltip said different things, and the arm was dead for its own case | New `statusTitle` shares `isUngraded` with the pill and mirrors its arm order |
| 3 | ADR: "**Readers read.** No surface recomputes a verdict" — the deploy ladder does | Scoped to the *badge* verdict; deploy ladder named as an open seam, with why it is left alone |
| 4 | ADR contradicted itself about curated ids, 30 lines apart | Scoped to generated ids; the curated disagreement stated as an expected consequence |
| 5 | `docs/agent-quickstart.md`: "cannot disagree about one id" — false for every curated id | Scoped to generated, with the curated exception named |
| 6 | MCP `archimedes_passport`: an invariant and its counterexample in one string | First clause qualified to generated strategies |
| 7 | ADR `Accepted` while naming an unreviewed quant lane, index said "None open" | `Accepted, pending quant sign-off` + a review-debt row (`num-trials-self-containment.md` precedent) |
| 8 | Stale comment describing the deleted read-time derivation | Deleted |
| 9 | Two comments this PR had made false | Corrected |
| 10 | Test name "the curated tab's existing labels are unchanged" asserted one tuple and claimed a tab — curated labels **do** move | Renamed to what it proves; the movement pinned as intended behaviour and recorded in the ADR |
| 11 | `classify_return_source` read the stored boolean while the badge derived the four-state — two sources for one fact, in one function | Both read the derivation; pinned by a forced mixed-vintage row |

Nothing was pushed back on: all eleven reproduced. Defect 1 was worse than reported — the
same rewrite also fired on **`degenerate`** rows, giving them a tooltip announcing that no
backtest had run over returns that had one. Defect 10 was broader too: the movement covers
every store status, not only `validated`.

---

## Review nit 1, landed (`5fbece3c`) — raised in the 2026-09-02 review session; no formal review is recorded on this PR

> **`StrategyPassport.jsx` still has local two-arg `statusTag`/`statusLabel`.** Only
> `GATE_FAILED_LABEL` is shared. After boolean↔four-state coupling, `pending`/`degenerate`
> are `passes_rigor_gate: false`, so a `live` passport can still paint the status pill
> **"Reference only — gate failed"** while the rigor chip correctly says pending/unevaluable.
> Library fixed this; Passport didn't finish the job.

Correct, and the gap was exactly where the review put it: **importing the shared LABEL bought
byte-identical words over an unshared DECISION.** The passport's local pair took
`(status, passesRigor)` and never saw `rigor_gate_status`, so once the Library moved to the
four-state the passport was the last place the fail-closed boolean could still stand in for
the verdict. `passes_rigor_gate` is `false` for `pending` AND `degenerate`, so the tuple
those helpers saw — `("live", false)` — was indistinguishable from a real `fail`.

The passport now calls `statusTag` / `statusLabel` / `statusTitle` from
`ui/src/libraryStatus.js` with the full four-state — the same single argument list the
Library's desktop row and mobile card use. One vocabulary, one module. A pill says "gate
failed" only when `rigor_gate_status === "fail"`; `pending` → "Not yet graded",
`degenerate` → "Unevaluable — flat returns". It also picks up the tooltip the Library already
renders, from the same call, so class, words and reason come from one row's one verdict.

**Two pill colours move as a result, deliberately** — recorded in the ADR beside the curated
label movement rather than left to be discovered: `validated` was **green** on the passport
and is `tag-accent` in the shared helper; an unknown/`candidate` status was `tag-accent` and
is `tag-muted`. Green is reachable from exactly one place now — a `live` row with a literal
`true` verdict — which is the whole point of the shared module; a second green door on the
passport would re-open the gap this PR closes. **No graded state's wording changes.**

The guard that existed here was part of the problem: it asserted the passport *imported the
shared label*, which was true the entire time the bug was live. Replaced by two tests
(`ui/test/library-status.test.js`) — one pinning the wiring (all three helpers imported, no
local definition of any of them, no two-argument call, each called exactly once with
`(s.status, s.passes_rigor_gate, s.rigor_gate_status)`, no live copy of the demotion literal)
and one pinning the behaviour for the served shape behind the bug. Mutations in the
adversarial section below (**P1**, **P2**).

Nit 2 (`docs/api/strategies-and-rigor.md`'s intro + `GET /api/strategies/` blurb still
claiming badge+numbers come from the same live gate as `/api/selection-bias/gate`) is **not**
in this commit and remains open.

## Root cause

`strategy_passports.passes_rigor_gate` was **mixed-vintage**, and no surface agreed about
where a strategy's verdict even lived.

**Two writers, one column, no way to tell them apart.**

| Writer | When | What it actually meant |
|---|---|---|
| `generation_pipeline._persist_candidate` | at synthesis, before any backtest exists | the **fusion/debate** gate's opinion (`c.rigor_verdict["passing"]`) |
| `_refresh_passport_real_metrics` / `_backtest_and_persist` | after a real backtest — *if* it ran and did not throw | the **real** gate over the real return series |

Both post-backtest paths swallow their exceptions into a `logger.warning`, so "the fusion
verdict is still sitting in that column" was a common, invisible state.

**Three readers, three different sources.**

- `GET /api/strategies/{id}` **derived** a four-state badge at read time (`_passport_rigor_status`)
  from the stored aggregate plus a freshly-loaded return series.
- `GET /api/strategies/passports/{id}` served the raw row — which carried **no
  `rigor_gate_status` key at all**, so the two payloads were not even comparable field-for-field.
- `GET /api/strategies/generated` — the Library's Generated tab — **never read
  `strategy_passports`**. It served `StrategyRecord.to_dict()`, whose `status` and
  `rigor_verdict` are *both* written from the same synthesis blob. So the tab's own honesty
  guard (`status === 'live' && passesRigor === false` → "Reference only — gate failed") was
  **structurally unreachable**: the two halves of that condition came from one field.

That is #1747's twenty-one rows reading "Live ✓" in the Library while their own passports
read "Reference only — gate failed", with no shared field between the two answers.

And two states that mean opposite things were being served identically: every curated
passport row carries a hardcoded `passes_rigor_gate = False` — the deliberate #821
**fail-closed placeholder**, not a gate result — which the passport route served as a
verdict. That is #1746's half: a placeholder disagreeing with a real gate run, and the
placeholder winning on one surface.

Two more defects found on the way, both fixed here:

- `passport_loader._update_record` never wrote `win_rate` / `calmar_ratio` /
  `correlation_to_spy` / `total_trades` / `backtest_start` / `backtest_end` / `n_obs_daily`
  / `sharpe_ci_lower` / `sharpe_ci_upper` on the `force_update` path — only the INSERT
  branch did. Every post-backtest refresh takes `force_update`, so on any row that already
  existed those stayed frozen at the first ingest's values (usually NULL) while
  `_passport_to_strategy_response` served them beside a fresh Sharpe from a different run.
- `passes_rigor_gate = True` beside `sharpe_ratio IS NULL` was a **reachable pair** (the
  fusion verdict wrote the boolean; nothing wrote a status). The old read path called that
  row `pending` while `vaults_routes._strategy_rigor_status` read the boolean as a pass.

---

### Model + migration
- `strategy_passports` gains four columns: `rigor_gate_status` (`String(16)`, **NOT NULL**,
  server default `'pending'`), `graded_at` (`DateTime`), `gate_version` (`String(64)`),
  `cohort_n` (`Integer`).
- New service `services/rigor_gate_version.py`. `gate_version()` is a short, stable digest
  (`gate-v1-<16 hex>`) of everything that can move a verdict **without the strategy's own
  returns moving**: the whole strictness ladder, the always-on floors (`DSR_P_FLOOR`,
  `OOS_ABS_FLOOR`, `CPCV_MIN_POSITIVE_FRACTION`), `STRICTEST_LEVEL`,
  `_MIN_RETURNS_FOR_GATE`, `MIN_LIBRARY_N_FOR_PBO_GATING`, the DSR/rf convention constants,
  and a hand-bumped `GATE_CODE_REVISION`. Its docstring documents what goes in **and what
  deliberately does not**: the git SHA (a digest that changed every deploy would mark every
  verdict stale and stop being read) and board FDR (#1654 option 1 — it never flips
  `passes_all`, so it cannot move a stored verdict).
- `_update_record` now also writes the eight metric columns named above on `force_update`.

### Single writer
- `passport_loader.RigorVerdictWrite` is the only thing `ingest_passport` accepts as a
  verdict. `passes` is a **derived property** (`status == "pass"`), not a field — the
  boolean and the four-state are unconstructible apart. `__post_init__` fills `graded_at`
  and `gate_version`, so a verdict *without* provenance cannot be built, and rejects a
  status outside the four states.
- `ingest_passport` reads `passport.passes_rigor_gate` **nowhere**. With no
  `rigor_verdict=`: a new row inserts ungraded (`pending` / `False` / no provenance); an
  existing row **keeps the verdict it has** — a refresh that carries no grade neither
  erases one nor silently rewrites one.
- `_persist_candidate` no longer sends the fusion verdict or its DSR/PBO/OOS numbers to the
  passport. The fusion verdict stays on `StrategyRecord.rigor_verdict` as the **debate
  record** — worth keeping precisely because the real gate can disagree with it.
- `_refresh_passport_real_metrics` / `_backtest_and_persist` now take the live
  `RigorGateVerdict` (not a bool) and write `RigorVerdictWrite.from_verdict(live, cohort_n=1)`.
  Storing `verdict.status` verbatim is what keeps `degenerate` from collapsing into `fail`
  at the write, where no read could recover it.
- Both swallowed exceptions around the post-backtest refresh are now `logger.error(...,
  exc_info=True)` naming the strategy id and saying the passport stays UNGRADED. This is
  the swallow that decides whether a strategy is ever graded at all.

### Readers (generated)
- `_passport_to_strategy_response` reads `record.rigor_gate_status` and derives
  `passes_rigor_gate = (status == "pass")`. The `daily_returns` parameter and the whole-cohort
  `get_all_daily_returns` in `_passport_responses` are **deleted** — a verdict recomputed on
  read is a second gate run whose answer can differ from the stored one. (Side effect: the
  unbounded `/passports` list route stopped projecting and deserializing every row's
  `artifact_json` per page. A saving, not the motive.)
- `_passport_rigor_status` is retained **off the request path** as the oracle the migration's
  backfill rule is tested against, and as documentation of what the states meant before they
  were stored.
- `list_generated_strategies` overlays the stored verdict + the four rigor numbers from the
  same grading event, in **one `IN` query per page** (`_passport_verdicts_for`). No passport
  row → `passes_rigor_gate: null`, `rigor_gate_status: "pending"`, numbers `null` — never
  `false`, which would claim a gate ran and the strategy lost.
- `to_dict()` exposes `rigor_gate_status` / `graded_at` / `gate_version` / `cohort_n`.
- Docs: `docs/api/strategies-and-rigor.md` (all three routes), `docs/agent-quickstart.md`,
  and the `archimedes_passport` MCP tool description — which previously called the route
  "the unified record carrying the gate result" while it served a placeholder.

### UI
- New importable `ui/src/libraryStatus.js` (plain `.js` — `.jsx` is not importable under
  `node --test`) owning `statusTag` / `statusLabel` / `statusTitle`, consuming the four
  states: `pass` → green "Live"; `fail` → muted **"Reference only — gate failed"**;
  `pending`/`null` → muted "Not yet graded"; `degenerate` → muted "Unevaluable — flat returns".
- **Both surfaces call those helpers — the Library AND the Strategy Passport** (`5fbece3c`,
  review nit 1). The passport first shipped importing only the demotion *constant* while
  keeping its own two-argument `statusTag` / `statusLabel`, which shared the words and not the
  decision: with `passes_rigor_gate` false for `pending` and `degenerate` alike, a live
  passport whose gate had never run painted "Reference only — gate failed" beside its own
  chip reading "rigor gate pending". Both surfaces now pass
  `(s.status, s.passes_rigor_gate, s.rigor_gate_status)`. Two passport pill colours move with
  it, on purpose: `validated` green → `tag-accent`, unknown/`candidate` `tag-accent` →
  `tag-muted`. Green is reachable from one place only.
- **The boolean check was deliberately NOT widened to `!== true`.** That widening re-commits
  #1358 in mirror image: it tells an UNGRADED row it failed, contradicting the clock icon the
  same row renders ("Not yet evaluated"). The four-state gets its own arm instead.
- `coerceGenerated` takes `passes_rigor_gate` only as a literal boolean
  (`typeof === 'boolean'`, null preserved), carries `rigor_gate_status`, and sources
  DSR/PBO/OOS/dsr_p from the served row. It reads `row.rigor_verdict` nowhere — a badge from
  one gate beside numbers from another is the shape #1187/#1340 removed from the curated path.
- The mobile lib-card uses the same helpers with the same four-state input. It has no rigor
  icon of its own, so its pill is the only verdict signal on a phone.
- Stale comment at `StrategyPassport.jsx` corrected: the `/api/selection-bias/gate` 404 is
  no longer "expected" (`_generated_strategy_rigor` answers 200 with a live ladder). It now
  names the seam: **the badge is the verdict, the ladder is the deploy check.**
- **The invented `pending_backtest` status is retired.** `coerceGenerated` used to rewrite a
  store status of `"rejected"` to a client-side `pending_backtest` whenever no metric number
  was present. That workaround predates the four-state and asks the wrong question — "is
  there a NUMBER?" instead of "has a GATE run?" — and the first version of this PR turned
  that into a live mislabel: `/generated` serves no `sharpe_ratio` key at all, so the guard
  collapsed to `deflated_sharpe_ratio != null`, and DSR is Optional on a `BacktestResult`. A
  strategy the real gate graded **`fail` with no DSR** rendered as an amber **"Pending
  Backtest"** — a badge asserting no gate had run, on a row a gate ran and failed. A
  `degenerate` row got the same rewrite plus a tooltip announcing no backtest had run, over
  flat returns that had one. The store status now travels unchanged; `rigor_gate_status` is
  the only answer to "has a gate graded this?". Its last reader in `metricDomain.js` goes
  with it (those rows are caught by `is_backtest_placeholder`, set on every generated row).
- **The pill's tooltip now comes from the same condition as the pill.** It was keyed on
  `s.status === 'pending_backtest'` while the pill read the four-state, so the two disagreed
  in both directions — silent on an ungraded `candidate` row, firing on a degenerate one. New
  `statusTitle` shares `isUngraded` with `statusLabel` and mirrors its arm order; all three
  helpers now take the same three arguments at both call sites, pinned by test.
- **Curated labels move on the Examples tab, on purpose.** A curated row whose live verdict is
  `pending` or `degenerate` used to render its store status ("Validated", "Candidate") and now
  renders "Not yet graded" / "Unevaluable — flat returns"; a `validated` row also drops from
  `tag-accent` to `tag-muted` in those two states. A curated row with a real `pass` is
  unchanged. Intended — a `validated` status beside an ungraded gate is the same
  false-confidence shape as the generated side — and now stated in the ADR's Consequences and
  pinned by its own test, rather than hidden behind a control test whose name
  ("the curated tab's existing labels are unchanged") asserted one tuple and claimed a tab.

### Curated in PR-A
The curated detail route keeps its live computation, unchanged — PR-B replaces it.
`GET /api/strategies/passports/{id}` for a curated strategy now serves `"pending"` from the
migration, and the docs and the MCP tool description **say so explicitly**, including where
to get the live verdict instead.

---

## Migration + backfill

`b3f19d6c47ae` — one revision on the serial chain, `down_revision = c5e81a4f7b32`.

| Rows | `rigor_gate_status` | `gate_version` |
|---|---|---|
| `generation_method = 'curated'` | `pending` | NULL |
| other, `sharpe_ratio IS NULL` | `pending` | NULL |
| other, graded, stored boolean true | `pass` | `legacy-derived` |
| other, graded, stored boolean false | `fail` | `legacy-derived` |

Then `passes_rigor_gate` is rewritten to `(status == 'pass')` on **every** row — the
invariant the loader now enforces on writes, and a repair of the `true`-beside-NULL-sharpe
pair. `graded_at` stays NULL everywhere: no gate ran at a time we know, and inventing one
would be a fabricated provenance stamp. Downgrade drops the four columns (and deliberately
does not restore the decoupled booleans — that would be restoring a bug).

**The visible consequence, stated plainly.** The old read path had a fifth input SQL cannot
reach: it loaded each row's persisted return series and reported `degenerate` for a
zero-variance one. So a legacy generated row that today reads "Unevaluable — flat returns"
will read "Reference only — gate failed" until PR-C re-grades it. Both are non-pass, neither
renders green — the direction is fail-closed — but it *is* a claim getting less precise for
those rows, and the honest fix is the re-grade, not a recompute on read. Going forward the
writer stores `degenerate` as itself, so this is a one-time cost on existing rows only.

**And the intended one:** curated strategies stop saying "fail" on `/passports/{id}` and
start saying "not yet graded", which is true.

Verified on scratch sqlite: `upgrade head` → `downgrade base` → `upgrade head`, all exit 0.

---

## Adversarial pass — every new guard shown red

Each mutation was applied, the suite run, the output captured, and the file restored.

**M1 — `ingest_passport` writes the fusion verdict again**
(`passes_rigor_gate=passport.passes_rigor_gate` in the insert branch)
```
FAILED backend/tests/test_rigor_verdict_of_record.py::TestTheFusionVerdictNeverReachesThePassportVerdict::test_a_passing_fusion_verdict_lands_as_pending_not_pass - assert True is False
FAILED backend/tests/test_passport_loader.py::TestIngestPassport::test_basic_ingest - assert True is False
2 failed, 14 passed
```

**M2 — the read path derives the verdict again**
(`_rigor_status, _ = _passport_rigor_status(record, [])`)
```
FAILED backend/tests/test_multipaper_passport.py::TestTheStoredVerdictIsServedVerbatim::test_a_stored_degenerate_is_served_as_degenerate
FAILED backend/tests/test_multipaper_passport.py::TestTheStoredVerdictIsServedVerbatim::test_the_list_path_agrees_with_the_detail_path - AssertionError: assert {'bulk-degen'...g': 'pending'} == {...}
FAILED backend/tests/test_multipaper_passport.py::TestAFlatSeriesIsGradedDegenerate::test_a_long_flat_series_grades_degenerate - AssertionError: assert 'pending' == 'degenerate'
FAILED backend/tests/test_multipaper_passport.py::TestAFlatSeriesIsGradedDegenerate::test_a_real_varied_series_does_not_grade_degenerate - AssertionError: assert 'pending' == 'fail'
8 failed, 13 passed
```

**M3 — the Generated list drops the passport overlay** (restore the bare `r.to_dict()`)
```
FAILED backend/tests/test_rigor_verdict_of_record.py::test_the_three_read_surfaces_serve_one_verdict - KeyError: 'rigor_gate_status'
FAILED backend/tests/test_rigor_verdict_of_record.py::test_a_generated_row_with_no_passport_is_pending_never_green - KeyError: 'passes_rigor_gate'
2 failed
```

**M4 — the migration derives curated rows from the placeholder boolean**
(`'fail'` + `legacy-derived` for `generation_method = 'curated'`)
```
FAILED backend/tests/test_alembic_migrations.py::test_alembic_rigor_verdict_backfill_derives_the_documented_verdicts
E   AssertionError: assert ('fail', 0, '...erived', None) == ('pending', 0, None, None)
E     At index 0 diff: 'fail' != 'pending'
1 failed, 2 passed, 29 deselected
```

**M7 — the grading event writes no provenance** (`gate_version = None`, `cohort_n = None`)
```
FAILED backend/tests/test_rigor_verdict_of_record.py::TestTheGradingEventWritesACoupledVerdictWithProvenance::test_the_post_backtest_refresh_stores_the_real_gates_verdict - AssertionError: assert None == 'gate-v1-4d5c5472e3bc6d8d'
FAILED backend/tests/test_rigor_verdict_of_record.py::test_the_passport_payload_carries_its_verdicts_provenance - AssertionError: assert None == 'gate-v1-4d5c5472e3bc6d8d'
FAILED backend/tests/test_gate_equivalence.py::TestPipelineCallSiteReadsTheLiveVerdict::test_live_gate_verdict_reaches_the_persisted_passport[pass|fail|degenerate] - AssertionError: assert None == 'gate-v1-4d5c5472e3bc6d8d'
5 failed, 22 passed
```

**M8 — `gate_version` stops reacting to its inputs** (hash only the schema constant)
```
FAILED backend/tests/test_rigor_gate_version.py::test_the_digest_moves_when_a_gate_threshold_moves - AssertionError: assert 'gate-v1-67d6291b301412df' != 'gate-v1-67d6291b301412df'
FAILED backend/tests/test_rigor_gate_version.py::test_the_digest_moves_when_an_always_on_floor_moves
FAILED backend/tests/test_rigor_gate_version.py::test_the_digest_moves_when_the_pending_boundary_moves
3 failed, 8 passed
```

**M5 — `coerceGenerated` reads the fusion verdict again**
(`passes_rigor_gate: row.rigor_verdict ? Boolean(row.rigor_verdict.passing) : null`)
```
✖ coerceGenerated no longer reads rigor_verdict for the badge or its numbers
  AssertionError [ERR_ASSERTION]: passes_rigor_gate must be taken as a LITERAL boolean, with null preserved as null
ℹ pass 8   ℹ fail 1
```

**M6 — the label widens to `passesRigor !== true`** (an ungraded row is told it FAILED)
```
✖ an ungraded row says so, and never says it failed
    actual: 'Reference only — gate failed',
    expected: 'Not yet graded',
ℹ pass 8   ℹ fail 1
```

**M6b — the mobile lib-card keeps the old two-argument call**
```
✖ the pill helpers are imported, not redefined, and get the four-state
  AssertionError [ERR_ASSERTION]: both the table row and the lib-card must pass the four-state
    actual: 1, expected: 2,
ℹ pass 8   ℹ fail 1
```

---

### Second round — the guards added by the verifier fixes

The verifier's FIX-FIRST pass found a **new** mislabel this PR had introduced (below), plus a
cluster of false claims in the ADR/docs/MCP. Each fix that changes behaviour got its own test
and its own mutation.

**N1 — restore the `honestStatus` rewrite in `coerceGenerated`**
```
✖ coerceGenerated invents no status of its own, and the retired one has no readers
  coerceGenerated must not rewrite a store status to an invented one
ℹ pass 13   ℹ fail 1
```

**N2 — re-add the `pending_backtest` arms to `statusTag` / `statusLabel`**
```
✖ a strategy the gate FAILED never reads as pending, with or without a DSR
  no state may render the retired amber pill (fail)
✖ coerceGenerated invents no status of its own, and the retired one has no readers
ℹ pass 12   ℹ fail 2
```

**N3 — key the tooltip back on `s.status === 'pending_backtest'`**
```
✖ the pill's three helpers are called with the same argument list
  statusTitle must be called with the full four-state on BOTH the table row and the lib-card
✖ coerceGenerated invents no status of its own, and the retired one has no readers
ℹ pass 12   ℹ fail 2
```

**N4 — swap the two arms of `statusTitle` (a degenerate row gets pending's words)**
```
✖ the tooltip fires on exactly the rows the pill calls ungraded
  Expected values to be strictly equal: DEGENERATE_TITLE !== NOT_GRADED_TITLE
ℹ pass 13   ℹ fail 1
```

**N5 — drop the `isUngraded` + degenerate arms from `statusLabel`**
```
✖ curated rows with a pending or degenerate verdict change label, on purpose
✖ an ungraded row says so, and never says it failed
✖ a degenerate row gets its own words, not pending's and not fail's
✖ the tooltip fires on exactly the rows the pill calls ungraded
ℹ pass 10   ℹ fail 4
```

**N6 — restore the `row.status === "pending_backtest"` operand in `metricDomain.js`**
```
✖ coerceGenerated invents no status of its own, and the retired one has no readers
  metricDomain.js still has a live reader of the retired status
ℹ pass 13   ℹ fail 1
```

**N7 — restore `passes_rigor_gate=bool(record.passes_rigor_gate)` for `classify_return_source`**
```
backend/tests/test_rigor_verdict_of_record.py:570: in test_the_detail_route_derives_the_boolean_even_when_the_row_has_them_apart
    assert seen == [False], f"classify_return_source was handed {seen}, not the derived verdict"
E   AssertionError: classify_return_source was handed [True], not the derived verdict
E   assert [True] == [False]
1 failed, 12 passed
```

Each restored and byte-compared against a pre-mutation copy (`cmp -s`) before moving on.

### Third round — review nit 1 (`5fbece3c`)

Both mutations restore the shape that WAS on the branch when the review found it, so these
are the guards the old test could not have been: it asserted the passport imported
`GATE_FAILED_LABEL`, which was true throughout.

**P1 — restore the local two-argument `statusTag` / `statusLabel` in `StrategyPassport.jsx`
and call them with two arguments** (the branch's pre-fix state)
```
✖ the passport renders the Library's pill, not a private two-argument copy
  AssertionError [ERR_ASSERTION]: StrategyPassport.jsx must take all three pill helpers from the shared module
ℹ tests 15   ℹ pass 14   ℹ fail 1
```

**P2 — keep the shared import, drop only the third argument at the call site** (the subtler
regression: it still type-checks, still renders, and silently reinstates the mislabel because
`isUngraded` falls through to `passesRigor == null` → `false` and the fail arm answers)
```
✖ the passport renders the Library's pill, not a private two-argument copy
  AssertionError [ERR_ASSERTION]: no two-argument call may remain — it would silently drop the four-state
ℹ tests 15   ℹ pass 14   ℹ fail 1
```

`StrategyPassport.jsx` restored from a pre-mutation copy after each; full suite green below.

All files restored; `git status` clean of mutations before the final run below.

---

## Verification

Re-run in full after the verifier's fixes, on the branch merged up to `origin/main` @
`03397108` (#1766, backtests-are-frozen, landed mid-review and conflicted on the two doc
indexes — resolved as unions, both ADR rows kept, ADR count bumped to twenty-five).

```
# Full backend unit suite
$ pytest -p no:cacheprovider -q backend/tests
5649 passed, 6 skipped, 334 warnings in 678.36s (0:11:18)
EXIT=0

# The named regression net, on its own
$ pytest -p no:cacheprovider -q backend/tests/test_rigor_verdict_of_record.py \
    backend/tests/test_rigor_gate_version.py backend/tests/test_passport_loader.py \
    backend/tests/test_multipaper_passport.py backend/tests/test_gate_equivalence.py \
    backend/tests/test_unify_curated_generated.py backend/tests/test_alembic_migrations.py
109 passed in 116.74s
EXIT=0

# UI (node:test, archimedes env's node)
$ cd ui && npm ci --no-audit --no-fund --silent && node --test
tests 927   pass 927   fail 0        (+5 more in ui/test/library-status.test.js)
EXIT=0
# re-run at 5fbece3c (nit 1): one guard replaced, one added
tests 928   pass 928   fail 0
$ npx eslint src/components/StrategyPassport.jsx test/library-status.test.js   EXIT=0

# MCP server
$ cd mcp-server && PYTHONPATH=src:../cli/src pytest -p no:cacheprovider -q
213 passed in 4.71s
EXIT=0

# Migration round-trip on scratch sqlite  (single head b3f19d6c47ae, zero forks)
$ alembic upgrade head / downgrade base / upgrade head    EXIT=0 / 0 / 0
  rigor_gate_status VARCHAR(16) NOT NULL DEFAULT 'pending' present after the round trip

# Lint
$ ruff format --check <14 touched .py>   -> 14 files already formatted, EXIT=0
$ ruff check <14 touched .py>            -> 3 errors, all PRE-EXISTING on origin/main
    (SIM905 x2 generation_pipeline.py:236,245; SIM105 strategies_routes.py _rollback_quietly
     — each reproduced by running ruff, with the repo's own ruff.toml, over the origin/main
     copy of the same file)
$ cd ui && npx eslint .                  -> EXIT=0
```

**On the backend count.** It reads 5649 here against 5672 in the first round. Every unit of
that drop is `origin/main`, not this branch:

* #1774's dead-module audit deleted four test files — `test_sharpe_statistics.py`,
  `services/test_dynamodb_paper_index.py`, `services/test_regime_weight_schedule.py`,
  `services/test_s3_artifact_store.py` (54 test functions) — and #1772 added
  `test_cloudfront_session_paths_uncached.py`: **5672 → 5653**.
* #1766 deleted `test_backtest_scheduler.py` (9 tests) with the scheduler it retires and
  added `test_backtests_are_frozen.py` (5): **5653 → 5649**, exactly −4.

This branch's own delta is **+1** (the mixed-vintage detail-route test). All three runs are
EXIT=0 with zero failures. #1766 touches the backtest write path — `main.py`,
`backtest_repository.py`, `paper_trading.py` — but not `passport_loader.py`,
`generation_pipeline.py` or `strategies_routes.py`, and the alembic chain is still the single
head `b3f19d6c47ae` (it ships no migration).

The `test_health_always_answers[corpus_meta]` flake named in the first round
(`/health took 2.13s` against a 2s budget) did **not** recur on either full run here.

---

## PR-B / PR-C

- **PR-B — curated grading.** Replace the `passes_rigor_gate = False` placeholder in
  `services/strategy_provider.py` with a real stored grade (the curated cohort already has a
  live gate run on the detail route; that computation moves to the write side), and retire
  the detail route's read-time computation. After it, `/passports/{id}` for a curated
  strategy stops saying `pending`.
- **PR-C — the explicit re-grade.** A versioned event that replaces every
  `gate_version = 'legacy-derived'` verdict — and every `pending` this backfill could not
  resolve, including the flat-series rows that lost their `degenerate` label — with a real
  gate run. This is where `graded_at` and a current `gate_version` land on history.

## Open concerns for the owner

1. **Curated rows read `pending` on `/passports/{id}` between this PR and PR-B.** Honest,
   and the docs + MCP description say where to get the live verdict, but it is a visible
   change on a documented, anonymous, agent-facing route. The verifier found three places
   that had claimed the opposite — the ADR's "same four-state for the same id, by
   construction", `docs/agent-quickstart.md`'s "cannot disagree about one id", and the
   `archimedes_passport` MCP description's "agrees ... by construction" — each of which then
   documented the curated exception a few lines later. All three are now scoped to generated
   ids, so an agent reading top-to-bottom no longer gets an invariant and its counterexample
   in one string.
2. **The degenerate → fail regression for legacy rows** (named above). If you would rather
   not ship that window, PR-C can be pulled forward to land in the same train.
3. **`GATE_CODE_REVISION` is a human obligation.** A digest over constants cannot see a
   logic change in `run_rigor_gate`; the module docstring says bump it in the same commit,
   and `test_rigor_gate_version.py` pins the value so the bump is reviewed. There is no way
   to enforce it mechanically, and I would rather say so than pretend otherwise.
4. **The badge and the deploy ladder now have a named seam.** The passport badge is the
   stored verdict; `/api/selection-bias/gate/{id}` (the Deploy ladder, and the vault deploy
   gate through `vaults_routes._strategy_rigor_status`) still runs `run_rigor_gate` live.
   They can differ in vintage. The ADR asserted the opposite — "**Readers read.** No surface
   recomputes a verdict" — while this PR's own body named the seam; the ADR is now scoped to
   the *badge* verdict and states the deploy ladder as an open seam, with the reason it is
   deliberately left alone: **admission is the one place where recomputing against the
   current gate is arguably the safer behaviour.** Whether deploy admission may read a stored
   answer is still your call, and is not made here.
5. **`Leaderboard.jsx` still has only pass/fail/no-backtest arms** — a row the passport calls
   `degenerate` is asserted as "Gate failed" there. Same claims-must-be-true class, different
   surface, out of scope here.
6. **`main.py` seeds one `StrategyRecord` per curated strategy with `status="live"` and no
   `rigor_verdict`** — kept off the Generated tab only by the `is_example.is_(False)` filter.
   Untouched here; worth a look in PR-B.
7. **The ADR is `Accepted, pending quant sign-off`, not plain `Accepted`.** It names Önder
   Akkaya as quant reviewer of record and that review has not happened, which is exactly the
   case `docs/adr/README.md` defines that status string for. The index row and an
   "Open review debt" entry now say so (the entry names the two things that want a quant's
   eye: that `gate_version` covers every input that can move a verdict without the returns
   moving, and that the migration's `legacy-derived` backfill reproduces the old read-time
   derivation). Following the `num-trials-self-containment.md` precedent. Say the word and I
   will flip it to plain `Accepted` with the reviewer parenthetical dropped instead — this
   was your sole owner call, and the reviewer line may simply not be wanted.
8. **`classify_return_source` was the last two-sources-for-one-fact site**, inside the same
   function as the badge. It read `bool(record.passes_rigor_gate)` while the badge four lines
   below derived `_rigor_status == "pass"` — coupled today only by the migration and the
   loader. Now both read the derivation, pinned by a test that forces a real mixed-vintage
   row (`rigor_gate_status='pending'` beside `passes_rigor_gate=1`) apart with a raw UPDATE
   and asserts the route serves them together anyway.

---

Closes #1747

Part of #1746


danielscoffee:
## Improvements (honesty / provenance / polish)

Lane fit: yes. Closing the #1746/#1747 split on Generated + making `pending` mean "ungraded" is exactly this lane. Comment only — **not approving** (owner-vets; migration is a one-way coupling rewrite).

### What lands cleanly
- `ui/src/libraryStatus.js` + tests: four-state pills, shared demotion string, no `!== true` widening. Library desktop + mobile both pass the third arg.
- `coerceGenerated` keeps `null` as null and stops reading fusion `rigor_verdict` for the badge/numbers — overlay from `_passport_verdicts_for` is the verdict of record.
- Curated `/passports/{id}` → `pending` until PR-B is documented honestly (API doc + MCP contract + ADR). Null-not-false on ungraded Generated rows is correct.

### Two honesty nits worth fixing (here or a tiny follow-up)
1. **`StrategyPassport.jsx` still has local two-arg `statusTag`/`statusLabel`.** Only `GATE_FAILED_LABEL` is shared. After boolean↔four-state coupling, `pending`/`degenerate` are `passes_rigor_gate: false`, so a `live` passport can still paint the status pill **"Reference only — gate failed"** while the rigor chip correctly says pending/unevaluable. Library fixed this; Passport didn't finish the job. Prefer importing the four-state helpers (or stop demoting the status pill when `rigor_gate_status` is pending/degenerate).
2. **`docs/api/strategies-and-rigor.md` intro + `GET /api/strategies/` blurb** still claim badge+numbers come from the **same live gate** as `/api/selection-bias/gate` and cannot disagree. That contradicts the stored-verdict body (and the named badge-vs-ladder seam). Update the intro so agents don't trust the outdated sentence.

### Noted, out of scope (owner already listed)
- Leaderboard degenerate → "Gate failed"
- Legacy flat series → `fail` until PR-C
- Don't merge without owner vet (migration + curated pending window)

dbrowneup:
Both nits accepted and will land in this PR: (1) the passport status pill goes four-state (import the Library helpers; a pending/degenerate row must never paint "Reference only — gate failed"), (2) the API doc intro and the list-endpoint blurb stop claiming the badge and the ladder come from the same live gate. The branch has its adversarial verifier pass in flight from the owner's session; the fix round after it carries these two so we do not write to the branch twice. Not merging — owner vets, and the migration is his call. — via the owner's session

dbrowneup:
**Nit 1 is landed — `5fbece3c`.**

You had it exactly right, including *why* it hid: importing only `GATE_FAILED_LABEL` bought byte-identical **words** over an unshared **decision**. The passport's local pair took `(status, passesRigor)` and never saw `rigor_gate_status`, so once the Library moved to the four-state the passport was the last place the fail-closed boolean could still stand in for the verdict — and since `passes_rigor_gate` is `false` for `pending` *and* `degenerate`, the tuple those helpers saw, `("live", false)`, was indistinguishable from a real `fail`.

Took the first option you offered rather than the second: `StrategyPassport.jsx` now imports `statusTag` / `statusLabel` / `statusTitle` from `ui/src/libraryStatus.js` and calls them with `(s.status, s.passes_rigor_gate, s.rigor_gate_status)` — the same single argument list the Library's desktop row and mobile card use. No local definitions remain. "Stop demoting when pending/degenerate" would have fixed the symptom and left two vocabularies; a pill now says "gate failed" only when `rigor_gate_status === "fail"`, `pending` → "Not yet graded", `degenerate` → "Unevaluable — flat returns". The pill also picks up the tooltip the Library already renders, from the same call, so class, words and reason come from one row's one verdict.

**Two passport pill colours move, deliberately** — flagging it because it is not strictly the mislabel you reported: `validated` was **green** on the passport and is `tag-accent` in the shared helper; an unknown/`candidate` status was `tag-accent` and is `tag-muted`. Green is now reachable from exactly one place, a `live` row with a literal `true` verdict. Keeping a second green door on the passport would have re-opened the gap. No graded state's wording changes. Recorded in `docs/adr/rigor-verdict-of-record.md` beside the curated label movement, so it is a decision rather than a surprise.

The guard that was there was part of the problem — it asserted the passport *imported the shared label*, which was true the whole time the bug was live. Replaced with one test pinning the wiring (all three imported, no local definition, no two-argument call, each called once with the full four-state, no live copy of the demotion literal) and one pinning behaviour for the served shape: `("live", false, "pending")` → "Not yet graded", never the demotion label.

Mutations, both red (full output in the PR body, "Third round"):

```
P1  restore the local two-arg helpers + two-arg call   (the pre-fix branch state)
    ✖ the passport renders the Library's pill, not a private two-argument copy
      StrategyPassport.jsx must take all three pill helpers from the shared module
    ℹ tests 15   pass 14   fail 1

P2  keep the import, drop only the third argument at the call site
    ✖ the passport renders the Library's pill, not a private two-argument copy
      no two-argument call may remain — it would silently drop the four-state
    ℹ tests 15   pass 14   fail 1
```

`cd ui && node --test` → **tests 928, pass 928, fail 0**. `npx eslint src/components/StrategyPassport.jsx test/library-status.test.js` → EXIT 0.

**Nit 2 is still open** — `docs/api/strategies-and-rigor.md`'s intro and the `GET /api/strategies/` blurb are untouched by this commit. Not merging; owner vets first.

dbrowneup:
Nit 2 landed in 9ac7d38a: the API doc intro no longer claims the badge and `/api/selection-bias/gate` come from the same live gate run. It now states the real relationship — the strategies routes serve the stored verdict of record with its provenance, the ladder route recomputes live, the two can differ in vintage until an explicit re-grade, and curated rows serve `pending` until PR-B. Both of your nits are now on the branch (nit 1 in 5fbece3c). — via the owner's session

PR #1734, opened 2026-09-01
[docs] Point fusion-spec future pin at the hash-only ADR

danielscoffee:
## What

Follow-up to the non-blocking #1731 review note: `docs/specs/strategy-fusion-spec.md` still listed an IPFS pin as a future hop. It now points at [`docs/adr/ipfs-pinning-not-live.md`](https://github.com/aprin-labs/archimedes/blob/main/docs/adr/ipfs-pinning-not-live.md) so a later agent does not rebuild the deleted Pinata client.

The hop stays "vision, not v1." Re-enablement remains owner-gated (JWT in SSM + `ecs.tf` `secrets{}` + rebuilt pin client + CID on a public gateway).

## Why

#1731 review: a pointer on the fusion-spec future hop would stop a later agent from reconstructing the dead path. #1731 already merged without this sentence.

## Issues

Part of #1526 (already closed by #1731). Does not close anything new.

## Verification

Docs-only. The future-hop bullet and the v1 acceptance checkbox now name the ADR instead of "IPFS pin." `docs/README.md` index row updated (2026-09-01).

```
git grep -n "IPFS-pin the full reasoning" docs/specs/strategy-fusion-spec.md
# (no matches)

git grep -n "ipfs-pinning-not-live.md" docs/specs/strategy-fusion-spec.md
# docs/specs/strategy-fusion-spec.md:268
# docs/specs/strategy-fusion-spec.md:292
```

## What a reviewer should push back on

- **Not rewriting the rest of the fusion spec.** Only the future pin hop changed. The vision (public, falsifiable novelty claim) stays; the implementation path is no longer "rebuild Pinata."
- **`docs/infrastructure-provider-analysis.md` still naming Pinata** in the excluded-SaaS cost list. That is not a product claim; left as-is, matching the #1731 review.

PR #1795, opened 2026-09-02
[infra][deploy] Post-rollout probe: retry transport flakes, judge on an observed streak + good final sample (#1791)

### LINKED ISSUE #1791: [infra][deploy] Post-rollout probe fails the run on ONE transient sample — 2 of the last 4 main deploys red while prod was serving the new version
Two of the last four `deploy.yml` runs on main ended **red** at "Assert the deployed app actually answers (post-rollout probe)" while the rollout had already succeeded and prod was serving the expected version:

| run | head | what the probe saw |
|---|---|---|
| 33566302291 (22:39Z, #1765) | 25dc64dc | samples 1–7 = 200 in ~0.4 s; sample 8 `http=000 time=15.0s curl_exit=28` (client timeout); samples 9–10 = 200 → "only 2 consecutive trailing" → **exit 1** |
| 33573075995 (00:06Z, #1734) | 7a6c5b73 | samples 1–9 = 200 in ~0.4 s; sample 10 `http=000 time=0.025s curl_exit=35` (TLS handshake reset, 25 ms); → "only 0 consecutive trailing" → **exit 1** |

In both cases `/api/health` reported the expected version immediately afterwards (verified 00:45Z: `version=7a6c5b73…`, task-def 225, 2/2 running) and the CloudFront invalidation step still ran (#1765's `set +e` did its job). The run is red anyway, so the deploy badge/notifications say "failed" for a deploy that succeeded. That trains everyone to ignore red deploys, which is the one outcome a probe exists to prevent.

## Root cause
The probe's verdict is "the **last 5** samples must be consecutive 200s under 5 s". A single transport-level flake (curl exit 28/35/56, `http=000`) at any of the last five positions resets the streak and fails the run; the two flakes above are edge/TLS transients, not the app.

## Fix (proposed)
1. Transport failures (`http=000`, curl exit ≠ 0) are **retried** once or twice after a short pause before they count — a real outage still fails because the retries fail too.
2. The verdict becomes "≥ 5 consecutive good samples were observed **and** the final (post-retry) sample is good", so a transient inside an otherwise healthy window does not fail the run; every sample and retry is still printed so a real degradation is visible.
3. A 5xx or a wrong `.version` remains an immediate failure (no retry) — those are the app.
4. Hermetic guard: extract the probe script from `deploy.yml` in a test and run it against a local fake `/api/health` that injects one transport reset among good responses → green; inject five consecutive failures → red; wrong version → red (mutations pasted in the PR).

Related: #1762 (the `set +e` fix to this step), #1765 (invalidation step), #1714/#1713 (Daniel's rollout-window work — a retry here must not mask the 503 windows those track; print every sample).

dbrowneup:
Closes #1791.

## The bug

Two of the last four `deploy.yml` runs on main ended **red** at *Assert the deployed app actually answers (post-rollout probe)* while the rollout had already succeeded and prod was serving the expected version:

**run 33566302291** (22:39Z, #1765, head `25dc64dc`)
```
probe  1/10: http=200 time=0.4s ... -> ok     (consecutive good: 1)
...
probe  7/10: http=200 time=0.4s ... -> ok     (consecutive good: 7)
probe  8/10: http=000 time=15.0s curl_exit=28 -> NOT-OK (consecutive good: 0)   # client timeout
probe  9/10: http=200 ... -> ok     (consecutive good: 1)
probe 10/10: http=200 ... -> ok     (consecutive good: 2)
::error::post-rollout answer probe: only 2 consecutive trailing samples answered 200 under 5s (needed 5)...
```

**run 33573075995** (00:06Z, #1734, head `7a6c5b73`)
```
probe  1/10 .. 9/10: http=200 time=~0.4s -> ok  (consecutive good: 9)
probe 10/10: http=000 time=0.025s curl_exit=35 -> NOT-OK (consecutive good: 0)  # TLS handshake reset, 25 ms
::error::post-rollout answer probe: only 0 consecutive trailing samples answered 200 under 5s (needed 5)...
```

`/api/health` reported the expected version immediately afterwards in both cases (verified 00:45Z: `version=7a6c5b73…`, task-def 225, 2/2 running). The deploys succeeded; the probe reported them failed. That trains everyone to ignore red deploys, which is the one outcome a probe exists to prevent.

Root cause: the verdict was "the **last 5** samples must be consecutive 200s under 5 s", so one transport-level flake (curl exit 28/35/56, `http=000`) at any of the last five positions failed the run.

## The change

**1. The probe body moves into `.github/scripts/post_rollout_probe.sh`**, and the step calls it with the same env. It was inline, which is why nothing could ever execute it — the two runs above were the first real test of a verdict no test could reach. The step keeps `set +e -uo pipefail` (#1762) and the script sets the same for itself; a test asserts the step invokes the script and does not reimplement a `curl` loop inline, so the two cannot drift.

**2. A request that never got an answer is retried. Anything the app answered is not.**

| what came back | treated as | retried? |
|---|---|---|
| `http=000` — no HTTP answer at all (connect / DNS / TLS reset / client timeout) | transport failure | **yes** — `TRANSPORT_RETRIES=2` after `RETRY_PAUSE=3s` |
| 5xx, any non-200 | NOT-OK immediately | no — this is the app answering (#1714/#1713's rollout-window 503s must stay visible) |
| a 5xx **truncated mid-body** (real status line, `curl_exit=18/56`) | NOT-OK immediately | no — it is still the app answering; see defect 1 below |
| 200 slower than `MAX_SECONDS=5` | NOT-OK immediately | no — this is the #1532 latency failure mode |

The predicate is `[ "$sample_code" = "000" ]` and nothing else. It is deliberately **not** `|| [ "$sample_rc" -ne 0 ]`: both #1791 production shapes report `http=000`, so keying on the status line alone loses nothing real and keeps every 5xx unretried.

**3. The verdict is four conjuncts** — all must hold:
- `NEED_STREAK=5` consecutive good samples were **observed somewhere** in the window (was: in the *last* five positions), **and**
- the **final (post-retry) sample is good**, **and**
- that final sample reports `.version == EXPECTED_VERSION`, **and**
- at most `MAX_NOT_OK=2` samples were NOT-OK after their retries.

The "final sample" conjunct is what keeps this fail-closed: a cold-start tail — slow-then-fast, the #1532 shape this step was built for — still fails, because a window that is broken **at the end** fails no matter how healthy its start was.

`MAX_NOT_OK` is the owner ruling from review. "Observed somewhere" **alone** tolerates `SAMPLES - NEED_STREAK - 1` = **4 of 10** samples being 503s, which is far more than the "one transient" this PR was arguing for. Two NOT-OK samples now pass with a `::warning`; **three go red**, naming the count, even when a 5-streak was observed and the final sample is good.

Every sample and every retry is still printed with `http/time/curl_exit`. A passing run warns when it had any NOT-OK sample **or spent any retry** — a sample that failed and recovered on retry never reaches the NOT-OK tally, and both #1791 red runs are exactly that shape, so it would otherwise be the one degradation the log hid. `::error` lines name exactly which conjunct failed.

**4. `MAX_TOTAL_SECONDS=300` is a real wall-clock ceiling**, not a check between samples: curl's `--max-time` is capped to the budget the probe has left (floored at 1s — `--max-time 0` means *no* timeout), a retry is refused once the budget is gone, and both sleeps are truncated to what remains. deploy.yml's poll step reserves 420s of the deploy-ecs timeout for "the answer probe and the CloudFront invalidation"; overrunning it would turn a clean `::error` into an opaque job timeout *and* eat the invalidation's share. Hitting the ceiling always **fails**: a probe that did not finish measuring cannot report a deploy as verified. Tests assert `DEPLOY_ROLLOUT_BUDGET_SECONDS + MAX_TOTAL_SECONDS < DEPLOY_JOB_TIMEOUT_MINUTES`, that the ceiling is smaller than the poll step's 420s reserve, and — twice — the measured wall clock.

**5. Integer parameters are validated before use.** `set -e` is off by design (#1762), so `[ 3 -gt "two" ]` prints a bash error, returns 2, and reads as **false** — a typo'd `MAX_NOT_OK` would have silently disabled the cap. `SAMPLES`, `NEED_STREAK`, `MAX_NOT_OK`, `TRANSPORT_RETRIES` and `MAX_TOTAL_SECONDS` must be non-negative integers or the probe exits 1 before a single request.

**6. Hermetic test** `backend/tests/test_post_rollout_probe.py` (33 tests, ~56 s) starts a scripted `/api/health` on `127.0.0.1` (`http.server` in a thread) that can drop the connection without a response, answer 503, answer slowly, or answer with the wrong commit — plus a **raw socket** that answers a real `503` and hangs up mid-body — and runs **the exact script CI runs** with `INTERVAL`/`RETRY_PAUSE` at 0.1 s. No AWS, no network, no CloudFront.

## Guards shown red on the input they reject

Every block below is real output from this branch, trimmed only where marked `…`. Mutations A-D are the shipped verdict; E-G are one per review defect.

### Mutation A — remove the transport retry (`if is_transport_failure && [ "$attempt" -lt "$TRANSPORT_RETRIES" ]` → `if false`)

Input: run 33573075995's shape — nine good samples, the connection dropped on sample 10 of 10.

Baseline (correct — the flake is retried, the deploy is reported green, and the retry is warned on):
```
probe  9/10: http=200 time=0.000878s curl_exit=0 -> ok     (consecutive good: 9, best so far: 9)
probe 10/10: http=000 time=0.000621s curl_exit=52 -> transport failure (curl: (52) Empty reply from server) — retrying in 0.1s
probe 10/10 retry 1/2: http=200 time=0.000729s curl_exit=0 -> ok     (consecutive good: 10, best so far: 10)
::warning::post-rollout probe passed with 0 of 10 samples NOT-OK (cap MAX_NOT_OK=2) and 1 transport retry attempt(s) spent. …
post-rollout probe OK: longest good streak 10/10 (needed 5), 0 NOT-OK (cap 2), 1 transport retry attempt(s), final sample 200 in 0.000729s under 5s, serving version 7a6c5b73…
$ echo $?
0
```

Mutated (red — this is the bug being fixed):
```
probe  9/10: http=200 time=0.000799s curl_exit=0 -> ok     (consecutive good: 9, best so far: 9)
probe 10/10: http=000 time=0.001097s curl_exit=52 -> NOT-OK (consecutive good: 0, best so far: 9)
::error::post-rollout answer probe: the FINAL sample taken (10/10, after up to 2 transport retries) did not answer 200 under 5s — http=000 time=0.001097s curl_exit=52. …
post-rollout version check: not attempted — the final sample did not answer (see the error above).
$ echo $?
1
```
```
FAILED backend/tests/test_post_rollout_probe.py::TestTransportRetry::test_one_transport_reset_mid_window_is_retried_and_passes
FAILED backend/tests/test_post_rollout_probe.py::TestTransportRetry::test_a_sample_that_recovered_on_retry_is_still_annotated
FAILED backend/tests/test_post_rollout_probe.py::TestTransportRetry::test_transport_reset_on_the_final_sample_is_retried_and_passes
FAILED backend/tests/test_post_rollout_probe.py::TestTransportRetry::test_mid_window_outage_burst_passes_on_an_observed_streak_and_a_good_final_sample
FAILED backend/tests/test_post_rollout_probe.py::TestTransportRetry::test_total_transport_outage_fails
FAILED backend/tests/test_post_rollout_probe.py::TestVersionAssertion::test_version_is_read_from_the_final_sample_not_an_extra_request
6 failed, 27 passed, 1 warning in 51.94s          # pytest exit 1
```

### Mutation B — verdict back to the trailing streak (`[ "$best_streak" -lt "$NEED_STREAK" ]` → `[ "$streak" -lt … ]`)

Input: sample 6 fails through both retries; samples 1-5 and 7-10 are good, so the trailing run is only 4.

Baseline (correct — an observed 5-streak plus a good final sample; the transient is logged loudly and tolerated, 1 NOT-OK is inside the cap):
```
probe  5/10: http=200 time=0.001778s curl_exit=0 -> ok     (consecutive good: 5, best so far: 5)
probe  6/10: http=000 time=0.000642s curl_exit=52 -> transport failure (curl: (52) Empty reply from server) — retrying in 0.1s
probe  6/10 retry 1/2: http=000 time=0.000654s curl_exit=52 -> transport failure (curl: (52) Empty reply from server) — retrying in 0.1s
probe  6/10 retry 2/2: http=000 time=0.000642s curl_exit=52 -> NOT-OK (consecutive good: 0, best so far: 5)
probe  7/10 … 10/10: http=200 -> ok     (consecutive good: 4, best so far: 5)
::warning::post-rollout probe passed with 1 of 10 samples NOT-OK (cap MAX_NOT_OK=2) and 2 transport retry attempt(s) spent. …
post-rollout probe OK: longest good streak 5/10 (needed 5), 1 NOT-OK (cap 2), 2 transport retry attempt(s), final sample 200 in 0.000763s under 5s, serving version 7a6c5b73…
$ echo $?
0
```

Mutated (red):
```
probe 10/10: http=200 time=0.000742s curl_exit=0 -> ok     (consecutive good: 4, best so far: 5)
::error::post-rollout answer probe: the longest run of consecutive good samples anywhere in the window was 5 (needed 5) out of 10. …
$ echo $?
1
```
(The message still interpolates `best_streak`, hence the odd "was 5 (needed 5)" — an artifact of the mutation, not of the shipped script.)
```
FAILED …::TestTransportRetry::test_mid_window_outage_burst_passes_on_an_observed_streak_and_a_good_final_sample
FAILED …::TestNotOkCap::test_two_not_ok_samples_with_a_streak_and_a_good_final_sample_pass_with_a_warning
FAILED …::TestNotOkCap::test_three_not_ok_samples_fail_even_with_a_streak_and_a_good_final_sample
FAILED …::TestBudgetsAndMisconfiguration::test_probe_stops_at_its_wall_clock_ceiling_and_fails
4 failed, 29 passed, 1 warning in 57.06s          # pytest exit 1
```

### Mutation C — drop the version comparison (`elif [ "$version" != "$EXPECTED_VERSION" ]` → `elif false`)

Input: a perfectly healthy window where the app reports a different commit.

Baseline (correct — red):
```
probe 10/10: http=200 time=0.000766s curl_exit=0 -> ok     (consecutive good: 10, best so far: 10)
::error::post-rollout version check: …/api/health reports version=25dc64dc… but this job deployed 7a6c5b73…. The rollout completed and the app answers — but it is answering with different code than this run built, so this deploy did not take.
$ echo $?
1
```

Mutated (a deploy that did not take is reported green — exactly what the guard exists to stop):
```
probe 10/10: http=200 time=0.001842s curl_exit=0 -> ok     (consecutive good: 10, best so far: 10)
post-rollout probe OK: longest good streak 10/10 (needed 5), 0 NOT-OK (cap 2), 0 transport retry attempt(s), final sample 200 in 0.001842s under 5s, serving version 25dc64dc…
$ echo $?
0
```
```
FAILED …::TestVersionAssertion::test_wrong_version_fails_and_the_error_names_both_versions
FAILED …::TestVersionAssertion::test_version_is_read_from_the_final_sample_not_an_extra_request
2 failed, 31 passed, 1 warning in 57.72s          # pytest exit 1
```

### Mutation D — remove the NOT-OK cap (`if [ "$not_ok" -gt "$MAX_NOT_OK" ]` → `if false`)

Input: samples 1-5 good, 6-8 = 503, 9-10 good — a 5-streak *was* observed and the final sample *is* good, so the other three conjuncts all hold.

Baseline (correct — three NOT-OK samples is a degraded window, not a transient):
```
probe  6/10: http=503 time=0.000771s curl_exit=0 -> NOT-OK (consecutive good: 0, best so far: 5)
probe  7/10: http=503 time=0.000790s curl_exit=0 -> NOT-OK (consecutive good: 0, best so far: 5)
probe  8/10: http=503 time=0.000831s curl_exit=0 -> NOT-OK (consecutive good: 0, best so far: 5)
probe  9/10: http=200 time=0.000861s curl_exit=0 -> ok     (consecutive good: 1, best so far: 5)
probe 10/10: http=200 time=0.000798s curl_exit=0 -> ok     (consecutive good: 2, best so far: 5)
::error::post-rollout answer probe: 3 of the 10 samples were NOT-OK after retries (MAX_NOT_OK=2). A 5-sample good run was allowed to be observed ANYWHERE in the window (#1791) so that one transient cannot fail a healthy deploy — this cap is what keeps that from tolerating a pockmarked one. …
$ echo $?
1
```

Mutated (the pockmarked window passes — the tolerance the owner ruled against):
```
probe  8/10: http=503 time=0.001217s curl_exit=0 -> NOT-OK (consecutive good: 0, best so far: 5)
probe 10/10: http=200 time=0.001051s curl_exit=0 -> ok     (consecutive good: 2, best so far: 5)
::warning::post-rollout probe passed with 3 of 10 samples NOT-OK (cap MAX_NOT_OK=2) and 0 transport retry attempt(s) spent. …
post-rollout probe OK: longest good streak 5/10 (needed 5), 3 NOT-OK (cap 2), 0 transport retry attempt(s), final sample 200 in 0.001051s under 5s, serving version 7a6c5b73…
$ echo $?
0
```
```
FAILED …::TestNotOkCap::test_three_not_ok_samples_fail_even_with_a_streak_and_a_good_final_sample
FAILED …::TestNotOkCap::test_the_cap_can_be_tightened_to_zero
2 failed, 31 passed, 1 warning in 60.40s          # pytest exit 1
```

### Mutation E — restore `|| [ "$sample_rc" -ne 0 ]` (review defect 1)

Input: a raw socket that answers `503` with `Content-Length: 100` and hangs up after 10 bytes — curl reports `http=503` **and** exit 18. `SAMPLES=3`.

Baseline (correct — a truncated 5xx is still the app answering; 3 requests for 3 samples, no retry):
```
[requests the origin received: 3]
probe  1/3: http=503 time=0.000732s curl_exit=18 -> NOT-OK (consecutive good: 0, best so far: 0)
probe  2/3: http=503 time=0.000691s curl_exit=18 -> NOT-OK (consecutive good: 0, best so far: 0)
probe  3/3: http=503 time=0.000758s curl_exit=18 -> NOT-OK (consecutive good: 0, best so far: 0)
::error::… the longest run of consecutive good samples anywhere in the window was 0 (needed 2) out of 3. …
::error::… 3 of the 3 samples were NOT-OK after retries (MAX_NOT_OK=2). …
::error::… the FINAL sample taken (3/3, after up to 2 transport retries) did not answer 200 under 5s — http=503 time=0.000758s curl_exit=18. …
$ echo $?
1
```

Mutated (the shipped-in-b78ab9db bug: a 503 is retried three times over — 9 requests for 3 samples):
```
[requests the origin received: 9]
probe  1/3: http=503 time=0.000728s curl_exit=18 -> transport failure (curl: (18) transfer closed with 90 bytes remaining to read) — retrying in 0.1s
probe  1/3 retry 1/2: http=503 time=0.000723s curl_exit=18 -> transport failure (curl: (18) transfer closed with 90 bytes remaining to read) — retrying in 0.1s
probe  1/3 retry 2/2: http=503 time=0.000602s curl_exit=18 -> NOT-OK (consecutive good: 0, best so far: 0)
…
$ echo $?
1
```
```
FAILED …::TestAppLevelFailuresAreNotRetried::test_a_truncated_5xx_is_not_retried
1 failed, 32 passed, 1 warning in 57.41s          # pytest exit 1
```
(The mutated run still exits 1 here because *every* attempt fails. It fails open the moment one retry lands a 200: the 503 then leaves the verdict **and** the NOT-OK tally.)

### Mutation F — ceiling back to a between-samples check (review defect 2)

Input: every connection dropped, `MAX_TOTAL_SECONDS=4`, `RETRY_PAUSE=4`, `INTERVAL=4`.

Baseline (correct — the retry is refused at the ceiling, the sleeps are truncated, wall clock ≈ the ceiling):
```
probe  1/10: http=000 time=0.001510s curl_exit=52 -> transport failure (curl: (52) Empty reply from server) — retrying in 4s
probe  1/10 retry 1/2: transport failure (curl: (52) Empty reply from server) — NOT retried: the probe has used all 4s of its wall-clock ceiling; the sample counts as it stands
probe  1/10 retry 1/2: http=000 time=0.003797s curl_exit=52 -> NOT-OK (consecutive good: 0, best so far: 0)
probe 2/10: not taken — the probe has already run 4s of its 4s wall-clock ceiling.
::error::post-rollout answer probe: ran out of its 4s wall-clock budget after 1 of 10 samples. …
$ echo $?
1
[wall clock 4.7s]
```

Mutated (3.2× the "ceiling" — the arithmetic deploy.yml asserts against becomes fiction):
```
probe  1/10 retry 1/2: http=000 time=0.005582s curl_exit=52 -> transport failure … — retrying in 4s
probe  1/10 retry 2/2: http=000 time=0.002191s curl_exit=52 -> NOT-OK (consecutive good: 0, best so far: 0)
probe 2/10: not taken — the probe has already run 12s of its 4s wall-clock ceiling.
$ echo $?
1
[wall clock 12.7s]
```
```
FAILED …::TestBudgetsAndMisconfiguration::test_the_ceiling_bounds_the_whole_probe_not_just_the_gap_between_samples
FAILED …::TestBudgetsAndMisconfiguration::test_a_single_slow_request_cannot_run_past_the_ceiling
2 failed, 31 passed, 1 warning in 64.67s          # pytest exit 1
```

### Mutation G — warn on `not_ok` alone (review defect 3)

Input: sample 7 drops the connection and recovers on its first retry — the modal shape of both #1791 red runs.

Baseline (correct — the run passes, and the retry is on the record):
```
probe  7/10: http=000 time=0.000673s curl_exit=52 -> transport failure (curl: (52) Empty reply from server) — retrying in 0.1s
probe  7/10 retry 1/2: http=200 time=0.000965s curl_exit=0 -> ok     (consecutive good: 7, best so far: 7)
::warning::post-rollout probe passed with 0 of 10 samples NOT-OK (cap MAX_NOT_OK=2) and 1 transport retry attempt(s) spent. …
post-rollout probe OK: longest good streak 10/10 (needed 5), 0 NOT-OK (cap 2), 1 transport retry attempt(s), final sample 200 in 0.000725s under 5s, serving version 7a6c5b73…
$ echo $?
0
```

Mutated (the origin hung past the client deadline and the summary says `10/10` with no warning at all):
```
probe  7/10 retry 1/2: http=200 time=0.000773s curl_exit=0 -> ok     (consecutive good: 7, best so far: 7)
post-rollout probe OK: longest good streak 10/10 (needed 5), 0 NOT-OK (cap 2), 1 transport retry attempt(s), final sample 200 in 0.000809s under 5s, serving version 7a6c5b73…
$ echo $?
0
```
```
FAILED …::TestTransportRetry::test_a_sample_that_recovered_on_retry_is_still_annotated
1 failed, 32 passed, 1 warning in 56.18s          # pytest exit 1
```

### Fail-closed cases that are green in this PR

| input | result |
|---|---|
| every sample and retry dropped (30 requests) | exit 1 — no streak **and** bad final sample **and** over the cap |
| healthy for 5 samples, then broken from sample 6 to the end | exit 1 — a 5-streak *was* observed; the final sample still decides |
| one 503 mid-window that leaves no 5-streak (6-sample window) | exit 1 — `was 3 (needed 5)` |
| 3 of 10 samples 503 with a 5-streak and a good final sample | exit 1 — `3 of the 10 samples were NOT-OK after retries (MAX_NOT_OK=2)` |
| 2 of 10 samples 503 with a 5-streak and a good final sample | exit 0 + `::warning` — the tolerated case, named in the log |
| 503 on the final sample | exit 1, and **no retry burned** (10 requests) |
| a 503 truncated mid-body on every sample | exit 1, **no retry burned** (3 requests for 3 samples) |
| 200 in 1.6 s against `MAX_SECONDS=1` on the final sample | exit 1, **not retried** (10 requests) |
| wall-clock ceiling hit while samples were good | exit 1 — an unfinished probe verifies nothing |
| `MAX_NOT_OK=two` (or a non-integer `SAMPLES`/`NEED_STREAK`/`TRANSPORT_RETRIES`/`MAX_TOTAL_SECONDS`) | exit 1 before a single request — a garbage cap must not silently disable itself |
| `NEED_STREAK > SAMPLES` | exit 1 before a single request |
| `HEALTH_URL`/`EXPECTED_VERSION` unset | exit 1 before a single request |

## Verification

Run on macOS with `bash 3.2.57` (the script uses no bash-4 construct — no `[[`, arrays, `<<<`, `mapfile` or `date -d`).

```
$ pytest -p no:cacheprovider backend/tests/test_post_rollout_probe.py
33 passed, 1 warning in 55.72s                    # exit 0

$ pytest -p no:cacheprovider backend/tests/test_cloudwatch_alarms.py backend/tests/test_single_openssl.py \
    backend/tests/test_ecs_paper_advance_deploy_pin.py backend/tests/scripts/test_alembic_migrate_preflight.py \
    backend/tests/services/test_paper_advance_kill_switch.py
97 passed, 2 skipped, 1 warning in 2.40s          # exit 0 — every existing test that reads deploy.yml

$ pytest -p no:cacheprovider backend/tests/test_feature_flag_fliplist_drift.py backend/tests/test_docs_site.py \
    backend/tests/test_infra_gate_workflow.py backend/tests/test_local_mode_contract.py
69 passed, 1 warning in 3.74s                     # exit 0 — every other test that reads .github/workflows

$ ruff check backend/tests/test_post_rollout_probe.py            # exit 0
$ ruff format --check backend/tests/test_post_rollout_probe.py   # exit 0
$ bash -n .github/scripts/post_rollout_probe.sh                  # exit 0 (also asserted as a test)
```

**`actionlint` is not available on this machine, so it was not run** — flagging that rather than claiming it, and `shellcheck` is not installed either. Nothing in `.github/workflows/` runs either tool and `main-format-guard` is ruff-only, so this shell script has no linter in CI at all. The workflow is parsed with PyYAML and the probe step is located and asserted through the parsed YAML (name, `if`, `env`, `run`) rather than by text grep.

## Concerns for the reviewer

- **The tolerance is real, and now bounded twice.** Up to two NOT-OK samples — including two dropped connections that failed through both retries — pass if a 5-streak was observed and the final sample is good. That is deliberate (#1791's whole complaint). It is bounded by the final-sample conjunct and by `MAX_NOT_OK`, and every tolerated sample is a `::warning`, not a silent pass. If those warnings become routine, that is a finding about post-rollout availability, not a reason to loosen further.
- **`MAX_NOT_OK` counts what a sample *answered with*, not what version it reported.** `.version` is deliberately read only from the final sample (no extra request after the window), and a mismatch fails the run outright — it can never be "capped away". I did **not** add a per-sample version check: during ALB deregistration draining an old task can still legitimately answer a sample after `rolloutState=COMPLETED`, and counting that NOT-OK would newly red-flag healthy deploys. Say the word if you want the stricter reading.
- **`MAX_TOTAL_SECONDS=300` is new policy**, chosen to fit the 1200s rollout budget + the probe inside the 30-minute deploy-ecs timeout with room for the CloudFront invalidation. Tests pin the arithmetic (including against the poll step's 420s reserve); the number itself is a judgement call.
- **This adds ~56s of mostly-sleep to `quality-gate`'s hard-blocking `backend-tests` job** (no xdist in the plugin set). The sleeps are `INTERVAL`/`RETRY_PAUSE` at 0.1s plus three deliberately slow cases.
- The script depends on `jq`, `awk`, `curl`, `seq`, `date` — all present on `ubuntu-latest` and used by neighbouring steps already. Missing `jq` degrades to "could not read .version" → exit 1 (fail-closed).

### DIFF
--- a/.github/scripts/post_rollout_probe.sh
+++ b/.github/scripts/post_rollout_probe.sh
@@ -22,25 +22,35 @@
 #   Both were edge/TLS transients, not the app. A probe that cries wolf
 #   trains everyone to ignore red deploys, which is the one outcome a probe
 #   exists to prevent. So:
-#     * TRANSPORT-level failures (http=000, or curl exit != 0 — timeout,
-#       reset, TLS handshake) are RETRIED up to TRANSPORT_RETRIES times
+#     * TRANSPORT-level failures — NO HTTP ANSWER CAME BACK AT ALL, which
+#       curl reports as http_code 000 (connect refused, DNS, TLS handshake
+#       reset, client timeout) — are RETRIED up to TRANSPORT_RETRIES times
 #       after RETRY_PAUSE before they count. A real outage still fails,
 #       because the retries fail too.
-#     * A 5xx, any non-200, or a 200 slower than MAX_SECONDS counts
-#       IMMEDIATELY as NOT-OK with no retry — those are the app answering,
-#       and #1714/#1713's rollout-window 503s must not be retried away.
+#     * ANYTHING THE APP ANSWERED counts IMMEDIATELY as NOT-OK with no
+#       retry: a 5xx, any non-200, or a 200 slower than MAX_SECONDS. Those
+#       are the app talking, and #1714/#1713's rollout-window 503s must not
+#       be retried away.
 #     * The verdict is "a run of NEED_STREAK consecutive good samples was
 #       OBSERVED SOMEWHERE in the window, AND the final (post-retry) sample
-#       is good, AND that final sample reports EXPECTED_VERSION".
+#       is good, AND that final sample reports EXPECTED_VERSION, AND no more
+#       than MAX_NOT_OK samples were NOT-OK".
 #       "Observed somewhere" is what stops one mid-window transient from
 #       failing an otherwise healthy deploy; "final sample is good" is what
 #       keeps it fail-closed — a run that is broken AT THE END fails no
 #       matter how healthy its start was, so a cold-start tail (the #1532
-#       failure mode this probe was built for) still goes red.
+#       failure mode this probe was built for) still goes red; and
+#       MAX_NOT_OK is what stops "observed somewhere" from tolerating a
+#       pockmarked window. Without that cap the streak rule alone would pass
+#       a window with SAMPLES - NEED_STREAK - 1 = 4 bad samples out of 10
+#       (owner ruling, #1791 review): a transient is one or two samples, a
+#       third is a degraded deploy and goes red.
 #   Every sample and every retry is printed with http/time/curl_exit: the
 #   timings ARE the evidence (#1532 had to be diagnosed by hand-curling
 #   because the pipeline kept none), and a degradation that this verdict
-#   tolerates must still be visible in the log.
+#   tolerates must still be visible in the log — including a sample that
+#   FAILED AND THEN RECOVERED ON RETRY, which is counted and warned on
+#   separately (it does not appear in the NOT-OK tally, because it recovered).
 #
 # Parameters (all via environment):
 #   HEALTH_URL         required — probed URL (deploy.yml uses /api/health
@@ -49,21 +59,29 @@
 #   EXPECTED_VERSION   required — the commit this run deployed
 #   SAMPLES            default 10
 #   NEED_STREAK        default 5
+#   MAX_NOT_OK         default 2   — most NOT-OK samples (after retries) a
+#                                    passing run may contain. 3+ is a
+#                                    degraded window, not a transient.
 #   MAX_SECONDS        default 5   — client-realistic deadline, deliberately
 #                                    below #1436's recorded p99 of 16.8s
 #   INTERVAL           default 3   — seconds between samples
 #   TRANSPORT_RETRIES  default 2   — retries per sample, transport failures only
 #   RETRY_PAUSE        default 3   — seconds before each retry
 #   CURL_MAX_TIME      default 15  — per-request curl --max-time
-#   MAX_TOTAL_SECONDS  default 300 — wall-clock ceiling for the whole probe.
-#                      Retries widen the worst case (3 attempts x
-#                      CURL_MAX_TIME + pauses per sample), and deploy.yml's
-#                      poll step asserts the deploy-ecs job has headroom for
-#                      "the answer probe and the CloudFront invalidation".
-#                      Blowing through that headroom would turn a clean
-#                      ::error into an opaque job timeout, so the probe stops
-#                      itself first — and stopping early always FAILS, it can
-#                      never pass a deploy it did not finish measuring.
+#   MAX_TOTAL_SECONDS  default 300 — REAL wall-clock ceiling for the whole
+#                      probe, not a between-samples check: no request is
+#                      started with a deadline that reaches past it (curl's
+#                      --max-time is capped to the remaining budget) and no
+#                      sleep runs past it. The only possible overshoot is
+#                      sub-second rounding plus the 1s floor this keeps on
+#                      --max-time (`curl --max-time 0` means NO timeout).
+#                      It matters because deploy.yml's poll step reserves 7
+#                      minutes of the deploy-ecs job timeout for "the answer
+#                      probe and the CloudFront invalidation": a probe that
+#                      overran that reserve would turn a clean ::error into
+#                      an opaque job timeout AND eat the invalidation's
+#                      share of it. Stopping early always FAILS — the probe
+#                      can never pass a deploy it did not finish measuring.
 #
 # Exit 0 = the app answers. Exit 1 = it does not, or it is answering with
 # code this run did not build.
@@ -81,6 +99,7 @@ HEALTH_URL="${HEALTH_URL:-}"
 EXPECTED_VERSION="${EXPECTED_VERSION:-}"
 SAMPLES="${SAMPLES:-10}"
 NEED_STREAK="${NEED_STREAK:-5}"
+MAX_NOT_OK="${MAX_NOT_OK:-2}"
 MAX_SECONDS="${MAX_SECONDS:-5}"
 INTERVAL="${INTERVAL:-3}"
 TRANSPORT_RETRIES="${TRANSPORT_RETRIES:-2}"
@@ -92,6 +111,21 @@ if [ -z "$HEALTH_URL" ] || [ -z "$EXPECTED_VERSION" ]; then
   echo "::error::post-rollout probe: HEALTH_URL and EXPECTED_VERSION are both required (got HEALTH_URL='${HEALTH_URL}', EXPECTED_VERSION='${EXPECTED_VERSION}'). Refusing to report a deploy as verified without probing it."
   exit 1
 fi
+
+# Every parameter used in integer arithmetic or an integer test is validated
+# BEFORE it is used. `set -e` is off by design (see above), so a `[ "$a" -gt
+# "$b" ]` against a non-integer would print a bash error, return 2, and be
+# read as FALSE — i.e. a garbage MAX_NOT_OK would silently DISABLE the cap
+# and pass a window it must fail. Fail-closed on the misconfiguration instead.
+for _int_param in SAMPLES NEED_STREAK MAX_NOT_OK TRANSPORT_RETRIES MAX_TOTAL_SECONDS; do
+  eval "_int_value=\${${_int_param}}"
+  case "$_int_value" in
+    '' | *[!0-9]*)
+      echo "::error::post-rollout probe misconfigured: ${_int_param}='${_int_value}' is not a non-negative integer. Refusing to run with a parameter that would silently disable a check."
+      exit 1
+      ;;
+  esac
+done
 if [ "$NEED_STREAK" -gt "$SAMPLES" ]; then
   echo "::error::post-rollout probe misconfigured: NEED_STREAK=${NEED_STREAK} exceeds SAMPLES=${SAMPLES}, so no run of samples could ever satisfy it and this step could never pass."
   exit 1
@@ -104,16 +138,43 @@ trap 'rm -f "$BODY_FILE" "$ERR_FILE"' EXIT
 sample_code=""
 sample_secs=""
 sample_rc=0
+started_at="$(date +%s)"
+
+# Whole seconds of MAX_TOTAL_SECONDS left, never negative. Every piece of work
+# the probe starts — request or sleep — is bounded by this, which is what makes
+# MAX_TOTAL_SECONDS a ceiling rather than a suggestion.
+remaining_budget() {
+  local left
+  left=$(( MAX_TOTAL_SECONDS - ( $(date +%s) - started_at ) ))
+  [ "$left" -gt 0 ] || left=0
+  printf '%s' "$left"
+}
+
+# Sleep, truncated to the remaining budget (and skipped entirely when there is
+# none). RETRY_PAUSE/INTERVAL may be fractional, hence awk rather than `[ -lt ]`.
+capped_sleep() {
+  local want left
+  want="$1"
+  left="$(remaining_budget)"
+  [ "$left" -gt 0 ] || return 0
+  sleep "$(awk -v w="$want" -v l="$left" 'BEGIN { print (w + 0 < l + 0) ? w + 0 : l + 0 }')"
+}
 
 # Takes one sample. The body of the LAST request made lands in $BODY_FILE, so
 # the version assertion below is made against the very response that closed
 # the window rather than an extra request taken afterwards.
 take_sample() {
-  local out
+  local out max_time
+  # Per-request deadline: CURL_MAX_TIME, but never more wall clock than the
+  # probe has left. Floored at 1 because `curl --max-time 0` means NO timeout —
+  # the exact opposite of a cap (the caller does not start a sample with zero
+  # budget left, so this floor is only ever the sub-second rounding tail).
+  max_time="$(awk -v c="$CURL_MAX_TIME" -v l="$(remaining_budget)" \
+    'BEGIN { m = (c + 0 < l + 0) ? c + 0 : l + 0; if (m < 1) m = 1; print m }')"
   # Declared and assigned separately on purpose: `local out="$(curl ...)"`
   # would make $? the exit status of `local`, not of curl.
   out="$(curl -sS -o "$BODY_FILE" -w '%{http_code} %{time_total}' \
-    --max-time "$CURL_MAX_TIME" "$HEALTH_URL" 2>"$ERR_FILE")"
+    --max-time "$max_time" "$HEALTH_URL" 2>"$ERR_FILE")"
   sample_rc=$?
   # curl writes the -w line even on timeout/connect failure (http_code 000);
   # this only backfills the case where it wrote nothing at all, so an empty
@@ -123,10 +184,20 @@ take_sample() {
   sample_secs="${out##* }"
 }
 
-# Transport level: no HTTP answer came back at all. http=000 covers connect /
-# TLS / timeout failures; a non-zero curl exit covers the rest.
+# Transport level: NO HTTP ANSWER CAME BACK AT ALL. http_code 000 is curl's
+# report for connect / DNS / TLS / client-timeout failures, and it is the only
+# thing retried here.
+#
+# Deliberately NOT `|| [ "$sample_rc" -ne 0 ]` (#1791 review, defect 1): curl
+# also exits non-zero on a TRUNCATED answer that DID carry a real status line —
+# exit 18 "transfer closed with N bytes remaining", exit 56 "recv failure" —
+# and an origin that 503s while dropping the body under load is exactly the
+# #1714/#1713 rollout-window shape. Retrying those retries a 5xx, which is the
+# one thing this probe promises never to do. Both #1791 production shapes
+# (curl 28 client timeout, curl 35 TLS reset) report http=000, so nothing real
+# is lost by keying on the status line alone.
 is_transport_failure() {
-  [ "$sample_code" = "000" ] || [ "$sample_rc" -ne 0 ]
+  [ "$sample_code" = "000" ]
 }
 
 # Good: the app answered 200 inside the client deadline.
@@ -145,12 +216,12 @@ curl_error_text() {
 streak=0
 best_streak=0
 not_ok=0
+retried=0
 taken=0
 final_good=0
 deadline_hit=0
-started_at="$(date +%s)"
 
-echo "Probing $HEALTH_URL — ${SAMPLES} samples every ${INTERVAL}s; need ${NEED_STREAK} consecutive good (200 under ${MAX_SECONDS}s) anywhere in the window AND a good final sample reporting version ${EXPECTED_VERSION}. Transport-level failures are retried ${TRANSPORT_RETRIES}x after ${RETRY_PAUSE}s; 5xx and slow 200s are never retried."
+echo "Probing $HEALTH_URL — ${SAMPLES} samples every ${INTERVAL}s; need ${NEED_STREAK} consecutive good (200 under ${MAX_SECONDS}s) anywhere in the window AND a good final sample reporting version ${EXPECTED_VERSION} AND at most ${MAX_NOT_OK} NOT-OK samples. ONLY transport-level failures (no HTTP answer at all, http=000) are retried, ${TRANSPORT_RETRIES}x after ${RETRY_PAUSE}s; anything the app answered — a 5xx, any non-200, a 200 slower than ${MAX_SECONDS}s — counts immediately and is NEVER retried. Whole-probe wall-clock ceiling ${MAX_TOTAL_SECONDS}s; hitting it fails the run."
 
 for i in $(seq 1 "$SAMPLES"); do
   elapsed=$(( $(date +%s) - started_at ))
@@ -168,10 +239,19 @@ for i in $(seq 1 "$SAMPLES"); do
       prefix="$(printf 'probe %2d/%s retry %d/%s:' "$i" "$SAMPLES" "$attempt" "$TRANSPORT_RETRIES")"
     fi
     if is_transport_failure && [ "$attempt" -lt "$TRANSPORT_RETRIES" ]; then
+      if [ "$(remaining_budget)" -le 0 ]; then
+        # The retry is work, and work stops at the ceiling. The sample counts
+        # as it stands (NOT-OK), and the run fails on the ceiling regardless.
+        printf '%s transport failure (%s) — NOT retried: the probe has used all %ss of its wall-clock ceiling; the sample counts as it stands\n' \
+          "$prefix" "$(curl_error_text)" "$MAX_TOTAL_SECONDS"
+        deadline_hit=1
+        break
+      fi
       printf '%s http=%s time=%ss curl_exit=%s -> transport failure (%s) — retrying in %ss\n' \
         "$prefix" "$sample_code" "$sample_secs" "$sample_rc" "$(curl_error_text)" "$RETRY_PAUSE"
       attempt=$(( attempt + 1 ))
-      sleep "$RETRY_PAUSE"
+      retried=$(( retried + 1 ))
+      capped_sleep "$RETRY_PAUSE"
       continue
     fi
     break
@@ -195,7 +275,7 @@ for i in $(seq 1 "$SAMPLES"); do
     "$prefix" "$sample_code" "$sample_secs" "$sample_rc" "$verdict" "$streak" "$best_streak"
 
   if [ "$i" -lt "$SAMPLES" ]; then
-    sleep "$INTERVAL"
+    capped_sleep "$INTERVAL"
   fi
 done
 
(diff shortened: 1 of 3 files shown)


dbrowneup:
Review fixes pushed as `dec7ea8e` — all three defects, plus the owner ruling on the tolerance (item 4). Nothing merged.

### Defect 1 — `is_transport_failure` retried 5xx (`post_rollout_probe.sh`)

The predicate is now `[ "$sample_code" = "000" ]` and nothing else; the `|| [ "$sample_rc" -ne 0 ]` half is gone, with the reason written above it. Both #1791 production shapes (curl 28 client timeout, curl 35 TLS reset) report `http=000`, so keying on the status line loses nothing real.

Guard test added (`test_a_truncated_5xx_is_not_retried`): a **raw socket** answers `HTTP/1.1 503` with `Content-Length: 100` and hangs up after 10 bytes, so curl reports `http=503` with exit 18 — `http.server` cannot produce this shape. Result: exactly 3 requests for 3 samples, no `retry N/` line, exit 1.

Adversarial pass — restoring `|| [ "$sample_rc" -ne 0 ]` (mutation E) makes it red, and reproduces the defect exactly:
```
[requests the origin received: 9]
probe  1/3: http=503 … curl_exit=18 -> transport failure (curl: (18) transfer closed with 90 bytes remaining to read) — retrying in 0.1s
probe  1/3 retry 1/2: http=503 … -> transport failure … — retrying in 0.1s
FAILED …::TestAppLevelFailuresAreNotRetried::test_a_truncated_5xx_is_not_retried
1 failed, 32 passed          # pytest exit 1
```

Every sentence that claimed 5xx are never retried is now true, and all of them were fixed: the script header, the runtime banner, the deploy.yml step comment (which now names the truncated-5xx case explicitly), the PR body's retry table (a fourth row for it), and the test that asserted it — that assertion was also a whole-log `"retry" not in stdout` substring check, which is now `re.search(r"retry \d+/", …)` so a reworded banner cannot turn a real guard into a false red (your nit 5).

### Defect 2 — `MAX_TOTAL_SECONDS` was not a ceiling

Taken as "make the number true", not "correct the claim":

- curl's `--max-time` is capped to the remaining budget, floored at 1s (`--max-time 0` means **no** timeout — worth knowing);
- a retry is refused once the budget is gone, with its own log line, and the sample counts as it stands;
- both `sleep`s are truncated to what is left (awk, since `RETRY_PAUSE`/`INTERVAL` may be fractional).

The only overshoot left is sub-second rounding plus that 1s floor. Measured on your scaled-down settings (`MAX_TOTAL_SECONDS=4 RETRY_PAUSE=4 INTERVAL=4`): **12.7s → 4.7s**, and reverting to the between-samples check (mutation F) puts it back to 12.7s and turns two tests red. A second test drives a single `slow:8` response against a 3s ceiling with `CURL_MAX_TIME=20`: 3.6s wall clock, because curl's own deadline was capped.

Three assertions instead of the one that encoded the understated 300: the job-timeout fit, the measured wall clock (twice), and — new — `probe_ceiling < the 420s the poll step reserves`, so the "reserve 7 min for the probe + the invalidation" comment is now checked rather than trusted. deploy.yml's two comments say "a REAL ceiling, not a check between samples" and name what the reserve is split into.

### Defect 3 — a retried-and-recovered sample was invisible

`retried` is counted, the `::warning` fires on `not_ok > 0 || retried > 0`, and both counts appear in the summary line (`… 0 NOT-OK (cap 2), 1 transport retry attempt(s) …`). Both #1791 red runs are that shape, so it was the *modal* tolerated degradation passing as "10/10, no warnings". Gating the warning on `not_ok` alone (mutation G) is red.

### Owner ruling on item 4 — `MAX_NOT_OK`

Tightened as instructed. A run passes only if **best streak ≥ NEED_STREAK AND the final post-retry sample is good AND it reports this commit AND `not_ok ≤ MAX_NOT_OK` (default 2)**. Your 4-of-10 window now exits 1.

- 2 NOT-OK + a 5-streak + a good final sample → **exit 0** with `::warning::post-rollout probe passed with 2 of 10 samples NOT-OK (cap MAX_NOT_OK=2)…`
- 3 NOT-OK + a 5-streak + a good final sample → **exit 1**: `::error::… 3 of the 10 samples were NOT-OK after retries (MAX_NOT_OK=2). …`
- Mutation D (cap removed) → that same 3-NOT-OK window exits **0**, and two tests go red.

`MAX_NOT_OK` is in the env parameter block, the runtime banner, the deploy.yml step comment (asserted by the workflow-wiring test) and the PR body table. One thing it does *not* fold in: **wrong-version**. `.version` is read only from the final sample by design (no extra request after the window) and a mismatch fails the run outright, so it can never be capped away — but I did not add a per-sample version check, because during ALB deregistration draining an old task can legitimately answer a sample after `rolloutState=COMPLETED` and counting that NOT-OK would newly red-flag healthy deploys. Say the word if you want the stricter reading.

While adding the cap I also closed a fail-open next to it: `set -e` is off by design (#1762), so `[ 3 -gt "two" ]` returns 2 and reads as **false** — a typo'd `MAX_NOT_OK` would have silently disabled the cap. `SAMPLES`, `NEED_STREAK`, `MAX_NOT_OK`, `TRANSPORT_RETRIES` and `MAX_TOTAL_SECONDS` are now validated as non-negative integers before the first request, with a test.

Nit 6 also taken: the wall-clock test uses `need_streak=1` so a slow sample on a loaded runner cannot trip the ceiling before the streak assertion.

### Re-run

```
new suite                        33 passed              exit 0
deploy.yml readers (5 files)     97 passed / 2 skipped  exit 0
other workflow readers (4 files) 69 passed              exit 0
ruff check / ruff format --check exit 0 / exit 0
bash -n                          exit 0                 (bash 3.2.57)
```
Mutations, each re-run red with pytest exit 1: **A** retry removed (6 failed), **B** trailing streak (4 failed), **C** version check dropped (2 failed), **D** `MAX_NOT_OK` check removed (2 failed), plus one per defect: **E** (1 failed), **F** (2 failed), **G** (1 failed). Full baseline-vs-mutated logs are in the PR body.

`actionlint` is not available on this machine and was not run; `shellcheck` is not installed either, and — as you found — neither runs in CI, so this script has no linter there at all.

PR #1797, opened 2026-09-02
[deploy] Strip the retired BACKTEST_REFRESH_* env vars from the task-def on deploy (#1760 follow-up)

dbrowneup:
## What

`.github/scripts/ecs_rewrite_task_def.py` now **strips** a declared list of retired env names from the `backend` container before registering the new revision.

```python
RETIRED_BACKEND_ENV = (
    "BACKTEST_REFRESH_ENABLED",
    "BACKTEST_REFRESH_INTERVAL_HOURS",
    "BACKTEST_MAX_AGE_HOURS",
    "BACKTEST_REFRESH_STARTUP_DELAY_S",
)
```

## Why

#1766 (merged `03397108`, deployed as task-def **228**) deleted `services/backtest_scheduler.py` and its lifespan wiring. Nothing on main reads any of those four names. But the live task definition still carries `BACKTEST_REFRESH_ENABLED=false` on the backend container — the owner pinned it by hand as **revision 216** during the #1760 storm, and `deploy.yml` clones the *currently registered* task-def forward on every deploy. So it rides along on 217, 218, … 228 and every revision after, forever.

The flip-list called removing it "a **deliberate operator action**". Per `docs/CONVENTIONS` / the prod-ops principle, cleanups and migrations ship with the deploy — not as a ritual someone has to remember. So the script that actually ships does it.

**The first deploy after this merge** registers the revision that drops them. Every deploy after that is a no-op (the strip is idempotent).

## Rebased onto main — main wins on `PAPER_ADVANCE`

This branch was cut from `03397108`, which predates **#1778** (the arming lift). It carried `PAPER_ADVANCE_VALUE = "false"`, the old function name `pin_backend_paper_advance_off(...)`, and tests asserting `"false"` — all three now contradict production. `origin/main` (`e81b60f0`) is merged in and **main's side is taken for everything about `PAPER_ADVANCE`**: the pinned value is `"true"`, the function is `pin_backend_paper_advance`, and main's assertions stand unchanged.

What survives from this branch is only the additive feature:

- the `RETIRED_BACKEND_ENV` tuple and `strip_retired_backend_env()`;
- the module-docstring item, **renumbered to 6** to fit main's list (image retag → `PAPER_ADVANCE` pin → `FREE_GENERATIONS` pin → Tiingo secret → readiness `healthCheck` + `HEALTH_STALE_UNREADY_S` pin → **strip** → drop describe-only fields);
- the call site, chained so the strip runs **after** the three pins, on the already-pinned environment.

Every assertion about the tick in the new test class now reads `ecs_rewrite_task_def.PAPER_ADVANCE_VALUE` rather than a literal, so a future flip in either direction moves one constant and not this class.

### The ordering, and the guard it needs

The strip runs **last**, so `RETIRED_BACKEND_ENV` is the final word on what the backend container may carry: a knob nothing reads can never ship, whatever else the file asked for. The cost of that order is that a name landing on both the pinned list and the retired tuple would silently lose its pin — so `test_no_pinned_name_is_also_retired` holds `PAPER_ADVANCE_ENABLED` / `FREE_GENERATIONS_PER_ACCOUNT` / `HEALTH_STALE_UNREADY_S` disjoint from the tuple and names the fix in its failure message. With the two disjoint, the strip cannot reach a pin at all.

## Verification that all four are dead

Repo-wide grep for each name over `*.py`/`*.tf`/`*.yml`/`*.yaml`/`*.js`/`*.jsx`/`*.ts`/`*.tsx`: outside `docs/` and this PR's own files, every surviving hit is prose under `backend/tests` and none of them is a reader — a tombstone *comment* in `backend/tests/test_backtests_are_frozen.py`, and three mentions in `backend/tests/services/test_paper_advance_isolation.py` (two docstrings and one assertion message) that name `BACKTEST_REFRESH_ENABLED=false` as the #1760 mitigation the #1741 arming used to hang off. **Zero readers under `backend/archimedes`** (241 modules walked), zero in `infra/`, zero in `.github/workflows/`. All four stay on the tuple.

## Guards, shown red

**1. Delete the strip → red.** Mutation: `environment, removed = pinned, []` at the call site.

```
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_the_live_pin_is_stripped_and_nothing_else_moves
  - AssertionError: assert 'BACKTEST_REFRESH_ENABLED' not in {'APP_ENV': 'produ...
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_every_retired_name_is_stripped
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_stripping_is_idempotent
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_it_says_what_it_removed
4 failed, 14 passed
```

**2. Re-add a reader without un-retiring the name → red.** Mutation: `os.getenv("BACKTEST_MAX_AGE_HOURS", "168")` in `backend/archimedes/main.py`.

```
E   AssertionError: retired env names are read again under backend/archimedes:
E   ['backend/archimedes/main.py:13: BACKTEST_MAX_AGE_HOURS']. The CI deploy strips every
E   name in RETIRED_BACKEND_ENV from the backend container, so this reader can never be set
E   in production. Take the name off RETIRED_BACKEND_ENV (and re-add it to the task
E   definition) or delete the reader. Policy: docs/adr/backtests-are-frozen-evidence.md.
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_no_retired_name_has_a_reader_left_in_the_backend
1 failed, 17 passed
```

That scan is AST-based over committed sources (string constants only), so a tombstone *comment* naming the knob does not trip it, an `os.getenv("…")` does, and a reader in a module the suite never imports still trips it.

**3. Retire a name that is also pinned → red.** Mutation: add `"FREE_GENERATIONS_PER_ACCOUNT"` to `RETIRED_BACKEND_ENV`.

```
E   AssertionError: ['FREE_GENERATIONS_PER_ACCOUNT'] is both pinned and retired — the
E   strip runs last, so the pin would never reach the registered revision. Take the name
E   off RETIRED_BACKEND_ENV or stop pinning it.
FAILED ...::TestRewriteStripsRetiredBackendEnv::test_no_pinned_name_is_also_retired
```

…and, exactly as the message predicts, the pin itself stops shipping: `test_ecs_free_generations_pin.py` goes **8 failed / 10 passed** on `KeyError: 'FREE_GENERATIONS_PER_ACCOUNT'`, and this file adds two more (the disjointness guard plus `test_every_retired_name_is_stripped`, which asserts the strip did not eat a pin) — 10 failures across the two files. Each mutation was reverted by editing it back, never by `git checkout --`.

Other tests in the new class: everything else byte-identical after the strip (`json.dumps(sort_keys=True)` against a clone that never had the pin, with the pin inserted **mid-list** so a reorder is caught); the removal is logged once and the *next* deploy logs nothing; a clean clone logs nothing; `nginx`/`auth` containers are untouched.

The idempotence test feeds the registered revision back the way the next deploy clones it — `describe-task-definition` returns `taskDefinitionArn`, which main's Tiingo pin reads to derive account and region — and compares the environment as a **mapping**, because main's pin chain is stable in content but not in entry position across two passes.

## Tests

```
backend/tests/test_ecs_paper_advance_deploy_pin.py (18)
backend/tests/test_ecs_free_generations_pin.py (18)
backend/tests/test_ecs_backend_secrets.py (40)
backend/tests/test_ecs_readiness_deploy_pin.py (21)
→ 97 passed
ruff check + ruff format --check (774 files) → clean
.github/scripts/docs_links.py → 1857 links, 0 broken
.github/scripts/docs_index.py → 182 docs, 0 orphaned
```

## Docs

- `docs/operations/feature-flag-fliplist.md:153` — "Operator action outstanding" → "No operator action — the deploy strips the pin", naming the first deploy after merge as the revision that drops them, and the guard.
- `docs/adr/backtests-are-frozen-evidence.md` and `docs/runbooks/curated-backtests.md` carried the same now-false "still sitting on the live task definition / deliberate operator action" claim; both corrected.

## Concerns

- The strip only fires on the path `deploy.yml` takes (clone-and-retag). A `terraform apply` that rewrites the task-def from `infra/ecs.tf` is a separate path — but none of these four names is in `ecs.tf`, so it produces the same result.
- If someone ever *needs* one of these names back, adding a reader alone goes red with an actionable message; they must take it off the tuple too. That is the intent, but it is one extra step.


dbrowneup:
Verifier defect 1 (vacuous reader-ban walk) fixed in 13254707 by the owner's session: the test now requires the walk to see 20+ modules and exactly one `PAPER_ADVANCE_ENABLED` string constant. Mutation M3 (glob → `*.nope`) is red: `walk saw only 0 modules`. Full file green after restore. — via the owner's session

Issue #1802, opened 2026-09-02
[paper-trading][design] Drill-down for a deployed strategy: daily ledger + trade tape + positions from data that already exists; mount Önder's Quant Lab panels; clarify the two passes and the 15-minute marks job that is not deployed


dbrowneup:
Owner direction (Dan, 2026-09-02): paper trading must be "fun, interesting, exciting, useful, engaging"; users cannot click into a deployed strategy today; consider bolting Önder's hidden Quant Lab onto the drill-down; clarify replay vs agent pass vs the intraday cadence.

**Facts (main @ 9cb868eb):**
- The ledger tick is DAILY (`paper_trading.py:1202`, `PAPER_ADVANCE_INTERVAL_HOURS` default 24, 240 s startup delay). It replays the frozen `spec_json` deterministically through the DSL engine (no LLM; `paper_trace.py:27-31` forbids one) and appends one `paper_daily_returns` row per trading day, publishes a hashed decision trace, caches positions.
- A 15-minute cadence EXISTS but only for mark-to-market (`paper_marks.py:103-109`, `run_paper_marks.py`, decided in `docs/plans/2026-08-30-intraday-paper-trading.md:14-17`: marks yes, intraday signals no — "re-deciding the position more often is a different strategy from the one the gate graded"). **No scheduler for `run_paper_marks` exists in infra/** — whether marks are written in prod is unverified; the sparkline's dashed intraday tail may be empty.
- The agent pass (`paper_agent_execution.py:192`) is deterministic too (evaluate → targets → 15 % drift gate → trades) and writes ONLY `paper_agent_trades` — no route, no UI reads it. Complementary by design (`paper_store.py:361-366`), not redundant; the execution-agent SOCIETY in `docs/specs/execution-trading-agent-society-spec.md` is DRAFT and not built.
- `PaperTrading.jsx:308-420` shows name (links to the passport), days, ACTIVE/DRIFT chips, sparkline, total return, live mark, Stop. `GET /api/paper/deployments/{id}` (`paper_routes.py:157`) exists and is called by nothing. Unshown data: every daily row, `trace_coverage`, `ledger_engine_versions`, `unversioned_rows`, the published traces, `position_cache_json`, the whole agent trade tape.
[...]

PR #1805, opened 2026-09-02
[paper-trading][product-truth] The 15-minute re-pricing sentence is conditional on a mark actually existing (#1802 PR 0)

dbrowneup:
## The claim

`ui/src/components/PaperTrading.jsx` told every reader, unconditionally:

> The **live value** beneath each total return re-prices the strategy's asset basket **every 15 minutes** — it is unsettled, carries the time it was observed, and never changes what the strategy does.

## Why it was not true today

The marks job exists in the backend — `backend/archimedes/services/paper_marks.py` and `backend/archimedes/scripts/run_paper_marks.py` — but **nothing under `infra/` schedules it**:

```
$ grep -rn paper_marks infra/
(no output)
```

So in production no marks are written, and a sentence asserting a 15-minute cadence is a promise about a job that does not run. The graded truth is the **daily settle**: `paper_trading.py`'s advance loop runs on `PAPER_ADVANCE_INTERVAL_HOURS` (default 24) and appends one real-data return per trading day.

## The change (one logical change)

`paperCadenceCopy(mark, now, intervalMinutes)` in `ui/src/paperCopy.js` derives the intro copy from the marks actually in the payload:

| state | copy |
|---|---|
| no usable mark | the daily-settle sentence **alone** — no cadence claimed |
| fresh mark | settle sentence **+** the 15-minute sentence |
| stale mark | settle sentence **+** the existing `marksStalenessNote` observation-age wording |

- `PAPER_SETTLE_CADENCE` is the unconditional sentence: the ledger settles once per trading day from the graded replay, and a live value is shown *only when one has been marked*, always unsettled.
- Freshness is **not a new number**: `intervalMinutes` threads straight into `marksStalenessNote` (two cadence intervals), so the intro and the per-card live value can never disagree about the same mark. No new thresholds, no new wording invented — `MARK_BASIS_DISCLOSURE` and the staleness logic are reused as-is.
- `newestMark(deployments, marksById, errorsById)` computes the gate's input, mirroring `LiveValue`'s own precedence — **error first**, then the polled list, then `deployment_summary.latest_mark` (`paper_trading.py:1088`) — so the intro cannot claim a cadence the cards below it are not showing. Marks with an unusable timestamp are skipped.
- `noMarksNote('active')` no longer promises "the next 15-minute tick" either: it states the absence and points at the number that is graded.

## Verifier defects, fixed (commit `6967f2ee`)

A review pass ran four mutations against the first commit and found three real defects. All three are fixed here.

**1 — `paperCopy.js` `noMarksNote`: the identical 15-minute promise survived on every card.** The first commit removed the claim from one paragraph a reader skims and left the same claim under *every active deployment* — arguably more prominent per-row than what was removed, and it renders in the exact state every production user is in today. Now:

```js
? 'No live value yet — none has been marked for this deployment. The daily settle is the graded number.'
```

New guard `noMarksNote does not promise a tick no deployed job produces`; the existing `paper-marks.test.js` expectation that pinned `/next 15-minute tick/` is rewritten to pin the absence.

**2 — `paperCopy.js` `newestMark` ignored the marks-error map.** `LiveValue` checks `error` **first** (`PaperTrading.jsx:153`), then `marks === undefined`. `newestMark` checked neither, and on a marks-endpoint failure `loadMarks` sets `nextErrors[id]` and leaves `nextMarks[id]` unset — so every card rendered "Live value unavailable" with no number while the intro fell back to the summary's `latest_mark` and printed "re-prices … every 15 minutes". A total feed outage left the intro as the one line on the page still asserting a live cadence — the exact property this PR body previously asserted was impossible. `newestMark` now takes `errorsById` and skips any deployment whose marks fetch failed; `PaperTrading.jsx:314` threads `marksErrors` in, and the wiring guard pins the three-argument call.

**3 — false file path.** The comment in `paperCopy.js` and the header comment in `paper-cadence-copy.test.js` cited `scripts/run_paper_marks.py`. A top-level `scripts/` directory exists and does **not** contain it, so the citation actively misdirected. All occurrences (source, test, and this body, twice) now read `backend/archimedes/scripts/run_paper_marks.py`.

## Guard

`ui/test/paper-cadence-copy.test.js` — 14 tests: the three copy states, the shared staleness threshold, `newestMark` precedence/ordering, the marks-error skip, the `noMarksNote` negative, two source-text wiring tests that the unconditional claim is gone from `PaperTrading.jsx`, and (added on the merge with main, below) the settle sentence's #1822 wording.

**Mutations** (whole-suite counts as of commit `6967f2ee`, before main was merged in — the suite is 1054 tests now; each reverted, tree clean):

```
M1  restore the unconditional sentence in the intro   ✖ PaperTrading.jsx no longer states the 15-minute cadence as page prose
                                                      pass 928  fail 1
M2  paperCadenceCopy gate always passes               ✖ no mark: the daily-settle sentence ONLY
                                                      ✖ a mark with no usable value or timestamp is no mark at all
                                                      ✖ stale mark: the staleness note, and NOT the cadence sentence
                                                      ✖ newestMark ignores a deployment whose marks fetch failed
                                                      pass 925  fail 4
M3  newestMark drops the polled-list preference       ✖ newestMark: falls back to the summary's latest_mark…
                                                      pass 928  fail 1
M4  staleness threshold 2x -> 100x                    ✖ stale mark: the staleness note, and NOT the cadence sentence
                                                      ✖ marksStalenessNote: the freshness window follows the configured cadence
                                                      pass 927  fail 2
```

The three new guards, red against the pre-fix code and green after:

```
A   noMarksNote promises the tick again    ✖ noMarksNote does not promise a tick no deployed job produces
                                           ✖ noMarksNote: an active deployment is told the absence, not a cadence
                                           pass 927  fail 2
B   drop `if (errorsById[id]) continue`    ✖ newestMark ignores a deployment whose marks fetch failed
                                           pass 928  fail 1
B2  un-thread marksErrors in the JSX       ✖ PaperTrading.jsx derives the intro cadence from the marks in the payload
                                           pass 928  fail 1
restored                                   pass 929  fail 0
```

## Merge with main (#1822) — the settle sentence is reworded

Main now carries #1822, which retracted "the track record that carries to mainnet" from all four paper-trading copy surfaces: the Arc mainnet cutover was cancelled by owner call (#1240) and no date is scheduled, so the sentence promised an unscheduled event.

**The conflict** was the intro `<p>` in `PaperTrading.jsx`: #1822 edited that hardcoded sentence in place, and this PR deletes it outright in favour of `cadence.sentences.map(...)`. Resolved to this PR's structure — the prose #1822 fixed no longer exists on this branch.

**The silent half** is what mattered. `paperCopy.js` auto-merged clean, and `PAPER_SETTLE_CADENCE` — the constant this PR moves that copy into — still said the settled series "is the track record that carries to mainnet". Taking git's merge as produced would have re-landed the exact claim #1822 removed, now pinned in a constant. So the settle sentence is **reworded to match #1822**:

> The ledger settles once per trading day from the graded replay — that settled series is a paper track record on Arc testnet, with no real funds. A live value is shown beneath it only when one has been marked, and it is always unsettled.

`PAPER_INTRADAY_CADENCE` is untouched and still gated on a fresh mark; the three-state behaviour above is unchanged.

**New guard** — `the settle sentence names what the series is today, not a cancelled cutover` — pins the property on the exported **value**, complementing #1822's source-text scan of `ui/src`. Two mutations, each reverted by editing:

```
M-A  restore "the track record that carries to mainnet"
     ✖ no paper-trading surface names mainnet (#1807, cutover cancelled by #1240)
     ✖ no file under ui/src says the ledger "carries to mainnet"
     ✖ no file under ui/src pairs mainnet with the paper ledger
     ✖ the paper-trading copy says what is true instead: Arc testnet, no real funds
     ✖ the settle sentence names what the series is today, not a cancelled cutover
                                                            pass 17  fail 5

M-B  the same claim, wrapped so no raw line holds the word
     ('…carries to main' + 'net.')
     ✖ no file under ui/src says the ledger "carries to mainnet"
     ✖ no file under ui/src pairs mainnet with the paper ledger
     ✖ the paper-trading copy says what is true instead: Arc testnet, no real funds
     ✖ the settle sentence names what the series is today, not a cancelled cutover
                                                            pass 18  fail 4
```

Under M-B the raw-line test `no paper-trading surface names mainnet` goes **green** — the banned word is split across the concatenation and no single line holds it. That is precisely why the value-level pin was added rather than left to the source scan alone.

## Verification

```
$ git grep -n -i "carries to mainnet" -- ui/src
(empty)

$ cd ui && node --test test/no-mainnet-track-record.test.js test/paper-cadence-copy.test.js
ℹ tests 22
ℹ pass 22
ℹ fail 0

$ cd ui && node --test
ℹ tests 1054
ℹ pass 1054
ℹ fail 0

$ npx eslint src test
(clean)

$ ruff format --check .
764 files already formatted
```

## Out of scope — follow-up

This PR does **not** deploy the marks job. That is #1802 decision A (schedule `backend/archimedes/scripts/run_paper_marks.py` in `infra/`, or retire the marks surface) — once a marks job actually runs, this copy starts telling the truth about it on its own, with no further edit.

~~Also left alone … `PAPER_SETTLE_CADENCE` says the settled series "carries to mainnet" … it is Dan's call whether it changes here or in the #1802 copy sweep~~ — **no longer true.** #1822 landed on main and made that call: the settle sentence is reworded in the merge above, and `ui/test/no-mainnet-track-record.test.js` now fails this branch if it ever comes back.

Still open, and unchanged by the merge: the constant stutters slightly with the surrounding JSX paragraph.

Part of #1802


dbrowneup:
All three verifier defects fixed in `6967f2ee` (pushed). Per defect:

**Defect 1 — `ui/src/paperCopy.js:219`, `noMarksNote` kept the identical 15-minute promise on every card.** Correct and the more serious of the three: the first commit removed the claim from one skimmable paragraph and left the same claim rendering under *every active deployment*, in the state every production reader is in today. Now:

```js
export function noMarksNote(status) {
  return status === 'active'
    ? 'No live value yet — none has been marked for this deployment. The daily settle is the graded number.'
    : 'No live value — marks stop when a deployment is stopped.'
}
```

There *was* an existing test pinning the old string — `ui/test/paper-marks.test.js:128` asserted `/next 15-minute tick/`. Rewritten to pin the absence instead (`daily settle is the graded number`, plus `doesNotMatch(/15-minute|15 minutes|next tick/)`), so the old copy cannot come back through that file. New guard added in `paper-cadence-copy.test.js`: `noMarksNote does not promise a tick no deployed job produces`. Red against the old string: 2 fail (the new guard and the rewritten one), 927 pass.

**Defect 2 — `newestMark` ignored `marksErrors`.** Confirmed by reading the two branches: `LiveValue` checks `error` first (`PaperTrading.jsx:153`), `loadMarks` deliberately does not carry stale marks forward (`:266-275`), so on a total marks outage every card shows "Live value unavailable" with no number while `newestMark` fell back to the summary's `latest_mark` and the intro printed the cadence. Exactly the property the body claimed was impossible. Applied as specified:

```js
export function newestMark(deployments, marksById = {}, errorsById = {}) {
  …
    const id = dep?.deployment_id
    // Same precedence as LiveValue, error branch included: a deployment whose
    // marks fetch failed shows NO number on its card, so it must not supply
    // the intro's cadence claim either.
    if (errorsById[id]) continue
    const polled = marksById[id]
```

`PaperTrading.jsx:314` now reads `paperCadenceCopy(newestMark(deployments, marks, marksErrors))`, and the wiring guard pins that three-argument form. New test `newestMark ignores a deployment whose marks fetch failed` also asserts the skip is scoped to the failure (`newestMark(deps, {}, {}) === FRESH`), so it cannot pass by returning null unconditionally. Two red runs: dropping the `continue` line → 1 fail; un-threading `marksErrors` in the JSX → 1 fail (the wiring guard).

**Defect 3 — false file path.** Fixed in all four places: `paperCopy.js` comment, `paper-cadence-copy.test.js` header comment, and both PR-body occurrences → `backend/archimedes/scripts/run_paper_marks.py`. You were right that a top-level `scripts/` exists and does not contain it, which made the citation misdirect rather than merely be imprecise.

**Not taken (notes 4-7).** Note 4 (`carries to mainnet` vs the closed #1240) is a real wording problem but pre-existing copy this PR only promotes into a constant — left for Dan's call and recorded in the body's out-of-scope section rather than changed silently under a copy-truth PR. Note 5 (stutter) same bucket. Notes 6 and 7 need no action, agreed.

**Re-run.** `node --test`: **929 pass, 0 fail** (927 → 929, two new guards). `npx eslint src test`: clean, exit 0. M1-M4 re-run whole-suite: M1 928/1, M2 925/4, M3 928/1, M4 927/2 — M4 trips one extra (`marksStalenessNote: the freshness window follows the configured cadence`) because the whole suite runs, not just this file. All mutations reverted from backups, not `git checkout` (an early `git checkout --` in this session clobbered an uncommitted fix; caught and reapplied before committing).

Issue #1798, opened 2026-09-02
[data][infra] Tiingo is in SSM but not wired: the task definition carries no TIINGO_API_TOKEN secret and MARKET_DATA_PROVIDER is unset, so every seam still runs on yfinance


dbrowneup:
Owner check 2026-09-02: `/archimedes/prod/TIINGO_API_TOKEN` exists in SSM (last modified 2026-08-31 16:58 -05:00). But on the live task definition (`archimedes-backend:228`) the backend container has **no** `TIINGO_*` secret and **no** `MARKET_DATA_PROVIDER` env, and `grep -ri tiingo infra/ .github/` on main returns nothing. `market_data_provider.py:537-539` says so itself ("NOT wired into … `/archimedes/prod/TIINGO_API_TOKEN` and adding the ecs.tf entry are cutover prerequisites"). PR #1627 split the sourcing (Tiingo for paid analysis, yfinance for the free Explore viewer) and PR #1455 added the provider, but the cutover step never got an issue — it slipped. Consequence: backtests, paper ledgers, daily returns and the oracle cross-check all run on yfinance today, and the claims-ledger row "paid analysis runs on licensed data" (docs/claims-ledger.md:212) stays PENDING.

## Fix (one PR + one owner apply + one proof)
1. `infra/ecs.tf`: add the `TIINGO_API_TOKEN` secret (`valueFrom` the SSM SecureString ARN) to the backend container and grant `ssm:GetParameters`/`kms:Decrypt` for it on the task execution role, mirroring the existing secrets block; add `MARKET_DATA_PROVIDER` for the paid seam per PR #1627's design (leave the free Explore seam on yfinance if that split is still the product call — confirm against docs/adr/market-data-sourcing.md).
2. Guard: extend `backend/tests/test_ecs_backend_secrets.py` so the secret's presence and ARN shape are pinned (mutation: remove it → red).
3. Health signal: a `/health` field or an ops script (`scripts/verify_market_data.py`) that pulls one known symbol for one known date range through the ACTIVE provider and reports vendor, row count and the first/last bar date — run once after apply and paste the output on this issue as the verified-pull record (dates + counts), which docs/ currently lacks.
4. Then flip the claims-ledger row from PENDING to its true state.

Owner steps: `infra/apply.sh --apply` after merge (expected plan: the task definition + IAM policy), then one deploy so the new revision carries the secret.

[...]

PR #1806, opened 2026-09-02
[data][infra] Wire the Tiingo token into the backend task definition on both paths + a market-data proof script (#1798)

### LINKED ISSUE #1799: [infra][terraform] aws_ecs_task_definition.backend has drifted from the pipeline-owned revisions — an untargeted apply would replace it from infra/ecs.tf
Seen 2026-09-02 03:00Z while applying #1786's WAF change: `terraform plan` on main (post-#1778) reports `aws_ecs_task_definition.backend must be replaced` (state holds revision 213 from the last terraform apply; the ECS service runs revision 228+ registered by `deploy.yml`'s clone-and-retag step; `infra/ecs.tf` changed in #1778 — `PAPER_ADVANCE_ENABLED` value). I applied with `-target=aws_wafv2_web_acl.main` to avoid it.

Two sources of truth for the backend task definition now exist: terraform's `ecs.tf` resource (env/secrets as declared) and the pipeline's cloned revision (env/secrets as accumulated + the rewrite script's pins/strips). An untargeted apply registers a new revision from `ecs.tf`; whether that is harmless depends on (a) whether the service resource ignores `task_definition` changes (check `lifecycle { ignore_changes }` in ecs.tf) and (b) whether the pipeline clones the SERVICE's revision or the family's LATEST (if latest, the terraform-built one — possibly missing pipeline-only pins — becomes the base of the next deploy).

## Fix (decide, then encode)
Pick one owner: either the pipeline (terraform keeps `ecs.tf` as the declared baseline but `ignore_changes = [container_definitions]` on the task-def resource, and the pipeline is documented as the only registrar), or terraform (the pipeline stops cloning and instead applies terraform per deploy). Add a guard so `terraform plan` on a clean main shows zero task-def changes, and a note in docs/runbooks/cloudfront-cache-behaviour-apply.md (or a new terraform runbook) that infra applies are targeted until this lands.

Related: #1778, #1797 (strips retired env in the pipeline path), #1798 (Tiingo secret must land in BOTH sources or the chosen one).

dbrowneup:
Wires `/archimedes/prod/TIINGO_API_TOKEN` into the backend task definition on both registration paths, and adds the verified-pull script #1798 asks for.

> **Premise correction (2026-09-02).** The first version of this PR — and #1798 itself — said the token was unreachable: *"nothing has ever read it, because no container was handed it."* **That was false**, caught by the adversarial verifier and confirmed against the code. `main.py:47-48` calls `secrets_service.load_ssm_secrets()` whenever `PUBLIC_DOMAIN` is set, and that loader `GetParametersByPath`es the whole `/archimedes/prod/` prefix (`AWS_SSM_PATH_PREFIX`, `infra/ecs.tf:523` on the backend container) into `os.environ`. The task role (`ecs.tf:250-280`) carries `ssm:GetParametersByPath` on both the path and child ARNs plus `kms:Decrypt` via `ssm.<region>`. So the token has reached the app process since 2026-08-31, and the daily-bar path would have answered before this PR. Correction posted on the issue: [#1798 (comment)](https://github.com/aprin-labs/archimedes/issues/1798#issuecomment-5504420216). Commit `2324f735` fixes the same claim in all six places it appeared in the tree.

## What this PR actually buys

The credential arrived as a **soft** dependency. `load_ssm_secrets` catches every error and boots degraded by design, so nothing declared the token, nothing verified it, and its absence produced a log line rather than a failed deploy. A `secrets` entry converts it into a **task-launch** dependency — which is the thing the runbook's `execute-command` check can actually observe, and the thing a deploy can actually fail on. That is defensible hygiene, and it is the honest justification; it is not "the wiring", and it is **not** what was blocking the flip.

**What blocks the flip** is `TiingoProvider.get_intraday_quote`, `get_intraday_quotes_batch` and `get_series` raising `NotImplementedError` — see the § below.

## The secret lands on both registrars

#1799 established that there are two sources of truth for the backend task definition, and the pipeline is the one that ships between applies. A declaration in only one of them is not a deployment.

| Path | Change | Live when |
|---|---|---|
| `infra/ecs.tf` → `aws_ecs_task_definition.backend` | `TIINGO_API_TOKEN` in the backend `secrets` block | someone applies |
| `.github/scripts/ecs_rewrite_task_def.py` (clone-and-retag) | `ensure_backend_tiingo_secret()` on the backend container | next CI deploy |

**No IAM diff, and that is checked rather than asserted.** The execution role's SSM statement is already a prefix wildcard over `parameter/archimedes/prod/*`, and its `kms:Decrypt` statement already covers SecureStrings fetched via `ssm.<region>.amazonaws.com`. `test_execution_role_policy_is_a_prefix_wildcard` (pre-existing) fails if anyone narrows it to an enumeration, and names the secrets that would lose access.

**The ARN on the pipeline path is derived, not passed.** `tiingo_secret_arn()` reads partition/region/account out of the cloned definition's own `taskDefinitionArn` — which `describe-task-definition` always returns — and raises `RewriteError` if it cannot. Same reasoning the file already applies to `PAPER_ADVANCE_VALUE`: a workflow env var or CLI flag has a default, and that default silently decides production. A hardcoded account id would produce a syntactically perfect ARN that fails at launch with `ResourceInitializationError` — the whole service down.

**`aws_ecs_service.backend` has `ignore_changes = [task_definition, desired_count]`**, so there is no risky apply to schedule for this: a terraform apply registers a revision but does not change which one the service runs. Merging and letting `main` deploy is what puts the entry on the running revision. The `ecs.tf` entry is the baseline so a later apply cannot un-wire what the pipeline pinned.

## ⚠️ New coupling the owner should know about

A `secrets` entry is not free. `/archimedes/prod/TIINGO_API_TOKEN` becomes a **task-launch** dependency: deleting, renaming, or moving that parameter out of the prefix fails **every task start** with `ResourceInitializationError` — the whole service, not one feature. Before this PR it was a soft dependency reaching the process only via the best-effort bulk load, so its removal would have degraded one feature silently. Rotating the value in place is safe; removing the parameter is not. This is now stated explicitly in the runbook's Rollback section.

## `MARKET_DATA_PROVIDER` is NOT flipped, and the reason is not caution

I read the ADR and the #1627 diff (`git log -p --follow -- backend/archimedes/services/market_data_provider.py`) as instructed. **There is no separate variable for the paid seam.** `provider_name()` reads one global `MARKET_DATA_PROVIDER`, and the analytics-engine seam reads the same name. The ADR's split — Tiingo for paid analysis, yfinance for the free Explore viewer — is enforced today by *which methods exist*, not by a second variable:

> `TiingoProvider.get_intraday_quote`, `get_intraday_quotes_batch` and `get_series` raise `NotImplementedError`. So `MARKET_DATA_PROVIDER=tiingo` also takes the live oracle push's equity fetch (`oracle_updater._fetch_intraday` calls `get_intraday_quotes_batch` with **no handler** for that exception, so it raises rather than degrading), the VIX / S&P-MA regime reads, and the Explore history modal. The ADR names these under § Consequences as "not cutover-ready".

So the flip is left to the owner's proof step, per instruction, and this PR adds nothing that flips it. It also adds a guard (`FORBIDDEN_WHY`, which `FORBIDDEN_NAMES` derives from) so that setting `MARKET_DATA_PROVIDER` in `infra/ecs.tf` requires editing the guard first — the same anti-goal shape #1463 used for `REVENUE_SWEEP_ENABLED`. **This is the one judgement call in the PR that is worth a second opinion:** it makes the eventual flip a two-file change. I think that is right given the blast radius, but it is reversible in one line.

## The proof script

`scripts/verify_market_data.py` pulls SPY for a **fixed** window — `2026-06-01`..`2026-06-12` inclusive, ten US trading days with no market holiday in it — through `get_provider()`, and prints vendor, row count, first bar, last bar. Exit 0 only if all three match the known answer; exit 1 on a wrong count, wrong dates, no bars, or a raising provider.

The window is fixed rather than relative on purpose: a "last N days" window makes the expected answer a fresh calculation every run, so nothing can be asserted and the operator is left eyeballing plausible numbers. A custom `--start/--end` reports but explicitly withholds a verdict on the row count rather than printing a green tick on a number nobody checked.

`--no-cache` reaches past the `CachingMarketDataProvider` to the vendor it wraps. That matters twice: it forces a live vendor call rather than a warm-cache read, so a green run proves the vendor is reachable *now*; and it removes the database dependency, so the rehearsal runs from a laptop checkout with nothing but the token exported from SSM.

**Honest limit I hit:** the script does not run inside the backend container. `backend/Dockerfile` does `COPY backend /app`, so `/app/scripts` is `backend/scripts` and the repo-root `scripts/` tree is not in the image. The runbook therefore splits the proof: an `aws ecs execute-command` check that the *container process* carries the variable (length only, never the value), plus the pull run from a checkout. Moving the script under `backend/scripts/` would make it exec-able in-container; I kept the path #1798 asked for.

## Guards, shown red before landing

Every mutation was run, not reasoned about. Re-run after the premise-correction commit, because mutation 1's assertion text changed.

**1. Remove the entry from `infra/ecs.tf`** — `7 failed, 30 passed`:

```
E   AssertionError: TIINGO_API_TOKEN is missing from the backend container's `secrets` block.
    Without it the token is a SOFT dependency only — it reaches the process via main.py's
    best-effort SSM bulk load, which catches every error and boots degraded — so nothing
    declares or verifies it, and MARKET_DATA_PROVIDER=tiingo can raise
    TiingoAPIKeyMissingError at provider construction with no prior signal. The provider
    refuses to fall back to yfinance on purpose (docs/adr/market-data-sourcing.md).
E   AssertionError: the header comment says ['eight'] but the block has 7 entries
FAILED tests/test_ecs_backend_secrets.py::TestTiingoTokenOnTheTerraformPath::test_the_secret_is_declared
FAILED ...::test_it_resolves_from_ssm_not_a_literal                      - KeyError: 'TIINGO_API_TOKEN'
FAILED ...::test_it_points_at_its_own_parameter                          - KeyError: 'TIINGO_API_TOKEN'
FAILED ...::test_the_wired_name_is_the_one_the_provider_reads_first
FAILED ...::TestTiingoTokenOnThePipelinePath::test_both_paths_agree_on_the_parameter_path
FAILED ...::TestTheCommentsTellTheTruth::test_comment_states_the_real_count[HEADER]
FAILED ...::TestTheCommentsTellTheTruth::test_comment_states_the_real_count[INLINE]
```

**2. Remove the `ensure_backend_tiingo_secret` call from the rewrite** — `3 failed, 34 passed`:

```
FAILED ...::TestTiingoTokenOnThePipelinePath::test_a_clone_that_predates_the_wiring_gains_the_secret
  - AssertionError: assert 'TIINGO_API_TOKEN' in {'DATABASE_URL': 'arn:aws:ssm:...'}
FAILED ...::test_a_stale_arn_on_the_clone_is_corrected - AssertionError: assert False
FAILED ...::test_the_arn_is_derived_from_the_clones_own_account_and_region - KeyError: 'TIINGO_API_TOKEN'
```

**3. Remove the row-count check from the proof script** — `1 failed, 17 passed`:

```
E   AssertionError: assert 'expected 10 rows, got 9' in 'provider     : tiingo\n...
FAILED tests/test_verify_market_data.py::TestItSaysNo::test_a_short_frame_is_rejected
```

The verifier added a fourth of its own — drifting `TIINGO_SSM_PATH` prod→staging on the pipeline half only — and it went red in 5 tests including `test_both_paths_agree_on_the_parameter_path`.

## Checks run

- `pytest tests/test_ecs_backend_secrets.py tests/test_ecs_paper_advance_deploy_pin.py tests/test_verify_market_data.py tests/services/test_paper_advance_kill_switch.py tests/test_claims_ledger.py tests/test_docs_site.py tests/test_feature_flag_fliplist_drift.py tests/test_infra_gate_workflow.py tests/services/test_market_data_provider.py tests/services/test_tiingo_provider.py` — **267 passed**
- The `.env.example` readers, because `2324f735` edits that file: `tests/test_ipfs_pinning_absent.py tests/test_compose_no_prod_image_default.py tests/test_local_ollama_recipe_drift.py tests/test_main_ssm_prod_gate.py tests/test_local_setup_contract.py tests/test_local_mode_contract.py tests/chain/test_two_chain_config.py` — **93 passed**
- `ruff check` / `ruff format --check` on every touched Python file — clean
- `terraform fmt -check`, `terraform init -backend=false`, `terraform validate` in `infra/` — **Success! The configuration is valid.** (no plan, no apply)
- `.github/scripts/docs_index.py` — 179 indexed, 0 orphaned; `docs_links.py` — 1803 links, 0 broken

No Tiingo or yfinance call was made over the network at any point; every test is hermetic by construction.

## What this does not do

- Does not flip `MARKET_DATA_PROVIDER` (above), and does not fix #1799's underlying two-registrars problem — it works *with* it.
- Does not change `docs/claims-ledger.md`. The "paid analysis runs on licensed data" row stays `PENDING ADR MERGE`; it flips after a real proof run shows `provider : tiingo` **and** the deployed seam is reading it.
- Does not touch `docs/operations/feature-flag-fliplist.md` — its `MARKET_DATA_PROVIDER` and `TIINGO_MIN_REQUEST_INTERVAL_S` rows are still accurate.

Related: #1799, #1627, #1455, #1218.

## Review round 2 — `c8d5ee91`

Two prose defects found by the verifier, both fixed exactly as specified. No code, guard,
terraform, or behaviour change; the wiring itself verified correct on both paths.

1. **`infra/ecs.tf`** — the comment claimed `test_execution_role_policy_is_a_prefix_wildcard`
   "names this secret". It does not: its message enumerates `REQUIRED_CIRCLE_SECRETS` only.
   The comment now says what that guard actually pins, and points at the guard that *does*
   name this entry — `test_every_secret_sits_under_the_execution_roles_prefix`.
2. **`docs/runbooks/market-data-provider-proof.md`** — "confirm the *process* has it" collapsed
   the distinction `2324f735` exists to draw. The exec shell is a sibling process reading the
   container env, not the app process's `os.environ`. The step now says what it proves (the
   `secrets` entry has shipped) and what `MISSING` does not mean.

Adversarial pass on the new comment: moving the ARN to `/archimedes/staging/` fails
`test_every_secret_sits_under_the_execution_roles_prefix` with
`secrets outside the execution role's readable prefix: {'TIINGO_API_TOKEN': ...}` — so the
guard the comment now names really does name this secret.

Re-run: 55 passed on `test_ecs_backend_secrets.py` + `test_verify_market_data.py`; 4 passed on
the docs index/link guards; the three guard mutations plus the ARN-drift one all red, worktree
restored clean after each; `ruff check` / `format --check` clean; `terraform fmt -check
-recursive` and `validate` clean (no plan, no apply). No network call to Tiingo or yfinance.

## Round 3 — merged `main` in (`f7a0dded`)

Merged `origin/main` (`bf76999c`) into the branch — merge, not rebase, because the branch
is pushed. Two conflicts, both resolved **keep-both**; nothing from either side was dropped.

### `.github/scripts/ecs_rewrite_task_def.py` — 4 hunks, all unions

#1809 landed `FREE_GENERATIONS_PER_ACCOUNT=3` on the same two registrars this PR uses, so
both PRs edited the same three places in this file. The pins are independent, so every hunk
takes both sides:

| Hunk | Resolution |
|---|---|
| Module docstring | List **renumbered** rather than picking a winner: 3 = #1809's `FREE_GENERATIONS` pin, 4 = this PR's Tiingo secret, 5 = dropping the describe-only fields. The explanatory paragraph is re-pointed at **step 4**, since it explains the Tiingo step and that step moved. |
| Helpers | `pin_backend_free_generations` (#1809) **and** `tiingo_secret_arn` / `ensure_backend_tiingo_secret` (#1798), ordered to match the docstring's steps. |
| Function summary | Merged to *"pin the tick + allowance flags and the Tiingo secret"* — both clauses, because the function now does both. |
| Call site | `pin_backend_free_generations(pin_backend_paper_advance(env))` **and** `next_container["secrets"] = ensure_backend_tiingo_secret(...)`. This is the hunk that matters: the two env pins compose on one list while the secret lands on a different key, so neither can clobber the other. |

**#1811 does not touch this file** (`git log main -- .github/scripts/ecs_rewrite_task_def.py`
names only `b32d0417`, the #1809 commit; `gh pr view 1811 --json files` has no match). It
merged clean.

`docs/README.md`: both new register rows kept — this branch's
`runbooks/market-data-provider-proof.md` and main's
`incidents/2026-09-03-paper-advance-ddl-wedge.md`.

`infra/ecs.tf` auto-merged across six main-side commits, and
`backend/tests/test_ecs_backend_secrets.py` across one. Both are verified below rather than
assumed, since a clean auto-merge is not a correct one.

### The merged rewrite, run end-to-end

Not asserted — executed. Against a last-good `011b6bfc`-shaped task definition (no
`PAPER_ADVANCE_ENABLED`, no `FREE_GENERATIONS_PER_ACCOUNT`, no Tiingo entry, one
pre-existing `DATABASE_URL` secret), one pass produces all three pins together:

```
::notice::FREE_GENERATIONS_PER_ACCOUNT pinned to 3 (cloned revision did not carry the name)

backend environment:
  APP_ENV                          = 'production'
  PAPER_TRADING                    = 'true'
  PAPER_ADVANCE_ENABLED            = 'true'
  FREE_GENERATIONS_PER_ACCOUNT     = '3'

backend secrets:
  DATABASE_URL      <- arn:aws:ssm:us-east-1:037613907429:parameter/archimedes/prod/DATABASE_URL
  TIINGO_API_TOKEN  <- arn:aws:ssm:us-east-1:037613907429:parameter/archimedes/prod/TIINGO_API_TOKEN

describe-only fields still present: NONE (all seven dropped)
top-level keys kept: ['containerDefinitions', 'cpu', 'family', 'memory']
```

`DATABASE_URL` survives, the ARN is derived from the clone's own account/region, and the
`::notice::` fires exactly once — for the pin whose value actually changed.

Second pass, feeding that output back in as the next deploy's clone with a **wrong-account**
Tiingo ARN injected to simulate terraform drift: `PAPER_ADVANCE_ENABLED` 1×,
`FREE_GENERATIONS_PER_ACCOUNT` 1×, `TIINGO_API_TOKEN` 1× with the stale
`us-west-2:999999999999` ARN corrected back to `us-east-1:037613907429`, and **no
`::notice::`** — nothing changed, so nothing is claimed. Idempotent on all three.

### Adversarial pass on the resolution itself

A keep-both merge fails silently by dropping one side, so both mis-resolutions were written
and run:

| Mutation | Result |
|---|---|
| Drop `pin_backend_free_generations` from the merged call site (the "take ours" mis-resolution) | **7 failed** in `test_ecs_free_generations_pin.py` — `KeyError: 'FREE_GENERATIONS_PER_ACCOUNT'`, `assert '0' == '3'`, `assert 2 == 1` |
| Drop the `ensure_backend_tiingo_secret` assignment (the "take theirs" mis-resolution) | **3 failed** in `test_ecs_backend_secrets.py` — `assert 'TIINGO_API_TOKEN' in {'DATABASE_URL': ...}`, stale-ARN correction, ARN derivation |
| Remove the `TIINGO_API_TOKEN` line from the auto-merged `infra/ecs.tf` | **7 failed**, including both comment-count guards: `the header comment says ['eight'] but the block has 7 entries` |

Each reverted by editing; the tree is the merged content and green after each restore.

### Checks

- `test_ecs_backend_secrets.py`, `test_ecs_free_generations_pin.py`, `test_ecs_paper_advance_deploy_pin.py`, `test_run_generation_job.py`, `test_verify_market_data.py` — **98 passed**
- `test_docs_site.py`, `test_claims_ledger.py`, `test_feature_flag_fliplist_drift.py`, `test_infra_gate_workflow.py` — **58 passed**
- Full backend suite, `python -m pytest -m "not integration" --tb=short -q` from the repo root — **6243 passed, 4 skipped, 2 deselected, 0 failed** (18:49)
- `ruff format --check .` — 766 files already formatted; `ruff check` on the touched Python — clean
- `terraform fmt -check -recursive` + `validate` in `infra/` — **Success! The configuration is valid.** (no plan, no apply)
- `docs_index.py` — 181/181 indexed, 0 orphaned; `docs_links.py` — 1842 links, 0 broken

**`test_task_definition_ownership.py` is not on `main`** — it lives on the unmerged #1799
branch, which is where the 110-test union figure came from. It cannot run here; the 98 above
are the five named files that do exist.

No AWS call. No network call to Tiingo or yfinance.

### Note on the local full-suite run

The box this ran on was carrying a dozen concurrent suites (load average 27) and killed the
single-invocation run twice partway through, so it was first run in sequence by directory —
`{api,chain,marketplace,models,scripts}` **684 passed, 2 skipped**; `services` **1351 passed,
2 failed**; the 246 root-level files in four groups, **939 / 1064 / 924 / 1279 passed**. A
full single invocation then completed on the fourth attempt: **6243 passed, 0 failed**.

The 2 failures that appeared only in the sliced run —
`test_generation_pipeline.py::test_pipeline_writes_identity_ledger_generation_completed_for_owned_job`
and `..._for_anonymous_job`, both `sqlite3.OperationalError: no such table:` — are a
**collection-ordering artifact of slicing the suite, not a defect**, and were confirmed as
such twice: they reproduce identically on a clean `bf76999c` worktree (2 failed, 28 passed
either way), and they **pass in the full ordered run above**. `conftest.py` hands each pytest
*process* a fresh empty temp sqlite file, and these two tests call the real `get_session()`
expecting the identity tables to already exist, so they only pass once something earlier in
the run has created them. Slicing removes that something.

`ruff check .` repo-wide reports 46 non-blocking findings — **the same 46 on clean `main`**,
none in any file this PR touches — and the CI blocking subset (`E9,F63,F7,F40,F82`) is clean.

CI on `f7a0dded`: **21 pass, 2 skipping, 0 fail**, including `Backend — unit tests`.

Part of #1798


dbrowneup:
## Verifier defects — all nine applied, in `2324f735`

Re-ran the touched tests, the three mutations, ruff, terraform, and both docs scripts after every edit. Nothing merged.

---

**Defect 1 — `infra/ecs.tf`: "nothing read it, because the container was never handed it" is false.** Applied.

I confirmed the mechanism independently before rewriting, including the half the verdict did not state: `load_ssm_secrets` runs in the **app** process, so it authorizes against the **task** role, not the execution role. `infra/ecs.tf:250-280` grants that role `ssm:GetParameter`/`GetParameters`/`GetParametersByPath` on *both* `parameter/archimedes/prod` and `.../prod/*` plus `kms:Decrypt` conditioned on `kms:ViaService = ssm.<region>`, so `WithDecryption=True` over the prefix genuinely succeeds. `_PARAM_TO_ENV` is empty and `_extract_env_name` is identity-uppercase, so `/archimedes/prod/TIINGO_API_TOKEN` → `TIINGO_API_TOKEN`. The claim was false; the correction stands.

The block is rewritten to the verdict's text, with one deviation: `main.py:47` → `main.py:47-48`. Line 47 is the `if os.getenv("PUBLIC_DOMAIN"):`; the call is on 48. Also worth noting the file was already honest about exactly this shape 30 lines up — the `AURORA_MASTER_PASSWORD` comment says *"same as the box's runtime `secrets_service.load_ssm_secrets()` fetch today, but deterministic."* I had the right framing available in the same block and did not apply it. `TestTheCommentsTellTheTruth` still green (both anchors end before `secrets = [`, as predicted).

**Defect 2 — rewrite-script docstring.** Applied.

**Defect 3 — test module docstring.** Applied — and I extended it one place the verdict did not reach: `test_the_secret_is_declared`'s **failure message** carried the same overstatement (*"the task starts healthy and MARKET_DATA_PROVIDER=tiingo raises TiingoAPIKeyMissingError on the first fetch"*). With the bulk load in play, removing the `secrets` entry does not by itself produce that raise. That string is what the next engineer reads when the guard fires, so it now says the true thing: without the entry the token is a soft dependency only. Mutation 1's transcript in the PR body is updated to match.

**Defect 4 — runbook § "What was wrong" + § Step 3.** Applied. Step 3 gains an explicit paragraph: the credential was never what stood between here and the flip, and what does block it is the `NotImplementedError` surfaces.

**Defect 5 — runbook step 1's "Expected: the eight names".** Applied, and this was the most consequential of the batch. Step 1 runs after a CI deploy and before any apply, so the revision is a clone chain — not what `ecs.tf` declares. Line 66's diagnosis would have sent an operator to debug the rewrite script over a list that was never going to be eight. Now it expects `TIINGO_API_TOKEN` specifically and says why the rest varies.

**Defect 6 — the `ResourceInitializationError` coupling.** Applied as an explicit operator warning in § Rollback. I also had to fix the sentence above it — *"an unused secret changes no behaviour"* is flatly contradicted by "fails every task start", so it now scopes itself to runtime behaviour. Correct catch: I stated this hazard for a *wrong* ARN in the pipeline-path rationale and never for the *right* one.

**Defect 7 — the live `TiingoAPIKeyMissingError` text.** Applied. It still read *"NOT wired into infra/ecs.tf's task-definition secrets yet — … deliberately not in this PR"*, which this PR makes false in the one string an operator sees at the moment of failure. Now names both paths and says what a prod occurrence actually means: the value is blank, or the process is not the deployed container. No test asserts that substring (checked); `test_missing_key_error_never_contains_a_key_value` and `test_neither_var_set_fails_loud_and_names_the_canonical_var` both still pass.

**Defect 8 — `.env.example:400-401`.** Applied. Re-ran the seven suites that read that file: 93 passed.

**Defect 9 — PR body.** Rewritten. The correction is at the top as a quoted block rather than a silent edit, linked to the [#1798 comment](https://github.com/aprin-labs/archimedes/issues/1798#issuecomment-5504420216); the "what this buys" section now says soft→task-launch instead of impossible→possible; the coupling from defect 6 is its own ⚠️ section; the `--no-cache` rationale is restated on its real merits (forces a live vendor call, drops the DB dependency) rather than on the token's availability.

**Kept, per the owner's instruction and the verdict's own reasoning:** the explicit `secrets` entry. Task-launch dependency is the honest justification, and it is what makes the runbook's `execute-command` check observable rather than decorative.

---

### Re-run after the edits

| Check | Result |
|---|---|
| PR suite list (10 files) | **267 passed** |
| `.env.example` readers (7 files) | **93 passed** |
| Mutation 1 — drop the `ecs.tf` entry | **7 failed**, 30 passed |
| Mutation 2 — drop the `ensure_backend_tiingo_secret` call | **3 failed**, 34 passed |
| Mutation 3 — drop the row-count check | **1 failed**, 17 passed |
| `ruff check` / `format --check`, 5 files | clean |
| `terraform fmt -check` / `init -backend=false` / `validate` | Success — no plan, no apply |
| `docs_index.py` / `docs_links.py` | 179/179, 0 orphaned · 1803 links, 0 broken |

Mutations were applied to a backup-and-restore copy and reverted; `git status` showed only the six intended files, and the full 267 re-ran green afterwards.

### One thing I want on the record

This defect class was mine, not the verifier's to find. I asserted "nothing read it" from the absence of a `secrets` entry without checking whether any *other* path delivered the value — and then repeated it six times, into a live error string and an operator runbook, where each repetition made it look better-sourced. The single grep that would have caught it (`AWS_SSM_PATH_PREFIX` in the same `containerDefinition` I was editing) was eleven lines above my own diff.

Still open for the owner, unchanged from the original review: whether `MARKET_DATA_PROVIDER` belongs in `FORBIDDEN_NAMES` (makes the eventual flip a two-file change; reversible in one line), and whether `scripts/verify_market_data.py` should move under `backend/scripts/` to be exec-able in-container.

dbrowneup:
**Defect 1 — `infra/ecs.tf:803`, the guard did not name this secret.** Fixed in `c8d5ee91`. The comment asserted `test_execution_role_policy_is_a_prefix_wildcard` "fails and names this secret"; its message enumerates `REQUIRED_CIRCLE_SECRETS` (`CIRCLE_API_KEY`, `CIRCLE_ENTITY_SECRET`, `WALLET_ID`) only, and `TIINGO_API_TOKEN` is absent from it. The comment now says what that guard actually pins and names `test_every_secret_sits_under_the_execution_roles_prefix` as the one covering this entry. Adversarial pass before pushing: ARN moved to `/archimedes/staging/` → 3 failures, including `AssertionError: secrets outside the execution role's readable prefix: {'TIINGO_API_TOKEN': ...}`. The named guard does name it.

**Defect 2 — `docs/runbooks/market-data-provider-proof.md:83`, "the process".** Fixed in `c8d5ee91`, applied as written. The exec shell is a sibling process reading the container env, not the app process's `os.environ` that `load_ssm_secrets()` mutates — before this PR the app process held the token via the bulk load while that command would have printed `MISSING`. The step now states what it proves (the `secrets` entry has shipped) and what `MISSING` does not mean.

Both defects were prose only; no code, guard, terraform, or behaviour change. Re-ran: 55 passed on `test_ecs_backend_secrets.py` + `test_verify_market_data.py`, 4 passed on the docs index/link guards, the three guard mutations red as before (7 / 3 / 1 failures), `ruff check` + `format --check` clean, `terraform fmt -check -recursive` and `validate` clean — no plan, no apply, and no network call to Tiingo or yfinance.

Noted but not acted on, per the verdict: the flip's blast-radius list in the ADR omits `paper_marks.get_intraday_quotes_batch` (fail-soft per tick, no `NotImplementedError` handler) — an owner call, and no infra deploys paper marks today.

— via the owner's session

Issue #1800, opened 2026-09-02
[traces][P0] Preserve raw model completions: the debate engine discards the Bedrock text and stores only a scrubbed JSON extraction — replace the regex scrubber with pre-inference screening


dbrowneup:
Owner direction (Dan, 2026-09-02): "The raw completion should not be thrown away. What a horrible mistake." Model reasoning must be preserved exactly as returned, persisted, and surfaced — for the strategy engine and the execution engine.

**Facts (researched on main @ 9cb868eb):**
- `debate_engine.py:628-638`: `raw = backend.complete(...)` → `extract_json` → the raw string is dropped; only `{role, round, verdict(act|decline), claims[], discard[]}` survive. No log, no S3, not in the trace hash preimage (`models/trace.py:77-90`).
- The only durable copy is `debate_transcripts.transcript_json` AFTER `sanitize_transcript` (`models/debate_transcript.py:54-176`): a regex find-and-replace of `T\d\.\d`, `cutover`, `Phase-\w*` → `[redacted]` on model prose (verdict/claims/discard reasons/fusion_reasoning). It rewrites words, is irreversible, and false-positives on any `T<digit>.<digit>` token. Its original motive (prompt bleed-through of roadmap vocabulary, commit 8d8ef105) no longer holds: no system prompt contains that vocabulary today (only two Pydantic `description=` strings in `generate_schemas.py:96,101`, which ship into the public OpenAPI schema — fix those).
- The proposer (`strategy_fusion.py:1165`) and the brief validator (`generation_pipeline.py:354`) also discard raw text.

**Program (ADR + PRs):**
1. Persist the raw completion of EVERY LLM call in generation (validator, proposer, each debate turn, synthesis if any) as an immutable object: S3 `s3://<bucket>/traces/{job_id}/{call_seq}.json` (prompt + params + raw completion + model id + timestamps + hash), with a DB pointer row (`llm_calls`: job_id, candidate_id, role, seq, s3_key, sha256, bytes). Aurora stays small; S3 is cheap and append-only. Include the raw hash in the reasoning-trace preimage going forward.
2. Delete `sanitize_transcript`'s regex rewriting. Keep only structural normalisation (drop non-dict entries). Fix the two OpenAPI descriptions.
3. Surface: the passport's reasoning panel and the generation stream (#1782) get a "raw turn" expander reading the S3 object through an owner-gated route (`GET /api/strategies/{id}/llm-calls/{seq}`), never inline in the list payloads.
[...]

PR #1810, opened 2026-09-02
[traces] Record the verbatim provider response at every LLM seam; pointer-only stream event; no persistence yet (#1800 PR 3)

dbrowneup:
## What

Capture the **verbatim provider response** of every LLM call made inside a generation, at all six existing `record_llm_call` seams in `llm_backend.py` — post-merge `:155, 228, 316, 406, 478, 608`, with the new `record_llm_raw` beside each at `:162, 235, 323, 418, 488, 613`. New `backend/archimedes/services/llm_trace.py` holds the recorder; nothing is persisted yet.

## Why the capture cannot sit on `complete()`'s return value

The string callers receive is **already lossy**, in two independent ways:

- `_first_text_block` (`llm_backend.py:77-89`) returns `text.strip()` for the *first* block that carries text and discards every block after it.
- The Converse loop skips `reasoningContent` outright, so a reasoning model's chain of thought never reaches the caller at all.

So a trace built downstream of those lines would be a trace of the truncation. The recorder is called from *inside* `complete()`, beside the cost-meter seam, and is handed the **provider response object** — the Converse dict, the Anthropic SDK message, the parsed httpx body — before any extraction runs.

## What is recorded

Per call: `call_id, job_id, seq, model_requested, model_served, system, user, provider_response` (verbatim, JSON-safe), `completion_text` (what `complete()` returned, kept for diffing — never *instead of* the body), `usage`, `started_at`, `latency_ms`, `completion_bytes`, `completion_sha256`, plus `capture_error` and the still-unlabelled `stage` / `candidate_id` / `prompt_id`.

Four deliberate calls worth reviewing:

- **`completion_bytes` / `completion_sha256` digest the canonical JSON of `provider_response`, not `completion_text`.** Digesting the text would attest to the lossy string, which is the thing #1800 exists to stop trusting. Documented on the dataclass; a guard asserts the two digests differ.
- **`usage` comes from `cost_meter.extract_token_usage` on the same response object**, so a trace row and a cost row can never disagree about one call. `None` means the provider reported nothing — never a substituted zero.
- **`stage` / `candidate_id` / `prompt_id` are `None` on every record this PR produces.** No call site labels a call yet; the debate engine and the prompt registry fill them in later. Stated in the module docstring so the null reads as "nobody labelled it", not "the call had no stage" — the same caveat `cost_meter` carries about its own binding.
- **`model_served` is a served id only where the provider reports one.** The Converse response carries no model id, so on `bedrock_converse` — today's live backend — it repeats `model_requested`. That caveat now sits on `LLMCallRecord` beside the digest one, so a reader of a record sees it: equal values mean "the provider did not say", not "the requested model ran".

## Stream event: pointer only

`llm_call_recorded {call_id, seq, model_served, completion_bytes, completion_sha256}` through the existing `_Emitter`. **No body, ever**: the generation event log has no per-event owner gate, so whoever holds the stream reads everything pushed to it. (The stream *route* is owner-gated — `generate_routes.py:1011-1022` — so the sharp case is an anonymous job, where `_require_job_access` lets any authenticated caller with the id through.) Bodies are read through an owner-gated route in a later PR.

`complete()` runs off the event loop (`asyncio.to_thread` in the debate engine), so `_llm_pointer_sink` captures the job's loop and schedules the push with `call_soon_threadsafe`. The current UI subscribes only to the names in `GenerationStream.jsx`'s `EVENT_LABELS`, so this event is a deliberate no-op on the frontend today.

## Never raises, never blocks a generation

`record_llm_raw` swallows every failure and logs at DEBUG, exactly as `cost_meter.record_llm_call` does; the pointer sink is wrapped separately, and the record is buffered *before* the sink runs, so a dead stream costs the event and never the trace. An unserializable body is recorded as `capture_error` with `provider_response=None` and **no digest** — "body unavailable", never a fabricated empty completion.

`unbind` cannot raise either. It is the **first** statement in `run_generation`'s `finally`, above `meter.snapshot()`, `store.merge_result` and `_persist_generation_cost` (#1217/#1326) — so a `ContextVar.reset` that escaped there would mask the job's real exception, lose the cost row, *and* skip the `clear()`, leaving full prompts and bodies in memory. It is now guarded exactly as `cost_meter.unbind` guards its own (`ValueError` for a token from another context, `RuntimeError` for one already used), with the clear running either way.

## No persistence

In-memory only, bounded at `LLM_TRACE_MAX_CALLS` (default 512) with overflow **counted** in `stats()` rather than dropped silently. `llm_trace.unbind` clears the buffer in `run_generation`'s `finally` — nothing outlives the job. The S3 flush the next PR adds goes *above* that line.

## Guards — each run against the input that should break it

Every sabotage below was applied locally and confirmed red before pushing. The four marked **(re-run post-merge)** were applied again on the merge commit `309a91f9` — the merge touches `generation_pipeline.py`, so the wiring guard was re-proved against the merged file — and each was reverted by editing the mutation back out, leaving the tree byte-identical (`diff` clean):

| Sabotage | Result |
| --- | --- |
| record the returned string instead of the response object (`provider_response={"text": text}` at every seam) **(re-run post-merge)** | `E AssertionError: anthropic: the recorded body lost 'REASONING-MARKER: …'` — **1 failed** |
| remove the swallow in `record_llm_raw` and around the sink **(re-run post-merge)** | `E RuntimeError: recorder blew up mid-capture` / `E RuntimeError: stream is down` — **2 failed** |
| add `system` + `completion_text` to the pointer | both pointer-only tests — **2 failed** |
| drop the `clear()` in `unbind` | `E AssertionError: prompts + completions must not outlive the job` — **1 failed** |
| **delete the bind from `run_generation`** (`trace_token = llm_trace._CURRENT.set(None)`) | `E AssertionError: run_generation never bound a trace recorder` — **1 failed** |
| **delete the unbind from the `finally`** (`pass`) **(re-run post-merge)** | `E AssertionError: buffered prompts + bodies outlived the job` — **1 failed** |
| **delete both** (the review's original sabotage — previously left the whole suite green) **(re-run post-merge)** | `E AssertionError: run_generation never bound a trace recorder` — **1 failed** |
| **revert `unbind`'s guard to a bare `_CURRENT.reset(token)`** | `E RuntimeError: <Token …> has already been used once` — **1 failed** |

`backend/tests/services/test_llm_trace_recorder.py` — 13 tests, hermetic (no AWS, no network, no Postgres, no Redis).

- **G1** round-trips **all six** seams with a fake provider whose response carries a reasoning block *and* a second text block, and asserts the recorded body still contains both.
- **G5** (new) runs the **real `run_generation`** — the same brief-invalid entry `test_generation_cost_meter.py::test_meter_is_unbound_after_the_job_finishes` uses — with a fake store and a stage that records into whatever recorder is bound. It asserts the recorder was bound *during* the job, that the stage's call landed in *that* job's recorder, and that the binding and the buffer are both gone after. Without it, every guard bound its own recorder by hand and the production wiring was untested: "records every call" and "records nothing" were indistinguishable.

## Test + lint

Re-run on the merge commit `309a91f9` (branch + `origin/main` `e81b60f0`):

```
pytest backend/tests/services/test_llm_trace_recorder.py     → 13 passed
pytest backend/tests/services/test_generation_pipeline.py    → 2 failed, 28 passed  (both pre-existing, isolated-run only)
pytest -m "not integration" (whole backend suite, repo root) → 6620 passed, 4 skipped, 2 deselected
ruff format --check .                                        → 776 files already formatted
ruff check <the four files>                                  → All checks passed!
```

The `5725 passed` figure in an earlier revision of this body was pre-merge; `6620` is the count with `origin/main` merged in.

Two standing caveats, both **pre-existing on `origin/main`** and re-confirmed here by reverting all four PR files to `origin/main` and re-running:

- `backend/tests/services/test_generation_pipeline.py` run in isolation fails `test_pipeline_writes_identity_ledger_generation_completed_for_owned_job` and `…_for_anonymous_job` with `sqlite3.OperationalError: no such table: identity_events` — a fixture-ordering dependency, identical **2 failed, 28 passed** with the PR fully reverted. Does not appear in a full run.
- ~~`ruff check` reports 2 `SIM905` on `_FINANCE_WORDS` (`generation_pipeline.py:249-260`)~~ — **no longer true after the merge.** #1835 moved that block out of `generation_pipeline.py` into `services/brief_screen.py` and fixed it on the way; `ruff check` on all four files is now clean.

## Files shared with the concurrent #1800 PRs

- `backend/archimedes/agents/generation_pipeline.py` — **only** the module docstring, the import line, a new `_llm_pointer_sink` after `_Emitter`, and the bind/unbind pair beside the cost meter (`~:1256` and `~:1866`). **The brief-screening call site is untouched** — after #1835 that is `screen_brief` / `cheap_brief_reject` at `~:199-220`, delegating to `services.brief_screen`, and this PR does not go near it.
  `origin/main` (`e81b60f0`, carrying #1835/#1813) is merged in as `309a91f9`. The one conflict was the predicted `from archimedes.services import …` line, resolved by keeping **both** imports:
  ```python
  from archimedes.services import cost_meter, llm_trace
  from archimedes.services.brief_screen import Surface, Verdict, screen
  ```
  Confirmed post-merge that all **six** seams are still wired (`graft callers record_llm_raw` → six `complete()` callers, one per live backend; `CannedBackend` correctly has none) and that `_all_seams` in the guard still enumerates six.
- `backend/archimedes/services/llm_backend.py` — the six seams only. No prompt constants touched; `api/generate_schemas.py` and `services/brief_screen.py` are untouched.

## Not in this PR

S3 bodies, the `llm_call_index` table, the owner-gated read route, the scrubber deletion, and any stage/candidate labelling. Each is its own step in the ADR's PR sequence.

Part of #1800


dbrowneup:
## Review round applied — `f9c1251c`

All four items from the FIX-FIRST verdict, per item.

---

### 1. Pipeline wiring was unguarded (BLOCKING) — fixed

The finding was exact: deleting **both** the bind (`generation_pipeline.py:1269`) and the unbind (`:1809`) left all 5725 tests green. Every guard in the PR bound its own recorder by hand, so the half that makes this PR do anything in production had zero coverage.

Added `TestPipelineWiring::test_run_generation_binds_the_recorder_and_clears_it_after` to `backend/tests/services/test_llm_trace_recorder.py`. It runs the **real `run_generation`** through the narrowest real entry the cost-meter guard uses (`test_generation_cost_meter.py:465`'s brief-invalid path: fake store + `_llm_available` patched true + a patched `_validate_brief`), with the validator stage recording into whatever recorder is bound. It asserts the recorder was bound **during** the job, that the stage's call reached **that** job's recorder, and that the binding and the buffer are both gone after.

**Mutations, all run against the fixed tree and reverted from file backups afterwards:**

```
############ M1  delete the bind only  (trace_token = llm_trace._CURRENT.set(None)) ############
E   AssertionError: run_generation never bound a trace recorder
    assert None is not None
FAILED …::TestPipelineWiring::test_run_generation_binds_the_recorder_and_clears_it_after
=================== 1 failed, 12 passed, 1 warning in 2.02s ====================

############ M2  delete the unbind from the finally only  (pass) ############
E   AssertionError: buffered prompts + bodies outlived the job
    assert [LLMCallRecor...ompt_id=None)] == []
FAILED …::TestPipelineWiring::test_run_generation_binds_the_recorder_and_clears_it_after
=================== 1 failed, 12 passed, 1 warning in 1.91s ====================

############ M0  your exact sabotage — both deleted (was 5725 green) ############
E   AssertionError: run_generation never bound a trace recorder
    assert None is not None
FAILED …::TestPipelineWiring::test_run_generation_binds_the_recorder_and_clears_it_after
=================== 1 failed, 12 passed, 1 warning in 2.29s ====================
```

M1 neuters the bind while leaving the `finally`'s `unbind(trace_token)` a valid call, so the bind and the unbind fail **independently** rather than only as a pair.

---

### 2. `unbind()` could raise inside the `finally` (BLOCKING) — fixed

`llm_trace.py`'s reset is now guarded the way `cost_meter.unbind` guards its own, catching both cases you reproduced, with `clear()` running either way:

```python
    recorder = _CURRENT.get()
    try:
        _CURRENT.reset(token)
    except (ValueError, RuntimeError):
        logger.debug("llm-trace: token reset out of context", exc_info=True)
    if recorder is not None:
        recorder.clear()
```

The docstring now says why in the terms you used: first statement in `run_generation`'s `finally`, above `meter.snapshot()` / `merge_result` / `_persist_generation_cost` (#1217/#1326); an escape masks the primary exception, loses the cost row, and skips the clear.

I did **not** leave this guard itself unguarded — that would be the same defect class you just flagged. `TestUnbindClearsTheBuffer::test_unbind_survives_a_stale_token_and_still_clears` re-binds and unbinds with an already-used token, which is your `RuntimeError` reproduction, and asserts the buffer is still emptied. Red on a bare reset:

```
############ M3  revert the guard to a bare _CURRENT.reset(token) ############
E   RuntimeError: <Token used var=<ContextVar name='archimedes_llm_trace_recorder' default=None …> …> has already been used once
FAILED …::TestUnbindClearsTheBuffer::test_unbind_survives_a_stale_token_and_still_clears
=================== 1 failed, 12 passed, 1 warning in 1.44s ====================
```

Your honest caveat stands and is worth restating: not reachable on today's call graph. It is a latent hole in a `finally` above money-adjacent persistence, and it is now closed and held closed.

---

### 3. `stats()` docstring stated something the PR's own test disproves — fixed

Replaced verbatim with your wording. `clear()` breaks `recorded == buffered + dropped` by design and `test_llm_trace_recorder.py:456-457` asserts exactly that state, so the docstring now reads:

> Until `clear` runs, `recorded == buffered + dropped`; after it `buffered` falls to 0 while the counts stand, so a shortfall means "the buffer was cleared", and `dropped > 0` is the only honest way to say "this trace is missing its oldest calls".

---

### 4. `model_served` is not a served model on the live backend — fixed

Added verbatim after `llm_trace.py:161`, next to the digest caveat on `LLMCallRecord`, so it sits where the reader of a record is rather than only in a test comment:

> `model_served` is a served id only where the provider reports one. The Converse response carries no model id, so on `bedrock_converse` — the live backend — it repeats `model_requested`. Equal values there mean "the provider did not say", not "the requested model ran".

No issue filed: the verdict asked for the caveat to move to the dataclass, not for a tracking issue, and the behaviour is pre-existing and now honestly pinned where it is read. Say the word if you'd rather it also carry an issue number.

---

### Re-run

```
backend/tests/services/test_llm_trace_recorder.py    → 13 passed   (was 11)
backend/tests/test_generation_pipeline.py            → 18 passed
backend/tests/services/test_generation_pipeline.py   →  2 failed, 28 passed
backend/tests/test_generation_cost_meter.py          → 53 passed
all four together                                    →  2 failed, 112 passed
ruff format --check <the four PR files>              → 4 files already formatted
ruff check <the four PR files>                       → 2 errors, both SIM905
```

Both residuals re-confirmed pre-existing under this exact tree: reverting all four PR files to `origin/main` and re-running `backend/tests/services/test_generation_pipeline.py` alone gives an **identical 2 failed, 28 passed** (`sqlite3.OperationalError: no such table: identity_events`, a fixture-ordering dependency that does not appear in a full run), and both `SIM905` hits are `_FINANCE_WORDS` at `generation_pipeline.py:249-260`, untouched by this PR.

Your advisories on `to_jsonable`'s `except TypeError`, the count-vs-byte cap, the missing meta-guard for newly added backends, and the un-awaited pointer pushes are not addressed here — all were marked "no change required" and none is a truth defect. The pointer-only framing note **is** addressed: the PR body now names the anonymous-job case rather than overstating the ungated read.

PR body updated. **Do not merge — owner vets first.**

PR #1812, opened 2026-09-02
[access] Signed-out visitors browse Explore and Corpus only; leaderboard and passports are gated (#1753)

### LINKED ISSUE #1753: [product][decision] Signed-out visitors can browse nothing — /app/* redirects to sign-in over public APIs
Dogfood 2026-09-01 (human-signup): the client router redirects anonymous users to /sign-in for all /app/* even though underlying reads are public (e.g. /api/explore/assets, 281 assets, 190ms). Combined with the activation blockers this compounds into one bounce funnel: can't look → sign up → free tier gated → only ask left is $2+wallet. DECISION: which read surfaces (explore, library, passports, leaderboard) should be anonymous-browsable? The honest-validation pitch benefits from letting skeptics read verdicts without an account.

dbrowneup:
Owner decision on **#1753**: Explore and Corpus stay open to signed-out visitors; **Library, passports and the leaderboard are gated.** This narrows #1194 revision d, which had four anonymous pages.

## What changed

| Layer | Before (#1194 rev d) | After (#1753) |
|---|---|---|
| `nginx/nginx.conf` | ungated `= /app`, `^~ /app/explore`, `^~ /app/leaderboard`, `^~ /app/corpus`, `^~ /app/strategy` | ungated `= /app`, `^~ /app/explore`, `^~ /app/corpus` |
| `ui/src/routes.js` `ANON_APP_PAGES` | `{explore, leaderboard, corpus, strategy}` | `{explore, corpus}` |
| `ui/src/routes.js` `ANON_NAV_IDS` | `{explore, corpus, leaderboard, generate}` | `{explore, corpus, generate}` |
| `ui/public/sitemap.xml` | 6 `<loc>` incl. `/leaderboard` | 5 `<loc>` |

The server-side half is the one that matters: `/app/leaderboard` and `/app/strategy/*` now fall through to the gated `^~ /app` block, so nginx answers an anonymous request before a line of client JS runs.

## Fail-closed, verified against real nginx

Not asserted — measured. The rendered `nginx/nginx.conf` was mounted into `nginx:alpine` with the `/_auth_session` subrequest stubbed, once as `401` (anonymous) and once as `204` (signed in).

**Anonymous (`auth_request` → 401):**

```
anonymous GET /app                     -> 200
anonymous GET /app/explore             -> 200
anonymous GET /app/corpus              -> 200
anonymous GET /app/leaderboard         -> 302 /sign-in?next=/app/leaderboard&
anonymous GET /app/strategy/abc123     -> 302 /sign-in?next=/app/strategy/abc123&
anonymous GET /app/strategy            -> 302 /sign-in?next=/app/strategy&
anonymous GET /app/library             -> 302 /sign-in?next=/app/library&
anonymous GET /app/insights            -> 302 /sign-in?next=/app/insights&
anonymous GET /app/nonsense            -> 302 /sign-in?next=/app/nonsense&
```

The two newly-gated paths are **indistinguishable from `/app/nonsense`** — no existence oracle, the property #1437 established for `/app/insights`.

Deep link and query string survive, and the redirect stays uncacheable (#1767/#1768):

```
$ curl -sD- -o/dev/null "…/app/strategy/abc123?tab=brief&ref=twitter"
HTTP/1.1 302 Moved Temporarily
Location: /sign-in?next=/app/strategy/abc123&tab=brief&ref=twitter
Cache-Control: private, no-store
```

**Signed in (`auth_request` → 204):** `/app/leaderboard`, `/app/strategy/abc123`, `/app/library`, `/app/explore` all `200`. `nginx -t`: `syntax is ok`.

The round trip through sign-in is pinned in `ui/test/routes.test.js` — `postAuthPath("?next=%2Fapp%2Fstrategy%2Falpha")` → `/app/strategy/alpha`, query preserved, off-site `next` still refused (gating a page must not turn the redirect into an open redirect).

> **Correction, merge round.** The "query preserved" half of that claim was not actually pinned, and I found it by mutating rather than by re-reading. The assertion above encodes `?tab=brief` **inside** the `next` value; nginx emits `return 302 /sign-in?next=$uri&$args`, i.e. the deep link's args arrive as **sibling params**, which a different branch of `postAuthPath` reassembles onto the destination. No assertion reached that branch — deleting it whole left the test green. Two assertions were added and both mutations are red (5 and 6 below). The behaviour was always correct; only the guard was hollow.

## New guard: the lockstep is no longer only a comment

Both files carried "Must stay in lockstep with…". A comment cannot fail. `backend/tests/test_nginx_anonymous_carve_outs.py` parses `ANON_APP_PAGES` + `APP_PATHS` out of `routes.js`, derives the nginx location set they imply, and compares — **in both directions**, so it holds no third copy of the list.

### Guards shown red (adversarial pass)

**Mutation 1 — re-add the `/app/leaderboard` carve-out to nginx:**

```
FAILED test_the_ungated_app_locations_are_exactly_the_anonymous_pages
E   AssertionError: nginx/nginx.conf serves these /app paths with no auth_request, but
E   ui/src/routes.js's ANON_APP_PAGES does not mark them anonymous: ['/app/leaderboard'].
FAILED test_the_gated_pages_have_no_ungated_location[/app/leaderboard]
E   AssertionError: /app/leaderboard is served without auth_request — it is a gated page (#1753)
2 failed, 5 passed
```

**Mutation 2 — re-add `'leaderboard'` to `ANON_APP_PAGES` only (the other direction):**

```
FAILED test_the_ungated_app_locations_are_exactly_the_anonymous_pages
E   AssertionError: ui/src/routes.js marks these pages anonymous-OK but nginx gates them,
E   so a cold load 302s to /sign-in before the SPA ever runs: ['/app/leaderboard'].
FAILED test_the_anonymous_set_is_explore_and_corpus
E   AssertionError: {'corpus', 'explore', 'leaderboard'}
2 failed, 5 passed
```

The decision pin catches the case both mutations are applied together — the quiet reversal the symmetric test alone would miss.

**Mutation 3 — re-open `ANON_APP_PAGES` on the UI side:**

```
✖ ANON_APP_PAGES is Explore and Corpus — the owner's decision, pinned (#1753)
✖ the leaderboard and the strategy passport are gated (#1753)
✖ canNavigateTo gates on ANON_APP_PAGES for a null user … (#1364)
✖ ui/src/routes.js still backs: Two browse pages are deliberately anonymous
pass 78, fail 4
```

**Mutation 4 — re-add `'leaderboard'` to `ANON_NAV_IDS`:**

```
✖ anonymous nav shows browse + the Generate conversion path, nothing stateful
pass 25, fail 1
```

**Mutation 5 — delete the sibling-arg reassembly in `postAuthPath` (`return next`):**

```
✖ a gated deep link survives the round trip through sign-in (#1753)
    actual:   '/app/strategy/abc123',
    expected: '/app/strategy/abc123?tab=brief&ref=twitter',
pass 0, fail 1
```

Against the guard as originally written, this same mutation left the test **green**. That is the hole the added assertion closes.

**Mutation 6 — remove the off-site early return, so a refused `next` keeps its sibling args:**

```
✖ a gated deep link survives the round trip through sign-in (#1753)
    actual:   '/app?tab=x',
    expected: '/app',
pass 0, fail 1
```

**Mutation 7 (re-run after the merge) — re-add the `/app/leaderboard` carve-out to nginx:**

```
FAILED test_the_ungated_app_locations_are_exactly_the_anonymous_pages
FAILED test_the_gated_pages_have_no_ungated_location[/app/leaderboard]
2 failed, 5 passed
```

## Backend APIs: unchanged, and here is what I found

I grepped `strategies_routes.py` and `leaderboard_routes.py` for `is_example` / `anonymous` / public short-circuits. **No route is justified only by the anonymous page**, so none was closed:

- `GET /api/strategies/passports/{id}` and `GET /api/strategies/{id}` — public for *published/curated* rows by the #1557 visibility matrix (`services/strategy_visibility.py`), with private rows still 404-hiding existence. Called by `mcp-server` `archimedes_passport` and by the CLI.
- `GET /api/leaderboard` — its own docstring: "public, no wallet required — never 401s"; anonymous callers get the curated cohort. Called by `mcp-server` `archimedes_leaderboard`.
- `GET /api/leaderboard/live-paper` — public, returns an honest empty `scope='anonymous'` board.

Closing these would break the CLI and the MCP tools for an unrelated reason. The page gate and the route gate are separate decisions, and `docs/api/strategies-and-rigor.md` now says so instead of leaving "the library, passports … are public" to be read as a statement about the browser.

**No feature flag exists for this** — the gate is nginx config plus a JS constant, not an env read — so there is no `feature-flag-fliplist.md` row to add (`test_feature_flag_fliplist_drift.py` still green).

## Claims made true

- `ui/src/components/Security.jsx`: "**Four** browse pages are deliberately anonymous" → "**Two**". Its ledger row is rewritten with live line numbers, and the "kept in lockstep" clause it makes is now backed by the new guard rather than by a comment.
- `docs/claims-ledger.md`: the sitemap row goes from six `<loc>` to five and re-cites `ANON_APP_PAGES`.
- `ui/public/sitemap.xml`: `/leaderboard` removed, and the header says why (a crawler following it gets a 302 to `/sign-in`).
- `infra/cloudfront.tf` + `docs/runbooks/cloudfront-cache-behaviour-apply.md` + `backend/tests/test_cloudfront_session_paths_uncached.py`: carve-out lists narrowed to `/app`, `/app/explore`, `/app/corpus`; the two gated paths moved into `SESSION_DEPENDENT_PATHS`.
- `cli/src/archimedes_cli/cli.py` `_passport_url`: said the passport link "is deliberately reachable without a session, so this link works for whoever the user forwards it to". False as of this PR — rewritten to say a recipient with no session gets `302 /sign-in?next=<this URL>` and lands there after signing in.
- `ui/src/routes.js` `passportBackPage`: no longer describes an anonymous passport flow; the `user == null` branch is documented as the fail-safe it now is.

**CloudFront behaviour is unchanged** and no `terraform apply` is needed: `/app*` was already `CachingDisabled` for gated and public pages alike. That two pages crossed the line with zero edits to `cloudfront.tf` is the owner's ruling on #1772 demonstrated rather than argued — comments updated to say so.

## Test runs

All figures below are from the post-merge tree (`76e60ce2`), re-run after `origin/main` came in.

| Suite | Result |
|---|---|
| backend, full CI command (`pytest -m "not integration"`) | **6216 passed, 4 skipped, 0 failed** (18m17s local); CI's "Backend — unit tests" green in 15m02s |
| `ui` node --test | **1043 passed, 0 failed** |
| `npx eslint .` | clean |
| `npm run check:routes` (orphans + sitemap) | OK — 5 public routes |
| nginx / cloudfront / claims / breadcrumbs / api-docs / fliplist / local-setup guards | **116 passed** |
| `cli/tests` | **164 passed** |
| `ruff check` on changed files, `ruff format --check` | clean (the repo's 46 pre-existing findings are identical on `origin/main`) |
| `docs_links.py` / `docs_index.py` | 0 broken, 0 orphaned |
| `nginx -t` (rendered conf, `nginx:alpine`) | syntax is ok |
| GitHub Actions on `76e60ce2` | **21 pass, 0 fail** (2 conditional skips: S3/CloudFront publish, agent-only coverage) |

## Merged `main` (2026-09-03)

`origin/main` at `bf76999c` merged in (never rebased). One conflict, in `docs/claims-ledger.md`, on two adjacent rows:

| Row | Resolution |
|---|---|
| `POST /api/rigor/verify` — `not_evaluable` / capped quorum | **main's**. #1815 moved both cited lines; the refreshed `:787`/`:803` are correct and this branch's `:273`/`:289` are stale. Verified by reading — and worth saying that `test_claims_ledger.py` would **not** have caught the stale pair: `test_every_cited_line_is_inside_its_file` range-checks a citation, it does not check that the line says what the row claims. |
| `ui/public/sitemap.xml` — anonymous-only `<loc>` list | **this branch's**. Five `<loc>`, `ANON_APP_PAGES` re-cited at `:49`. This is the substance of #1753. |

The ledger's net diff against main is now exactly the two rows this PR means to change. Every other file auto-merged; the four both sides touched (`cli.py`, `strategies-and-rigor.md`, `Leaderboard.jsx`, the ledger) were each read after merging to confirm the two changes compose rather than merely apply — e.g. `Leaderboard.jsx` keeps main's "paper deployments record it forward on Arc testnet" copy under this PR's rewritten gating docstring.

## Concerns for the owner

1. **Share links are the real cost.** A passport URL forwarded to someone without an account now lands on `/sign-in` instead of the strategy. The deep link survives the round trip, so they arrive after signing up — but it converts a "look at this result" link into a signup wall, which is the opposite of how a skeptic-facing validation layer usually earns trust. If passport sharing is meant to be a distribution channel, the alternative is a **read-only public passport** (published rows only, reasoning fields stripped) rather than re-opening the whole page.
2. **SEO.** `/leaderboard` left the sitemap; `/app/strategy/*` was never in it (no canonical URL for a dynamic route). Anything already indexed for `/leaderboard` will now be crawled as a 302 to `/sign-in` and drop out. Two of six indexable URLs remain application pages (`/explore`, `/corpus`) — the marketing surface (`/`, `/architecture`, `/security`) carries proportionally more of the site's crawlable weight now.
3. **`/leaderboard` (top-level alias) still returns 200.** It is not under `/app`, so nginx's `location /` serves the SPA shell and `routes.js` bounces it to `/app/leaderboard`, which then 302s. Fail-closed for content — no board data is served — but the *shell* answers 200 where `/app/leaderboard` answers 302. Same for `/strategy/<id>`. Closing that would mean gating the legacy aliases at nginx too; I left it as pre-existing behaviour rather than widening the diff.
4. **`Leaderboard.jsx`'s signed-out banner is now unreachable** ("Sign in to rank your strategies", `!user` branch). I left it as defensive code rather than deleting UI in a gating PR; say the word and it goes in a follow-up.
5. **`passportBackPage(null)` / `passportBackLabel(null)`** are likewise no longer reachable in production. Kept and documented as a fail-safe — deleting them would remove the thing that stops a sessionless render from ejecting the visitor if the carve-out is ever re-opened.
6. **`ANON_NAV_IDS` was my judgment call, not an explicit instruction.** Leaderboard could instead stay in the anonymous sidebar as a second conversion path (clicking it → sign-in, exactly like Generate). I removed it because the set's own comment defines it as "the browsable pages plus Generate". One line to reverse if you want the funnel.

Closes #1753


dbrowneup:
Verifier defect 1 (stale `Leaderboard.jsx` docstring claiming the page is never auth-gated) fixed in cc60f44d by the owner's session with the verifier's exact text; eslint clean, ui suite green. Concerns 1 (share links hit a sign-up wall — a read-only public passport is the alternative) and 5 (`leaderboard` dropped from the anonymous nav; one line to reverse) are for the owner's morning read. — via the owner's session

PR #1814, opened 2026-09-02
[data] Per-seam market-data routing: daily bars may come from Tiingo while intraday/history stay on yfinance (#1798, ADR amended; default unchanged)

dbrowneup:
## What #1798 actually blocked

The issue was filed as "Tiingo is in SSM but not wired". The owner's correction comment (confirmed against the code) found something else: the token has reached the app process since 2026-08-31 via the `/archimedes/prod/` SSM path loader, and what actually blocks the flip is the **shape of the switch**.

`MARKET_DATA_PROVIDER` was ONE variable read by every seam (`provider_name()`), and `TiingoProvider` serves daily bars only — `get_intraday_quote`, `get_intraday_quotes_batch` and `get_series` raise `NotImplementedError`. So the flip that was supposed to move backtests onto licensed data would, in the same breath, have pointed the live oracle push, the paper-marks loop, the VIX/S&P regime reads and the Explore history modal at three refusals. **The ADR's own decision was not expressible in one variable.**

## The change

Route by **seam**, where a seam is the set of reads one feature makes inside one run.

| Seam | Env var | Serves | Refuses |
| --- | --- | --- | --- |
| `daily` | `MARKET_DATA_DAILY_PROVIDER` → `MARKET_DATA_PROVIDER` → `yfinance` | `get_daily_close_batch`, `get_daily_ohlcv` | the intraday methods — always, whatever vendor is configured |
| `intraday` | `MARKET_DATA_PROVIDER` → `yfinance` | the whole interface, daily bars included | nothing |

- **`get_provider(seam=…)` and `provider_name(seam)` are required arguments.** After the split there is no such thing as "the active provider"; a call site that did not name its seam would be picking a vendor by accident. All **ten** `get_provider(seam=…)` invocations in `backend/archimedes` name theirs — across six modules, which is the count the module docstring uses (`oracle_updater` accounts for four of the ten). An `ast`-based test fails on any bare `get_provider()` left in the package, and since `4977997d` the per-call-site seam table is `ast`-based too: it used to substring-match `inspect.getsource`, which includes the DOCSTRING, and `portfolio_backtester._fetch_price_panel`'s docstring quotes `get_provider(seam="daily")` verbatim — so that row passed on prose and a real `seam="intraday"` mutation left the suite green.
- **`SeamRoutedProvider`** dispatches by method and refuses off-seam ones *before anything is fetched*. The refusal does not depend on today's flag values, so a daily-seam call site cannot pass review on yfinance and break on the day of the flip.
- **`_VENDOR_SEAMS`** declares what each vendor can serve (yfinance: both; tiingo: daily). A vendor named for a seam it cannot serve resolves to yfinance and **logs the substitution**. That is not the silent fallback § 3 of the ADR forbids — the returned name is the vendor that actually answered, so every `source` stamp on a price row, a paper mark or an `asset_daily_bars` row stays true. The missing-token case still raises `TiingoAPIKeyMissingError` at construction.

**Net effect:** the pre-existing global flip (`MARKET_DATA_PROVIDER=tiingo`) now moves daily bars and leaves the live surfaces alone, instead of breaking them. With nothing set, behaviour is byte-for-byte unchanged.

## The one judgement call worth reviewing

**The oracle's ^GSPC moving averages are on the `intraday` seam, not `daily`** — even though they are daily bars. `fetch_market_snapshot` reads ^VIX (intraday) and the ^GSPC 50/200-day MAs in ONE run, and the ADR's "never mix vendors inside one run" is exactly what pins both to the same vendor. ^GSPC is also an index ticker, which Tiingo does not serve at all (`TiingoUnsupportedSymbolError`). The issue text listed "regime daily reads" under the daily column; this PR deviates, with that reasoning written into the code and the ADR table. Say the word if you want it the other way.

## ADR amendment

`docs/adr/market-data-sourcing.md` gains § "Amendment: per-seam routing":

- **unchanged:** a single RUN never mixes vendors (both § 3 guarantees stand);
- **new:** different FEATURES may use different vendors, with a table of every feature → call site → seam → vendor today → vendor after the flip;
- the stale "the live oracle push … is not cutover-ready" consequence is annotated rather than deleted — those surfaces no longer gate the daily cutover;
- the last table row is the one to read twice: **analytics-engine's standalone CLI seam still reads `MARKET_DATA_PROVIDER` only and registers no Tiingo adapter.** Pointing it at the daily variable would make it log "unknown provider" and serve yfinance under a "tiingo" label — the exact lie the no-silent-fallback rule exists to prevent. Its module docstring now says so, and moving it is a named follow-up.

## Guards, shown red before green

```
$ # mutation 1: seam dispatch removed (_route returns getattr unconditionally)
FAILED tests/services/test_market_data_seams.py::TestSeamDispatch::test_daily_seam_refuses_intraday_methods_without_touching_the_vendor[get_intraday_quote]
FAILED tests/services/test_market_data_seams.py::TestSeamDispatch::test_daily_seam_refuses_intraday_methods_without_touching_the_vendor[get_intraday_quotes_batch]
FAILED tests/services/test_market_data_seams.py::TestSeamDispatch::test_daily_seam_refuses_intraday_methods_without_touching_the_vendor[get_series]
FAILED tests/services/test_market_data_seams.py::TestSeamDispatch::test_the_refusal_does_not_depend_on_which_vendor_is_configured
4 failed, 33 passed

$ # mutation 2: default flipped to tiingo
FAILED tests/services/test_market_data_seams.py::TestDefaultsAreUnchanged::test_both_seams_default_to_yfinance
FAILED tests/services/test_market_data_seams.py::TestDefaultsAreUnchanged::test_get_provider_defaults_to_a_yfinance_vendor_on_both_seams
FAILED tests/services/test_market_data_seams.py::TestDefaultsAreUnchanged::test_a_blank_value_is_not_a_vendor_name
FAILED tests/services/test_market_data_seams.py::TestSeamDispatch::test_the_refusal_does_not_depend_on_which_vendor_is_configured
FAILED tests/services/test_market_data_provider.py::TestProviderSelection::test_default_is_yfinance
FAILED tests/services/test_market_data_provider.py::TestProviderSelection::test_get_provider_wraps_vendor_in_caching_provider
6 failed, 70 passed

$ # mutation 3: _VENDOR_SEAMS capability check dropped (the #1798 breakage restored)
FAILED tests/services/test_market_data_seams.py::TestTheSplitHolds::test_global_flip_to_tiingo_moves_daily_and_leaves_intraday_alone
FAILED tests/services/test_market_data_seams.py::TestTheSplitHolds::test_intraday_provider_under_a_global_tiingo_flip_is_a_working_yfinance_one
FAILED tests/services/test_market_data_seams.py::TestTheSplitHolds::test_marks_and_pushes_stamp_the_vendor_that_actually_served_them
3 failed, 34 passed
```

Each mutation was reverted and the suite re-run green.

## Tests run

```
backend: tests/services/ tests/chain/ tests/test_feature_flag_fliplist_drift.py
  → 1724 passed, 2 skipped, 2 failed
backend docs guards (api-docs-drift, asset-universe, claims-ledger, docs-site,
  env-requirements, quant-docs, fliplist-drift) → 89 passed
analytics-engine: tests/ → 482 passed
ruff check (my files) → clean; ruff format --check (repo root) → 738 files already formatted
ruff check backend/archimedes --select F821 (the CI import-guard) → clean
```

The 2 failures are `test_generation_pipeline.py::test_pipeline_{writes,does_not_ledger}_identity_ledger_*`, which fail in `_cleanup()` on `sqlite3.OperationalError: no such table: identity_events` — a local-DB fixture gap, before the pipeline runs, in a test that never touches market data. Not caused by this change.

No network in any test: the Tiingo "refuses before any request" test uses an `httpx.MockTransport` that raises if it is ever called and asserts zero requests. No infra touched, no terraform, nothing flipped.

## What this does NOT do

- **Does not flip anything.** `MARKET_DATA_DAILY_PROVIDER` is unset everywhere; both seams resolve to yfinance on deploy.
- Does not touch `infra/ecs.tf`. ~~It does correct the now-stale "NOT wired into ecs.tf … cutover follow-ups" text in the credential error to say what is actually true (arrives via the SSM path loader, not as a task-def secret).~~ **Superseded by the merge of `main` (`b7c57a5a`)**: #1806 landed, so the token *is* a task-definition secret on both registrars, and the credential error now says so. It does touch one file in that lane — `backend/tests/test_ecs_backend_secrets.py`'s `FORBIDDEN_WHY`, which forbade `MARKET_DATA_PROVIDER` in `ecs.tf` for a reason this PR makes false and did not forbid `MARKET_DATA_DAILY_PROVIDER`, the actual cutover switch, at all. Both are now listed with true reasons. `infra/ecs.tf` itself is byte-identical to `main`.
- Does not flip `docs/claims-ledger.md`'s "paid analysis runs on licensed data" row. It stays PENDING because it is still true: nothing runs on Tiingo until you flip the variable and prove a pull.

## Verification pass (`4977997d`)

Three findings from the review of this branch, all fixed on it:

| # | Finding | Fix |
| --- | --- | --- |
| 1 | **Mixed-vendor bar survived the flip, via the write path.** `_write_cached_series` overwrote `close` + `source` on another vendor's row and left its `open/high/low/volume` in place. Pre-existing (byte-identical to `main`), but this PR is what makes the flip a supported operation, and the body asserted the opposite. | Cross-vendor close-only writes clear the four unknown columns and log the displaced vendor. `_write_cached_ohlcv` needed no change — it assigns every column, so it replaces a bar rather than blending one; its docstring now says so. |
| 2 | **The seam table could not catch a wrong seam at the backtester** — substring match over `inspect.getsource`, whose docstring quotes the call. `seam="daily"` → `"intraday"` at `portfolio_backtester.py:86` left all 43 tests green. | `_seams_requested_by` parses the function with `ast` and matches the `seam` keyword on the real `ast.Call` node. Structural, so a row whose docstring later quotes its own call stays honest. |
| 3 | **The flip recipe named no proof command.** | ADR § "Proving the flip" + the fliplist row point at it (step 3 below). |

New guards, shown red before green:

```
$ # mutation 4: restore the partial overwrite in _write_cached_series
FAILED tests/services/test_market_data_seams.py::TestAFlipNeverProducesAMixedVendorBar::test_the_daily_seam_refetches_from_the_new_vendor
FAILED tests/services/test_market_data_seams.py::TestAFlipNeverProducesAMixedVendorBar::test_no_row_survives_the_sweep_with_a_foreign_vendors_ohlv
FAILED tests/services/test_market_data_provider.py::TestDailyOhlcvCache::test_cross_vendor_close_write_clears_the_old_vendors_ohlv
E   AssertionError: MIXED-VENDOR BAR: source='tiingo' close=400.0 but open=900.1 volume=11.0 came from another vendor
E   assert (900.1, 901.1, 899.1, 11.0) == (None, None, None, None)

$ # mutation 5: portfolio_backtester call site -> seam="intraday"  (GREEN before this pass)
FAILED tests/services/test_market_data_seams.py::TestEveryCallSiteNamesItsSeam::test_function_asks_for_its_seam[archimedes.services.portfolio_backtester-_fetch_price_panel-daily]
E   AssertionError: archimedes.services.portfolio_backtester._fetch_price_panel asks ['intraday'], expected ['daily']
1 failed, 46 passed

$ # mutation 1 re-run: seam dispatch removed (_route returns getattr unconditionally)
4 failed: TestSeamDispatch (3 x refuses-intraday-methods, + the vendor-independence case)
```

Anti-vacuity in both directions: a **same-vendor** close write must NOT clear the bar (`test_a_same_vendor_close_write_does_not_clear_the_bar`), or every universe sweep would cold-start the OHLCV cache; and the read-side `source` filter is pinned separately (`test_the_read_filter_still_rejects_the_other_vendors_row`), so the two halves cannot cover for each other.

Re-run after the fixes: `tests/services/ tests/chain/ tests/test_feature_flag_fliplist_drift.py` → **1730 passed, 2 skipped, 2 failed** (the same two pre-existing `test_generation_pipeline` `identity_events` fixture failures — unrelated, and they fail identically in isolation). Docs guards → 87 passed. `ruff check` + `ruff format --check` clean. No network, no infra.

## Owner steps when you want the cutover

1. Merge this.
2. Set `MARKET_DATA_DAILY_PROVIDER=tiingo` (task-def env or SSM) with `TIINGO_API_TOKEN` present. Putting it in `infra/ecs.tf` now fails `test_ecs_backend_secrets.py` until you edit `FORBIDDEN_WHY` — deliberately, so the flip is a stated decision rather than a line in a config diff.
3. Prove one pull on the daily seam and paste it on #1798. **Since the `main` merge there is a packaged proof:** `python scripts/verify_market_data.py --no-cache` (from a checkout, token in the env), driven by `docs/runbooks/market-data-provider-proof.md`. It reads a fixed ten-bar window whose right answer is a constant, prints **seam + vendor** and exits non-zero on a wrong one, so it can say NO. The by-hand form below still works and is the one that runs from inside a task (the repo-root `scripts/` tree is not in the backend image): exec into a serving backend task (`enable_execute_command = true`), then

   ```sh
   python -c "from datetime import date,timedelta; from archimedes.services.market_data_provider import get_provider,provider_name; p=get_provider(seam=\"daily\"); e=date.today(); s=e-timedelta(days=30); d=p.get_daily_ohlcv(\"SPY\",s.isoformat(),e.isoformat()); print(\"vendor:\",provider_name(\"daily\"),\"| rows:\",len(d),\"| first:\",d.index[0].date(),\"| last:\",d.index[-1].date(),\"| last close:\",float(d[\"Close\"].iloc[-1]))"
   ```

   and a second one-liner that prints the written row's `source` / `fetched_at` — the provenance half of the proof. `vendor: tiingo` + `sources: ['tiingo']` + a fresh `fetched_at` is the flip proven.
4. Then, and only then, flip the claims-ledger row.

Expect a cold `asset_daily_bars` cache on the first run after the flip. **Both halves of the cache are vendor-safe, and they are separate guards** (the write half landed in the verification pass, `4977997d`):

- the `source` filter does its job on **reads** — a `yfinance` row is never served to a Tiingo caller;
- the **write** path is now vendor-safe too. `asset_daily_bars` is unique on `(symbol, trade_date)`, so the close-only writer (`get_daily_close_batch` → `_write_cached_series`) lands *on* the previous vendor's row and knows only `close`. It used to assign `close` + `source` and leave `open/high/low/volume` alone, producing a bar whose Close was Tiingo's and whose OHLV was still yfinance's, stamped `source='tiingo'` — a blend the read filter cannot see, because the row now claims to *be* the vendor being asked for, and one `portfolio_backtester._fetch_price_panel` would have graded (it consumes `Volume` as well as `Close`). A cross-vendor close-only write now **clears** the four columns it does not know and logs which vendor it displaced; the existing partial-bar guard turns that into a miss and the next OHLCV read refetches from the new vendor. Same-vendor refreshes are untouched, so the routine sweep does not evict the OHLCV cache.

## Reconciliation with `main` (merge `b7c57a5a`)

`main` now carries #1806, which wired `TIINGO_API_TOKEN` as a task-definition secret on both registrars and shipped `scripts/verify_market_data.py` + `docs/runbooks/market-data-provider-proof.md`. Two conflicts and one thing git could not see:

| What | Resolution |
| --- | --- |
| `.env.example` + the `TiingoAPIKeyMissingError` text | Kept this PR's seam wording (`MARKET_DATA_DAILY_PROVIDER`, or `MARKET_DATA_PROVIDER` on the daily seam when unset); replaced this PR's "NOT a task-def secret yet (#1806)" with #1806's now-true claim (wired on both paths → a task-launch dependency, not only the best-effort boot load). |
| **`scripts/verify_market_data.py` was broken by the merge, silently** | It calls `get_provider()` / `provider_name()`; this PR makes `seam` required. Git merged both files cleanly and the operator's proof command died on line one with a `TypeError` — while all 18 of its tests stayed green (verified: `pytest --collect-only` on `main`'s copy of the file collects 18), because they stub `get_provider` with `lambda: provider` and a stub accepts whatever it is handed. Fixed: the script names the daily seam and prints it; `--no-cache` unwraps *all* wrappers (the nesting is two deep now, so one `_inner` hop returned the **cache** while printing "bypassed"); `TestItSpeaksTheRealSeamApi` runs the script's own calls against the real module with only the vendor registry swapped, so the next signature change fails in CI rather than in prod. |
| The runbook's § Step 3 | It said "there is no way to flip only the paid seam with the variables that exist" and that the flip "needs Tiingo's intraday methods implemented first, **or a second variable so the two seams can diverge**". This PR is that second variable. Rewritten — the two superseded sentences are quoted and dated rather than deleted — along with § Step 2's rehearsal command, § Rollback, § Related, and the `docs/README.md` / `docs/runbooks/README.md` rows. |

Guards for the fix, shown red then restored green (22 passed): dropping the seam argument from either call → 11 and 12 failed on `TypeError`; single-`_inner` unwrap → 2 failed **and only in the new class**, which is why it was added; `SEAM = intraday` → 2 failed. Adding `MARKET_DATA_DAILY_PROVIDER` to `ecs.tf`'s backend `environment` → the new forbidden-name row fails; reverted by edit, `infra/ecs.tf` unchanged from `main`.

Part of #1798


dbrowneup:
### Blocker 1 — fixed in `4977997d`: a vendor flip could still produce a mixed-vendor bar, via the **write** path

`backend/archimedes/services/market_data_provider.py` — `_write_cached_series`

**What was wrong.** `asset_daily_bars` is unique on `(symbol, trade_date)`, so the close-only writer behind `get_daily_close_batch` lands *on* the previous vendor's row rather than beside it — and it knows only `close`. It assigned `close` + `source` + `fetched_at` and left `open/high/low/volume` untouched. After `MARKET_DATA_DAILY_PROVIDER=tiingo`, the close-only universe sweep (`strategy_signal_evaluator`, daily seam) runs before the generation panels, on the same seam and the same tickers, so every warm prod row became **Close = Tiingo, OHLV = yfinance, `source='tiingo'`**.

`_read_cached_ohlcv`'s `source` filter cannot catch that: the row now claims to *be* the vendor being asked for. Its partial-bar guard only rejects NULLs, so the blend read back as a valid hit and `fusion_market_data._fetch_one` / `portfolio_backtester._fetch_price_panel` were served it without ever calling Tiingo. `_fetch_price_panel` consumes `Volume` as well as `Close`, so this reached graded numbers — precisely the failure `_read_cached_ohlcv`'s own docstring says the seam exists to prevent.

Pre-existing (the function was byte-identical to `main`), and a blocker for this PR anyway, because this PR is what makes the flip a supported operation and the body asserted the opposite.

**The fix.** A cross-vendor close-only write now clears the four columns it does not know, and logs which vendor it displaced:

```python
        if row.source != source:
            # A different vendor wrote this row and we only know `close`.
            # Drop the old vendor's OHLV rather than leave a bar stitched
            # from two vendors under one `source` label — see the docstring.
            displaced_vendors.add(row.source)
            row.open = row.high = row.low = row.volume = None
```

The row becomes the honest partial bar it is, the *existing* partial-bar guard turns it into a miss, and the next OHLCV read refetches the whole range from the new vendor. **"Write a new row" is not available** — the unique constraint forbids it — and "skip" would leave a stale close, so clearing is the only shape that never blends. `_write_cached_ohlcv` needed no change: it assigns every column, so it *replaces* a bar rather than blending one; its docstring now says that explicitly rather than leaving a future reader to re-derive it.

`_read_cached_ohlcv` already filtered on `AssetDailyBar.source == source` — **confirmed**, both readers do — and that half is now pinned by its own test rather than assumed.

**The probe, made permanent.** Your stub-only probe is now `TestAFlipNeverProducesAMixedVendorBar` in `backend/tests/services/test_market_data_seams.py`: two stub vendors and a throwaway SQLite behind the *real* `get_provider` → `SeamRoutedProvider` → `CachingMarketDataProvider` stack, driving prime-on-yfinance → flip → close-only sweep → OHLCV read. It asserts no mixed bar *and* that the daily seam actually calls the daily vendor for OHLCV after the flip. Plus a write-path unit guard in `test_market_data_provider.py`.

Mutation — restore the partial overwrite:

```
FAILED tests/services/test_market_data_seams.py::TestAFlipNeverProducesAMixedVendorBar::test_the_daily_seam_refetches_from_the_new_vendor
E   AssertionError: the OHLCV read was served from cache after a vendor flip — the close-only sweep
    left a row stamped source='tiingo' whose OHLV is still yfinance's
E   assert []
E    +  where [] = <_TwoColumnVendor object>.ohlcv_calls

FAILED tests/services/test_market_data_seams.py::TestAFlipNeverProducesAMixedVendorBar::test_no_row_survives_the_sweep_with_a_foreign_vendors_ohlv
E   AssertionError: MIXED-VENDOR BAR: source='tiingo' close=400.0 but open=900.1 volume=11.0 came from another vendor
E   assert (900.1, 901.1, 899.1, 11.0) == (None, None, None, None)

FAILED tests/services/test_market_data_provider.py::TestDailyOhlcvCache::test_cross_vendor_close_write_clears_the_old_vendors_ohlv
```

Anti-vacuity in both directions, because a one-sided guard here is easy to write and worthless: `test_a_same_vendor_close_write_does_not_clear_the_bar` fails if the clearing is not scoped to a vendor *change* (otherwise every routine sweep would evict the OHLCV cache), and `test_the_read_filter_still_rejects_the_other_vendors_row` pins the read half separately, so the two cannot cover for each other.

— via the owner's session

dbrowneup:
### Blocker 2 — fixed in `4977997d`: the seam table passed on prose, so it could not catch a wrong seam at the backtester

`backend/tests/services/test_market_data_seams.py` — `TestEveryCallSiteNamesItsSeam`

**What was wrong.** `test_function_asks_for_its_seam` was `assert f'get_provider(seam="{seam}")' in inspect.getsource(func)` — and `getsource` includes the **docstring**. `portfolio_backtester._fetch_price_panel`'s docstring quotes the call verbatim at `portfolio_backtester.py:77`, so that row was satisfied by prose, not by the call at `:86`.

You proved it: mutating `:86` to `seam="intraday"` left the module fully green (43 passed). Nothing else covered it either — `TestCallSiteSeamsAtRuntime` reaches only `fusion._fetch_one` and `oracle._fetch_sp500_moving_averages`, and at runtime the intraday seam legitimately serves `get_daily_ohlcv`, so the generation-path backtester would have run on the intraday vendor with no error and no log line.

I took the structural option rather than the cheap one, for the reason you gave: adding a runtime case for `_fetch_price_panel` would have left the other four rows exposed to the same docstring accident the moment someone documents their own call site.

**The fix.** `_seams_requested_by(func)` parses `textwrap.dedent(inspect.getsource(func))` with `ast` and collects the `seam` keyword's value off each real `ast.Call` node — which a docstring cannot reach. Both parametrized tables (`test_function_asks_for_its_seam` and `test_oracle_updater_is_entirely_on_the_intraday_seam`) use it. The assertion is `set(seams) == {seam}` plus non-empty, so it is red on three distinct defects: the wrong seam, no `get_provider` call at all, and a function asking **two** seams — which is the vendor mix inside one run the ADR forbids, and belongs red in the table rather than discovered in a panel.

Mutation — your exact one, `portfolio_backtester.py:86` → `seam="intraday"`, which was **green before this commit**:

```
FAILED tests/services/test_market_data_seams.py::TestEveryCallSiteNamesItsSeam::test_function_asks_for_its_seam[archimedes.services.portfolio_backtester-_fetch_price_panel-daily]
tests/services/test_market_data_seams.py:389: in test_function_asks_for_its_seam
E   AssertionError: archimedes.services.portfolio_backtester._fetch_price_panel asks ['intraday'], expected ['daily']
E   assert {'intraday'} == {'daily'}
E     Extra items in the left set: 'intraday'
1 failed, 46 passed
```

`test_no_backend_call_site_asks_without_a_seam` (the sibling at `:357`) is untouched — it was already `ast`-based and already adversarial; this just brings the per-row table up to the same standard. The module docstring's mutation list now names both this mutation and the mixed-bar one, so the next reader sees what turns each guard red without re-deriving it.

— via the owner's session

dbrowneup:
### Gap — fixed in `4977997d`: the flip recipe now names a proof command

`docs/adr/market-data-sourcing.md` § "Proving the flip" (new), and the `MARKET_DATA_DAILY_PROVIDER` row in `docs/operations/feature-flag-fliplist.md` now points at it.

You were right that "prove one pull and paste vendor + row count + first/last bar date" left the owner to invent a live-vendor proof against prod, and that no script exists (`grep -rl tiingo scripts/` is still empty — I did not add one; a one-liner that needs no maintenance beats a script nobody will re-read).

The recipe leans on the fact already established in [`docs/runbooks/curated-backtests.md`](https://github.com/aprin-labs/archimedes/blob/dbrowneup/1798-provider-seams/docs/runbooks/curated-backtests.md) § Alternative: the backend service has `enable_execute_command = true`, and that runbook's rule is "exec is for reading state, not for a library run" — a single-symbol daily pull is squarely the reading-state case.

```bash
CLUSTER=archimedes
TASK=$(aws ecs list-tasks --cluster "$CLUSTER" --service-name archimedes-backend \
  --desired-status RUNNING --query 'taskArns[0]' --output text)
aws ecs execute-command --cluster "$CLUSTER" --task "$TASK" --container backend \
  --interactive --command /bin/sh
```

Then inside the task, the pull itself — through the real seam, so it proves the routing and not just the adapter:

```sh
python -c "from datetime import date,timedelta; from archimedes.services.market_data_provider import get_provider,provider_name; p=get_provider(seam=\"daily\"); e=date.today(); s=e-timedelta(days=30); d=p.get_daily_ohlcv(\"SPY\",s.isoformat(),e.isoformat()); print(\"vendor:\",provider_name(\"daily\"),\"| rows:\",len(d),\"| first:\",d.index[0].date(),\"| last:\",d.index[-1].date(),\"| last close:\",float(d[\"Close\"].iloc[-1]))"
```

and the row it wrote, which is the provenance half — the part that distinguishes "Tiingo answered" from "the cache answered":

```sh
python -c "from archimedes.db import get_session; from archimedes.models.asset_daily_bars import AssetDailyBar as B; s=get_session(); r=s.query(B).filter(B.symbol==\"SPY\").order_by(B.trade_date).all(); print(\"cached rows:\",len(r),\"| sources:\",sorted({x.source for x in r}),\"| first:\",r[0].trade_date,\"| last:\",r[-1].trade_date,\"| newest fetched_at:\",max(x.fetched_at for x in r))"
```

`vendor: tiingo` + `sources: ['tiingo']` + a fresh `fetched_at` is the flip proven. The ADR also says what a *mixed* `sources: ['tiingo', 'yfinance']` reading means — the sweep has not reached every symbol yet, which is expected mid-rollout and is **not** a mixed bar, since after blocker 1 no single row can hold two vendors' columns.

Both one-liners are single-quote-free by construction (checked, so the `sh` wrapper cannot break them), every symbol in them was import-resolved locally, and the shell quoting was executed against a stub. Nothing was run against Tiingo or yfinance from this session.

**Also corrected in the PR body, from your accuracy notes:** "all nine call sites" → **ten** `get_provider(seam=…)` invocations across six modules (`oracle_updater` is four of them; the module docstring's "Six call sites" counts modules and stands); and `ruff format --check` → **738** files from the repo root, not 683.

The last minor note — analytics-engine logging `unknown MARKET_DATA_PROVIDER='tiingo'` under a *global* flip — I left as-is deliberately. It is log noise on a path no intraday call site reaches, the ADR's last table row already says the CLI seam registers no Tiingo adapter, and the recommended `MARKET_DATA_DAILY_PROVIDER=tiingo` recipe avoids it entirely. Say the word if you want it silenced rather than documented.

— via the owner's session

Issue #1448, opened 2026-08-20
[frontend][docs] Stale '#588 pending' claims in .well-known/agent.json and agent-api.md


dbrowneup:
Follow-up to PR #1447, which fixed the same class of stale claim in `agent_manifest_routes.py` (the served manifest). Two sibling surfaces still assert that #588 (commit-reveal ABI) is pending/open, which has been false since the T3.2 redeploy (2026-07-09) and #588's closure (2026-07-14):

- `ui/public/.well-known/agent.json` — the static discovery document references "#588" stale text in **5 places**.
- `docs/agent-api.md` ~line 363 — states #588 "is still open".

**Fix:** align both with the statuses PR #1447 established in the served manifest (deploy/marketplace/monitor = live; no payment-scheme advertisement until the settle proof exists). Keep the static agent.json and the served manifest consistent — consider a test asserting the two agree on group statuses so they cannot drift apart again.

**Acceptance:** `grep -rn '588' ui/public/.well-known/ docs/agent-api.md` returns no pending/open claims; consistency test passes and is mutation-checked (change one side → test fails).

PR #1815, opened 2026-09-02
[verify][security] Strict, chronological, finite input for /api/rigor/verify with explicit rejection codes (#1803)

### LINKED ISSUE #1803: [verify][security] /api/rigor/verify: dates are never used for ordering or dedupe (a shuffled series games the 70/30 OOS split), trials is unbounded, NaN reads as 'unavailable' — harden + document before promotion
Owner direction (Dan, 2026-09-02): the bring-your-own-returns endpoint is a large attack surface; harden and document before promoting; nobody uses it today.

**Facts (main @ 9cb868eb, tested in-process):** schema `RigorVerifyRequest` (`rigor_verify_routes.py:153-172`): `returns` 1–2,600 rows, `trials: int ge=1` with NO upper bound; `ReturnPoint.date: str` with no format/ordering/uniqueness check. Nothing eval'd/pickled/shelled/persisted; payload not logged (one malformed date is `%r`-logged via `rf_series.py:253,288`). Compute is ~10 ms at the cap (DSR+HAC 3.9 ms, OOS 2.0 ms) — not a CPU DoS. Auth = cookie or `archim_` key; 5/min rate limit. Findings: (1) unsorted/duplicate dates are accepted and the 70/30 split is POSITIONAL (`_rigor_helpers.py:594`), so a caller can shuffle rows to put the best 30 % in the holdout → `oos_consistency: pass`; (2) JSON `NaN`/`Infinity` literals are accepted; DSR/OOS legs catch them (`not_evaluable`) but `in_sample_sharpe` has no isfinite guard and renders `null` — reads as "unavailable" instead of "poisoned input"; (3) `trials=10**18` → `dsr=-inf` → `not_evaluable` (no crash); (4) the response reveals thresholds (fine — they are public in the docs). CLI `archimedes verify RETURNS_CSV` (`cli.py:569`) and MCP `archimedes_rigor_verify` (`tools.py:87`) call it; no UI; docs at `docs/api/strategies-and-rigor.md` describe it accurately but nobody links to it.

**Hardening (each one line-ish):** strict ISO date parsing + sort + reject duplicates + reject non-finite (pydantic validators on `ReturnPoint`/`RigorVerifyRequest`); `trials: le=10_000`; a chronological OOS split keyed on the parsed dates; explicit `input_rejected` reason codes; keep the 5/min limit; document minimum useful length (DSR needs a meaningful T — state it) and a worked CLI example; decide whether 1 year is enough to be worth a verdict (product call).

dbrowneup:
Closes #1803 (owner decision Q7: harden first; the last open item — "is one trading year enough to be worth a verdict" — is answered below).

**The 250-bar rule (owner decision, round 3).** The minimum evaluation window is **250 daily bars — one trading year**. Under it `POST /api/rigor/verify` returns a **typed refusal** — `422`, `detail.reason: "window_too_short"`, with `bars_received` and `bars_required` as fields as well as in the sentence — and **never a verdict**, not even a verdict with a warning attached: `passes` is a field the CLI exits on and CI gates on, and a caveat in prose beside it is read by none of them. The CLI and the MCP tool mirror the floor and refuse locally with the same code and the usage exit code (`2`), spending no request.

> ## ⚠️ DEPLOY PRECONDITION — do not ship this before the #1749 WAF apply
>
> **This PR takes `POST /api/rigor/verify` from partially working in prod to answering nothing, unless the terraform lands first.** Round 2 of the vet caught that the PR body never said so.
>
> The ALB's `AWSManagedRulesCommonRuleSet` blocks any body over `SizeRestrictions_BODY`'s **8,192-byte** inspection limit. Before this change a caller could send ~160 bars and get a 200; measured at the edge on 2026-09-01: `160 rows / 8,076 B -> 200`, `165 rows / 8,326 B -> 403` (HTML from `awselb/2.0`), `252 rows -> 403, every time`. With the 250-bar floor the **smallest body the schema accepts** — 250 rows, most-compact JSON, no whitespace — is **10,774 B**. There is no longer any legal request that survives the ALB without the exemption. The endpoint is not degraded without it; it is dead.
>
> The exemption exists in `infra/waf.tf` (PR #1786 for #1749, merged to main 2026-09-02) — **but terraform state is not the repo**, and that PR's own body says it: *"The terraform half is **not live until applied**"* and *"nobody has seen the fixed path go green at the edge yet."*
>
> **Ordering:**
> 1. `infra/apply.sh` — plan first; expect **one in-place change to `aws_wafv2_web_acl.main`**.
> 2. `infra/apply.sh --apply`.
> 3. Confirm with a real 250-row `POST /api/rigor/verify` through `https://archimedes-arc.com`. A `403` whose body is HTML from `awselb/2.0` is the WAF, not FastAPI.
> 4. Then deploy this branch. **Or hold the merge until step 3 is confirmed** — either order is fine, this one is not.
>
> The coupling is now discoverable from the repo rather than only from a PR body: `backend/tests/test_waf_verify_body.py::test_the_waf_exemption_is_load_bearing_for_every_request_not_just_long_ones` pins the fact that makes the ordering matter (minimum accepted body > the edge limit), and its docstring carries the runbook above. No test can see terraform state, and it does not pretend to.

`POST /api/rigor/verify` is the one route that takes a caller's raw numbers and hands back a verdict on them. Nobody uses it today, which is exactly why the input contract gets tightened now rather than after promotion.

## What was wrong

The schema carried `ReturnPoint.date` but never used it for anything. The walk-forward split is **positional** — `_rigor_helpers.compute_oos_sharpe` slices `arr[split:]` — so a caller could sort their own series by return, park the best 30% in the holdout and collect `oos_consistency: pass` on numbers that fail chronologically. Proved in the test, on the same numbers:

```python
assert compute_oos_sharpe(series) <= 0.0       # chronological: FAIL
assert compute_oos_sharpe(sorted(series)) > 0.0 # shuffled:      PASS
```

`trials` had no upper bound, so `trials=10**18` drove the DSR deflation to `-inf` and the leg to `not_evaluable` — a caller-controlled way to erase a FAIL. JSON `NaN`/`Infinity` literals parsed; the DSR/OOS legs caught them, but `in_sample_sharpe` had no isfinite guard and rendered `null`, which reads as "unavailable" rather than "poisoned input".

## What this does

The series is now made to **be a series** before it is graded, and every refusal carries a machine-readable reason: `invalid_date`, `duplicate_date`, `unsorted_dates`, `non_finite`, `out_of_range`, `window_too_short`, `too_many_rows`, `trials_out_of_range`.

| Rule | Code | Why here |
|---|---|---|
| strict `YYYY-MM-DD` (parsed to `datetime.date`) | `invalid_date` | pydantic's lax `date` also takes epoch seconds, and `date.fromisoformat` takes `20240102` / `2024-W01-1` on py≥3.11 — the rf-series resolution would have to guess |
| no repeated dates | `duplicate_date` | two bars on one date is not a daily series, and a repeat can be inserted anywhere without breaking monotonicity |
| strictly ascending | `unsorted_dates` | **rejected, not sorted** — see below |
| finite returns | `non_finite` | a non-finite bar cannot be graded, only refused |
| `abs(r) <= 1.0` | `out_of_range` | an unlevered daily bar cannot lose >100%, and >+100% in a day means the column is percentages/annualized/prices — it silently inflates the Sharpe the verdict rests on |
| `>= 250` rows (one trading year) | `window_too_short` | the owner's minimum evaluation window: under it there is no verdict at all, only a refusal naming `bars_received`/`bars_required`. A **product** floor, deliberately above both arithmetic ones (DSR needs 4 bars, the walk-forward split ~70) — at 250 **both** runnable legs actually run |
| `<= 2600` rows | `too_many_rows` | the #1749 cap, unchanged in value and behaviour, now with a code |
| `1 <= trials <= 10000` | `trials_out_of_range` | keeps the deflation finite |

**Round 2 closed two ways the `trials` bound answered in the wrong shape** (it always failed closed — it just did not carry the documented code). `_trials_bounds` ran on the RAW value and tested `isinstance(value, int)` only, so `1e18` and `"10000000000"` — both ordinary spellings; `json.dumps(10**18)` emits `1e+18` — sailed past it onto the declarative `le`, whose type is `less_than_equal`, *not* one of this route's codes, so the refusal degraded to the generic list shape and the MCP mapped it to `invalid_request`. The guard now widens the value the way pydantic's lax mode is about to, before bounding it. And `trials: true` returned **200** with `trials: 1` (pydantic reads `bool` as `int`); a route that refuses `20240102` rather than guessing must not guess a boolean into a count, so it is refused.

Plus: `in_sample_sharpe` gets the same isfinite guard the other legs carry (the leg reports `not_evaluable`, never a pass/fail beside a `null`), and the endpoint threads `pt.date.isoformat()` — the parsed, canonical, ascending index — into `_resolve_gate_rf`, never the caller's raw text.

**Why REJECT rather than sort.** Sorting server-side would return a verdict on a series the caller never sent. "These are my returns, in order" is the claim the OOS leg is grading, so a shuffled body has to be refused, not repaired. The property is pinned both ways: an adversarial test that a shuffle which *would* have flipped the verdict is refused, and a general one that a permutation is either rejected or yields its sorted twin's verdict.

**`DSR_MIN_BARS`.** There was no named constant — the floor was the literal `4` repeated across three DSR guards in `_rigor_helpers.py`. It is now named there, used by those three guards, and imported by the route. The route's own floor is now the 250-bar window rather than that constant, but it is written `max(_MIN_WINDOW_BARS, DSR_MIN_BARS)`, so the product floor can never sit *below* the gate's arithmetic one — by construction, not by comment.

## The 422 had to be fixed to be deliverable

FastAPI's default validation error echoes the offending `input` verbatim. That is (a) **unserialisable for the exact input this change exists to reject** — `JSONResponse` uses `allow_nan=False`, so a NaN payload would have 500'd on its way to its own 422 — and (b) a 122 KB echo of a 122 KB body. Caught by writing the NaN test first:

```
E  fastapi.exceptions.RequestValidationError: 1 validation error:
E    {'type': 'non_finite', 'loc': ('body','returns',3,'daily_return'), ..., 'input': nan}
E  ValueError: Out of range float values are not JSON compliant: nan
```

A `route_class` **scoped to this router** (an app-wide exception handler would rewrite every other route's 422, which is not this change's business) renders instead:

```json
{"detail": {"error": "input_rejected", "reason": "unsorted_dates",
            "reasons": ["unsorted_dates"], "message": "…", "loc": ["body", "returns"]}}
```

Nothing is echoed back.

## Surfaced end to end — and refused before the request when it can be

`archimedes verify` prints the server's own sentence plus a remedy, and carries `reason` in `--json` (real run against a mocked 422):

```
the API rejected the input (unsorted_dates): returns must be in ascending date order; row 4 (2025-03-03) precedes row 3 (2025-11-18). The walk-forward out-of-sample split is positional, so row order IS the time order it grades — a shuffled series is refused rather than sorted, because sorting would return a verdict on a series you did not send.
Sort the CSV by date, oldest first (`sort -t, -k1,1 returns.csv`). The out-of-sample split is positional, so row order IS the time order it grades; the server refuses to re-sort for you because that would grade a series you did not send.
exit=2
```

**Round 2: the CLI now holds the CSV to the same contract before it builds the request.** It had to. `float()` accepts `nan`/`inf`/`Infinity` case-insensitively, and httpx serialises with `json.dumps(allow_nan=False)`, so a poisoned CSV never reached the server at all — it raised inside the request build, printed a traceback, and exited **1**, the one code `exits.py` reserves for a real failing verdict. The `non_finite` code this PR added was unreachable from the CLI, and its test mocked a server 422 — i.e. mocked past the broken step. Now (real run, no mock, no session needed to reach it):

```console
$ archimedes verify poison.csv --json ; echo "exit=$?"
{"ok": false, "command": "verify", "error": "non_finite_returns", "message": "'poison.csv' row 2 has daily_return 'nan', which is not a finite number and cannot be sent as JSON", "reason": "non_finite"}
exit=2
```

Six of the eight codes are mirrored locally. `error` is the CLI's own stable string; `reason` is the SERVER's code for the identical refusal — the same split `--trials` already used, so a CI job branches on one `reason` whichever side caught the row:

| Bad row in the CSV | local `error` | `reason` (server's code) | exit |
|---|---|---|---|
| fewer than 250 data rows | `window_too_short_returns` | `window_too_short` | 2 |
| `nan` / `inf` / `-inf` / `Infinity`, any case | `non_finite_returns` | `non_finite` | 2 |
| `abs(r) > 1.0` | `out_of_range_returns` | `out_of_range` | 2 |
| `20260102`, `2026-W01-1`, `2026-1-2`, `2026-02-30` | `invalid_date_format` | `invalid_date` | 2 |
| the same date twice | `duplicate_date_rows` | `duplicate_date` | 2 |
| a row out of order | `unsorted_date_rows` | `unsorted_dates` | 2 |

**Round 3 adds the window to the mirror, and it is checked FIRST** — the order the server's own validators run in (`_row_count_bounds` is `mode="before"`), so a file that is both short *and* malformed gets the same code from both sides. Real run, no mock, no session needed to reach it:

```console
$ archimedes verify short.csv --json ; echo "exit=$?"
{"ok": false, "command": "verify", "error": "window_too_short_returns", "message": "'short.csv' has 249 data rows; the minimum evaluation window is 250 daily bars (one trading year). A shorter series is refused, not graded with a caveat", "reason": "window_too_short"}
exit=2
```

**Still NOT mirrored: `too_many_rows`.** The ceiling tracks the edge's payload budget (#1749) — infrastructure, not a published promise — so a CLI that hard-coded it could refuse a series a newer server accepts. The FLOOR is mirrored because it is the opposite kind of number: a product rule, printed in `--help`, and making a caller spend one of five requests a minute to be told it helps nobody. An EMPTY file keeps its own `empty_returns` answer: zero rows almost always means the wrong file, and that is the more useful thing to say. `--trials` keeps its local bound because it is an argument the CLI itself defines, not a property of the caller's file. And this is a mirror, never a replacement: the server re-checks all of it, its sentence is the one printed when the request does go out, and nothing is sorted, deduplicated or coerced client-side either.

### The full exit-code contract for `verify`

`exits.py`'s split is 1 vs everything else, and this is the table that has to hold:

| Exit | Name | Means |
|---|---|---|
| `0` | `OK` | Every runnable leg ran and passed (capped: PBO and look-ahead never run here) |
| `1` | `GATE_FAILED` | **A leg actually failed.** A real verdict about the strategy — and nothing else may produce this code |
| `2` | `USAGE` / `AUTH` | The input was refused (locally or by the server), or there is no session. No verdict was produced |
| `3` | `NOT_IMPLEMENTED` | `verify --local` — the local engine is not published |
| `4` | `INCOMPLETE` | **No leg failed** and not every runnable leg ran. Round 3 removes SHORTNESS as a cause of this — the 250-bar window is above the ~70 the walk-forward split needs — so it now means a leg that could not run on the NUMBERS (a zero-variance series, say) |

Before this push a malformed CSV produced `1`. It now produces `2`, which is the whole point of the change.

The MCP tool gets the same pre-check, for a sharper reason: `client.request` catches `httpx.HTTPError` only, so a non-finite bar raised `ValueError` straight out of the handler — and that module's stated contract is that handlers never raise. It returns the server's own code (`non_finite`, `out_of_range`, `invalid_date`, `duplicate_date`, `unsorted_dates`) with the same remedy line the 422 path uses, now shared through `errors.input_rejected_remedy()`.

## Guards shown red

Every one reverted after measuring.

| Mutation | Result |
|---|---|
| remove the duplicate-date check | `test_reason_code_duplicate_date` FAILED — `assert 200 == 422` |
| remove the ordering check | `test_reason_code_unsorted_dates` FAILED, `test_a_shuffle_that_would_have_flipped_the_oos_verdict_is_refused` FAILED |
| remove the `trials` upper bound | `test_reason_code_trials_out_of_range[10001]` FAILED, `[1000000000000000000]` FAILED |
| remove the length floor (validator **and** `min_length`) | `test_reason_code_window_too_short` FAILED |
| remove the `in_sample_sharpe` isfinite guard | `test_non_finite_in_sample_sharpe_is_not_evaluable_never_a_null_metric` FAILED |
| relax the strict ISO regex | `test_reason_code_invalid_date[20240102]` FAILED, `[2024-W01-1]` FAILED |
| remove only the declarative `le` (validator intact) | 45 passed — the backstop is a backstop, the validator is the primary |

Round 2's three, same rule — each applied, measured, reverted:

| Mutation | Result |
|---|---|
| remove the CLI's local non-finite check | **8 CLI tests FAILED**. The `nan` cases fail as `assert 1 == 2` — the crash, and the wrong exit code, reproduced exactly; the `inf` cases fall through to the range check and answer `out_of_range_returns` |
| disconnect the MCP row pre-check | **11 MCP tests FAILED**, three of them with `UnexpectedToolError: Error executing tool` — the handler really does raise |
| revert `_trials_bounds` to the `isinstance(int)`-only form | **9 backend tests FAILED**: `1e18`, `1e9`, `"10000000000"`, `"0"`, `" 10001 "` and the three non-finite literals all degrade to the generic `less_than_equal` list shape, and `trials: true` returns **200** |

Round 3's four, same rule — each applied in a scratch worktree at this commit, measured, reverted by editing, and the tree re-verified clean and green afterwards:

| Mutation | Result |
|---|---|
| remove the server window guard (validator **and** `min_length`) | **5 backend tests FAILED**: `test_reason_code_window_too_short`, `test_one_bar_under_the_window_is_refused_and_names_both_counts`, `test_the_four_bar_one_evaluable_body_is_now_refused_outright` and `test_empty_returns_rejected_422` all `assert 200 == 422` — including the empty body, which is graded rather than refused — plus `test_length_bounds_are_declared_on_the_field_as_the_backstop` with `KeyError: 'minItems'` |
| stop surfacing `bars_received`/`bars_required` as fields | **3 backend tests FAILED** with `KeyError: 'bars_received'`. The English sentence still says both numbers; the fields a client branches on are gone, which is the difference the test exists for |
| disable the CLI's local window check | **3 CLI tests FAILED** (`[1]`, `[30]`, `[249]`) as `assert 1 == 2` — the `_forbid_network` guard fires, i.e. a short series really does reach the request builder, and the exit code becomes `1`: a short file reported to CI as a failing verdict |
| disable the MCP window pre-check | **2 MCP tests FAILED** with `KeyError: 'error'` — the 3-bar and 249-bar calls spend a request instead of refusing |
| revert the `passes` quorum to the pre-#1481 rule | **1 backend test FAILED**: `test_a_leg_that_could_not_run_never_produces_a_pass` — `assert True is False`. The #1481 regression guard survived the window change; it is just reached by a forced-unevaluable leg now that a 4-bar body cannot be submitted |

And the **red CI check**, measured the same way (deptry 0.25.1, the exact invocation `.github/workflows/import-guard.yml` runs):

```
backend/archimedes/api/rigor_verify_routes.py:96:1: DEP001 'pydantic_core' imported but missing from the dependency definitions
Found 1 dependency issue.
```

With `pydantic-core` declared in `backend/requirements-base.txt`: `Success! No dependency issues found.`

(Precision, from the vet: the code shown above is what **CI** reports, because the import-guard job installs only `ruff deptry`. In a dev env where `pydantic-core` is present as a transitive of `pydantic`, the same invocation reports **DEP003** — `'pydantic_core' imported but it is a transitive dependency` — rather than DEP001. Red either way, and the declaration fixes both.)

Every real refusal was also exercised against the running app; each returned 422 with the right code and a message naming the offending row.

## The red import-guard check (#1815's CI failure)

`from pydantic_core import PydanticCustomError` tripped deptry's **DEP001**: the module is imported directly and the package was not declared. Both alternatives were checked before choosing:

- **Reach it through pydantic's public surface** — there isn't one. `pydantic.PydanticCustomError` does not exist in 2.13.4 (`ImportError: cannot import name 'PydanticCustomError' from 'pydantic'`); the only re-export is `pydantic.types.PydanticCustomError`, which is a private implementation detail that would satisfy the guard by hiding the import rather than declaring it.
- **Ignore the rule** (`--per-rule-ignores 'DEP001=…|pydantic_core'`) — that weakens a guard for a real direct dependency.

So the package is declared, which is what DEP001 asks for. `pydantic` pins `pydantic-core` EXACTLY (2.13.4 → `pydantic-core==2.46.4`), so the floor never selects the version — pydantic's own pin does; it exists so the dependency is visible to every consumer of `requirements-base.txt` (Docker/CI via `requirements.txt`, local dev via `environment.yml`, both by `-r`). deptry resolves `pydantic-core` → `pydantic_core` with its default dash→underscore guess, no map entry needed — confirmed by running the CI invocation locally, red then green.

## Docs

`### POST /api/rigor/verify` now exists in `docs/api/strategies-and-rigor.md` (and came **off** `test_api_docs_drift.py`'s undocumented allowlist), with the limits table, a worked 252-bar CLI example whose numbers were produced by running this code — the series is disclosed as synthetic, seed 1803, which is why its Sharpe is implausibly good — and a paragraph on what a capped/incomplete verdict can and cannot claim. `docs/agent-quickstart.md` gets the agent-facing version plus an error-table row.

Two true-claims fixes found while reading, both flagged rather than folded in silently:

- `docs/agent-api.md` still stated the **pre-#1481** `passes` rule ("no evaluable check failed *and* at least one check was evaluable"). That has been false since the quorum landed. Corrected.
- `docs/claims-ledger.md` cited `rigor_verify_routes.py:273` / `:289`, which this change moves. Re-pointed. (`test_claims_ledger.py` only checks the line exists, so it stayed green while being wrong.)

**Round 2 fixed two things round 1's docs got wrong about this very change:**

- **Both new `curl` examples sent 2 rows — which this PR's own floor refuses.** `{"returns": [2 rows]}` → `422 reason=too_short, "returns has 2 rows; the minimum is 4"`, in a block sitting ~10 lines above the table that states the floor. Copy-paste failed. Round 2 fixed them to four bars — and **round 3's 250-bar window refuses those too**, which is the same lesson landing twice. So neither doc hand-writes a body any more: both build it from the CSV with a short `python` heredoc that was **run** against the checked-in series (252 rows in, `200 passes=true` out), and the illustrative shape is shown as one row rather than as a payload you could paste and be refused for.
- **The documented exit-code rule was wrong.** Both docs said `4` = "not every runnable leg ran", and the quickstart told the reader an under-70-bar result is "neither a pass nor a fail; do not report it as either". Measured on a 30-bar series (DSR p 0.3026 < 0.50, OOS `not_evaluable`, legs 1/2): the CLI prints `FAILS` and exits **1**. `_verify_verdict` returns `"fail"` as soon as any leg failed, *before* the completeness test — which is the right behaviour (a short series must not launder a FAIL), so the docs were corrected to it, not the code: `4` is "**no leg failed** and not every runnable leg ran".

**Round 3's doc changes**, all of them measured rather than asserted:

- the limits table row, the OpenAPI `description` on `returns`, the endpoint's own docstring (which IS the OpenAPI description — the handler had none), the MCP tool description and the CLI `--help` text all state the 250-bar window and what happens below it;
- the 422 example in `docs/api/strategies-and-rigor.md` is a **real** response, captured from a 249-bar request through the app, `bars_received`/`bars_required` included;
- the CLI refusal shown there is a **real** run against a 249-row CSV — no mock, no session needed to reach it;
- the paragraph claiming an under-70-bar answer is an INCOMPLETE evaluation was corrected **in `docs/api/strategies-and-rigor.md` only** — round 1's own new section in `docs/agent-quickstart.md` kept the old rule forty lines under the new one, and this body claimed otherwise. Both files are corrected in review round 2 below, along with the two exit-code paragraphs and the CLI's `exits.INCOMPLETE` docstring. Shortness can no longer produce that state through this route (the window is above the split's own requirement); the remaining cause — a leg that cannot run on the numbers, e.g. zero variance — is what all of them now say;
- `docs/claims-ledger.md`'s line citations moved again (`:668`/`:684` → `:787`/`:803`) and were re-pointed by reading the lines, not by arithmetic;
- the two **agent-facing** surfaces (`ui/public/llms.txt`, `ui/public/.well-known/agent.json`) now state the window, the refusal shape and the ordering/finiteness rules — an agent building a body reads those, not the docs tree. Fixing them turned up a live false claim next door: `llms.txt` still ended its verify entry with the **pre-#1481** rule ("`passes` is true only when at least one check was evaluable and none failed"), directly under the line that correctly calls `passes` a quorum. Two contradictory rules in one paragraph, the wrong one last. Removed. `skills/archimedes-cli/SKILL.md` carried the same stale rule plus a citation to a test name that no longer exists, and used "a series too short for either DSR or OOS to run" as its example of the zero-evaluable state — which this window makes unreachable. Both corrected in a separate commit so the vet can read them apart from the code.

## Tests changed, not deleted

Two existing tests encoded the old contract:

- the T=3 "nothing evaluable" case is now a `window_too_short` refusal, so the zero-evaluable invariant ("vacuous truth is not honesty") is asserted on a zero-variance series a full window long instead — the state still exists, it is just reached by degeneracy rather than by shortness;
- `minItems` in the published schema moved `1` → `250`, because advertising a length the server rejects is its own small lie;
- **#1481's four-bar regression test changed shape, and did not lose its guard.** That body is now a `window_too_short` refusal, so the one-evaluable case cannot be *submitted* — asserted as such, with the precondition (DSR computable at 4 bars, OOS not) proved on the gate functions themselves. The quorum it was protecting is re-pinned on a full-length series with the walk-forward leg forced `not_evaluable`: `legs_evaluated: 1`, `passes: false`. Reverting the quorum still fails.

One fixture built 300 far-future dates as a `% 28` cycle — duplicated and unsorted, which was never a legitimate way to reach the rf fallback — and now builds consecutive days.

## Verification

- **full backend suite, the way CI runs it** (`pytest -m "not integration" --tb=short -q` from the repo root): **5,810 passed**, 4 skipped, 2 deselected, 0 failed
  - (Run first from `backend/` instead, per the task instructions, it reports **2 failed / 5,808 passed** — `test_generation_cost_meter.py::test_passport_write_is_counted_inside_the_success_branch_not_at_the_caller` and `test_generation_cost_persistence.py::test_the_pipeline_does_not_tally_the_durable_write`, both `FileNotFoundError: 'backend/archimedes...'`. Those two open a source file by a path relative to the REPO ROOT, so they are cwd-sensitive and unrelated to this change; both pass from the root, which is where CI runs them. Reported rather than quietly dropped.)
- backend `-k "rigor or dsr or sharpe or gate or pbo"`: **976 passed** (round 2: 967)
- `test_rigor_verify_routes.py` + `test_waf_verify_body.py` + `test_api_docs_drift.py` + `test_claims_ledger.py`: **99 passed**; the agent-surface suites (`test_agent_discovery.py`, `test_agent_manifest_static_consistency.py`, `test_canonical_repo_links.py`, `test_claims_ledger.py`): **66 passed**
- CLI: **144 passed** (round 2: 139) · MCP (`PYTHONPATH=src:../cli/src`): **233 passed** (round 2: 232)
- `python .github/scripts/docs_links.py`: 1,795 links in 184 files, **0 broken** · `docs_index.py`: 179 docs, **0 orphaned** · `mkdocs build --strict`: **exit 0**
- import-guard's deptry invocation: **Success! No dependency issues found.**
- `ruff check` / `ruff format --check`: **clean on every file this PR touches.** Correcting round 1's claim: the repo-root run finds **45**, not 6 — and one of those 45 *was* in a file this PR touches (a stale `# noqa: ARG001` at `cli/tests/test_cli.py:81`, which the root config's `cli/tests/** = ["ARG"]` ignore already makes redundant). It is removed here. The remaining 45 are in files this PR does not touch at all, so this change cannot have moved them.

## Review round 2 — six defects from the vet

Every one addressed; each fix carries a guard, and each guard was shown **red under mutation** before this was pushed (mutations applied in this worktree, measured, reverted **by editing**, tree verified clean).

| # | Sev | Defect | Fix | Guard, shown red |
|---|---|---|---|---|
| 1 | blocker | The 250-bar floor makes every valid request exceed the WAF's 8,192 B limit, and the #1749 exemption is merged but **not applied**. The body listed ten concerns and never mentioned the coupling. | The deploy precondition is now the first thing in this body, with the ordering and the confirming request. | `test_the_waf_exemption_is_load_bearing_for_every_request_not_just_long_ones` — pins that the *minimum* accepted body (10,774 B) exceeds the edge limit, so the exemption is load-bearing for **every** request, not just long ones. Red on `_MIN_WINDOW_BARS = 100`: `AssertionError: the minimum accepted body is 4713 B, at or under the 8192 B edge limit.` |
| 2 | major | `mcp-server/tests/test_no_credential_leak.py` went **vacuous** for `archimedes_rigor_verify`: its 1-row fixture is refused by the tool's own window pre-check, so no credential was resolved, no request was built, and 16 leak assertions passed with nothing to search. This PR is what removed the coverage. | Fixture raised to 252 well-formed bars (`_rows()`, mirroring `test_tools_http.py`'s `_ROWS()`). | `_drive` now asserts a request actually reached the wire, on **every** call in the matrix — the file's existing control proves the search *finds* a leak, this proves there was something to search. Red on the 1-row fixture: **16 failed, 132 passed**, e.g. `AssertionError: archimedes_rigor_verify never issued a request — ARGUMENTS['archimedes_rigor_verify'] was refused by a local pre-check, so this run proves nothing about credential handling. Fix the fixture.` |
| 3 | major | `docs/agent-quickstart.md:694` still said "Below about 70 bars the OOS leg cannot run at all … the CLI exits `1` there, not `4`" — forty lines under this PR's own new "under 250 bars there is no verdict". Unreachable state, wrong exit code, and this body claimed it was fixed. | Rewritten the way `strategies-and-rigor.md:330` was: the remaining cause is a leg that cannot run on the numbers (zero variance); a short series is **refused**, `422 window_too_short` / CLI exit `2`. The body's round-3 claim above now names both files. | New module `backend/tests/test_verify_window_docs.py` (23 tests) — see below. Red on the shipped paragraph: `paragraph at line 690: 'Below\nabout 70 bars'` and `still offers shortness as something the gate weighs against a FAIL: ['short series is not an excuse']`. |
| 4 | minor | The two discovery surfaces disagreed: `.well-known/agent.json` states the window, the **served** `GET /api/agent/manifest` did not. The #1448 drift again, on the surface the product actually serves. | `rigor.verdict_note` carries the window, sourced from `rigor_verify_routes._MIN_RETURN_ROWS` rather than a literal, so it cannot drift from what the route enforces. | Two tests in `test_agent_manifest_static_consistency.py`, both directions (surface silent / surface quotes a number the route does not enforce). Red on dropping the sentence: `the served discovery surface does not state the 250-bar minimum evaluation window: '…'` plus `the served surface no longer states a bar count`. |
| 5 | minor | `requirements-base.txt` declares **33** packages; three headers said 32. | All three corrected — and the two stale counts in `test_env_requirements_alignment.py`'s own docstrings, which the vet did not catch, reworded to carry no number. | `test_no_file_states_a_package_count_the_base_file_does_not_have` reads the count out of each file's prose (handling `environment.yml`'s two-line `#` continuation) and compares it to the list. Red on restoring 32: `backend/requirements-base.txt declares 33 packages, but environment.yml says [32]; requirements-base.txt says [32].` Paired with a guard-on-guard so a regex that matched nothing cannot pass vacuously. |
| 6 | minor | `docs/api/strategies-and-rigor.md:369-374` still framed shortness as a reachable state ("a CI job can never read 'too few bars' as 'strategy rejected'"), 30 lines under the paragraph that says it is not. | Reworded: the "a leg could not run" example is a **degenerate** series; "too few bars" is explicitly named as no longer an input to that choice — it exits `2` with `window_too_short`. | Same new module. |

**Found while fixing, same class, same round:** the CLI's own `exits.INCOMPLETE` docstring — the line a CI author reads to learn what a `4` means — documented exit 4 as "typically too few bars for the walk-forward OOS split (~70) while DSR only needs 4". Unreachable through this route since the floor. Corrected, and `cli/src/archimedes_cli/{exits,cli}.py` are scanned by the new guard alongside the five docs. Shown red on the stale docstring: `cli/src/archimedes_cli/exits.py names shortness as a cause without saying it is a refusal: paragraph at line 33: 'too few'`.

### How the new docs guard works, and why it is not a word list

`backend/tests/test_verify_window_docs.py` checks a **relation**, not forbidden strings:

- any claim about a bar count **below the enforced floor** must sit in a paragraph that names the refusal;
- "too few bars" / "too short" offered as a **cause** must name it too;
- the floor is read from `rigor_verify_routes._MIN_RETURN_ROWS`, so moving it moves the guard rather than retiring it.

Two design decisions worth the vet's eye. The unit is the **paragraph**, not a character window: a fixed window either splits a markdown paragraph (a correction three lines below the claim reads as absent — this produced a false positive on `strategies-and-rigor.md:332` at 350 chars) or spills into a neighbouring docstring (an unrelated "refused" reads as present). And the cause check uses a **narrower** vocabulary (`refus`, `window_too_short`) than the numeric one, because the very paragraph that made exit 4 mean "too few bars" also contains the words *"strategy rejected"* — the wide list was fooled by it, which is how the `exits.py` instance was nearly missed. A control test drives both readers over the pre-fix prose and over correctly-framed prose, so neither can pass by matching nothing.

Deliberately **not** a ban on mentioning short series: the docs must keep explaining *why* the floor is 250 (the split needs ~70 bars, the DSR wants a year), and a guard that forbade the number would delete the explanation.

## Concerns for the vet

1. **`minItems` 1 → 250 is a published-schema change, and the biggest one here.** Any client generated from the OpenAPI document sees the new floor, and any caller who was sending a few months of bars starts getting a 422. That is the owner's call as given; flagging it because it is the one change in this PR that can break a working integration rather than a broken one. The `returns` field also carries a `description` in the schema now, so the window is legible to a generated client and not only in the refusal.
2. **The 422 shape for this route changed** from FastAPI's list to a single object. The CLI handles both (an uncoded 422 still takes the old path, tested); a third-party caller parsing `detail[0].loc` would need the new shape. It is the shape #1803 asked for and the one the CLI/MCP already parse elsewhere.
3. **`|r| <= 1.0` is a product call, not just a safety one.** It refuses a genuinely levered daily series that exceeded +100% in one bar. I judged loud refusal better than a silently inflated Sharpe, and the message says which; say the word if you want the ceiling raised or made opt-in.
4. **The window is 250, not 252.** One trading year is 252 US sessions; 250 leaves two days for a holiday-short year or a series that starts mid-week, and it is the number in the message, the schema, the CLI, the MCP tool and the docs. Say the word if you want it at 252 exactly — it is one constant (`_MIN_WINDOW_BARS`), and the tests read it from the module rather than restating it, so the change is one line plus whatever the docs say.
5. **`window_too_short` no longer collides with pydantic's own `min_length` type (`too_short`), and that is a small loss.** Round 1 leaned on the collision: remove the validator and the declarative backstop still answered with the right code. It cannot now — `too_short` is not one of this route's published codes any more, and the generic error cannot produce the `bars_*` fields anyway — so removing the validator degrades to the generic list shape, exactly as `max_length`/`too_long` already did. It still REFUSES; it just refuses less usefully. Documented inline. (`too_short` is kept in the CLI's and the MCP's remedy maps, deliberately: `--api-url`/`ARCHIMEDES_API_URL` can point at an older host, and its `too_short` should still render a remedy rather than a bare HTTP error.)
6. **The ordering guard closes the ROW-ORDER form of the holdout attack, not relabelling.** A caller who writes ascending dates onto return-sorted values sends a body no server can distinguish from a real series, and gets `200 passes=true`. No input rule can detect that; the response says `self_attested: true` and `verdict_capped: true` for exactly this reason. Round 1's docs read as though it were fully closed — a clause in both docs now says which half is closed.
7. **Unknown fields are still dropped, and I did NOT change that.** `{"returns": [...], "trails": 100000}` returns 200 with `trials: 1`. `model_config = ConfigDict(extra="forbid")` is the fix, but it needs a new published reason code (pydantic's own type is `extra_forbidden`, which is not one of ours and would degrade to the generic shape) and it narrows the accepted-request contract for any existing caller. That is a product call, not a review fix — say the word and it is a small follow-up.
8. **`error` strings are new API surface.** The six local codes above (`non_finite_returns`, `window_too_short_returns`, …) are names a CI job can branch on and therefore cannot be redefined later, same rule as the exit codes. They are deliberately distinct from the server's `reason` values so a reader can tell which side caught the row; if you would rather the local path reused `error: "input_rejected"` verbatim, that is a one-line change and now is the time.
9. **The CLI now checks the row count BEFORE the per-row rules, which changes which code a doubly-bad file gets.** A 3-row CSV with a `nan` in it used to answer `non_finite`; it now answers `window_too_short`, because that is the order the server's own validators run in and the mirror's whole value is one file → one code. Both answers are true and both exit `2`; flagging it because it is a visible behaviour change in a path that already shipped in this PR's round 2.
10. **An EMPTY file still answers `empty_returns`, not `window_too_short`.** Zero rows almost always means the wrong file or the wrong columns, and saying "you need 250 bars" to someone who pointed at a JSON file is worse advice than saying "nothing parsed". The server has no such split — it answers `window_too_short` with `bars_received: 0` — so this is the one place the two sides deliberately word the same rejection differently. Say the word if you would rather they match exactly.

11. **`agent_manifest_routes` now imports a private name from `rigor_verify_routes`** (`_MIN_RETURN_ROWS`, aliased). The alternative was a second literal `250` on a discovery surface, which is the exact drift the fix exists to close — the same reasoning as `_CHAIN_ID` reading `ARC_CHAIN_ID` rather than restating it. Same package, no cycle, `test_agent_discovery.py` and the manifest suite green. Say the word if you would rather the constant were promoted to a public name first.
12. **The two CLI modules are now scanned by a `backend/tests/` module.** `test_verify_window_docs.py` lives in the backend suite because that is where the enforced constant is importable from, and it reads `cli/src/…` off disk. If you would rather the CLI's half moved into `cli/tests/`, it is a straight lift — but then the floor has to be restated there, which is the thing the module exists to prevent.


dbrowneup:
## Review round 2 — every numbered defect, addressed

Pushed as `23c672cd`. Each item below names what changed, the file, and the mutation that proves the new guard is load-bearing (all three mutations applied, measured, reverted).

---

### 1. The CLI crashed on a NaN/inf CSV and exited `1` = `GATE_FAILED` — **fixed**

Confirmed exactly as reported, before touching anything:

```console
$ archimedes verify poison.csv --json ; echo "exit=$?"
  File ".../httpx/_content.py", line 177, in encode_json
ValueError: Out of range float values are not JSON compliant: nan
exit=1        # stdout: empty
```

`cli/src/archimedes_cli/cli.py` now parses the CSV **strictly, before building the request**. It went further than the one check you specified, because the same argument covers all five: a row the CLI can already see is wrong is refused locally with the **server's own reason code** and `exits.USAGE`.

| Bad row | local `error` | `reason` | exit |
|---|---|---|---|
| `nan` / `NaN` / `inf` / `-inf` / `Infinity` (case-insensitive — `float()` takes them all) | `non_finite_returns` | `non_finite` | 2 |
| `abs(r) > 1.0` | `out_of_range_returns` | `out_of_range` | 2 |
| `20260102`, `2026-W01-1`, `2026-1-2`, `02/01/2026`, epoch, `2026-02-30` | `invalid_date_format` | `invalid_date` | 2 |
| a repeated date | `duplicate_date_rows` | `duplicate_date` | 2 |
| a row out of order | `unsorted_date_rows` | `unsorted_dates` | 2 |

Same `error`-local / `reason`-server split `--trials` already used. `--json` emits a parseable object on every one of these paths:

```console
$ archimedes verify poison.csv --json ; echo "exit=$?"
{"ok": false, "command": "verify", "error": "non_finite_returns", "message": "'poison.csv' row 2 has daily_return 'nan', which is not a finite number and cannot be sent as JSON", "reason": "non_finite"}
exit=2
```

Three deliberate limits, in case you disagree with any of them:

- **The server stays the authority.** This is a mirror. It repairs nothing — no sorting, no dedup, no coercion — and the server re-checks every rule; when a request does go out, the server's sentence is the one printed.
- **`too_short` / `too_many_rows` are NOT mirrored.** They are the gate's own moving policy (the floor *is* `DSR_MIN_BARS`; the ceiling tracks the edge's payload budget, #1749). A CLI that hard-coded either would refuse a series a newer server accepts. `--trials` keeps its local bound because it is an argument the CLI defines, not a property of the file. A test pins this: a 1-row file really is sent, and comes back `too_short` from the server.
- **Row numbers count DATA rows**, not file lines, so a skipped header/comment does not shift them (tested).
- **Not stricter than the server**: `|r| == 1.0` and `-1.0` still go through (tested).

**MCP got the same pre-check** (`mcp-server/src/archimedes_mcp/tools.py`). It does not build the payload from a file, so by your literal condition it was out of scope — but `client.request` catches `httpx.HTTPError` only, so a non-finite bar raised `ValueError` straight out of the handler, and that module's stated contract is *"Handlers never raise."* Same codes, and the remedy line is now shared with the 422 path through `errors.input_rejected_remedy()`.

**Mutation — remove the local non-finite check:**

```
FAILED test_non_finite_return_is_refused_locally_never_as_a_failing_verdict[nan]  - assert 1 == 2
FAILED ...[NaN] ...[NAN]                                                          - assert 1 == 2
FAILED ...[inf] ...[-inf] ...[Infinity] ...[-Infinity]  - assert 'out_of_range_returns' == 'non_finite_returns'
FAILED test_local_refusal_row_numbers_count_data_rows_not_file_lines              - assert 1 == 2
8 failed, 131 passed
```

The `nan` cases reproduce the original defect exactly — exit 1, not 2. (The `inf` cases fall through to the range check, which is why they fail on the code rather than the exit.)

**Mutation — disconnect the MCP pre-check:** `11 failed, 39 passed`, three of them `UnexpectedToolError: Error executing tool archimedes_rigor_verify` — the handler really does raise.

---

### 2. Both `curl` examples sent 2 rows, which this PR's own floor refuses — **fixed**

`docs/agent-quickstart.md` and `docs/api/strategies-and-rigor.md` now send the first **four** bars of the same disclosed seed-1803 series. I regenerated it rather than trusting the numbers: `rng(1803).normal(0.0011, 0.0085, 252).round(5)` over `bdate_range("2025-01-02")` gives `2025-01-06, -0.02371` and `2025-01-07, -0.00077` — your rows, verbatim. Each block now carries the clause you specified: four is the *accepted* minimum, and it still answers `legs_evaluated: 1` / INCOMPLETE until ~70 bars.

### 3. The documented exit-code rule did not match the code — **docs corrected, code untouched**

Confirmed: `_verify_verdict` returns `"fail"` as soon as any leg failed, *before* the completeness test, so a 30-bar series with a failing DSR exits `1`, not `4`. The code is right — a short series must not launder a FAIL — so the docs moved to it: `4` is now "**no leg failed** and not every runnable leg ran", and the quickstart's "neither a pass nor a fail" is now conditioned on no leg having failed. The PR body carries the full five-row exit table.

---

### 4 & 5. `_trials_bounds` — **both applied** (you marked them owner's call; they are cheap and they make the documented envelope actually reachable)

The guard ran on the raw value and tested `isinstance(value, int)` only, so `1e18` / `"10000000000"` landed on the declarative `le` and came back as `less_than_equal` in the generic list shape — the promised `trials_out_of_range` never reached the caller, and the MCP mapped it to `invalid_request`. It now widens the value the way pydantic's lax mode is about to (integral floats, numeric strings), before bounding it, and refuses a `bool` outright instead of reading `true` as `1`. Non-finite literals (`Infinity`, `NaN` — Python's own `json.dumps` emits them) are refused with the same code. A non-numeric `trials` deliberately keeps pydantic's own error: it has no range to be out of.

**Mutation — revert to the `isinstance(int)`-only form:** `9 failed, 51 passed`, including `test_trials_true_is_refused_rather_than_read_as_one - assert 200 == 422`.

### 6. `extra="forbid"` — **not applied, and I would like your call**

`{"returns": [...], "trails": 100000}` still returns 200 with `trials: 1`. The fix needs a **new published reason code** (pydantic types it `extra_forbidden`, which is not one of ours, so it would degrade to the generic shape — the exact defect 4 was about) and it narrows the accepted-request contract for any existing caller. That is a product decision, not a review fix. It is a small follow-up if you want it.

---

### Claim corrections in the PR body

- **ruff.** Round 1 said "the 6 repo-root findings"; it is **45**. Worse, one of them *was* in a file this PR touches — a stale `# noqa: ARG001` at `cli/tests/test_cli.py:81` that the root config's own `cli/tests/** = ["ARG"]` ignore makes redundant. Removed, so "clean on every file touched" is now true and checkable. The other 45 are in files this PR does not touch at all.
- **The ordering guard closes the row-order form only.** Both docs now say so explicitly: a caller who relabels ascending dates onto return-sorted values sends a body no server can detect and gets a 200 — which is what `self_attested: true` / `verdict_capped: true` are for.

### Counts

CLI **139** (was 113; the original 113 unchanged and green) · MCP **232** (was 219) · `test_rigor_verify_routes` + `test_waf_verify_body` + the five drift/ledger suites **139** · backend `-k "rigor or dsr or sharpe or gate or pbo"` **967** (was 952) · `ruff check` + `ruff format --check` clean on all nine touched files.

dbrowneup:
## Fix round — all six defects addressed

Pushed as **[`bdf2a47b`](https://github.com/aprin-labs/archimedes/commit/bdf2a47b9decb12a6077c3e56ed825459d308cec)** (every fix and guard), **[`9078423e`](https://github.com/aprin-labs/archimedes/commit/9078423e8b7cded3c043b6900e2d80e88d42b12f)** (line re-wrap of the two corrected paragraphs — no claim change, so the sentence diff reads as sentences) and **[`3e75907c`](https://github.com/aprin-labs/archimedes/commit/3e75907c0a3882e298a6bdf482d8041e38fe2350)** (a fourth stale count, in a docstring, reworded to carry no number). Branch head `3e75907c`. PR body rewritten: the deploy precondition is now the **first** thing in it.

Thank you for the round — defect 2 in particular. An adversarial file that had gone silently vacuous is the worst kind of green, and this PR is what did it.

| # | Sev | Defect | Commit | What changed |
|---|---|---|---|---|
| 1 | blocker | WAF coupling never stated | `bdf2a47b` | PR body ⚠️ **DEPLOY PRECONDITION** block at the top, with the ordering (`infra/apply.sh` plan → `--apply`, expect one in-place change to `aws_wafv2_web_acl.main`, then a real 250-row POST through the edge) and the alternative (hold the merge). Plus `backend/tests/test_waf_verify_body.py::test_the_waf_exemption_is_load_bearing_for_every_request_not_just_long_ones`, so the coupling is discoverable from the repo rather than only from a PR body. |
| 2 | major | `test_no_credential_leak.py` vacuous for `archimedes_rigor_verify` | `bdf2a47b` | Fixture raised to 252 well-formed bars, **and** `_drive` now asserts a request actually reached the wire on every call in the matrix. |
| 3 | major | `docs/agent-quickstart.md:694` still false | `bdf2a47b` | Rewritten the way `strategies-and-rigor.md:330` was. New guard module `backend/tests/test_verify_window_docs.py`. PR body's round-3 doc bullet corrected to name both files. |
| 4 | minor | Served manifest silent on the window | `bdf2a47b` | `rigor.verdict_note` carries it, sourced from `rigor_verify_routes._MIN_RETURN_ROWS` (not a literal). Two new tests in `test_agent_manifest_static_consistency.py`, both directions. |
| 5 | minor | Three headers said 32, the file declares 33 | `bdf2a47b` | All three corrected — **plus two you did not catch**, in `test_env_requirements_alignment.py`'s own docstrings (`:11`, `:200`), reworded to carry no number. New guard `test_no_file_states_a_package_count_the_base_file_does_not_have`. |
| 6 | minor | `strategies-and-rigor.md:369-374` stale framing | `bdf2a47b` | Reworded: the "a leg could not run" example is a degenerate series; "too few bars" is named as no longer an input to that choice. Same new guard module. |

**One instance you did not list, same class, found while fixing.** `cli/src/archimedes_cli/exits.py` documented exit `4` as *"typically too few bars for the walk-forward OOS split (~70) while DSR only needs 4"* — the line a CI author reads to learn what a `4` means, and unreachable through this route since the floor. Corrected, and both `cli/src/archimedes_cli/{exits,cli}.py` are scanned by the new guard alongside the five docs.

---

## Guards shown RED under mutation

Every mutation applied in `wt-1815`, measured, reverted **by editing**, `git status --porcelain` verified empty after each. Verbatim.

**R1 — `_drive` non-vacuity** (defect 2). Mutation: `ARGUMENTS["archimedes_rigor_verify"]` back to the shipped 1-row body.

```
E       AssertionError: archimedes_rigor_verify never issued a request — ARGUMENTS['archimedes_rigor_verify'] was refused by a local pre-check, so this run proves nothing about credential handling. Fix the fixture.
E       assert []
E        +  where [] = <conftest.Recorder object at 0x12352f860>.requests

tests/test_no_credential_leak.py:84: AssertionError
...
16 failed, 132 passed in 11.06s
```

Sixteen — the exact 16 cases you counted as vacuous. Green with the 252-bar fixture: `148 passed in 8.63s`.

**R2 — the docs guard** (defects 3 and 6). Mutation: `docs/agent-quickstart.md:694` restored to the paragraph this PR shipped.

```
E   AssertionError: docs/agent-quickstart.md makes a claim about a bar count below the 250-bar evaluation window without saying the route refuses it:
E       paragraph at line 690: 'Below\nabout 70 bars'
E     A series under 250 bars never reaches the gate — it answers 422 window_too_short (CLI exit 2), so it cannot produce legs_evaluated < legs_runnable, an INCOMPLETE verdict, or exit 1/4. Say refused, or delete the sentence.

E   AssertionError: docs/agent-quickstart.md still offers shortness as something the gate weighs against a FAIL: ['short series is not an excuse']. A short series is refused before the gate runs; the leg that can fail to run on the numbers is a degenerate one (zero variance).
```

**R3 — the same guard on the CLI surface.** Mutation: `exits.INCOMPLETE`'s docstring restored to the "too few bars" wording.

```
E   AssertionError: cli/src/archimedes_cli/exits.py names shortness as a cause without saying it is a refusal:
E       paragraph at line 33: 'too few'
E     Under 250 bars the route answers 422 window_too_short and the CLI exits 2. Say so in the same breath, or describe the cause that is still real (zero variance).
```

**R4 — both discovery surfaces** (defect 4). Mutation: window sentence dropped from `agent_manifest_routes.py`'s `verdict_note`.

```
E   AssertionError: the served discovery surface does not state the 250-bar minimum evaluation window: '`passes` requires BOTH runnable legs (DSR, OOS consistency) to have run and passed. … verdict_capped so the scalar is qualifiable without re-deriving leg statuses.'. An agent builds its body from this note, not from the docs tree.

E   AssertionError: the served surface no longer states a bar count — update or remove this guard
2 failed, 1 passed, 13 deselected, 1 warning in 16.61s
```

**R5 — the package-count guard** (defect 5). Mutation: `requirements-base.txt` and `environment.yml` headers back to 32.

```
E   AssertionError: backend/requirements-base.txt declares 33 packages, but environment.yml says [32]; requirements-base.txt says [32]. Update the sentence with the package, or reword it to carry no count.
E   assert not {'environment.yml': [32], 'requirements-base.txt': [32]}
```

**R6 — the deploy-precondition test** (defect 1). Mutation: `_MIN_WINDOW_BARS = 100`, i.e. the floor moved back below the WAF crossover.

```
E   AssertionError: the minimum accepted body is 4713 B, at or under the 8192 B edge limit. If the floor moved back below the crossover, some requests survive the ALB without the exemption again — re-read this test's docstring before relaxing it.
E   assert 4713 > 8192
```

At `_MIN_WINDOW_BARS = 250` the smallest legal body measures **10,774 B** — which is the fact that makes the deploy ordering mandatory rather than preferable.

---

## Two design notes worth your eye

**The docs guard is a relation, not a word list.** A claim about a bar count below the enforced floor must sit in a paragraph that names the refusal; "too few bars" as a *cause* must name it too; the floor is read from `rigor_verify_routes._MIN_RETURN_ROWS`, so moving it moves the guard rather than retiring it. The docs keep explaining *why* 250 (the split needs ~70, the DSR wants a year) — banning the number would have deleted the explanation.

Two things were learned by getting it wrong first:

- the unit has to be the **paragraph**, not a character window. At ±350 chars it false-positived on `strategies-and-rigor.md:332` (the correction sat four lines below the claim, outside the window) and it would have spilled across docstring boundaries in `exits.py`.
- the cause check needs a **narrower** refusal vocabulary (`refus`, `window_too_short`) than the numeric one. The very paragraph that made exit 4 mean "too few bars" also contains the words *"strategy rejected"* — the wide list was fooled by it, which is how the `exits.py` instance was nearly missed.

A control test drives both readers over the pre-fix prose *and* over correctly-framed prose, so neither can pass by matching nothing.

**`agent_manifest_routes` now imports a private name** (`_MIN_RETURN_ROWS`, aliased) from `rigor_verify_routes`. The alternative was a second literal `250` on a discovery surface — the exact drift the fix exists to close, and the same reasoning as `_CHAIN_ID` reading `ARC_CHAIN_ID` rather than restating it. Same package, no cycle. Say the word if you would rather the constant were promoted to a public name first (concern 11 in the body).

---

## Verification

- **full backend suite**, `pytest -m "not integration" --tb=short -q -p no:cacheprovider` from the repo root: **5,839 passed, 4 skipped, 2 deselected, 0 failed** in 25:13. No flakes this run — your three (`test_generation_credits` sqlite lock, two `test_security_hardening` subprocess timeouts) did not reproduce.
- targeted: `test_rigor_verify_routes` + `test_waf_verify_body` + `test_agent_manifest_static_consistency` + `test_env_requirements_alignment` + `test_verify_window_docs` + `test_agent_discovery` + `test_api_docs_drift` + `test_claims_ledger` + `test_docs_site` — **187 passed**
- CLI **144 passed** · MCP (`PYTHONPATH=src:../cli/src`) **233 passed**, including `test_no_credential_leak.py` **148 passed** with every case now actually issuing a request
- `ruff check` / `ruff format --check` clean on every file this round touches; the 45 repo-root findings are all in files this PR does not touch
- import-guard's exact deptry 0.25.1 invocation: `Success! No dependency issues found.`
- `docs_links.py` 1,795 links / 184 files, **0 broken** · `docs_index.py` 179 docs, **0 orphaned** · `mkdocs build --strict` exit 0
- CI on `3e75907c`: **18 pass, 2 skipping, 0 fail** — including Backend — unit tests, Ruff, import-guard, `mkdocs --strict`, Docs links + index, CodeQL ×5, Complexity, Frontend, Supply chain. The two `skipping` jobs are the deliberate ones.

Branch state, informational: `main` has moved again (this branch is now 21 behind / 8 ahead). `git merge-tree --write-tree HEAD origin/main` computes a **clean** merge — no conflicts, tree `e61be346` — checked without touching the worktree.

Worked in `~/.claude/jobs/09189ac8/tmp/clone/wt-1815`. The shared clone stayed on `main` and untouched; `wt-q7` was not touched.

**Do not merge — owner vets first.** And per the block at the top of the body: not deployed either, until the #1749 apply is confirmed.

Issue #1283, opened 2026-08-19
[api] Migrate rename_strategy's legacy-wallet ownership fallback onto account ownership


dbrowneup:
Flagged in the #1194 rename commit (3237e3e3): rename_strategy's legacy-wallet fallback is the single MUTATING call site on the relaxed get_linked_wallet_address lookup. Ruling at merge (2026-08-19): acceptable to ship — metadata mutation, not money-moving, so it does not breach the fresh-signature boundary — with this issue as the follow-up: migrate the fallback to canonical account ownership and delete the legacy path once pre-account rows are reclaimed. The boundary rule (docstring on get_linked_wallet_address): the relaxed lookup never gates a money-moving or irreversible action without a fresh signature.

dbrowneup:
PR #1336 ships the decidable half: rename_strategy's legacy-wallet fallback now reclaims the touched row (and every other unclaimed row for that wallet, via `claim_legacy_wallet_data`) onto canonical account ownership in the same transaction — self-healing pre-account rows as they're touched instead of re-deriving ownership from the wallet match forever.

Deleting the fallback branch stays open: it's explicitly gated on "once pre-account rows are reclaimed," which needs (1) verifying prod has zero rows with `owner_user_id IS NULL AND owner_wallet IS NOT NULL` across strategy_store/strategy_passports/strategy_proposals/vault_metadata, and (2) a decision on any row whose wallet was never linked to any account (those can't self-heal via a rename or link). Leaving this open until that verification + decision happens.

onder-akkaya:
Checked against `main` @ `b49609a6`. Half done, so leaving it open.

The migration part landed. `rename_strategy` no longer just trusts the relaxed lookup — a pre-account row (`owner_user_id` NULL) matched via the caller's linked wallet is now reclaimed onto canonical account ownership in the same transaction as the rename, via the same `claim_legacy_wallet_data` bulk claim a verified wallet link performs (`strategies_routes.py:2182-2192`). That addresses the reason this issue exists: the single mutating call site on the relaxed `get_linked_wallet_address` lookup now converts to canonical ownership rather than acting on the legacy match.

The second half hasn't happened. This issue asks to "delete the legacy path once pre-account rows are reclaimed", and the legacy path is still present — necessarily so, since it *is* the reclaim mechanism, and it can only go once no unclaimed pre-account rows remain. That's a data-state condition, not a code one, and I can't check it without prod DB access.

[...]

PR #1817, opened 2026-09-02
[data] #1283: adopt the 152 orphaned legacy strategy rows under a platform account (idempotent, pre-flight-guarded, dry-run first)

dbrowneup:
Closes the Q1 half of #1283 — the owner's 2026-09-02 decision: *"backfill the orphaned legacy rows to a platform account, or delete them if a clean backfill is not possible — and clean up everything they touch; the prod database must not break."*

**A clean backfill is possible, so nothing is deleted.** One column moves, on three tables, recorded row-by-row so every write can be undone or handed back.

---

## ⚠ Before merging: measure

**Run this first, from a read-only session on prod. It is the pre-merge gate, not a formality.** Merging arms the deploy: `deploy.yml`'s `migrate` job runs `alembic upgrade head` on the very next deploy, so the population has to be checked *before* the merge, not after it. This is the migration's own predicate, generated from `_candidate_sql` so the two cannot drift apart. It reads; it writes nothing.

```sql
SELECT 'strategy_store' AS tbl, COUNT(*) AS orphans FROM strategy_store t
 WHERE t.owner_user_id IS NULL AND t.owner_wallet IS NOT NULL AND t.owner_wallet <> ''
   AND NOT EXISTS (SELECT 1 FROM linked_wallets lw WHERE lower(lw.address) = lower(t.owner_wallet))
UNION ALL
SELECT 'strategy_passports', COUNT(*) FROM strategy_passports t
 WHERE t.owner_user_id IS NULL AND t.owner_wallet IS NOT NULL AND t.owner_wallet <> ''
   AND NOT EXISTS (SELECT 1 FROM linked_wallets lw WHERE lower(lw.address) = lower(t.owner_wallet))
UNION ALL
SELECT 'strategy_proposals', COUNT(*) FROM strategy_proposals t
 WHERE t.owner_user_id IS NULL AND t.owner_wallet IS NOT NULL AND t.owner_wallet <> ''
   AND NOT EXISTS (SELECT 1 FROM linked_wallets lw WHERE lower(lw.address) = lower(t.owner_wallet));

-- And the composition, so a refusal is diagnosable rather than surprising:
SELECT COUNT(*) FILTER (WHERE owner_user_id IS NULL AND owner_wallet IS NOT NULL) AS wallet_only,
       COUNT(*) FILTER (WHERE owner_user_id IS NULL AND owner_wallet IS NULL)     AS anonymous,
       COUNT(*) FILTER (WHERE owner_user_id IS NOT NULL)                          AS owned
  FROM strategy_store;   -- repeat for strategy_passports / strategy_proposals
```

**Read the three-table total:**

| Total | What it means | What to do |
|---|---|---|
| **in [140, 165]** | This is the population the 2026-09-02 decision was taken against. | Proceed. The migration will adopt them. |
| **anything else** | This is **not** that population. | The migration refuses, deliberately, and blocks the deploy. That is a **new decision to take**, not a band to widen. |
| **0** | No orphans here at all. | The migration is a no-op that creates the ledger table and nothing else. |

The most likely surprise is a total *below* 140: if prod's orphaned rows are mostly *wallet-less*, they are excluded by design (nothing identifies the author) and the count comes in short. That outcome is Concern 1 below, and it is a question for you, not something this PR pre-empts.

Optionally rehearse against the real container first — step 2 of *What the owner should run* below. **That task exits non-zero by design**, on any population including zero; that is what proves nothing was written, `alembic_version` included.

---

### The definition (and what it deliberately excludes)

`graft callers claim_legacy_wallet_data` / `owns_strategy` / `is_strategy_reasoning_visible` gives the whole ownership surface. An **orphaned legacy row** is a row on `strategy_store` / `strategy_passports` / `strategy_proposals` where:

```
owner_user_id IS NULL                         -- pre-account row
AND owner_wallet IS NOT NULL AND <> ''        -- it names a wallet
AND that wallet is in NO linked_wallets row   -- and nobody ever proved it
```

The third clause is what makes it *orphaned* rather than merely *unclaimed*. #1283 already shipped the self-heal (`claim_legacy_wallet_data`, `wallet_routes.py:236`): linking, re-verifying, or renaming through a wallet stamps `owner_user_id` onto every pre-account row for that address. That can only fire for a wallet an account has proven. **A wallet in no `linked_wallets` row has no account to fire it** — those rows sit at `owner_user_id IS NULL` forever, and meanwhile `owns_strategy`'s tier-2 fallback (`strategy_visibility.py:126-134`) hands them to *whoever* turns up controlling that address. That is the population the decision was about, and the remaining half of #1283's own stated deliverable.

Not orphans, and untouched:

| Excluded | Why |
|---|---|
| `owner_user_id IS NULL AND owner_wallet IS NULL` | Nothing identifies the author. The owner-scoped readers filter on an exact non-NULL match, so these are already returned to nobody; a stamp would attribute an anonymous artifact to the platform and buy nothing. Same call `docs/plans/2026-08-30-relations-phase2.md` § 2.7 made. **Curated / `is_example` rows are in this bucket** (house content carries no `owner_wallet`), so the library is untouched. |
| unclaimed, but the wallet IS linked | The existing reclaim reaches these behind a fresh signature. A migration must not race a mechanism that already works. |
| `vault_metadata` | `store_vault_metadata` 409s when `owner_user_id` is neither NULL nor the caller's, and its NULL case exists **specifically** so a legitimately transferred on-chain owner can still write metadata (`vaults_routes.py:487-489`). A platform stamp closes that door permanently. This is the same exclusion `rename_strategy`'s reclaim already makes, for the same reason. |
| `user_profiles` | PII, and `owner_user_id` is UNIQUE there — one platform account could structurally adopt at most one row. |
| `paper_deployments` | Not in the reclaim tuple, and its `owner_user_id` FK is `ON DELETE CASCADE`: adopting a ledger under an account and later deleting that account would destroy the ledger, not un-own it. |

### The touch map — the FK and soft references to these rows

*(Corrected: this was headed "every FK and soft reference." It is not exhaustive — `subscriber_liabilities.strategy_id` (`marketplace.py:170`), `subscriber_tick_log.strategy_id` (`:204`), `settlement_intents.strategy_id` (`:253`) and `paper_agent_trades.signal_strategy_id` (`paper_store.py:426`) are also soft references and are not listed below. Nothing is deleted, so they stay resolvable exactly like every row in the table; but this table is the stated evidence for choosing backfill over delete, so it should not claim completeness it does not have.)*

| Table | Reference to an adopted row | Kind | Under **backfill** (this PR) | Under **delete** (the rejected alternative) |
|---|---|---|---|---|
| `strategy_store` | the row itself (`id`) | — | `owner_user_id` NULL → `platform-legacy`. Every other column, `owner_wallet` included, untouched | the strategy disappears from the library, the leaderboard, and every row below |
| `strategy_passports` | the row itself (`id`), 1:1 mirror | — | same stamp; both mirrors move together (D1 refuses a split) | `passport_paper_refs` cascades with it |
| `strategy_proposals` | the row itself (`id`) | — | same stamp | the private brief + rigor verdict are destroyed, unrecoverably |
| `passport_paper_refs` | `passport_id → strategy_passports.id` | **FK, ON DELETE CASCADE** | untouched — no id changes | silently cascaded away: the paper-provenance chain, i.e. the product's central claim |
| `paper_deployments` | `strategy_id → strategy_store.id` | **FK** (validated in `9c2e7b5a1f4d`) | untouched, still resolves | the delete fails on the FK, or takes the deployment *and* its whole graded ledger (`paper_daily_returns`, `paper_decision_traces`, `paper_marks`, `paper_agent_trades` all cascade off `paper_deployments.id`) |
| `backtest_results` | `strategy_id` | soft (no FK, `backtest_store.py:31`) | untouched | dangles silently — no FK to stop it |
| `debate_transcripts` | `strategy_id` (nullable) | soft (`debate_transcript.py:182` says so explicitly) | untouched | dangles |
| `generation_costs` | `strategy_id` | soft (`generation_cost.py:62`) | untouched | dangles; cost/revenue provenance lost |
| `marketplace_agents` | `strategy_id` | soft (`marketplace.py:78`) | untouched | a publisher row pointing at nothing |
| `strategy_generators` | `strategy_id` + `wallet_address` | soft | untouched — publish authorization keys on the **wallet**, which this PR does not change | dangles |
| `vault_metadata` | `strategy_ids` (JSON array of ids) | soft, **textual** | untouched | a JSON array naming a deleted strategy: no FK, no cascade, no error, no way to notice |
| `strategy_daily_returns`, `strategy_backtest_fixtures` | `stem` | keyed on the **curated file stem**, not `strategy_store.id` | out of scope by construction | out of scope |
| `identity_events` | `wallet → wallet_identities` | FK on the wallet, not the strategy | untouched — `owner_wallet` is preserved | untouched |
| `linked_wallets` | `address` | the claim source | read-only in the pre-flight — it *is* the definition of "orphan" | — |
| `auth_users` | `owner_user_id` target | FK | **one new row**, created idempotently | — |
| `legacy_row_adoptions` | new | — | one row per adopted row, carrying the prior wallet | — |

**Nothing can dangle under backfill.** This revision deletes nothing and changes no primary key, so every row in that table still points at exactly what it pointed at before. The only column that moves is `owner_user_id`. That is the whole argument for backfill over delete, and the table above is the evidence: five of those references have no FK at all, so a delete would go quietly wrong.

---

## STEP 2 — Design

**The platform account** (`archimedes/models/account.py`): a real `auth_users` row with a fixed, documented id `platform-legacy`, email `platform-legacy@archimedes.invalid`. It is an **owner, never a login**:

* no `auth_accounts` credential row and no `auth_sessions` row is ever written, so nothing can authenticate as it — adopted rows are therefore readable by nobody, the fail-closed direction;
* `.invalid` is RFC 6761-reserved and resolves nowhere, so no email flow reaches it — and because `auth_users.email` is UNIQUE, a signup attempting that address is rejected by the database rather than taking the account over;
* **every published count of accounts, or of accounts-who-did-something, excludes it by id** — a bookkeeping row must never inflate a user number the product publishes. Three filters, three tests, all red without them: `user_stats._query_distinct_user_count` (the public "users" figure), `engagement_metrics.get_account_metrics` (total / new_7d / new_30d), and `engagement_metrics.get_repeat_generation_metrics`. *(Corrected: this said "`user_stats.py`, `engagement_metrics.py`" and claimed two tests. The third filter did not exist — see* ***Defect 2*** *below. The adjacent `is_example` filter does **not** cover adopted rows: an adopted row carries a real `owner_wallet`, which is exactly what house content does not.)*

**Ownership semantics match the readers.** `owns_strategy` is two-tier: a non-NULL `owner_user_id` is the *only* thing that grants ownership; the wallet fallback is reachable only when it is NULL. Stamping the platform id therefore moves the row out of tier 2 — which is exactly the point (no unproven wallet-holder can claim it) and exactly the risk (**neither can the real one**, since `claim_legacy_wallet_data` filters on `owner_user_id IS NULL`).

**So adoption is made reversible per row.** New table `legacy_row_adoptions` records `(table_name, row_pk, prior_owner_wallet, adopted_by_user_id, adopted_at)`, UNIQUE on `(table_name, row_pk)`. It buys four things: an exact downgrade, idempotency, auditability, and — the one that matters — **release**: `release_platform_adopted_rows` is called from `claim_legacy_wallet_data`, so the first time the real holder of that wallet proves control with a fresh signature, every row the platform took for it becomes theirs. Authorization is identical to the claim it runs inside, and it grants nothing the caller would not have had under the tier-2 fallback the adoption replaced. Scoped to the tables the caller passed, so `rename_strategy`'s narrowed claim stays narrow.

Nothing had to be deleted, so the "delete instead" branch was not taken. The touch map above is why that is the better outcome, not just the easier one.

---

## STEP 3 — The migration (`d3a71f5c9e28`)

**(a) Pre-flight.** `population = candidates + rows already adopted by this revision`.

* `population == 0` → no-op (every fresh DB, CI, dev clone). It still creates the ledger table, because the ledger is schema, not data — otherwise an alembic-built DB and a `create_all()` one diverge. **Under `ORPHAN_MIGRATION_DRY_RUN` this branch raises like every other**, so a rehearsal against a database with no orphans writes neither the table nor the version stamp (fixed this round — see *Defect 3* below; it used to return cleanly, which alembic commits).
* `0 < population < 140` or `> 165` → **RAISE**. The decision was taken against a measured 152; the band is that ±≈8%, wide enough to absorb rows that self-healed or were written between the measurement and the deploy, narrow enough that a materially different number means *this is not that population*. The message says so, and says explicitly: do not widen the band to make a deploy pass.
* Three structural refusals, each naming a way adoption could leave the DB worse: **D1** split mirror (a candidate whose store/passport twin already has a real owner — an ownership rule that disagrees with itself is an authorization bug); **D2** parentless passport (the relations plan's Q1 population — fix via #1438, not an ownership stamp); **D3** non-lowercase `owner_wallet` (the release path matches by exact lowercase equality, so such a row would be adopted and then impossible to hand back).

**(b) Dry run.** `ORPHAN_MIGRATION_DRY_RUN=1` logs the full plan and then **raises**, aborting the transaction. That is the mechanism, not a bug: a dry run that returned cleanly would stamp `alembic_version` and the real run would skip the revision. **A dry run exits non-zero by design — read the log, not the exit code.** The stronger guarantee (a dry run or a refusal writes *nothing*, on any dialect) comes from **ordering**: every raise happens before the first DDL or DML. That is load-bearing rather than stylistic — pysqlite does not open a transaction for DDL, so a `CREATE TABLE` before the raise would survive the abort on SQLite even though it rolls back on Postgres. **There is no exception to that ordering.** (There was: the zero-population early return created the ledger and returned, so a dry run on a database with no orphans committed the DDL *and* stamped `alembic_version`. Fixed and pinned — *Defect 3*.)

**(c) Idempotency.** Second run: every adopted row now has a non-NULL owner so it is no longer a candidate, the ledger carries the population, the band still passes, the write phase finds nothing. The `auth_users` insert is existence-guarded, and each `UPDATE` re-checks `owner_user_id IS NULL` so a row claimed between pre-flight and write is left to its real owner.

**(d) Downgrade** re-orphans **ownership only**: `owner_user_id` back to NULL, but only where the platform still owns it — a row released to its real owner in the meantime is left alone. Then it drops the ledger and removes the platform account **only if nothing anywhere still names it** (checked across all eleven declared `auth_users.id` FKs; **seven** are `ON DELETE CASCADE` — `paper_deployments`, `user_profiles`, `api_keys`, `auth_sessions`, `auth_accounts`, `linked_wallets`, `wallet_link_challenges` — so a blind `DELETE` would destroy rows rather than un-own them). No row is deleted by this revision, so **there is no deletion to be irreversible about** — and the docstring says that a future revision that *does* delete an orphan must say so in its own downgrade.

Single alembic head: `d3a71f5c9e28`.

---

## STEP 4 — Tests, and the guards shown red

**19 new tests** (`--collect-only`: `test_orphan_row_adoption.py` 10 + `test_legacy_row_release.py` 9) + 3 metrics tests. *(Corrected: this said "13 new tests" — it was 14 at that point, miscounted by one; the fix round below added five more.)* Every guard was verified by writing the input that *should* fail it.

**Every refusal, red (real output, scratch SQLite):**

```
### A_band_300_rows   exit=1
RuntimeError: #1283 orphan adoption REFUSED: found 300 orphaned legacy row(s)
(strategy_store=140, strategy_passports=140, strategy_proposals=20, already adopted 0),
outside the expected band [140, 165]. ... do not widen the band to make a deploy pass.
-- AFTER: rows owned by platform-legacy = 0; ledger table exists = False; auth_users platform row = 0; alembic_version = c5e81a4f7b32

### B_D1_split_mirror   exit=1
RuntimeError: #1283 orphan adoption REFUSED — structural pre-flight failed:
  - D1 split mirror: 1 candidate id(s) already have a real owner on strategy_passports (first: ['orph00000']).
-- AFTER: rows owned by platform-legacy = 0; ledger table exists = False; auth_users platform row = 0; alembic_version = c5e81a4f7b32

### C_D2_parentless_passport   exit=1
RuntimeError: #1283 orphan adoption REFUSED — structural pre-flight failed:
  - D2 parentless passport: 1 candidate strategy_passports row(s) have no strategy_store row (first: ['orph00059']).
-- AFTER: rows owned by platform-legacy = 0; ledger table exists = False; auth_users platform row = 0; alembic_version = c5e81a4f7b32

### D_D3_mixed_case_wallet   exit=1
RuntimeError: #1283 orphan adoption REFUSED — structural pre-flight failed:
  - D3 non-lowercase owner_wallet on 152 candidate row(s) ... adopted and then impossible to hand back.
-- AFTER: rows owned by platform-legacy = 0; ledger table exists = False; auth_users platform row = 0; alembic_version = c5e81a4f7b32

### E_dry_run_152_rows   exit=1
OrphanMigrationDryRun: ORPHAN_MIGRATION_DRY_RUN is set: plan logged, transaction aborted, nothing written.
-- AFTER: rows owned by platform-legacy = 0; ledger table exists = False; auth_users platform row = 0; alembic_version = c5e81a4f7b32
```

**The band is load-bearing, not decorative** — widen it to `[0, 10000]` and the *same* 300-row database is adopted:

```
exit: 0
rows adopted with the band widened: 300
ledger rows: 300
```

(reverted immediately; `EXPECTED_MIN = 140` / `EXPECTED_MAX = 165` on the branch).

**The dry-run plan the owner will see:**

```
#1283 orphan adoption — candidates: strategy_store=60, strategy_passports=60, strategy_proposals=32 (total 152); already adopted: 0; population: 152; band [140, 165]
#1283 DRY RUN — would adopt 60 row(s) on strategy_store; first ids: ['orph00000', ...]
#1283 DRY RUN — would adopt 60 row(s) on strategy_passports; first ids: ['orph00000', ...]
#1283 DRY RUN — would adopt 32 row(s) on strategy_proposals; first ids: ['prop00000', ...]
#1283 DRY RUN — would stamp owner_user_id='platform-legacy' (auth_users row created if absent, email platform-legacy@archimedes.invalid), write 152 ledger row(s) to legacy_row_adoptions, and change nothing else. Aborting the transaction now so NOTHING is written, including the alembic_version stamp; a non-zero exit here is expected.
```

**The metrics exclusion is load-bearing too** — delete the `AuthUser.id != PLATFORM_LEGACY_USER_ID` filters and both account-count tests go red (`2 failed`); restore them and they pass. The third exclusion, on `get_repeat_generation_metrics`, was missing entirely until this round — *Defect 2*.

**And the missing-ledger-table guard** — `release_platform_adopted_rows` runs on every wallet link, including on a database that has not reached this revision. Remove its existence check and:

```
E   sqlite3.OperationalError: no such table: legacy_row_adoptions
FAILED test_legacy_row_release.py::test_release_survives_a_database_that_has_no_ledger_table
```

That guard also caught a real bug in review: written as `sa.inspect(session.get_bind())` it checks out a **second** connection, outside the transaction the caller's pending `UPDATE`s live in — four wallet-linking tests went red (`Could not refresh instance <LinkedWallet>`, and an already-claimed row reading back as still unclaimed). It inspects `session.connection()` instead, and the comment says why.

### Review round 2 — three defects found and fixed, each pinned red first

An adversarial re-verification found three real defects, two of them user-visible. All three are fixed on this branch; each has a test that fails on the un-fixed code, and the failure output below is the real mutation run.

**Defect 1 (blocking, user-visible) — adoption silently switched off the only UI that tells the orphan's real owner to come and get it.**

`release_platform_adopted_rows` was real, reachable and correct. Nobody was ever told about it. `_wallet_has_unclaimed_legacy_data` (`wallet_routes.py`) mirrors the claim loop's `owner_user_id IS NULL` filter — and the adoption stamp is precisely what makes that column non-NULL, so from the moment the migration ran the predicate answered `False` for exactly the wallets whose rows were sitting in the ledger. That predicate is `GET /api/wallets/check`, which is the **sole** gate on the relink banner (`ui/src/AuthenticatedApp.jsx`) — the one surface built for these ~10 pre-account wallet users in #1194 rev e. Measured on a scratch DB, before vs after the real revision: `has_legacy_data` went `True` → `False` while 152 rows waited to be handed back. A release door with a key and no sign; reversible in code, a one-way door in practice.

Fixed by mirroring the *other* half of the claim: new `platform_adopted_row_count` (`models/legacy_adoption.py`, same `session.connection()` ledger guard and for the same reason), consulted by `_wallet_has_unclaimed_legacy_data` after the NULL-owner loop. The claim path, the banner and its copy are unchanged and still reachable — an adopted row reports `has_legacy_data: true`, the banner still says *"Your earlier strategies are waiting… nothing is deleted, and only you can claim it,"* and linking still fires the release. Four tests, in `test_legacy_row_release.py`:

```
FAILED test_legacy_row_release.py::test_the_relink_prompt_still_fires_for_an_adopted_wallet - AssertionError: assert False is True
FAILED test_legacy_row_release.py::test_the_prompt_and_the_release_agree_end_to_end       - AssertionError: assert False is True
2 failed, 7 passed        <- with the ledger check removed
9 passed                  <- restored
```

The other two pin the boundaries: proving a *different* address must not light the banner (it would advertise a stranger's adopted rows), and a database with no ledger table must answer, not 500.

**Defect 2 (blocking, wrong published number) — the platform account was counted as a real user, and as a *repeat* one.**

`get_repeat_generation_metrics` grouped `strategy_store` by `owner_user_id` with no platform exclusion, and the adjacent `is_example` filter does not reach adopted rows (they carry a real `owner_wallet`; house content does not). It ships in the same `get_engagement_snapshot` dict as `account_metrics`, which *was* filtered. Measured with one real account plus orphan rows spread over two calendar days — the shape ~60 real prod rows will have — `generating_users` went 1 → 2 and **`repeat_users` went 0 → 1 with no human doing anything.** That falsified this PR's own "every account count excludes it by id," and is verbatim the failure that function's docstring says its `is_example` filter exists to prevent: *"leaving the two adjacent tiles reading different populations."*

Fixed with `.filter(StrategyRecord.owner_user_id != PLATFORM_LEGACY_USER_ID)`. The tile's published `note` now says so too. Third metrics test, which also asserts the adjacent tile agrees:

```
FAILED test_engagement_metrics.py::test_repeat_generation_metrics_never_counts_the_platform_legacy_account - assert 2 == 1
1 failed, 2 passed        <- with the filter removed
24 passed                 <- restored
```

**Defect 3 — a dry run on a zero-population database wrote DDL and stamped `alembic_version`.**

The `population == 0` branch ran `_create_ledger` and returned **before** the `_dry_run_requested()` check. So `ORPHAN_MIGRATION_DRY_RUN=1` against a DB with no orphans exited **0**, created the ledger, and marked the revision applied — meaning the real deploy would skip it, while the rehearsal reported success. This is not a corner case: it is exactly the branch *Concern 1* below says is plausible for prod, and it made this PR's promise that a dry run "proves nothing was written, `alembic_version` included" false at the moment it mattered. The writes are benign; the broken contract is not.

Fixed by re-checking the flag at the top of that branch, so **no** path through `upgrade()` writes before a possible raise. Pinned:

```
FAILED test_orphan_row_adoption.py::test_dry_run_on_a_zero_population_database_also_writes_absolutely_nothing
E   AssertionError: a dry run must abort, not apply:
E   assert 0 != 0
1 failed, 9 deselected    <- with the dry-run block removed
10 passed                 <- restored
```

Two overstated claims in this body are corrected in place above (marked *Corrected:*), plus the `_AUTH_USER_REFERENCES` comment in the migration, which said "every (table, column) that points at `auth_users.id`" and lists only *declared* FKs — `payment_receipts.user_id`, `generation_credits.user_id` and `free_generation_grants.user_id` also do, with no FK. The downgrade's `DELETE` is still safe (this account cannot authenticate, so it can never hold a receipt, credit or grant) and the comment now says which word is load-bearing and what a future reader must add.

**Suites run** (`python -m pytest -p no:cacheprovider`):

* `test_orphan_row_adoption.py` — **10 passed** (152 seeded orphans + 5 controls + 5 touch-map rows; adopt / band×2 / D1 / D2 / D3 / dry-run / **zero-population dry-run** / second-run-no-op / downgrade)
* `test_legacy_row_release.py` — **9 passed** (release to the real owner; a different wallet gets nothing; a narrowed claim cannot widen itself; no-op where nothing was adopted; survives a database with no ledger table; **and four on discoverability** — the prompt still fires for an adopted wallet, prompt and release agree end to end, a stranger's address lights nothing, and a pre-migration database answers rather than 500s)
* `test_engagement_metrics.py` — **24 passed** (3 of them the platform-account exclusions)
* `test_alembic_migrations.py` — 28 passed (unchanged; single head, empty-DB replay, `create_all()` parity)
* The four suites above together — **71 passed** (was 65 before this round)
* The wider ownership/wallet/metrics/alembic subset (`-k "wallet or identity or metrics or strategy_store or ownership or proposal or visibility or legacy or orphan or alembic or user_stats"`) — **588 passed, 2 skipped** (both Postgres-only), up from 582
* `alembic heads` → `d3a71f5c9e28 (head)`, single
* `ruff check` on every touched file — **All checks passed!**; `ruff format --check` clean on all of them (this round also formatted three files that were already unformatted on the branch, plus one my edit made unformatted)

---

## What the owner should run, and in what order

**1. Measure first, from a read-only session — this is the pre-merge gate.** → **[Before merging: measure](#-before-merging-measure)**, at the top of this description: the read-only SQL, and what each outcome means. Do this before the merge, because the merge is what arms the deploy.

**2. Dry-run the real container, once the branch image exists** (after merge + build, before letting the deploy's `migrate` step run for real — or as a standalone one-off):

```bash
aws ecs run-task \
  --cluster archimedes-cluster \
  --task-definition archimedes-migrate \
  --launch-type FARGATE \
  --network-configuration '{"awsvpcConfiguration":{"subnets":["subnet-0f412b89a025ca15b","subnet-010efb75b1093ba92"],"securityGroups":["sg-0f21c2773901a14e7"],"assignPublicIp":"DISABLED"}}' \
  --overrides '{"containerOverrides":[{"name":"migrate","command":["python","-m","archimedes.scripts.alembic_migrate_preflight"],"environment":[{"name":"ORPHAN_MIGRATION_DRY_RUN","value":"1"}]}]}' \
  --query 'tasks[0].taskArn' --output text
```

Then read `/archimedes/app`, stream prefix `ecs-migrate-*`, for the `#1283 DRY RUN` lines. **The task will exit non-zero. That is the design** — it is what proves nothing was written, `alembic_version` included. That now holds for *every* population, zero included; it did not before (*Defect 3*), which is what made this rehearsal worth fixing rather than trusting.

---

## Concerns — read before merging

1. **I have not touched the production database and cannot verify the 152.** The figure comes from the deck; the band is the mechanism that makes a wrong figure loud instead of silent, and step 1 above is how it gets verified. If prod's orphans are mostly *wallet-less* rows, the count will come in under 140 and the migration will refuse — correctly — and the owner then has a **new** decision to take (adopt anonymous rows too, or leave them). I did not pre-empt that decision.
2. **`claim_legacy_wallet_data` is auth-adjacent and I edited it — and now `_wallet_has_unclaimed_legacy_data` beside it.** The release grants nothing a caller would not have had under the tier-2 wallet fallback before adoption, and it is scoped to the tables the caller passed. The new predicate branch is read-only and returns a bare boolean behind the existing session gate, and a test pins that a *stranger's* address lights nothing — but both deserve your eyes, not just mine.
3. **Merging arms the deploy.** `deploy.yml`'s `migrate` job runs `alembic upgrade head` on the next deploy. If the population is out of band, that job fails and blocks the deploy. That is the intended safety behaviour and the reason step 1 comes before the merge.
4. **Every test here is SQLite.** The pre-flight SQL is dialect-neutral; the one real difference is DDL rollback (pysqlite autocommits DDL), handled by ordering every raise before the first write.
5. **`strategy_proposals.owner_wallet` has no FK to `wallet_identities`**, so a prod orphan there may name a wallet that is not a known identity. That does not block adoption (the pre-flight does not test it) but it is worth knowing when reading the counts.
6. **Downgrade leaves released rows with their real owner.** Deliberate: taking a row back from someone who proved the wallet would be worse than an incomplete reversal.

---


dbrowneup:
## Defect 1 (blocking, user-visible) — adoption silently switched off the only UI that tells the orphan's real owner to come and get it

Fixed in 2d9229d1.

`release_platform_adopted_rows` was real, reachable and correct. **Nobody was ever told about it.**

`_wallet_has_unclaimed_legacy_data` (`backend/archimedes/api/wallet_routes.py`) mirrors the claim loop's `owner_user_id IS NULL` filter — and the adoption stamp is *precisely* what makes that column non-NULL. So from the moment the migration ran, the predicate returned `False` for exactly the wallets whose rows were sitting in the ledger. That predicate is `GET /api/wallets/check`, which is the **sole** gate on the relink banner in `ui/src/AuthenticatedApp.jsx` — the one surface built for these ~10 pre-account wallet users in #1194 rev e.

Measured on a scratch DB, before vs after the real revision: `has_legacy_data` went `True` → `False` while 152 rows waited to be handed back. A release door with a key and no sign — reversible in code, a one-way door in practice.

The function's own docstring stated the invariant the diff broke: *"This predicate mirrors the claim loop's models and filters one-for-one, so the relink prompt fires iff linking would actually attach something."* The diff added a second effect to the claim loop and did not mirror it here.

**The fix mirrors the other half of the claim.** New `platform_adopted_row_count` in `backend/archimedes/models/legacy_adoption.py` — same `session.connection()` ledger guard as the release, and for the same reason (this runs on every `/check`, including on a database that has not reached the revision, where raising would turn "you have nothing to claim" into a 500). `_wallet_has_unclaimed_legacy_data` consults it after the NULL-owner loop.

**The claim path stays reachable and the copy still invites the claim.** The release is still triggered by the same flow that previously surfaced the prompt — `claim_legacy_wallet_data`, behind a fresh signature. An adopted row now reports `has_legacy_data: true`, the banner renders unchanged (*"Your earlier strategies are waiting… Connect and sign with that wallet to bring it into this account — nothing is deleted, and only you can claim it"* — still true of an adopted row), and linking hands it back. The only UI change is a comment recording why the predicate must keep counting the ledger.

Four tests in `backend/tests/test_legacy_row_release.py`, shown red first:

```
FAILED test_the_relink_prompt_still_fires_for_an_adopted_wallet - AssertionError: assert False is True
FAILED test_the_prompt_and_the_release_agree_end_to_end       - AssertionError: assert False is True
2 failed, 7 passed        <- ledger check removed
9 passed                  <- restored
```

The other two pin the boundaries: proving a *different* address must not light the banner (that would advertise a stranger's adopted rows to anyone who can name an address), and a pre-migration database must answer rather than 500.

dbrowneup:
## Defect 2 (blocking, wrong published number) — the platform account was counted as a real user, and as a *repeat* one

Fixed in 2d9229d1.

`get_repeat_generation_metrics` (`backend/archimedes/services/engagement_metrics.py`) grouped `strategy_store` by `owner_user_id` with no platform exclusion. The adjacent `is_example` filter does **not** cover adopted rows: an adopted row carries a real `owner_wallet`, which is exactly what house content does not have.

Measured with one real seeded account plus orphan store rows spread over two calendar days — the shape ~60 real prod rows will have the moment the migration runs:

```
BEFORE: generating_users = 1   repeat_users = 0
AFTER : generating_users = 2   repeat_users = 1
        account_metrics.total = 1          <- the adjacent tile, correctly filtered
        distinct owner_user_id in scope: ['platform-legacy', 'user-real']
```

**`repeat_users` went 0 → 1 with no human doing anything** — the bookkeeping account became the product's repeat user. This ships in the same `get_engagement_snapshot` dict as `account_metrics`, which *was* filtered, and is served by `metrics_private_routes.py`. It falsified this PR's claim that "every account count excludes it by id," and the identical claim in `account.py`. It is also verbatim the failure that function's own docstring says the `is_example` filter exists to prevent: *"leaving the two adjacent tiles reading different populations."*

**Fix:** `.filter(StrategyRecord.owner_user_id != PLATFORM_LEGACY_USER_ID)`, plus the reason in the docstring and in the tile's published `note`. `account.py`'s comment now names all three filters rather than gesturing at two modules, and says explicitly that `is_example` does not reach adopted rows — so the next person who adds a count knows to join the list.

Third metrics test in `backend/tests/test_engagement_metrics.py`, which also asserts the adjacent tile agrees on the population:

```
FAILED test_repeat_generation_metrics_never_counts_the_platform_legacy_account - assert 2 == 1
1 failed, 2 passed        <- filter removed
24 passed                 <- restored
```

The existing two tests only covered `get_account_metrics` / `user_stats`, which is why this slipped.

dbrowneup:
## Defect 3 — a dry run on a zero-population database wrote DDL and stamped `alembic_version`

Fixed in 2d9229d1.

In `backend/migrations/versions/d3a71f5c9e28_adopt_orphaned_legacy_rows.py`, the `if population == 0:` branch ran `_create_ledger(bind)` and returned **before** the `if _dry_run_requested():` check below it. Run with `ORPHAN_MIGRATION_DRY_RUN=1` on a DB with no orphans:

```
DRY RUN exit code: 0   (PR body: "The task will exit non-zero. That is the design")
alembic_version after the DRY RUN : ('d3a71f5c9e28',)
legacy_row_adoptions table created: True
```

So the rehearsal **applied** the revision — marking it done, so the real deploy would skip it — while reporting success. This is not a corner case: it is exactly the branch Concern 1 says is plausible for prod (*"if prod's orphans are mostly wallet-less, the count comes in under 140"*). The writes themselves are benign. The broken contract is not: the PR promised a dry run "proves nothing was written, `alembic_version` included," and that was false at the one moment it mattered.

**Fix:** re-check the flag at the top of that branch and raise `OrphanMigrationDryRun`, so **no** path through `upgrade()` writes before a possible raise. The module docstring's *"The one exception is the zero-population early return"* is gone — there is now no exception — and the PR body's STEP 3(a)/(b) are corrected to match.

Test in `backend/tests/test_orphan_row_adoption.py`, shown red first:

```
FAILED test_dry_run_on_a_zero_population_database_also_writes_absolutely_nothing
E   AssertionError: a dry run must abort, not apply:
E   assert 0 != 0
1 failed, 9 deselected    <- dry-run block removed
10 passed                 <- restored
```

It asserts the exit is non-zero, the ledger table is absent, `alembic_version` is still `c5e81a4f7b32`, and no platform account exists — then that the real run afterwards still creates the ledger and stamps the revision, so the dry run changed the outcome of nothing.

---

### Also corrected (verifier item 5 — overstated completeness, no behaviour change)

- **The touch map** was headed *"every FK and soft reference to these rows"* but omits `subscriber_liabilities.strategy_id`, `subscriber_tick_log.strategy_id`, `settlement_intents.strategy_id` and `paper_agent_trades.signal_strategy_id`. Nothing is deleted so they stay resolvable, but that table is the stated evidence for choosing backfill over delete, so the heading no longer claims completeness it does not have, and names the four.
- **`_AUTH_USER_REFERENCES`** was headed *"Every (table, column) in this schema that points at `auth_users.id`"* and lists only declared FKs. `payment_receipts.user_id`, `generation_credits.user_id` and `free_generation_grants.user_id` also do, with no FK. The downgrade's `DELETE` is still safe — this account has no credential row and no session row, so nothing can authenticate as it and it can never hold a receipt, credit or grant — and the comment now says "every **declared** `ForeignKey`", names the three, and tells a future reader what kind of soft reference would have to be added.

### Verification after all three fixes

`test_orphan_row_adoption.py` 10 · `test_legacy_row_release.py` 9 · `test_engagement_metrics.py` 24 · `test_alembic_migrations.py` 28 → **71 passed** (was 65). Wider `-k "wallet or identity or metrics or strategy_store or ownership or proposal or visibility or legacy or orphan or alembic or user_stats"` → **588 passed, 2 skipped** (both Postgres-only), up from 582. `alembic heads` → `d3a71f5c9e28 (head)`, single. `ruff check` and `ruff format --check` clean on every touched file.

**Do not merge — owner vets first.** The pre-merge gate is unchanged and now at the top of the PR description: measure the population from a read-only session *before* merging, because the merge is what arms the deploy.

dbrowneup:
**Read-only prod orphan count — run by the owner on 2026-09-03 (SSM RunShellScript on the runner box, `psql` against the writer, the PR's own SQL):**

| table | orphans |
|---|---|
| strategy_store | 71 |
| strategy_passports | 71 |
| strategy_proposals | 10 |
| **total** | **152** |

152 is inside the migration's [140, 165] acceptance window, so the adoption migration will proceed rather than raise on the next deploy after merge. Command id `a770080a-cf18-42de-9a40-4b9bfa0c8c3d`.

Still held for the owner's merge word; migration `down_revision` must be re-pointed after #1792 lands (both fork from `c5e81a4f7b32`).

Issue #1818, opened 2026-09-03
P0 incident 2026-09-03: paper-advance child runs init_db() DDL every cycle → Postgres lock wedge, 94-min outage, replacements blocked at boot


dbrowneup:
## Summary

Production was unreachable for the first user of the day (Dan) from **13:29Z to 15:03Z on 2026-09-03** (ALB 504s, both targets unhealthy). Root cause: the paper-advance child runs `init_db()` (hand-rolled `ALTER TABLE … ADD COLUMN IF NOT EXISTS` schema patches) **on every cycle, in both ECS tasks at once**. At the 03:32Z cycle the two children interleaved so that one child's patch transaction queued an `AccessExclusiveLock` on `papers` behind the other child's open cycle transaction; the ticking child then needed `papers` from a second session and stalled behind its own sibling. PostgreSQL cannot see that cycle (the blocker is not waiting), so the fleet sat wedged for 10 hours while `/health` kept returning 200 from `stale_cached` probes. The first real page load hit the same lock queue and froze both web processes; the replacement tasks wedged at boot on the same DDL (`init_db()` at import, `main.py:800`) for 91 minutes. The wedge broke only when one old task was OOM-killed.

Not CPU starvation: ECS CPU was ~3% and Aurora CPU ~4% for the whole 11.5 h. This is a DDL lock chain.

## Timeline (UTC, 2026-09-03; app log `/archimedes/app`, ECS events, Aurora log, CloudWatch)

| time | evidence |
|---|---|
| 03:31:19 | task 9b58b8b9's child: `init_db: papers schema patches applied` → takes the fleet lock → replay running (dsl_to_backtrader warnings 03:31:59–03:32:33) |
| 03:32:08–03:32:19 | FIRST `HEALTH_PROBE_TIMEOUT` on **both** tasks (`corpus`, `corpus_db`, `corpus_meta`, `paper_rag`, then every probe); they stay `state=stale_cached` until the tasks die (cached_age reaches 35,797 s) |
| 03:31–03:33 | Aurora `DatabaseConnections` 16 → 33 and flat at 33 for 11.5 h; CPU ~4% (idle = waiting, not working) |
| 03:32:33 | last line ever from the ticking child (no `paper advance: {...}` summary, no `skipping this cycle` line from the sibling) |
| 03:32→13:28 | `/health` answers 200 in ~1.05 s (= the probe budget) every 10–30 s on both tasks; ECS + ALB stay green |
[...]

PR #1819, opened 2026-09-03
[paper-trading][db] The paper-advance child never runs DDL; boot schema patches take a lock_timeout (#1818 P1)

dbrowneup:
Part of #1818 (root cause, P1).

## What broke

On 2026-09-03 production served 504s from 13:29Z to 15:03Z and the replacement ECS tasks sat wedged at boot for 91 of those minutes. Nothing was slow — ECS CPU ~3%, Aurora ~4% for the whole 11.5 h. It was a DDL lock chain with two ends, and this PR closes both.

1. `paper_advance_loop` called `init_db()` **on every cycle, in both ECS tasks at once**.
2. `init_db()` ran all eight `ALTER TABLE … ADD COLUMN IF NOT EXISTS` patches in **one** transaction with **no `lock_timeout`**. Even a no-op `ADD COLUMN IF NOT EXISTS` holds `AccessExclusiveLock` on the table for the rest of that transaction, and a *waiting* exclusive-lock request queues every later reader behind it. So the three `ALTER TABLE papers` at the top of the list kept `papers` locked while `ALTER TABLE strategy_store` waited on somebody else — and the ticking child's own second session (`resolve_paper_hashes`, `SELECT … FROM papers`) queued behind its sibling's DDL while the sibling's DDL waited on the first session. PostgreSQL cannot see that cycle: the blocker is not itself waiting. No deadlock, no timeout, no log line, for ten hours.

Full timeline and the six-part mechanism: `docs/incidents/2026-09-03-paper-advance-ddl-wedge.md` (copied verbatim from #1818 — it is evidence).

## What this PR changes

**1. The child never runs DDL.** `paper_advance_loop` no longer imports or calls `init_db()`. Schema belongs to the migrate task (`alembic upgrade head`) plus the web process's boot-time call in `main.py` — and, today, the eleven request handlers listed under **P6** below; a 24-hourly ticker has no business asserting it. The docstring now says that, and the `"if init_db()/get_session() raised before the ask"` comment above the agent pass is corrected — it named a statement that no longer exists.

**2. Boot patches take a `lock_timeout`, one transaction per statement.** New `_apply_patch_statement` in `backend/archimedes/db.py` runs each patch in its own `engine.begin()` and issues `SET LOCAL lock_timeout = '5s'` first, **on the Postgres dialect only** (`DB_PATCH_LOCK_TIMEOUT` overrides the value). `SET LOCAL` was chosen over an engine-level connect option precisely because it is provable in a test — the statement shows up in the transaction, in order, ahead of the DDL.

A patch that cannot take its lock is a WARNING and boot continues. The patches were already declared non-fatal (#1028); this makes that declaration true under contention rather than aspirational. Crucially the statements *behind* a blocked one still run, which matters because a missing `strategy_store.is_example` is a 500 on every Generate.

`_ensure_ownership_columns()` gets the same treatment, including its `CREATE INDEX IF NOT EXISTS` — the sharpest case, since it takes a lock on `strategy_store` too and used to ride in the same transaction as the ALTERs above it. SQLite behaviour is unchanged: no `SET LOCAL` is issued there, and the columns still land on a pre-existing table.

**3. The per-statement timeout is bounded in aggregate too.** `lock_timeout` is per statement, and one `init_db()` issues 8 patch statements plus up to 5 more from `_ensure_ownership_columns` — 13 on a stale schema. At 5s each a fully contended boot would still burn 45–65s, past the **30s ALB health-check timeout the incident's own timeline names** (`TargetResponseTime max 30 s`). So the whole call shares one deadline (`DB_PATCH_LOCK_BUDGET`, default 10s), checked *before* each statement and threaded into `_ensure_ownership_columns` so it cannot start a fresh one; once past, the rest are skipped with a single WARNING naming what did not land. Worst case is the budget plus the one statement already in flight — 15s.

**4. `DB_PATCH_LOCK_TIMEOUT` is validated before it reaches the SQL.** `SET LOCAL` takes a literal, not a bind parameter, so this env value is interpolated. A typo — `5 s`, `abc`, a stray quote — makes the `SET LOCAL` a syntax error, which the deliberately broad non-fatal arm swallows into a WARNING, silently skipping **every** patch. That is the exact `strategy_store.is_example`-is-missing failure this block exists to prevent, so the value is now shape-checked against `\d+(us|ms|s|min|h|d)?` **and range-checked**, falling back to `5s` with a WARNING — at import *and* at the point of use.

The range check is not decoration. Two *well-formed* values would each put the outage straight back: `0`, which in Postgres **disables** `lock_timeout` altogether — the 2026-09-03 configuration exactly, a boot patch that waits forever — and anything over `INT_MAX` milliseconds, which Postgres rejects with "outside the valid range for parameter", swallowed by the same broad arm into the same skip-everything.

Net: nothing on a cycle path issues DDL, and the DDL that still runs — once at boot in the web process, **and again on every request that calls `init_db()`** (see P6 below) — can no longer wait longer than five seconds for any one lock, hold two tables at once, or spend more than the aggregate budget across the whole call.

## Tests — hermetic, no Postgres

`backend/tests/services/test_paper_advance_kill_switch.py`
- `archimedes.db.init_db` is now armed to **raise** in the shared `_drive_one_tick` harness instead of being a no-op stub, so every driven tick in that file is a standing guard. A re-introduced call shows up twice over: the counter moves *and* the loop's fail-soft arm eats the exception, so `advances` drops to 0.
- An AST walk over `paper_advance_loop` (ImportFrom / Name / Attribute), so the docstring can keep explaining *why* `init_db` is gone without satisfying — or breaking — a substring search.
- A mutation check that `init_db` still exists and `main.py` still calls it once, so "delete the function" cannot pass the class.

`backend/tests/test_db_boot_patches.py` (new, 43 tests)
- A recording fake engine: `transactions` is a list of statement-lists, one per `engine.begin()`. That structure is the whole point — the bug is invisible in a flat SQL log and obvious here. Asserts one transaction per statement, `SET LOCAL lock_timeout` inside it and before the DDL, and that **no transaction ever touches two tables**.
- A simulated `OperationalError` from the first `ALTER` is logged as a WARNING, does not propagate, and leaves the other seven statements applied (`7/8 statements` in the summary line).
- The SQLite case uses a **real** tmp-file engine — the columns and index still land, and `SET LOCAL` never reaches it.
- `TestTheAggregateLockBudget` installs a windable clock as `db.time` and an engine where every DDL statement burns its full 5s and times out, so thirteen timeouts can be tested without spending 65 real seconds. Asserts the run stops at the budget (2 statements, 10s — not 8 statements, 40s), that the give-up line appears **once** and names what it skipped, that `_ensure_ownership_columns` *shares* the deadline rather than starting a fresh one, and that an uncontended boot still applies everything.
- `TestTheLockTimeoutValueIsValidated` parametrises seven valid intervals, seven malformed ones (including `5 s`, `'5s'` and `5s'; DROP TABLE papers --`), the three spellings of zero, three over-range values, and the exact top of the range (`2147483647ms`, `1d` — so the bound cannot be off by one and silently override an operator widening the window for a real migration). Then it drives the consequence end to end against an engine that rejects the SQL a raw `5 s` would produce.

### Red proof (mutation, then revert)

Every guard was shown red on the input it rejects before this was pushed.

**A — re-introduce `init_db()` in the loop** (`from archimedes.db import get_session, init_db` + `init_db()` restored):

```
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestLoopHonoursTheSwitch::test_explicit_true_runs_the_work
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestOneTickerPerFleet::test_the_lock_is_taken_before_any_ledger_work
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestOneTickerPerFleet::test_a_held_lock_skips_the_whole_cycle_including_the_agent_pass
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestOneTickerPerFleet::test_the_winner_runs_both_passes
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestOneTickerPerFleet::test_a_broken_lock_check_fails_open
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestTheCycleNeverRunsDdl::test_driving_a_tick_never_reaches_init_db
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestTheCycleNeverRunsDdl::test_a_contended_cycle_also_runs_no_ddl
FAILED backend/tests/services/test_paper_advance_kill_switch.py::TestTheCycleNeverRunsDdl::test_the_loop_source_does_not_name_init_db
=================== 8 failed, 30 passed, 1 warning in 1.59s ====================
```

with, on the two new ones:

```
E   AssertionError: the cycle ran schema DDL — the #1818 root cause is back
    assert 1 == 0
     +  where 1 = _Tick(advances=0, sleeps=[0.0, 86400.0], agent_calls=1, statements=[], init_db_calls=1).init_db_calls
WARNING  archimedes.services.paper_trading: paper advance: cycle failed
         (AssertionError: paper_advance_loop called init_db() — the #1818 root cause) — will retry next tick

E   AssertionError: paper_advance_loop names init_db in executable code (#1818): line 55: imports init_db; line 85: references init_db
    assert not ['line 55: imports init_db', 'line 85: references init_db']
```

**B1 — drop `SET LOCAL lock_timeout`** (the no-timeout half of the 2026-09-03 state):

```
FAILED backend/tests/test_db_boot_patches.py::TestApplyPatchStatement::test_postgres_sets_lock_timeout_before_the_statement_in_one_transaction
FAILED backend/tests/test_db_boot_patches.py::TestApplyPatchStatement::test_the_timeout_value_is_configurable_and_defaults_to_five_seconds
FAILED backend/tests/test_db_boot_patches.py::TestInitDbPatchTransactions::test_every_patch_gets_its_own_transaction_and_its_own_lock_timeout
FAILED backend/tests/test_db_boot_patches.py::TestEnsureOwnershipColumnsOnPostgres::test_each_alter_and_the_create_index_get_their_own_guarded_transaction
=================== 4 failed, 11 passed, 1 warning in 1.41s ====================
```

**B2 — restore the single shared transaction for all eight statements** (the exact pre-fix code):

```
FAILED backend/tests/test_db_boot_patches.py::TestInitDbPatchTransactions::test_every_patch_gets_its_own_transaction_and_its_own_lock_timeout
FAILED backend/tests/test_db_boot_patches.py::TestInitDbPatchTransactions::test_no_transaction_ever_holds_two_tables_at_once
FAILED backend/tests/test_db_boot_patches.py::TestInitDbPatchTransactions::test_a_blocked_patch_does_not_stop_the_ones_behind_it
=================== 3 failed, 12 passed, 1 warning in 1.51s ====================
```

The middle one names the mechanism directly:

```
E   AssertionError: one transaction locks {'chat_messages', 'papers', 'strategy_store'}: ['ALTER TABLE papers ADD COLUMN IF NOT EXISTS cluster_id TEXT', ..., 'ALTER TABLE chat_messages ADD COLUMN IF NOT EXISTS verified BOOLEAN NOT NULL DEFAULT FALSE']
    assert 3 <= 1
```

and the third shows the blast radius of a single blocked patch under the old code:

```
E   AssertionError: a blocked patch swallowed the statements behind it
    Right contains 7 more items, first extra item: 'ALTER TABLE papers ADD COLUMN IF NOT EXISTS topic_label TEXT'
WARNING  archimedes.db: init_db: papers schema patch failed (non-fatal): (builtins.Exception) canceling statement due to lock timeout
INFO     archimedes.db: init_db: papers schema patches applied (idempotent) — 0/8 statements
```

**B3 — drop the dialect gate so `SET LOCAL` is issued on every engine** (the adversarial input for "SQLite unchanged"). The real SQLite engine rejects it and the columns never land:

```
FAILED backend/tests/test_db_boot_patches.py::TestApplyPatchStatement::test_sqlite_gets_no_set_local
FAILED backend/tests/test_db_boot_patches.py::TestEnsureOwnershipColumnsOnSqlite::test_missing_columns_and_the_index_still_land
=================== 2 failed, 13 passed, 1 warning in 1.52s ====================

E   AssertionError: assert {'is_published', 'owner_wallet'} <= {'id', 'name'}
WARNING  archimedes.db: init_db: ownership patch could not take its lock within 5s (non-fatal, boot continues):
         ALTER TABLE strategy_store ADD COLUMN owner_wallet VARCHAR(42) — (sqlite3.OperationalError) near "SET": syntax error
```

**M1 — remove the aggregate budget check from `init_db`** (back to `sum(_apply_patch_statement(...) for stmt in ...)`):

```
FAILED backend/tests/test_db_boot_patches.py::TestTheAggregateLockBudget::test_a_fully_contended_boot_stops_at_the_budget_not_after_every_statement
FAILED backend/tests/test_db_boot_patches.py::TestTheAggregateLockBudget::test_the_budget_gives_up_once_and_says_what_it_skipped
FAILED backend/tests/test_db_boot_patches.py::TestTheAggregateLockBudget::test_the_ownership_block_shares_the_budget_rather_than_starting_a_fresh_one
=================== 3 failed, 32 passed, 1 warning in 1.99s ====================

E   AssertionError: the budget did not stop the run: 8 statements attempted, 40.0s spent
E   AssertionError: expected exactly one give-up line, got []
E   AssertionError: init_db spent 40.0s, not the 10s budget
```

**M2 — `_ensure_ownership_columns` ignores the deadline it was handed and starts a fresh budget**:

```
FAILED backend/tests/test_db_boot_patches.py::TestTheAggregateLockBudget::test_the_ownership_block_shares_the_budget_rather_than_starting_a_fresh_one
=================== 1 failed, 34 passed, 1 warning in 1.80s ====================

E   AssertionError: the ownership block started a fresh budget: ['ALTER TABLE strategy_store ADD COLUMN owner_wallet VARCHAR(42)']
E   assert ['ALTER TABLE... VARCHAR(42)'] == []
```

**M3 — interpolate `PATCH_LOCK_TIMEOUT` raw instead of the sanitised value** (defect 3's bug, with the value set to `5 s`). Every patch is swallowed and nothing lands:

```
FAILED backend/tests/test_db_boot_patches.py::TestTheLockTimeoutValueIsValidated::test_a_malformed_value_cannot_silently_skip_every_patch
=================== 1 failed, 42 passed, 1 warning in 1.67s ====================

E   AssertionError: a malformed DB_PATCH_LOCK_TIMEOUT skipped the transitional columns
E   assert [] == ['ALTER TABLE...HAR(32)', ...]
E     Right contains 8 more items, first extra item: 'ALTER TABLE papers ADD COLUMN IF NOT EXISTS cluster_id TEXT'
WARNING  archimedes.db:db.py:267 init_db: papers patch could not take its lock within 5 s (non-fatal, boot continues): ALTER TABLE papers ADD COLUMN IF NOT EXISTS cluster_id TEXT — (builtins.Exception) near "SET": syntax error
...  (the identical WARNING for all eight)
INFO     archimedes.db:db.py:374 init_db: papers schema patches applied (idempotent) — 0/8 statements
```

**M4 — make `_sanitised_lock_timeout` accept anything** (the guard condition → `if True`):

```
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[5 s]        - AssertionError: assert '5 s' == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[abc]        - AssertionError: assert 'abc' == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[]           - AssertionError: assert '' == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[-5s]        - AssertionError: assert '-5s' == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[5seconds]   - AssertionError: assert '5seconds' == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds['5s']       - assert "'5s'" == '5s'
FAILED ...::test_a_malformed_value_falls_back_to_five_seconds[5s'; DROP TABLE papers --] - assert "5s'; DROP TABLE papers --" == '5s'
FAILED ...::test_zero_is_rejected_because_postgres_reads_it_as_no_timeout[0]
FAILED ...::test_values_past_int_max_milliseconds_are_rejected[2147483648ms]
FAILED ...::test_a_malformed_value_cannot_silently_skip_every_patch
=================== 14 failed, 29 passed, 1 warning in 2.04s ===================
```

**M5 — keep the shape check, drop the range check** (`if match and 0 < ... <= _PG_MAX_LOCK_TIMEOUT_MS` → `if match`). The two well-formed killers wave straight through:

```
FAILED ...::test_zero_is_rejected_because_postgres_reads_it_as_no_timeout[0]   - AssertionError: '0' would restore the #1818 unbounded wait
FAILED ...::test_zero_is_rejected_because_postgres_reads_it_as_no_timeout[0s]  - AssertionError: '0s' would restore the #1818 unbounded wait
FAILED ...::test_zero_is_rejected_because_postgres_reads_it_as_no_timeout[0ms] - AssertionError: '0ms' would restore the #1818 unbounded wait
FAILED ...::test_values_past_int_max_milliseconds_are_rejected[2147483648ms]   - AssertionError: assert '2147483648ms' == '5s'
FAILED ...::test_values_past_int_max_milliseconds_are_rejected[999999999s]     - AssertionError: assert '999999999s' == '5s'
FAILED ...::test_values_past_int_max_milliseconds_are_rejected[100d]           - AssertionError: assert '100d' == '5s'
=================== 6 failed, 37 passed, 1 warning in 1.67s ====================
```

Each mutation was reverted by inverse string-replace (never `git checkout --`) and the module re-run green — 43 passed — before the commit. M3, M4 and M5 were re-run against the final code after the range check landed.

## Runs

```
# targeted modules
$ pytest backend/tests/test_db_boot_patches.py \
         backend/tests/services/test_paper_advance_kill_switch.py \
         backend/tests/services/test_paper_advance_isolation.py \
         backend/tests/test_db.py -q -p no:cacheprovider
103 passed, 1 warning in 7.72s

# full backend suite, from the repo root (pytest.ini: testpaths = backend/tests)
$ pytest -q -p no:cacheprovider
5762 passed, 6 skipped, 328 warnings in 594.77s (0:09:54)

$ ruff format --check .        # 738 files already formatted
$ ruff check --select E9,F63,F7,F40,F82 .   # All checks passed  (the repo's blocking subset)

$ python .github/scripts/docs_index.py --root .
index: 179 docs under docs/; 179 indexed, 0 orphaned.
$ python .github/scripts/docs_links.py --root .
links: scanned 1793 in 184 markdown files; broken 0 (0.0%)
$ mkdocs build --strict      # Documentation built in 4.65 seconds
```

`pytest-xdist` is not installed in the env, so the suite ran serially.

One note on how to run it: `cd backend && pytest` fails
`test_generation_cost_meter.py::test_passport_write_is_counted_inside_the_success_branch_not_at_the_caller`
with `FileNotFoundError: 'backend/archimedes/agents/generation_pipeline.py'` — that test reads a
repo-root-relative path. It fails identically on unmodified `main` from that directory and passes
from the repo root, so it is unrelated to this PR; flagging it only because the obvious
"run it from backend/" invocation makes it look like a regression.

## What this PR does NOT change (P2–P6, tracked on #1818)

- **P2** — `resolve_paper_hashes` still opens a second session while the cycle transaction is open (mechanism item 3). With P1 in there is no DDL to queue between them, but the pattern is still there.
- **P3** — readiness. `/health` still answers 200 from `stale_cached` probe values, so a wedged task can stay in the target group indefinitely (mechanism item 6).
- **P4** — `list_generated_strategies` and siblings still run `session.query(...)` on the event loop (mechanism item 5).
- **P5 (ops)** — still no CloudWatch alarms on `UnHealthyHostCount ≥ 1`, `HTTPCode_ELB_5XX_Count`, or ECS `MemoryUtilization > 85%`.
- **P6** — **`init_db()` still runs on the SERVING path, not only at boot.** Eleven request handlers call it so the transitional columns exist before they read, and all of them are reached from `async def` endpoints with no threadpool hop, so this DDL executes **on the event loop on ordinary API traffic**: `api/paper_routes.py:107,146,163,210,227` (POST/GET/DELETE `/api/paper/deployments` and `.../marks`), `api/selection_bias_routes.py:333,736`, `api/leaderboard_routes.py:286`, `api/strategies_routes.py:419` (via `GET /api/strategies/`) and `:1661`, and `services/live_rigor_gate.py:273` (via the vaults routes). The incident's wedge SHAPE therefore survives P1: a web request's DDL can still queue an `AccessExclusiveLock` behind a long transaction and put every later reader of that table behind it. P1 **bounds** that (5s per statement, 10s per call) rather than removing it. The fix is to hoist these calls into the lifespan — the same argument this PR makes for the ticker — which is a behaviour change to eleven endpoints and does not belong in a P1.
- The unexplained linear memory ramp (39% → 100% over 90 min on the wedged tasks) is not addressed.


dbrowneup:
## Review round — all three defects addressed in `d0d361dd`

| # | Severity | Defect | Status | Where |
|---|---|---|---|---|
| 1 | major | PR body + incident doc claim the remaining DDL runs "once, at boot, in the web process"; 11 serving-path `init_db()` call sites say otherwise | **fixed** (docs + a third instance in code) | `d0d361dd` |
| 2 | minor | `lock_timeout` is per-statement with no aggregate budget — 9–13 statements × 5s can outlast the 30s health check | **fixed** (code + 5 tests) | `d0d361dd` |
| 3 | minor | `DB_PATCH_LOCK_TIMEOUT` interpolated into SQL unvalidated; a typo silently skips **every** patch | **fixed** (code + 20 tests) | `d0d361dd` |

---

### 1 (major) — the "once, at boot" claim. Fixed, and it was in three places, not two.

You were right, and the diff was the arbiter. I confirmed the 11 call sites and then checked the part the finding did not: whether they are actually on the event loop. They are. Seven sit directly in `async def` endpoints (`deploy_paper`, `list_paper_deployments`, `get_paper_deployment`, `get_paper_deployment_marks`, `stop_paper_deployment`, `evaluate_rigor_gate`, `get_strategy_returns`). The other four are sync helpers — `_live_paper_ledgers`, `_generated_strategy_rigor`, `_live_rigor_results_for_strategies`, `verdicts_for_strategies` — called **without** a threadpool hop from `async def` endpoints (`get_live_paper_leaderboard`, `evaluate_strategy_rigor`, `list_strategies`, the vaults routes). The only `asyncio.to_thread` calls in those files are `strategies_routes.py:863` and `vaults_routes.py:601`, on different paths. So "on the event loop" holds for all 11.

Reworded in **three** places — the finding named two, but `services/paper_trading.py`'s own docstring carried the same sentence ("plus the web process's single boot-time `init_db()`"), which would have been the most misleading of the three since it sits in the code a reader checks first:

- `docs/incidents/…-ddl-wedge.md` — the "what this PR changes" prose and the Net claim now read "the DDL that **still** runs — once at boot in the web process, **and again on every request that calls `init_db()`**".
- The PR body — same wording.
- `backend/archimedes/services/paper_trading.py` docstring.
- `backend/archimedes/db.py`'s `init_db` docstring now opens "This function is a BOOT step **in name, but NOT only at boot in fact**" and names the five modules.

Added **P6** to the incident doc's "what it does not change" (heading now P2–**P6**) and to the PR body, listing all 11 sites with line numbers, stating that the wedge SHAPE survives P1, that P1 *bounds* rather than removes it, and that the fix is hoisting them into the lifespan. No code change, for the reason you gave: that is a behaviour change to eleven endpoints and does not belong in a P1.

I did **not** touch the doc's Summary / Timeline / Mechanism sections — those are copied verbatim from #1818 as evidence. Re-verified after the edit: all three still byte-identical to `gh issue view 1818`, 32 timestamps in each, set difference empty both ways.

### 2 (minor) — aggregate budget. Fixed.

One deadline for the whole `init_db()` call — `DB_PATCH_LOCK_BUDGET`, default 10s — taken before the Postgres arm so the SQLite path can hand it on, and **threaded into `_ensure_ownership_columns(deadline=…)`** so the ownership block cannot start a fresh one (that was the trap: two independent 10s budgets would bound the call at 20s + 5s, back on the wrong side of the health check). Checked *before* each statement, so a statement in flight is never cut off; worst case is budget + one timeout = **15s**, under the 30s the timeline names. Skipping logs **one** WARNING naming the statements that did not land.

I took the second half of your suggestion too: the `db.py` comment now states the aggregate worst case explicitly rather than leaving "never held hostage" to carry it.

### 3 (minor) — validating `DB_PATCH_LOCK_TIMEOUT`. Fixed, and widened after an adversarial pass.

Your regex went in, then I asked what *else* passes it and still reproduces the failure. Two well-formed values do:

- **`0`** — in Postgres that does not mean "fail fast", it **disables `lock_timeout` entirely**. That is the 2026-09-03 configuration exactly: a boot patch that waits forever. A shape-only check waves it through.
- **anything over `INT_MAX` ms** (`999999999s`, `100d`) — Postgres answers "outside the valid range for parameter", the same broad arm swallows it, and again every patch is skipped.

So the value is shape-checked **and** range-checked (`0 < ms <= 2147483647`), falling back to `5s` with a WARNING, at import **and** at the point of use — the latter so no later assignment (a different env read, a test monkeypatch, a future edit) can put a syntax error inside the transaction guarding the DDL. `2147483647ms` and `1d` are pinned as still-accepted so the bound cannot go off by one and silently override an operator widening the window for a real migration.

---

### Red proof

`backend/tests/test_db_boot_patches.py`: **15 → 43 tests**. Every new guard shown red under mutation, reverted by inverse string-replace (never `git checkout --`), module green again before the commit. Full output in the PR body; the headline of each:

| Mutation | Result |
|---|---|
| M1 drop the budget check from `init_db` | 3 failed — `the budget did not stop the run: 8 statements attempted, 40.0s spent` |
| M2 `_ensure_ownership_columns` starts a fresh budget | 1 failed — `the ownership block started a fresh budget: ['ALTER TABLE strategy_store ADD COLUMN owner_wallet VARCHAR(42)']` |
| M3 interpolate the raw value (defect 3's bug) | 1 failed — all 8 swallowed, `init_db: papers schema patches applied (idempotent) — 0/8 statements` |
| M4 sanitiser accepts anything | 14 failed — `assert '5 s' == '5s'` and thirteen siblings |
| M5 shape checked but not range checked | 6 failed — `'0' would restore the #1818 unbounded wait` |

M3–M5 were re-run against the final code after the range check landed, not only against the intermediate version.

The budget tests install a windable clock as `db.time` and an engine where every DDL statement burns its full 5s and times out — thirteen timeouts, no real seconds spent.

### Runs

```
$ pytest backend/tests/test_db_boot_patches.py \
         backend/tests/services/test_paper_advance_kill_switch.py \
         backend/tests/services/test_paper_advance_isolation.py \
         backend/tests/test_db.py -q -p no:cacheprovider
103 passed, 1 warning in 7.72s

$ pytest -q -p no:cacheprovider          # repo root; pytest.ini testpaths = backend/tests
5762 passed, 6 skipped, 328 warnings in 594.77s (0:09:54)   # 5734 baseline + 28 new

$ ruff format --check .                          # 738 files already formatted
$ ruff check --select E9,F63,F7,F40,F82 .        # All checks passed
$ python .github/scripts/docs_index.py --root .  # 179 docs, 179 indexed, 0 orphaned
$ python .github/scripts/docs_links.py --root .  # 1793 links, 0 broken
$ mkdocs build --strict                          # built, INFO-level anchor notes only (pre-existing)
```

Issue #1527, opened 2026-08-30
[agent-native] ERC-8004 identity registration on Arc testnet (identity leg only)


dbrowneup:
## Why

We want Archimedes to be agent-native. ERC-8004 ("Trustless Agents" — identity/reputation/validation registries) is the emerging standard, and the canonical registries are **already deployed on Arc Testnet** at the vanity addresses:

- IdentityRegistry `0x8004A818BFB912233c491871b3d84c89A494BD9e`
- ReputationRegistry `0x8004B663056A597Dffe9eCcC1965A193B7388713`
- ValidationRegistry `0x8004Cb1BF31DAf7788923b405b754f57acEB4272`

(quickstart vendored in `submodules/context-arc`). We already satisfy the *spirit* of the discovery half — `GET /api/agent/manifest`, `.well-known/agent.json` (informal A2A card), machine-readable x402 402 quotes, SIWE `headless` wallet linking — but have **zero on-chain footprint** against the standard: no identity minted, no spec-typed registration file.

## Scope — Identity leg only (1–2 days, no new contracts)

1. Mint one ERC-721 agent identity for the platform: `register(metadataURI)` on the Arc IdentityRegistry from a dedicated platform wallet.
2. Author the spec-typed registration file (`"type": ".../eip-8004#registration-v1"`), reusing `.well-known/agent.json` content; pin it durably.
3. Add an `erc8004` block `{chain, identityRegistry, agentId, tokenURI}` to `/api/agent/manifest` + `.well-known/agent.json`.
4. Add `.well-known/agent-registration.json` for domain-ownership verification.

## Explicitly OUT of scope (and why)

- **Reputation/Validation legs**: only meaningful with a genuinely **independent** validator. Self-attestation would violate the claims-must-be-true rule — the registration is not "ERC-8004 validated" until an independent party posts against it. Follow-up issue when a validator identity exists.
- This does **not** resolve #787 (top-of-funnel conversion) or #1294 (faucet friction) — orthogonal problems; adjudicated 2026-08-30.

## Acceptance
- `IdentityRegistry.ownerOf(agentId)` returns the platform wallet on Arc testnet (cast call, command in PR body).
- `curl https://archimedes-arc.com/.well-known/agent.json | jq .erc8004` returns the block with the real agentId.
- The EIP is still Draft status: pin the ABI we compiled against in the PR, and note schema-change risk.
- Anti-goal: no claim of "validated" or "reputation" anywhere in UI/docs from this leg alone.

Issue #1525, opened 2026-08-30
[oracle] Weekend-blind freshness probe + Circle COMPLETE parsed as ERROR + wedged-tx retry loop — oracle_fresh false every weekend


dbrowneup:
## Symptom

`/api/health` has reported `oracle_fresh: false` since Saturday: "1/2 probed oracle(s) fresh, oldest age 159716s (of 281 in the universe)".

## Diagnosis (runner logs `/archimedes/runners` stream `oracle`, read 2026-08-30 16:23 UTC — entries minutes old)

The updater is **alive and cycling every minute**. 159,716s of staleness dates sSPY's last on-chain update to **Friday ~19:40 UTC — US market close**. sBTC (24/7 market) stays fresh. Three distinct defects, none of them "the updater is down":

1. **Weekend-blind freshness probe.** Equity synths cannot update when the market is closed; the probe applies one flat age threshold, so every weekend `oracle_fresh` flips false and `archimedes-oracle-stale` alarms (6 ALARM transitions this week). A "stale" signal that fires every weekend by design trains operators to ignore it — which is how a real weekday staleness will get missed.

2. **Circle `COMPLETE` parsed as an error.** Every cycle logs
   `ERROR archimedes.chain.oracle_updater Circle API error for sBTC (200): {'data': {'id': …, 'state': 'COMPLETE'}}`
   — a **successful terminal state logged as ERROR**. sBTC's tx id changes each cycle and its pushes land, so the parse failure is cosmetic there — but it means the code cannot distinguish success from failure on this path, and the ERROR volume buries real errors.

3. **A wedged tx is re-polled forever.** sSPY re-polls the **same** Circle tx id (`30a4c44b-a934-5b6b-bd6e-9f8668693908`) every cycle with no abandon/retry-new path. Caveat from review: the weekend explains the *timing* of sSPY's staleness but not the wedge itself — whether sSPY recovers at Monday's market open (2026-08-31 13:30 UTC) is the discriminating experiment. If it does not, defect 3 is the active cause, not just latent.

Also observed: `INFO … Prices fetched (no on-chain push — owner key not configured)` every cycle — either a dead code path that should be silenced or a real misconfiguration; adjudicate while in the file.

## Fix (probe-first, no new deps)

[...]

PR #1821, opened 2026-09-03
[chain] ERC-8004 register path: bounded log window, idempotent-replay 200, STUCK handling (#1527 pre-flight)

dbrowneup:
Part of #1527.

Pre-flight for the identity mint. This does **not** register anything: no transaction is sent, no AWS or on-chain write happens here. It fixes the three ways the register path turns a transaction that *landed* into an operator who believes it did not, plus the two Circle-API gaps the grounding review found.

Grounding: a local report (`~/.claude/jobs/09189ac8/tmp/erc8004-grounding.md`, not readable from GitHub) — §1 the call graph, §4 items 1–3 and 10 the pre-flight blockers, "What I would fix in code" 1–3. Everything it is cited for below is restated inline here. On-chain and RPC facts in it were read live on 2026-09-03 against `https://rpc.testnet.arc.network`.

### 1. The mint-log scan defaulted to block 0 — blocker

There is no owner→agentId lookup on the ERC-8004 registry, so discovery is an `eth_getLogs` scan for the mint. Arc's public RPC refuses a range wider than 10,000 blocks:

```
{"code":-32614,"message":"eth_getLogs is limited to a 10,000 range"}
```

The chain is ~60,294,500 blocks tall. So the confirming read after `--execute` was **guaranteed** to fail: the mint lands, the script reports `action: submitted`, and — before this — prints nothing further.

`--from-block` now defaults to `eth_blockNumber - 9,000`, resolved **once per run** and used by *both* chain reads around the mint (the idempotency check before, the confirming scan after), so what proved the wallet was empty is the same range that later finds what it minted. 9,000 rather than 10,000 leaves headroom for blocks mined between resolving the window and running the scan. An explicit `--from-block` is used verbatim, block tags included. An unreachable RPC falls back to 0 and says loudly that the scan will probably be refused — it never invents a height, because a manufactured "no identity found" is the input that decides whether a second, un-burnable identity gets minted.

### 2. `--verify` never forwarded `--from-block` (`:301`)

It called `find_agent_id(owner)` — no window at all. `balanceOf == 0` short-circuits the scan today, so the bug is invisible right up until there is an identity to find, which is the only moment `--verify` matters. Now forwarded, through the same resolver.

### 3. `action: submitted` was a dead end

The agentId is not lost in that state — it is `topics[3]` of the `Transfer(0x0, owner, tokenId)` log in the receipt of the transaction the script just printed. The follow-up is now printed anyway, behind a recovery block carrying the tx hash, the `cast receipt` command, the topic layout, and an explicit *do not reach for `--allow-second-identity`*. The registrations-entry template ships with an `<AGENT_ID>` placeholder as an integer, not a quoted string.

### 4. `circle_signer.py:248` — 201-only submit check

Circle documents **200** for a replayed `idempotencyKey` and 201 for a new transaction. Ours is deterministic *by design* (`uuid5` over wallet+contract+function+args), which makes a replay the shape of any retry rather than a rarity — so a landed transaction was being reported as `Circle contract execution failed (200)`, and the tempting next move after that reading is to force a second one.

Now `{200, 201}`. **The deterministic uuid5 key is unchanged**, and is now pinned by a test — accepting 200 is only correct while the key stays content-addressed.

### 5. `circle_signer.py:30` — `STUCK` unhandled

The poller spun its full 120-second budget and raised `Circle tx <id> timed out after 60 polls`, describing the poller rather than the transaction. Circle: a stuck transaction *"can block subsequent transactions from the same wallet"* — and that wallet also runs the oracle's price pushes. It now fails immediately, naming the Circle transaction id to accelerate or cancel in the Console. `STUCK` deliberately stays out of `_TERMINAL` (Circle does not call it a failure); the ordinary in-flight states `QUEUED`/`SENT`/`CONFIRMED` still poll through, and the generic timeout survives for what it was always for.

## Tests — hermetic, and each guard shown red first

`backend/tests/test_erc8004_log_window.py` (19 after the review round, 13 before) and `backend/tests/test_circle_signer.py` (27 after, 17 before). Targeted modules — those two plus `chain/test_circle_signer.py`, `test_erc8004_identity.py`, `test_erc8004_registration.py`, `test_agent_manifest.py` — **120 passed**. Full backend suite from the repo root, `eb7a3037`: **5834 passed, 6 skipped** in 27:03; three unrelated modules (`test_security_hardening.py` ×2, `test_generation_credits.py` ×1) flaked on a box running 12 concurrent suites — two `subprocess.TimeoutExpired … after 30 seconds` and one `sqlite3.OperationalError: database is locked` — and all 49 of their tests pass on a serial re-run. CI's **Backend — unit tests** is green on `eb7a3037` (15m48s). No network, no credentials, no key material — the "entity secret" in the signer tests is 32 bytes of `0x11` against a throwaway in-process RSA key, and the RSA-OAEP encryption runs for real.

`RangeCappedRPC` serves both transports the code uses — the script's `httpx` JSON-RPC (`eth_blockNumber`, how the window is resolved) and the backend's web3 `get_logs` (the scan itself). The **real** 10,000-block cap is enforced on the `get_logs` side, which is the only one that scans; the httpx side answers the height and rejects everything else. Every "the window works" test is paired with one proving the opposite input is refused — the unbounded scan the fix replaced, and (after the review round below) the too-narrow window the fix introduced.

Every guard was mutated red before this was pushed (mutation → run → capture → revert by editing):

| Mutation | Red |
|---|---|
| `start = 0` in `resolve_from_block` | `assert 0 == (60294500 - 9000)`, and `--verify` end-to-end: `ValueError: {'code': -32614, 'message': 'eth_getLogs is limited to a 10,000 range'}` |
| `find_agent_id(owner)` (drop the forward) | `assert [0] == [51000]`, plus the same `-32614` |
| delete the `submitted` follow-up | `AssertionError: assert 'cast receipt' in 'EXECUTE: … detail: register() was submitted but no identity is readable yet\n'` |
| `RangeCappedRPC.LIMIT = 100_000_000` | `Failed: DID NOT RAISE <class 'ValueError'>` — the cap is load-bearing, not decorative |
| `if resp.status != 201` | `RuntimeError: Circle contract execution failed (200): {'data': {'id': 'aaaa…', 'state': 'INITIATED'}}` |
| `if not (200 <= resp.status < 300)` (over-wide) | `KeyError: 'data'` on 202 and 204 — the widening must not become "any 2xx" |
| delete the `STUCK` branch | `AssertionError: assert 'STUCK' in 'Circle tx aaaaaaaa-… timed out after 60 polls'` |

## Also

`docs/runbooks/erc8004-identity-registration.md` gains the window explanation, the `submitted` recovery, and the 200/`STUCK` notes at the point in the runbook where each bites.

## Review round (commit `eb7a3037`)

A verification pass found one **blocker in this PR's own headline fix**, and three minors. All four are addressed here.

### Blocker — the bounded window converted a fail-safe into a fail-open double mint

`find_agent_id` returned `(None, detail)` for two different facts:

1. `balanceOf == 0` — definitively no identity.
2. `balanceOf > 0` but the scanned range names nothing.

With `from_block=0`, case 2 was **unreachable**: the Arc RPC answered `-32614`, the read raised, and `register_identity` returned `action: refused` ("refusing to mint blind"). With `--from-block` defaulting to head−9,000, any identity minted more than 9,000 blocks ago makes the scan **succeed** and name nothing — so `_resolve_existing` returned `_pending(...)`, `register_identity` fell through, and `register(string)` was submitted **for a wallet that already holds an identity**. Arc's block time measured live at 0.515 s/block, so the window ages out ~77 minutes after any real mint; from then on every `--execute` re-run double-minted. `_resolve_existing`'s own docstring states the rule that violated: *"'I could not check' and 'there is nothing there' must not share a return value."*

Proven live against `https://rpc.testnet.arc.network` on wallet `0xb7ac…1553` (which holds two identities — `ownerOf(1)` names it): the PR branch printed `status: registration_pending (no identity found for this wallet)` and exited **0**, where `origin/main` failed loudly with `-32614`. This PR traded main's *always refuses after a mint* — safe and annoying — for *usually correct, occasionally mints an identity that cannot be un-minted*.

**Fixed:**

- `find_agent_id` now **raises** on the `balance > 0` / nothing-in-range branch. `(None, detail)` is returned for exactly one fact: `balanceOf == 0`. `register_identity`'s existing "could not read the registry → refused" arm catches it, so nothing is submitted.
- `--verify` no longer prints `registration_pending (no identity found for this wallet)` and exits 0 for that state. It prints `status: undetermined`, names the way out (`--verify --agent-id <ID>`, which reads `ownerOf` and scans nothing; or a wider `--from-block`), and exits **1**. That pending line is what step 2 of the runbook reads as "step 3 is safe".
- The runbook now documents **three** answers at step 2, not two.

There was **no test on this path at all** — applying the one-line raise left all of the previous round's targeted tests green. Six now exist, each with its paired opposite so the fix cannot degenerate into "refuse always":

| Guard | Paired proof |
|---|---|
| `test_a_balance_the_window_cannot_name_raises_instead_of_answering_no` | `test_a_zero_balance_is_still_a_plain_no_identity_found` |
| `test_register_refuses_rather_than_minting_a_second_identity_it_cannot_see` (asserts `signer.submits == []`) | `test_an_empty_wallet_still_mints_exactly_once` |
| `test_verify_says_undetermined_not_pending_when_the_window_could_not_look` | `test_verify_still_reports_pending_for_a_wallet_that_really_holds_nothing` |

### Minors

- **`backend/tests/test_circle_signer.py` header claimed the oracle goes through `circle_signer`.** It does not: `oracle_updater.push_prices_on_chain` POSTs to `/developer/transactions/contractExecution` itself and polls with its own `_poll_circle_tx`. The header now enumerates the **real** callers — `TracePublisher.publish`/`.commit`/`.reveal`, `StrategyPublisher.anchor`, `ChainExecutor.execute_trades`/`.create_vault`/`._send_vault_admin_tx`/`.set_token_oracles`/`.set_target_allocations`, `bootstrap_vaults` (14 sites), `deploy_contracts`, `amm_bootstrap`, `marketplace_routes`, `erc8004_identity.register_identity` — and states that the oracle fixed this *identical* 200-vs-201 defect first, under **#1525**, grading on the payload rather than the status. The `_SUBMIT_ACCEPTED` comment records that as a known divergence rather than an accident. (Converging the two is its own change; `_poll_circle_tx` still has no `STUCK` case.)
- **`chain/test_circle_signer.py::test_non_201_submit_raises` documented the opposite of the shipped behaviour.** Renamed `test_a_rejected_submit_raises` (it uses 400, so it always passed), with a pointer from that canonical file to where `{200, 201}` is proven.
- **`body["data"]["id"]` was unguarded on the newly accepted 200 path.** A malformed 2xx — an error envelope, a proxy's stamp, a shape change — came out as `KeyError: 'data'` from inside the signer, which no caller on this seam catches: `register_identity` turns a `RuntimeError` into `action: refused` and would turn a `KeyError` into a traceback. Now a clean `RuntimeError`, with ten parametrized cases.
- **The `RangeCappedRPC` overclaim** in the tests-section paragraph above and in the test module's own `HERMETIC` note is corrected: the cap is enforced on the web3 side only.

### Review-round mutations (mutate → run → capture → revert by editing)

| Mutation | Red |
|---|---|
| `raise RuntimeError(…)` → `return None, (…)` in `find_agent_id` (i.e. undo the blocker fix) | `Failed: DID NOT RAISE <class 'RuntimeError'>`; `AssertionError: DOUBLE MINT: register() was submitted for a wallet already holding agentId 3`; `AssertionError: an undetermined read exited 0` — 3 failed, 16 passed |
| the lazy over-fix: `balance == 0` raises too | `RuntimeError: balanceOf(0x9257…) == 0 — this wallet holds no ERC-8004 identity`; `AssertionError: an empty wallet did not register — assert 0 == 1`; `assert 1 == 0` — 3 failed, 16 passed |
| restore `circle_tx_id = body["data"]["id"]` | `KeyError: 'data'` / `TypeError: 'NoneType' object is not subscriptable` / `KeyError: 'id'` — 10 failed, 17 passed |
| `--verify` keeps the `undetermined` wording but `return 0` | `AssertionError: an undetermined read exited 0 — assert 0 != 0` |
| `_SUBMIT_ACCEPTED = frozenset({200, 201, 400})` | `test_a_rejected_submit_raises`: `Regex pattern did not match. Expected regex: 'contract execution failed'` — the renamed test still guards what its new name claims |

**Out of scope, deliberately:** `oracle_updater._poll_circle_tx` is a second, independent poller with the same missing `STUCK` case (it returns states rather than raising, so the shape of the fix differs); the by-id `GET /transactions/{id}` poll; `setAgentURI` in the pinned ABI; and `sign_and_broadcast`, which the grounding review found cannot work as written (wrong endpoint, missing `entitySecretCiphertext`) and which nothing on this path calls.

Owner decisions from the grounding review — which wallet owns the identity, and whether `agentWallet` should be moved to the x402 recipient — are untouched here; they are §5 of that report and want an answer *before* signing.

### DIFF
--- a/backend/archimedes/chain/circle_signer.py
+++ b/backend/archimedes/chain/circle_signer.py
@@ -52,6 +52,14 @@ _MAX_POLLS = 60  # 2 minutes max
 # expected shape of any retry, not a rarity. Treating 200 as a failure told the operator a
 # transaction that had succeeded had failed, and the tempting next move after that is to
 # force a second one.
+#
+# KNOWN DIVERGENCE, deliberate: oracle_updater.py does NOT come through this method — it
+# POSTs to the same endpoint itself — and it fixed this identical defect first, under #1525
+# ("Circle API error for sBTC (200): {'state': 'COMPLETE'}" every cycle for a push that had
+# landed). Its fix grades on the PAYLOAD (a usable id whose state is not a failure state);
+# this one grades on the STATUS and then requires a usable id below. Same outcome for the
+# shapes Circle actually returns, two implementations, one seam apart. Converging them is
+# its own change; until then, a fix here is not a fix there.
 _SUBMIT_ACCEPTED = frozenset({200, 201})
 
 # Bounded wall-clock ceiling on the wallet-address lookup (#1412). This runs
@@ -272,7 +280,18 @@ class CircleSigner:
                 body = await resp.json()
                 if resp.status not in _SUBMIT_ACCEPTED:
                     raise RuntimeError(f"Circle contract execution failed ({resp.status}): {body}")
-                circle_tx_id = body["data"]["id"]
+                # A 2xx is not automatically a usable answer. ``body["data"]["id"]`` was an
+                # unguarded double subscript: a malformed success — an error envelope, a
+                # proxy's HTML, a 200 with no ``data`` — surfaced as ``KeyError: 'data'``
+                # from inside the signer instead of the RuntimeError every caller here is
+                # written to handle. Newly reachable now that 200 is accepted at all.
+                data = body.get("data") if isinstance(body, dict) else None
+                circle_tx_id = data.get("id") if isinstance(data, dict) else None
+                if not circle_tx_id:
+                    raise RuntimeError(
+                        f"Circle accepted the contract execution ({resp.status}) but the response carries "
+                        f"no transaction id to poll, so nothing can be confirmed: {body}"
+                    )
                 if resp.status == 200:
                     # Not a new transaction — Circle recognised the idempotency key and
                     # handed back the one it already has. Logged as such so a reader of
--- a/backend/archimedes/chain/circle_signer.py
+++ b/backend/archimedes/chain/circle_signer.py
@@ -26,10 +26,33 @@ logger = logging.getLogger(__name__)
 CIRCLE_API_BASE = "https://api.circle.com/v1/w3s"
 CIRCLE_BLOCKCHAIN = "ARC-TESTNET"
 
-# Transaction terminal states
+# Transaction terminal states. Circle's full enum is INITIATED, CLEARED, QUEUED, SENT,
+# STUCK, CONFIRMED, COMPLETE, FAILED, DENIED, CANCELLED (get-transaction API reference).
+# CONFIRMED is deliberately NOT terminal: Circle's blockchain-confirmations doc is explicit
+# that a transaction visible on a block explorer may still not be COMPLETE.
 _TERMINAL = {"COMPLETE", "FAILED", "DENIED", "CANCELLED"}
+
+# STUCK is its own case, and it is not in _TERMINAL because Circle does not call it a
+# failure: "STUCK is a signal for intervention, not a terminal failure" — and, crucially,
+# "if a transaction stays STUCK and you take no action, it can block subsequent
+# transactions from the same wallet" (transaction-limits-and-optimizations). For a caller
+# it is nonetheless the end of the road: nothing this process can do will unstick it, so
+# spinning out the full poll budget and then raising a generic timeout costs two minutes
+# and throws away the one fact the operator needs — WHICH transaction to accelerate or
+# cancel in the Circle Console, before it wedges the oracle's pushes from the same wallet.
+_STUCK = "STUCK"
+
 _POLL_INTERVAL = 2.0  # seconds
 _MAX_POLLS = 60  # 2 minutes max
+
+# Circle answers a contract-execution submit with 201 for a newly created transaction and
+# 200 when it recognises the idempotencyKey and replays the existing one
+# (create-developer-transaction-contract-execution). Our key is deterministic BY DESIGN —
+# uuid5 over wallet+contract+function+args (see execute_contract) — so a replay is the
+# expected shape of any retry, not a rarity. Treating 200 as a failure told the operator a
+# transaction that had succeeded had failed, and the tempting next move after that is to
+# force a second one.
+_SUBMIT_ACCEPTED = frozenset({200, 201})
 
 # Bounded wall-clock ceiling on the wallet-address lookup (#1412). This runs
 # inside the agent tick's reveal-reconciliation pass, so an unbounded read
@@ -192,7 +215,9 @@ class CircleSigner:
             The on-chain transaction hash.
 
         Raises:
-            RuntimeError: If Circle credentials are missing or the tx fails.
+            RuntimeError: If Circle credentials are missing, the submit is rejected
+                (anything outside :data:`_SUBMIT_ACCEPTED`), the transaction reaches a
+                failing terminal state, goes ``STUCK``, or the poll budget runs out.
         """
         if not self.is_configured:
             raise RuntimeError("Circle credentials not configured (CIRCLE_API_KEY / CIRCLE_ENTITY_SECRET / WALLET_ID)")
@@ -245,10 +270,16 @@ class CircleSigner:
                 },
             ) as resp:
                 body = await resp.json()
-                if resp.status != 201:
+                if resp.status not in _SUBMIT_ACCEPTED:
                     raise RuntimeError(f"Circle contract execution failed ({resp.status}): {body}")
                 circle_tx_id = body["data"]["id"]
-                logger.info("Circle tx submitted: %s", circle_tx_id)
+                if resp.status == 200:
+                    # Not a new transaction — Circle recognised the idempotency key and
+                    # handed back the one it already has. Logged as such so a reader of
+                    # the logs can tell one submission from two.
+                    logger.info("Circle tx replayed (idempotent, HTTP 200): %s", circle_tx_id)
+                else:
+                    logger.info("Circle tx submitted: %s", circle_tx_id)
 
             # Poll until terminal state
             return await self._poll_transaction(session, circle_tx_id)
@@ -313,7 +344,13 @@ class CircleSigner:
             return tx_hash
 
     async def _poll_transaction(self, session: aiohttp.ClientSession, circle_tx_id: str) -> str:
-        """Poll Circle transaction until terminal state."""
+        """Poll Circle transaction until terminal state.
+
+        Returns the on-chain hash only for ``COMPLETE``. Every other end — a failing
+        terminal state, ``STUCK``, or exhausting the budget — raises, and the ``STUCK``
+        message names the transaction because that is what an operator has to type into
+        the Circle Console to clear it.
+        """
         for _ in range(_MAX_POLLS):
             # Scope the list to this wallet's transactions via the walletIds
             # filter. An unfiltered GET /transactions returns the newest txs
@@ -331,6 +368,17 @@ class CircleSigner:
                         if tx.get("id") == circle_tx_id:
                             state = tx.get("state", "UNKNOWN")
                             tx_hash = tx.get("txHash", "")
+
+                            if state == _STUCK:
+                                raise RuntimeError(
+                                    f"Circle tx {circle_tx_id} is STUCK "
+                                    f"(txHash {tx_hash or 'not yet assigned'}, wallet {self._wallet_id}): "
+                                    "Circle accepted and broadcast it but it is not being mined, and a "
+                                    "stuck transaction blocks every later transaction from the same "
+                                    "wallet — including the oracle's price pushes. Accelerate or cancel "
+                                    f"transaction {circle_tx_id} in the Circle Console (Transactions), "
+                                    "then re-run. Do NOT submit a replacement first."
+                                )
 
                             if state in _TERMINAL:
                                 if state == "COMPLETE":
--- a/backend/archimedes/chain/erc8004_identity.py
+++ b/backend/archimedes/chain/erc8004_identity.py
@@ -316,8 +316,11 @@ async def find_agent_id(owner: str, client=None, from_block: int | str = 0) -> t
     candidate is then re-checked with ``ownerOf``, because a mint log proves the token was
     once minted to this wallet, not that it is still held.
 
-    ``(None, detail)`` means "no identity found"; a raised exception means "could not
-    look" and the caller must not read it as either.
+    ``(None, detail)`` means "no identity found", and it is returned for exactly ONE fact:
+    ``balanceOf == 0``. A raised exception means "could not look" — the RPC refused the
+    range, or the wallet holds something this window could not name — and the caller must
+    not read it as either. Those two must never share a return value: the caller's next
+    move on "no identity found" is to MINT, and a second identity cannot be un-minted.
     """
     from archimedes.chain.client import chain_client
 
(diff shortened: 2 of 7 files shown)


dbrowneup:
## Fix round — all four defects addressed in `eb7a3037`

Thanks for the live proof on `0xb7ac…1553`. The blocker was real and I have taken the fix you specified. Defect → commit map below; the whole round is one commit.

| # | Severity | Defect | Where it is fixed in `eb7a3037` |
|---|---|---|---|
| 1 | **blocker** | `find_agent_id` returns `(None, detail)` for both "balanceOf == 0" and "balanceOf > 0, window named nothing" → fail-open double mint | `backend/archimedes/chain/erc8004_identity.py` (raise), `scripts/register_erc8004_identity.py::verify` (`undetermined`, exit 1), `backend/tests/test_erc8004_log_window.py` (6 new tests), `docs/runbooks/erc8004-identity-registration.md` (three answers at step 2) |
| 2 | minor | test-module header claims the oracle's price pushes go through `circle_signer`; no #1525 cross-reference | `backend/tests/test_circle_signer.py` (header rewritten), `backend/archimedes/chain/circle_signer.py` (`_SUBMIT_ACCEPTED` comment) |
| 3 | minor | `test_non_201_submit_raises` documents the opposite of shipped behaviour | `backend/tests/chain/test_circle_signer.py` (renamed + pointer) |
| 4 | minor | PR-body `RangeCappedRPC` overclaim; unguarded `body["data"]["id"]` on the 200 path | PR body + `test_erc8004_log_window.py` HERMETIC note; `circle_signer.py:280-292` + 10 parametrized tests |

---

### 1 — blocker

Taken exactly as prescribed. `find_agent_id`'s final `return None, (…)` is now `raise RuntimeError(<same message>)`, and `(None, detail)` is returned for exactly one fact: `balanceOf == 0`. The docstring now states that as the contract rather than describing it loosely:

> `(None, detail)` means "no identity found", and it is returned for exactly ONE fact: `balanceOf == 0`. A raised exception means "could not look" — the RPC refused the range, or the wallet holds something this window could not name — and the caller must not read it as either. Those two must never share a return value: the caller's next move on "no identity found" is to MINT, and a second identity cannot be un-minted.

`register_identity`'s existing "could not read the registry → refused" arm catches it, so nothing is submitted. `--verify` no longer prints `registration_pending` for that state. Here is the real output on both sides of the line (hermetic, `RangeCappedRPC` at the live head 60,294,500):

```
=== A) mint 50,000 blocks back — OUTSIDE the default window ===
scan:       scanning blocks 60285500..latest — a 9,000-block window below head 60,294,500, because the public Arc RPC refuses eth_getLogs ranges wider than 10,000 (-32614)
discovery:  could not look (RuntimeError: balanceOf(0x9257…) == 1 but no mint log in range names a token this wallet still owns (scanned from block 60285500); pass the agentId explicitly)
status:     undetermined  — NOT 'no identity found'. Nothing was proved either way.
            Re-run as --verify --agent-id <ID> (reads ownerOf directly, scans no logs),
            or widen the scan with --from-block <block below the mint>.
            Do NOT run --execute off this answer.
exit=1

=== B) balanceOf == 0 — genuinely nothing ===
scan:       scanning blocks 60285500..latest — a 9,000-block window below head 60,294,500, because the public Arc RPC refuses eth_getLogs ranges wider than 10,000 (-32614)
discovery:  balanceOf(0x9257…) == 0 — this wallet holds no ERC-8004 identity
status:     registration_pending (no identity found for this wallet)
exit=0
```

You were right that nothing covered the path — six tests now do, each with its paired opposite so the fix cannot degenerate into "refuse always":

| Guard | Paired proof it did not over-fix |
|---|---|
| `test_a_balance_the_window_cannot_name_raises_instead_of_answering_no` | `test_a_zero_balance_is_still_a_plain_no_identity_found` |
| `test_register_refuses_rather_than_minting_a_second_identity_it_cannot_see` — asserts `signer.submits == []` | `test_an_empty_wallet_still_mints_exactly_once` |
| `test_verify_says_undetermined_not_pending_when_the_window_could_not_look` | `test_verify_still_reports_pending_for_a_wallet_that_really_holds_nothing` |

The `aged` fixture is the state you measured: mint 50,000 blocks back, i.e. any wallet ~77 minutes after its mint at the 0.515 s/block you clocked.

### 2 — the oracle is not on this seam

Correct, and the header was wrong. It now enumerates the real callers of `CircleSigner.execute_contract` — `TracePublisher.publish`/`.commit`/`.reveal`, `StrategyPublisher.anchor`, `ChainExecutor.execute_trades`/`.create_vault`/`._send_vault_admin_tx`/`.set_token_oracles`/`.set_target_allocations`, `bootstrap_vaults` (14 sites), `deploy_contracts`, `amm_bootstrap`, `marketplace_routes`, `erc8004_identity.register_identity` — and says plainly that the oracle is not among them: `oracle_updater.push_prices_on_chain` POSTs to the endpoint itself and polls with its own `_poll_circle_tx`.

#1525 is cited in both places. The `_SUBMIT_ACCEPTED` comment now carries:

> KNOWN DIVERGENCE, deliberate: `oracle_updater.py` does NOT come through this method … and it fixed this identical defect first, under #1525 … Its fix grades on the PAYLOAD (a usable id whose state is not a failure state); this one grades on the STATUS and then requires a usable id below. Same outcome for the shapes Circle actually returns, two implementations, one seam apart. Converging them is its own change; until then, a fix here is not a fix there.

Converging them (and giving `_poll_circle_tx` a `STUCK` case) stays out of scope for this PR, now visibly rather than silently.

### 3 — the misnamed contract

`test_non_201_submit_raises` → `test_a_rejected_submit_raises`, with a docstring saying why, and a pointer at the top of `backend/tests/chain/test_circle_signer.py` to where `{200, 201}` is proven.

### 4 — the fake, and the unguarded subscript

PR body reworded to your text: the cap is enforced on the web3 `get_logs` side, the only one that scans; the httpx side answers `eth_blockNumber` and rejects everything else. The test module's `HERMETIC` note made the same overclaim and is corrected too.

The id read is guarded as you specified, and ships with ten parametrized cases (`{}`, `{"data": null}`, `{"data": {}}`, `{"data": {"state": …}}`, an error envelope under a 2xx — each on both 200 and 201):

```python
data = body.get("data") if isinstance(body, dict) else None
circle_tx_id = data.get("id") if isinstance(data, dict) else None
if not circle_tx_id:
    raise RuntimeError(
        f"Circle accepted the contract execution ({resp.status}) but the response carries "
        f"no transaction id to poll, so nothing can be confirmed: {body}"
    )
```

---

### Mutations for this round — each shown red, reverted by editing

| Mutation | Red output (verbatim) |
|---|---|
| undo the blocker fix: `raise RuntimeError(…)` → `return None, (…)` | `Failed: DID NOT RAISE <class 'RuntimeError'>`<br>`AssertionError: DOUBLE MINT: register() was submitted for a wallet already holding agentId 3`<br>`AssertionError: an undetermined read exited 0 — assert 0 != 0`<br>**3 failed, 16 passed** |
| the lazy over-fix: make `balance == 0` raise too | `RuntimeError: balanceOf(0x9257…) == 0 — this wallet holds no ERC-8004 identity`<br>`AssertionError: an empty wallet did not register — assert 0 == 1`<br>`assert 1 == 0`<br>**3 failed, 16 passed** — the paired proofs reject it |
| restore `circle_tx_id = body["data"]["id"]` | `KeyError: 'data'` / `TypeError: 'NoneType' object is not subscriptable` / `KeyError: 'id'` — **10 failed, 17 passed** |
| keep the `undetermined` wording but `return 0` | `AssertionError: an undetermined read exited 0 — assert 0 != 0` |
| `_SUBMIT_ACCEPTED = frozenset({200, 201, 400})` | `test_a_rejected_submit_raises`: `AssertionError: Regex pattern did not match. Expected regex: 'contract execution failed'` — the renamed test still guards what its new name claims |

All five reverted by editing the source back; `git status --porcelain` empty; re-run green.

### Your required checks, re-verified after the fix

- **`--plan` byte-identical** — `--plan --offline` and `--plan --offline --from 0x9257…` and `--print-followup 42`, diffed against `wt-1527-main` at `df8379a6`: `PLAN IDENTICAL` / `PLAN2 IDENTICAL` / `FOLLOWUP IDENTICAL`.
- **Idempotency key** — still untouched; `circle_signer.py:238-249` is not in this round's diff either.
- **On-chain send semantics** — one addition only: a malformed 2xx now raises instead of `KeyError`. No new outbound call. No AWS writes, no on-chain sends made producing this round.
- **Suite** — targeted **120 passed**; full backend suite from the repo root **5834 passed, 6 skipped** in 27:03. Three unrelated modules flaked under 12 concurrent suites on this box (two `subprocess.TimeoutExpired … after 30 seconds` in `test_security_hardening.py`, one `sqlite3.OperationalError: database is locked` in `test_generation_credits.py`); all 49 of their tests pass on a serial re-run, and CI's **Backend — unit tests** is green on `eb7a3037` (15m48s).
- **ruff** — `All checks passed!` / `6 files already formatted`.

Not filed as defects but taken anyway: your cosmetic note about the local-path citation (the PR body now says it is a local file and restates its content inline), and the runbook's "step 3 is safe" line, which was the human-facing half of the blocker and is now a three-row table.

PR #1823, opened 2026-09-03
[rigor] One DSR bar: 0.95 everywhere, the 0.90 path retired (#1794)

### LINKED ISSUE #1794: [rigor][product-truth] Two DSR bars in code: _rigor_verdict_for hardcodes 0.95 while the strictest rigor profile (the badge) says 0.90 — and the public docs quote 0.95
Three places, two numbers:

| where | DSR bar |
|---|---|
| `rigor_profiles.get_profile(STRICTEST_LEVEL).dsr_p_min` — what the Library/passport badge grades against | **0.90** |
| `strategies_routes._rigor_verdict_for` (hardcoded) | **0.95** |
| `docs/rigor-methods.md:350` ("≥ 0.95"), `docs/quant/admission-criteria.md` | **0.95** |

Consequences already observed tonight: (a) the docs-site audit's refuters upheld "the public rigor page states a threshold that is not the live one" as a P0 (decision package for the owner); (b) PR #1788 (rejected-card reasons) had to invent an `unattributed` state for the [0.90, 0.95) band because a strategy can be `passing=false` under the 0.95 call site while every per-check threshold the badge profile publishes says pass; (c) the verifier on #1788 also found `_rigor_verdict_for` has zero production callers on main — so the 0.95 may be dead code, in which case the fix is deletion plus a docs correction.

## Fix
One bar, one source. Decide the number (Önder's call — the profile's 0.90 is what the badge has been grading with in prod), delete or repoint the hardcoded call site, correct the two docs pages, and add a guard that greps the repo for a literal DSR threshold outside `rigor_profiles` (mutation: reintroduce `0.95` in a route → red).

Related: #1788, #1792 (stored verdict writes `gate_version` — the bar must be part of the version digest), the docs-site decision package (artifact 762d994c).

dbrowneup:
Closes #1794.

## The defect

Two Deflated-Sharpe bars were written down in the tree, and which one graded a strategy depended on which code path reached it.

| where | bar before | status |
|---|---|---|
| `rigor_profiles.get_profile(STRICTEST_LEVEL).dsr_p_min` | **0.90** (PR #901) | live everywhere the badge is decided |
| `generation_pipeline._rigor_verdict_for` (hardcoded) | **0.95** | in the tree, no production caller |
| `generation_pipeline._patch_dsr_with_pool_correlation` (hardcoded) | **0.95** | called by `run_generation`, but a no-op on today's pipeline |
| every public rigor page + the app's own copy | **0.95 and 0.90**, inconsistently | shipped to readers |

### One correction to the issue's framing, verified in the code

The issue asks whether the 0.95 is dead code. I traced both call sites rather than assume:

- **`_rigor_verdict_for` has no production caller.** It moved from `strategies_routes` to `generation_pipeline:522`; every reference outside that definition is a test. The #1788 verifier was right about this function.
- **`_patch_dsr_with_pool_correlation` (the second literal, line 733) IS called from `run_generation`** — but its first statement is `agent_cands = [c for c in candidates if not c.has_real_rigor]`, and today's Phase-3 debate-only pipeline has exactly two runners: `_run_debate_leaderboard`, which stamps `has_real_rigor=True` on every entry (`debate_engine:862`), and `_run_fixture_candidate`, whose verdict carries `dsr: None` and no return series, so the loop `continue`s past it. So it reaches no candidate today either.
- **The bar that actually grades a live generated strategy is the profile**, via `fusion_evaluator`'s `_badge = get_profile(STRICTEST_LEVEL)` → `dsr_p >= _badge.dsr_p_min`, whose result becomes `passes_rigor` on the debate candidate.

So the shipped inconsistency was **badge/docs**, not badge/generator: the gate graded at 0.90 while every public page said 0.95. The two 0.95 literals were the loaded gun rather than the shot fired — a buy-and-hold path returning would have re-introduced the split silently. Both are now the constant.

Owner call, 2026-09-03: **0.95 everywhere, the 0.90 path deleted.**

## One definition

`rigor_profiles.DSR_P_BADGE_MIN = 0.95` is now the only place the number is written down.

- `_PROFILES[1].dsr_p_min` **is** that constant (identity, not a copy — the guard asserts `is`).
- `generation_pipeline` imports it at both former literal sites.
- Every other reader goes through a `RigorProfile` (`rigor_evaluator._passes_profile`, `gate_details["dsr"]`, `fusion_evaluator`, the `/strictness-ladder` disclosure).
- The two `analytics-engine` fixture-regen scripts, which re-grade fixtures against the gate, import the constant instead of restating `0.95`.
- Comments that quoted a bar now name the constant (`rigor_evaluator`, `_rigor_helpers`, `strategies_routes`, `models/backtest`).

Levels 2–5 of the ladder are untouched: they are a *personal deployment* risk knob, not a second badge bar. There was no flag selecting between the bars — the selection was the call site.

## Guards — every one shown RED under mutation before pushing

**1. Source scan — the constant is the only literal.** `backend/tests/test_single_dsr_bar.py::TestOneLiteral` scans the nine modules that compute or serve the gate. Outside `DSR_P_BADGE_MIN`'s own definition line, no `0.9` / `0.90` / `0.95` literal may sit within two lines of DSR code — in a comparison, a default, **or a comment** (a stale comment quoting a retired bar is how this drift started).

Mutation — re-hardcode `0.95` in `_rigor_verdict_for`:

```
E   AssertionError: A DSR-bar literal (0.9 / 0.90 / 0.95) appears next to DSR code outside rigor_profiles.DSR_P_BADGE_MIN. That is the #1794 defect: two bars in the tree, and which one graded you depended on which code path you reached. Import DSR_P_BADGE_MIN (or read profile.dsr_p_min); in a comment, name the constant instead of quoting its value:
E       backend/archimedes/agents/generation_pipeline.py:585: passing = bool(dsr_p is not None and dsr_p >= 0.95 and oos_pass and lookahead_passed)
FAILED backend/tests/test_single_dsr_bar.py::TestOneLiteral::test_no_dsr_bar_literal_outside_its_definition
```

Adversarial mutation A — hide the same literal behind the guard's one documented exemption (the ρ̄ pool-correlation prose), i.e. `... >= 0.95 and ...  # correlation-adjusted ρ̄`. The exemption does not apply to a line that itself names DSR, so it is still caught:

```
E   AssertionError: A DSR-bar literal (0.9 / 0.90 / 0.95) appears next to DSR code outside rigor_profiles.DSR_P_BADGE_MIN. ...
E       backend/archimedes/agents/generation_pipeline.py:585: passing = bool(dsr_p is not None and dsr_p >= 0.95 and oos_pass and lookahead_passed)  # correlation-adjusted ρ̄
```

Adversarial mutation B — name the constant mid-line to claim the definition exemption. Writing the probe is what found the hole: the first draft tested `"DSR_P_BADGE_MIN = " in line`, which this walks straight through. It is now `line.startswith(...)`, and the probe is red:

```
E   AssertionError: A DSR-bar literal (0.9 / 0.90 / 0.95) appears next to DSR code outside rigor_profiles.DSR_P_BADGE_MIN. ...
E       backend/archimedes/agents/generation_pipeline.py:585: passing = bool(dsr_p is not None and dsr_p >= 0.90 and oos_pass and lookahead_passed)  # DSR_P_BADGE_MIN = the real one
```

**2. Behaviour at the bar — DSR p = 0.92 fails, 0.96 passes**, on the badge gate, on `gate_details["dsr"]`, on `_rigor_verdict_for`, and on `_patch_dsr_with_pool_correlation`. Mutation — `DSR_P_BADGE_MIN = 0.95 → 0.90`:

```
E   AssertionError: DSR p=0.92 cleared the badge gate — the bar is not 0.9.
E   assert True is False
FAILED ...::TestGateBehaviourAtTheBar::test_badge_gate_fails_below_and_passes_above

E   AssertionError: PASS (p=0.9200)
E   assert 'PASS (p=0.9200)' == 'FAIL (p=0.9200, need ≥ 0.90)'
FAILED ...::TestGateBehaviourAtTheBar::test_badge_gate_detail_quotes_the_one_bar

E   assert True is False
FAILED ...::TestGateBehaviourAtTheBar::test_generation_verdict_uses_the_same_bar[0.92-False]
FAILED ...::TestGateBehaviourAtTheBar::test_pool_correlation_repatch_uses_the_same_bar[0.92-False]

E   assert 0.9 == 0.95
FAILED ...::TestOneLiteral::test_the_definition_is_singular_and_is_the_badge_bar
```

**3. The docs guard had to be INVERTED — and is now bound to the constant.** `test_quant_docs_conflicts._STALE_095_BAR` forbade `p ≥ 0.95` in `docs/quant/` — i.e. it forbade what is now the live bar, and would have started rejecting correct prose the moment the owner moved the bar back. It is now `_RETIRED_BAR` (the `p ≥ 0.90` form), plus `test_the_retired_bar_guard_is_not_the_live_bar`, which asserts the pattern does **not** match `DSR_P_BADGE_MIN`, and the usual reject/admit probe pairs. Same `→ 0.90` mutation:

```
E   AssertionError: _RETIRED_BAR matches the LIVE bar (p ≥ 0.90). The bar moved in rigor_profiles and this guard was not re-aimed — it now forbids docs from stating the truth.
FAILED backend/tests/test_quant_docs_conflicts.py::TestForbiddenClaims::test_the_retired_bar_guard_is_not_the_live_bar
```

Mutation — put a retired `p ≥ 0.90` gate condition back into `docs/quant/`:

```
E   AssertionError: A retired `p ≥ 0.90` gate condition survives in docs/quant/. The bar is 0.95 (#1794); narrate the old bar, do not state it as a condition:
E       docs/quant/admission-criteria.md:76: Admission is `p ≥ 0.90` at level 1.
FAILED backend/tests/test_quant_docs_conflicts.py::TestForbiddenClaims::test_no_retired_gate_bar
```

The live `_RETIRED_BAR` guard also caught a doc line I had missed by hand (`backtest-interpretation.md:70`) before any of this was pushed.

Every mutation was reverted by editing the value back, never by `git checkout`.

## What changes behaviour — stated plainly

This **tightens** the badge. It is not a no-op, and "nothing that passes today starts failing" does not survive contact with the code, so here is the honest version.

| path that reads the bar | before | after |
|---|---|---|
| `RigorGateResult._passes_profile` → `passes_all` / `passes_at_level` / `min_passing_level` | 0.90 | **0.95 — tightens** |
| `gate_details["dsr"]` (the passport PASS/FAIL string) | 0.90 | **0.95 — tightens** |
| `live_rigor_gate.RigorGateVerdict.from_result` → Library badge, passport, CANDIDATE→VALIDATED promotion | 0.90 | **0.95 — tightens** |
| `/api/selection-bias/gate`, `/gate/{id}`, `/strictness-ladder` (level-1 row) | 0.90 | **0.95 — tightens** |
| `vaults_routes` deploy gate (`min_passing_level`, `passes_all`) | 0.90 | **0.95 — tightens** |
| `fusion_evaluator` → `passes_rigor` on every debate candidate (the live Generate verdict) | 0.90 | **0.95 — tightens** |
| `generation_pipeline._rigor_verdict_for` / `_patch_dsr_with_pool_correlation` | 0.95 | 0.95 — **unchanged** |
| `analytics-engine` fixture-regen gate mirrors | 0.95 | 0.95 — **unchanged** |

Everything that passes at 0.95 today still passes. A strategy with DSR p in [0.90, 0.95) loses the badge and becomes a **level-2 (Balanced)** strategy — `min_passing_level == 2`, `blocked_by_floor == False`: deployable into a user's own vault at their chosen strictness, never "Archimedes Verified". `test_the_retired_090_band_no_longer_earns_the_badge` pins exactly that, replacing `test_recalibration_admits_dsr_between_090_and_095`, which asserted the opposite.

**Who is in the band.** The live gate regrades persisted returns from the DB, which is not in the tree, so I cannot enumerate prod from here. The only in-repo datapoint is the fixture snapshot: `maillard_2010_risk_parity` at `dsr_p_value = 0.943525`, inside the retired band. That snapshot is explicitly *not* what serves the badge (#1187), so it does not by itself flip a served verdict — but it is the one strategy I can name as a likely mover. **This is the check to run against prod before merge.**

Also relevant to #1792: the bar is now a single named constant, and as of the merge round below the `gate_version` digest **does** hash `DSR_P_BADGE_MIN` directly (`badge_bar`, schema 1 → 2) rather than trying to capture a number that lived in three places.

## Docs and UI

> **Corrected after review.** The first version of this section claimed the pass covered "every public rigor page". It did not, and the two sentences below are the corrected account. Review found three surfaces still stating the retired bar — see the fix-round commit.

20 pages and 5 UI files that quoted the bar now quote `0.95` or name the constant: the quant slice, `rigor-methods.md`, `architectural-principles.md`, `claims-ledger.md`, both specs, the demo script, the docs index, the repo `README.md`, the two live openwiki instruction pages, `Architecture.jsx`, `Landing.jsx`, `RigorExplainer.jsx`, `RigorStrictnessControl.jsx`'s fallback ladder, and the test that pins the Landing copy.

`Architecture.jsx` is the one review caught: it is the app's own published statement of the bar, on a shipped route, and it still said "positive at 90% one-sided confidence" — which two of the docs pages this PR edited quote as authoritative. That is #1794's headline symptom, so it shipped in the first pass. Fixed, and now bound to the constant by `TestTheAppQuotesTheLiveBar` rather than hand-copied a fifth time.

Left alone on purpose: dated audit/analysis/runbook records and the `second-wave` findings note's measurement table, which are historical evidence at their own vintage. Four *generated* openwiki pages — `rigor/admission-gate.md`, `rigor/reading-a-backtest.md`, `rigor/documented-conflicts.md`, `findings/measured-results.md` — carry a **stale-on-one-number banner** rather than a hand-patch that would falsify their `generated:` provenance. Two sentences were an exception and were corrected in place, because a banner does not repair a claim that asserts the *inverse* of the truth: `reading-a-backtest.md` called 0.95 "the retired bar", and `measured-results.md` said "the gate has since moved to 0.90". Each banner names which sentences were touched.

## Tests

> **Corrected after review.** The first version of this section omitted the `engine-tests` job, which this PR had broken. See the fix-round commit and the numbers in the fix-round PR comment.

- `pytest` (full backend suite, repo root) — see the fix-round comment for the numbers on the current head.
- `python -m pytest analytics-engine/tests` in a venv built exactly like the `engine-tests` CI job — the gate the first pass turned red. Now green.
- `cd ui && npm test` — **916 passed, 0 failed**.
- `ruff check backend analytics-engine` — 39 findings, identical to `origin/main` (all pre-existing); `ruff format --check` clean on every touched file.

One caveat worth stating: `backend/tests/services/test_generation_pipeline.py`'s two `identity_ledger` tests fail when that file is run as part of a small subset (`no such table: identity_events` — a fixture-ordering artifact). They fail identically on `origin/main` at 32158c79, and pass in the full suite on both.

## Merge round — `origin/main` (bf76999c) merged in

`git merge --no-edit origin/main` was CLEAN — no conflict hunks. It still turned the branch red in three places, all of them real collisions between "one DSR bar" and code that landed on main after this branch was cut. Fixed here.

**1. A stale docstring on main quoted BOTH retired bars.** `rigor_reasons.py:51-53` (from `a39488c0`, main) explained the `unattributed` state by narrating the 0.95-vs-0.90 divergence — the prose form of the exact defect this PR removes, and the single-literal guard caught it. Rewritten to name `DSR_P_BADGE_MIN` instead of quoting a number, and the *claim* needed correcting too: with one bar, no NEW write can open that gap, but the state stays reachable — rows STORED under the retired bar keep the `passing` they were graded with, and a leg with nothing on record blocks admission (fail-closed) while this report calls it `not_computed`, never `fail`, so such a row has no failing check to point at.

**2. `test_unattributed_when_nothing_on_record_falls_below_the_bar` — the FIXTURE assumed the retired bar; the payload logic dropped nothing.** Stated plainly, since the question was which of the two. The test fed `dsr_p_value = 0.92`, a number chosen precisely because it sat *between* the two bars. With one bar at 0.95 that row now has a genuinely failing DSR check, so `unattributed` is correctly `False`. The fixture is now DERIVED from the live bar — `(dsr_p_min + 1.0) / 2` — so it clears whatever the bar is, and a new assertion pins that all five legs read `pass`. Without it the test would keep "passing" on a report full of `not_computed`, which is a different state with a different surface.

The fixture had to move **above** the bar, not below it: `dsr_p_value` is a confidence p-value where `p >= bar` PASSES, so anything under 0.95 is exactly the newly-failing case.

**3. `test_the_digest_moves_when_a_gate_threshold_moves` — the mutation had become a no-op, and chasing it found a real hole.** The test recalibrated level 1 to a hardcoded `0.95`; #1794 moved the live bar onto that very number, so it mutated nothing and the digest correctly did not move. The recalibrated value is now derived from the live bar, with an assertion that it actually differs — a mutation test must not pin the value it mutates to.

The hole underneath: `_PROFILES[1].dsr_p_min` reached the digest only *because* it is `DSR_P_BADGE_MIN`, and this PR pins that with `is`. **`is` does not hold the line under mutation.** Re-writing level 1 as a bare `0.95` literal leaves `test_the_definition_is_singular_and_is_the_badge_bar` GREEN — equal float constants inside one module are the same object — so the bar could have been decoupled from the digest with both guards still passing. `DSR_P_BADGE_MIN` is therefore now hashed on its own, as `badge_bar`. That is independently correct rather than merely defensive: `generation_pipeline` compares against the constant directly, never through a profile, so it is a gate input in its own right. `GATE_VERSION_SCHEMA` goes **1 → 2** per its own documented rule (the input set gained an item).

Red/green by mutating the constant in source: `DSR_P_BADGE_MIN = 0.95 → 0.90` moves the digest `gate-v2-f8eaf291fcec899f` → `gate-v2-719565584d50f51f`. Reverted by editing the value back.

### Mutations run this round — every one reverted by editing, never `git checkout`

| mutation | what went red |
|---|---|
| a bar literal back in the rewritten `rigor_reasons` prose | `test_no_dsr_bar_literal_outside_its_definition` — `rigor_reasons.py:55` |
| `unattributed` also requires a `not_computed` leg | `test_unattributed_when_nothing_on_record_falls_below_the_bar` (`assert False is True`) — and only that test |
| a CLEARED DSR bar reported as `not_computed` | the new all-`pass` assertion in that same test |
| drop `dsr_p_min` from the digest's `profiles` | `test_the_digest_moves_when_a_gate_threshold_moves` |
| drop `badge_bar` from the inputs **and** re-write level 1 as a literal | `test_the_digest_moves_when_the_badge_bar_constant_moves`, `test_the_inputs_name_every_documented_ingredient` — while the `is` assertion in `test_the_definition_is_singular_and_is_the_badge_bar` stayed green, which is the whole reason `badge_bar` exists |
| `DSR_P_BADGE_MIN = 0.95 → 0.90` in source | 8 tests in `test_single_dsr_bar.py`, and the digest moved |

Also caught while merging: an earlier `unattributed`-guard mutation (dropping the `not any_failed` term) is already covered by `test_graded_rejection_names_every_failing_check_with_its_threshold`, so the mutation above was narrowed until only the changed test failed.

### Tests on the merged head (`origin/main` @ bf76999c)

- `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` (repo root) — **6267 passed, 4 skipped, 2 deselected, 0 failed** (20:18 on a loaded box).
- `cd analytics-engine && python -m pytest -q` — **486 passed** (the PR touches its fixture-regen scripts).
- `cd ui && npm test` — **1040 passed, 0 failed**. Supersedes the 916 below, which predates main's new UI tests.
- `ruff format --check .` — 766 files already formatted. `ruff check backend analytics-engine` — **40 findings, the same set as `origin/main` at bf76999c** (diffed; only line numbers shift). Supersedes the "39" below, which was measured against the older main.


dbrowneup:
## Fix round — all five defects addressed

Thanks for the review; the blocker was real and the two majors were the issue's own headline symptom still shipping. Every defect below is fixed in one commit, `05a96871`.

| # | Sev | Defect | Fix |
|---|-----|--------|-----|
| 1 | blocker | `regen_buy_hold_fixture.py` imports through `archimedes.services`, so `engine-tests` collects 2 errors / 0 tests | leaf module loaded by path with `importlib`; the package `__init__` never runs. New `analytics-engine/tests/test_engine_suite_is_standalone.py` keeps it that way |
| 2 | major | `Architecture.jsx:484` still said "positive at **90%** one-sided confidence" | `95%`, plus `TestTheAppQuotesTheLiveBar` — a backend guard that reads the number out of the shipped copy and compares it to `DSR_P_BADGE_MIN` |
| 3 | major | two served openwiki pages taught the retired bar; two sentences asserted the *inverse* of the truth | those two sentences corrected in place; STALE-ON-ONE-NUMBER banners on `reading-a-backtest.md`, `measured-results.md` and `documented-conflicts.md`; `TestServedWikiPagesDoNotTeachARetiredBar` scans every served page |
| 4 | minor | `consolidation_candidate1_verify.py:425` — "the badge threshold is 0.90" in a live comment, unguarded | comment names the constant; `TestOneLiteral`'s module set is now **derived** by walking `backend/`, `analytics-engine/`, `scripts/` instead of a 9-file allowlist |
| 5 | minor | four fresh UI hand-copies with nothing binding them | same guard as #2 — it binds all five UI sites, and the always-on floor in the same sentences to `DSR_P_FLOOR` |

Also corrected the two inaccurate claims in the PR body you flagged ("19 pages … every public rigor page", and the Tests section that omitted `engine-tests`).

---

### 1 — blocker: the `engine-tests` gate

Reproduced your finding byte-identically first, in a venv built with exactly the CI pip line:

```
$ pip install --quiet ./analytics-engine "pytest>=8.2" pytest-socket
$ python -m pytest analytics-engine/tests -q --override-ini="pythonpath=" --disable-socket
analytics-engine/scripts/regen_buy_hold_fixture.py:31: in <module>
    from archimedes.services.rigor_profiles import DSR_P_BADGE_MIN
backend/archimedes/services/__init__.py:11: in <module>
    from archimedes.agents import generation_pipeline as generation_pipeline
backend/archimedes/agents/generation_pipeline.py:49: in <module>
    from archimedes.api.generate_schemas import GenerateBrief
backend/archimedes/api/generate_schemas.py:15: in <module>
    from pydantic import BaseModel, Field, field_validator
E   ModuleNotFoundError: No module named 'pydantic'
ERROR analytics-engine/tests/test_curated_num_trials.py
ERROR analytics-engine/tests/test_library_pbo.py
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
2 errors in 29.13s
```

Fixed as you suggested, with one addition your sketch would have hit next: **the module must be registered in `sys.modules` before `exec_module`.** `@dataclass` resolves `sys.modules[cls.__module__]` while building `RigorProfile`, so the straight `module_from_spec` → `exec_module` version fails with `AttributeError: 'NoneType' object has no attribute '__dict__'`. I hit it, so the loader registers first and pops on failure.

Same command, same venv, after the fix — **`482 passed`**, and `486 passed` with the new guard module:

```
$ python -m pytest analytics-engine/tests -q --override-ini="pythonpath=" --disable-socket
486 passed in 54.63s
```

**Guard shown red.** Reverting the loader to the package import:

```
E       AssertionError: importing the engine's fixture scripts in a clean interpreter failed. In CI this is a COLLECTION error, so the whole engine job reports zero tests:
E         ModuleNotFoundError: No module named 'pydantic'
FAILED analytics-engine/tests/test_engine_suite_is_standalone.py::test_the_scripts_the_suite_imports_pull_in_no_backend_package
FAILED analytics-engine/tests/test_engine_suite_is_standalone.py::test_the_badge_bar_is_read_from_the_backend_not_copied
FAILED analytics-engine/tests/test_engine_suite_is_standalone.py::test_the_constant_is_reachable_from_every_re_exporting_script[regen_buy_hold_fixture]
FAILED analytics-engine/tests/test_engine_suite_is_standalone.py::test_the_constant_is_reachable_from_every_re_exporting_script[regen_fixtures]
4 failed in 1.35s
```

The guard imports in a **fresh subprocess** deliberately — by the time it runs in-process, conftest and sibling modules have already imported half the world, so an in-process `sys.modules` check would pass on a tree that fails in CI.

### 2 — major: `Architecture.jsx`

Changed to `95%`, and bound rather than hand-copied a fifth time. **Shown red** by reverting the string:

```
E   AssertionError: The app states a badge bar that is not the live one. That is #1794 verbatim: the public rigor surface quoting a threshold the gate does not use. The bar has ONE definition — rigor_profiles.DSR_P_BADGE_MIN:
E       ui/src/components/Architecture.jsx:484: 'positive at 90% one-sided' states 90, live bar is 0.95
E   assert not ["ui/src/components/Architecture.jsx:484: 'positive at 90% one-sided' states 90, live bar is 0.95"]
```

### 5 — minor: the other four UI hand-copies

The same guard covers `Landing.jsx`, `RigorExplainer.jsx` (4 sites) and `RigorStrictnessControl.jsx`'s fallback ladder. It reads the number *out of* the copy with a small set of shape patterns rather than asserting a fixed string is present, so a reworded sentence keeps its binding and a new hand-copy in a recognised shape is checked automatically. Per-file minimum counts stop a rewrite from silently unbinding a page.

Mutating `DSR_P_BADGE_MIN = 0.95 → 0.90` — **all four files red**:

```
E   assert not ["ui/src/components/Architecture.jsx:484: 'positive at 95% one-sided' states 95, live bar is 0.9"]
E   assert not ["ui/src/components/Landing.jsx:29: 'evidence at the 95% one-sided' states 95, live bar is 0.9", "ui/src/components/Landing.jsx:50: 'evidence at the 95% one-sided' states 95, live bar is 0.9"]
E   assert not ["ui/src/components/RigorExplainer.jsx:49: 'confidence ≥ 0.95 (badge)' states 0.95, live bar is 0.9", ... "RigorExplainer.jsx:323: 'Badge (Conservative/level-1) thresholds: DSR conf ≥ 0.95' states 0.95, live bar is 0.9"]
E   assert not ['ui/src/components/RigorStrictnessControl.jsx:16: "level: 1, label: \'Conservative\', dsr_p_min: 0.95" states 0.95, live bar is 0.9']
```

That whole mutation is **26 failed, 54 passed** across the two guard modules.

The `0.50`s sitting in the same sentences are bound to `DSR_P_FLOOR` rather than exempted — an exemption is a hole; a second binding is not.

### 3 — major: the served openwiki pages

Your "better" suggestion, taken. Two notes on how it landed:

**The naive extension does not work.** `_RETIRED_BAR` is `p\s*(?:≥|>=)\s*0\.90?` and matches *neither* offending sentence — not `"— is it ≥ 0.90?"` (no `p` adjacent) nor `"the gate has since moved to 0.90"`. So the wiki scan is a separate, broader pattern.

**Bind it to the constant, not to the digits.** My first draft matched `0\.9\d?` and my own adversarial cases caught it flagging the **live** bar — i.e. it would have demanded a "this page is stale" banner on a page telling the truth, which is the `_RETIRED_BAR` inversion one class up. It now captures the value and drops anything equal to `DSR_P_BADGE_MIN`, with `test_the_wiki_scan_never_flags_the_live_bar` pinning that.

The rule: **a served page either states the live bar, or carries a STALE-ON-ONE-NUMBER banner that names it.** That keeps your provenance argument — a generated page stays as generated — while making the disclosure something a reader actually sees. The banner is bound to the constant too, so the next bar move re-aims all of them instead of stranding them.

Two sentences were the exception and are corrected in place, because a banner does not repair a claim that asserts the inverse of the truth: `reading-a-backtest.md:148` ("the retired 0.95 bar") and `measured-results.md:184` ("the gate has since moved to 0.90"). Each banner says which sentences were touched, so `generated:` stays honest.

Scope is every markdown page under `openwiki/`, not a hand-named subset — a two-directory list is the same shape of mistake as the nine-module allowlist.

**Shown red on the current tree**, with the banners stripped:

```
E   AssertionError: A publicly served OpenWiki page states a DSR bar that is not 0.95 and carries no STALE-ON-ONE-NUMBER banner. That is #1794's headline symptom — a public rigor page quoting a threshold the gate does not use. Either correct the page or add the banner:
E       openwiki/rigor/reading-a-backtest.md:128: 2. **DSR p-value** — is it ≥ 0.90? If not, the excess Sharpe is not positive at 90%
E       openwiki/rigor/reading-a-backtest.md:148: > it is the one definition of the DSR bar. The 0.90 bar in step 2 is the RETIRED one; the
```

**And on a brand-new page** dropped into a section the old scan never looked at:

```
E       openwiki/library/_mut_probe.md:3: The admission gate requires a DSR p-value of at least 0.90 to earn the badge.
```

**And a banner that stops naming the live bar:**

```
E   AssertionError: A STALE-ON-ONE-NUMBER banner does not name the live bar (**0.95**). The banner exists to tell a reader what the number really is; one that omits it — or names a bar that has since moved again — is worse than no banner at all: ['openwiki/rigor/admission-gate.md']
```

### 4 — minor: the stale comment, and the allowlist that let it through

The comment now names `DSR_P_BADGE_MIN` instead of quoting `0.90`. More to the point, I took the second half of your fix: `GUARDED_MODULES` is no longer the scope. The scan **walks** `backend/`, `analytics-engine/` and `scripts/` for files that mention DSR (56 candidates today), and the old nine-file list is kept purely as an anti-shrink floor — `test_the_guarded_floor_is_still_covered` fails if any of them drops out of the derived set, and `test_the_scan_is_wider_than_the_backend` fails if the walk stops reaching outside `backend/`.

**Shown red on line 425**, restoring the original comment:

```
E   AssertionError: A DSR-bar literal (0.9 / 0.90 / 0.95) appears next to DSR code outside rigor_profiles.DSR_P_BADGE_MIN. That is the #1794 defect: two bars in the tree, and which one graded you depended on which code path you reached. Import DSR_P_BADGE_MIN (or read profile.dsr_p_min); in a comment, name the constant instead of quoting its value:
E       analytics-engine/scripts/consolidation_candidate1_verify.py:425: # 0.90; min_passing_level moves from None to 5, i.e. it only clears the
E   assert not ['analytics-engine/scripts/consolidation_candidate1_verify.py:425: # 0.90; min_passing_level moves from None to 5, i.e. it only clears the']
```

You also predicted the singleton check would miss a second `DSR_P_BADGE_MIN = ` outside the nine files. It would have. **Shown red** with one planted in an analytics-engine script:

```
E   AssertionError: DSR_P_BADGE_MIN is given a literal value in more than one place — that is two bars, which is #1794. Import it instead. Found: ['analytics-engine/scripts/gen_daily_returns_store.py:201', 'backend/archimedes/services/rigor_profiles.py:63']
```

Two refinements the widening forced, both worth flagging:

- The singleton check now distinguishes a **literal** from a **re-export**. `DSR_P_BADGE_MIN = _load_badge_bar()` (the blocker fix) is not a second bar — it writes no number down — but a `startswith` test counted it. It now matches `^DSR_P_BADGE_MIN\s*=\s*[\d.]`. A value laundered through `float("0.90")` still trips the literal scan, which does not care what syntax the number is wearing.
- `SCAN_ROOTS` is `backend`, not `backend/archimedes`, because the singleton walk includes tests — a second literal parked in `backend/tests` was invisible otherwise. **Shown red:**

```
E   AssertionError: DSR_P_BADGE_MIN is given a literal value in more than one place — that is two bars, which is #1794. Import it instead. Found: ['backend/archimedes/services/rigor_profiles.py:63', 'backend/tests/test_dsr_parity.py:127']
```

Test files are excluded from the *literal* scan only (pinning a number is a test's job) and never from the singleton scan.

---

### Tests on this head

- **`engine-tests`, CI-identical venv** (`pip install ./analytics-engine "pytest>=8.2" pytest-socket`, then `python -m pytest analytics-engine/tests -q --override-ini="pythonpath=" --disable-socket`) — `**486 passed** (482 baseline + the 4 new guard tests)`. This is the gate that was red.
- **Full backend suite**, `python -m pytest -q -p no:randomly` at repo root — `**5818 passed, 6 skipped, 1 failed in 34:54**. The one failure is `test_oracle_health_telemetry.py::TestOracleHealthChainReadFailure::test_partial_chain_read_failure_still_reports_not_fresh`, and its own reason string says why: `oracle_health probe_timeout: exceeded 1.5s budget` — a wall-clock budget assertion on a loaded machine, the same class as the two `test_post_rollout_probe` flakes you hit. Re-run alone: **20 passed in 5.90s**. It touches no rigor code, and nothing in this diff reaches it.`
- **Targeted** — `test_single_dsr_bar.py` + `test_quant_docs_conflicts.py` + `test_dsr_parity.py` + `test_docs_site.py`: `**107 passed**`
- **UI**, `cd ui && npm ci && npm test` — `**tests 916 / pass 916 / fail 0**`
- **Ruff** — `ruff check backend analytics-engine` = `Found 39 errors.`, identical to the `origin/main` baseline you measured; `ruff format --check .` = `740 files already formatted`.

Every mutation above was reverted by editing the value back, never `git checkout --`, and each file confirmed byte-identical afterwards.

PR #1825, opened 2026-09-03
[payments] The offload entrypoint releases what the enqueue spent (#1793)

### LINKED ISSUE #1793: [payments][free-tier] scripts/run_generation_job.py calls run_generation directly and bypasses _run_with_cleanup — a failed run on that path releases neither the free slot nor the paid credit
Found by the independent verifier on PR #1785 (free-slot void on undelivered runs), out of that PR's scope:

`backend/archimedes/scripts/run_generation_job.py:196` invokes `run_generation(...)` directly. The `finally` that voids a consumed credit (`_release_credit_if_undelivered`, already on main) and, after #1785, the free slot (`_release_free_slot_if_undelivered`) lives in `generate_routes._run_with_cleanup`, which that script never enters. On that path every post-enqueue failure (thin corpus, crash, cancel, timeout) leaves the credit AND the slot spent.

## To settle first
Is this script on a live path (runner task, scheduled ECS job, operator use)? `graft callers run_generation` / the ECS task definitions will say. If it is dead, delete it (audit P1-1 style, with the grep evidence). If it is live, route it through `_run_with_cleanup` (or lift the release logic into one function both paths call) and add a test that a failing run on the script path releases both.

Related: #1785 (the web path fix), the paid-credit void policy in `generate_routes._paywall_with_credit`, #1748.

dbrowneup:
Closes #1793.

## To settle first — is the script live or dead?

The issue asks this before anything else, because the answer picks the fix. **Live, and adopted on the record — so it gets the cleanup, not the delete.**

| Evidence | Reading |
|---|---|
| `graft callers run_generation` | two production callers: `_run_with_cleanup` and `run_job` (`run_generation_job.py`). Everything else is a test. |
| `graft callers run_job` → `handler` → `main` | the module is self-contained: a Lambda handler and an `argparse` CLI. |
| `grep -rn run_generation_job` (whole repo) | `infra/spike-1411/Dockerfile.lambda` (COPYs it into the image), `infra/spike-1411/spike_probe.py` (imports `handler`), the ADR, the streaming spec, its test. **No serving-path importer.** |
| `grep -rln run_generation_job infra/ .github/` | no ECS task definition, no terraform resource, no workflow invokes it today. |
| `docs/adr/lambda-generation-offload.md` (status **Proposed — DEFER**) | *"**ADOPT two of its parts now**, both of which are inert until something calls them: 1. `run_generation_job.py` … the lane is a deployment choice, not a code fork."* |

So it is not audit-P1-1-shaped dead code — it is a committed entrypoint a merged ADR adopted, runnable by an operator today (`python -m archimedes.scripts.run_generation_job --job-id … --brief …`) and the exact artifact any future offload lane is built on. Deleting it would reverse an ADR; leaving it as-is ships the defect into whichever lane wins.

## The defect

`run_job` awaited `run_generation` bare. The `finally` that hands a spent entitlement back when a run delivers nothing lives inside `generate_routes._run_with_cleanup` — a coroutine this entrypoint never enters. On the script path every post-enqueue failure (thin corpus, crash, cancel, timeout) left the caller charged for a generation they never received.

Nothing crashed. That is what made it survivable: the job ended, the summary came back, and the only trace was a `consumed` ledger row for a run that produced nothing.

## The change

One seam, `generate_routes.release_entitlements_if_undelivered(job_id, store)`, that **both** run paths call and that holds **both** refunds — the issue's second option ("lift the release logic into one function both paths call"), not the first.

```
run_job                      _run_with_cleanup
   │                                │
   └────────────┬───────────────────┘
                ▼
   release_entitlements_if_undelivered(job_id, store)
                │
                ├── _release_credit_if_undelivered      (paid credit)
                └── _release_free_slot_if_undelivered   (free tier, #1785)
```

Each caller's `finally` is one line. Routing the script through `_run_with_cleanup` itself was considered and rejected on three counts, each of which would have been a regression: it resolves its store with `get_job_store()` (the worker must use the store it bound from the live environment — this module's whole reason to exist), it swallows exceptions in an `except Exception` safety net (the worker must surface an infrastructure failure to its invoker as a failed invocation), and it has no `dual_regime` parameter.

`finally`, not `except`. The shape that made this invisible is the insufficient-corpus branch, which writes an `error` status and **returns** rather than raising — a handler keyed on exceptions would sail straight past the very failure that surfaced the bug. Both refunds are decided from the job's terminal status, the free one also from the event log, so a delivered run keeps paying and a run that persisted a strategy keeps its slot spent.

The store is passed in rather than resolved inside. `get_job_store()`'s singleton reads a `REDIS_URL` frozen at import time, which for a worker that loads its secrets from SSM is the localhost default — a refund decided from an empty localhost store would read every delivered run as undelivered.

## #1785 landed while this PR was open

**An earlier revision of this body said "#1785 is NOT on main, and this PR does not pretend otherwise". That is no longer true and the claim is withdrawn.** PR #1785 merged as `3b900870` at 2026-09-03T22:54:43Z, 89 minutes after this PR was opened, so every check that was green on the original head predates it. It put `_release_free_slot_if_undelivered` into `_run_with_cleanup`'s `finally` — exactly the place #1793 says a refund must not live — and the branch went `CONFLICTING`.

`origin/main` is merged in. The conflict was two hunks in `generate_routes.py`, and resolving them the obvious way (keep both sides) is **wrong**; this PR's own tripwire is what says so:

```
FAILED TestBothRunPathsReleaseTheSameThings::test_every_release_helper_is_reached_through_the_shared_seam
E   AssertionError: a release helper is not called by release_entitlements_if_undelivered — adding
    it to a caller's finally gives it to that path only, which is issue #1793
E     Right contains one more item: '_release_free_slot_if_undelivered'
```

So the resolution moves the free-slot call **into** the seam. Before that move, a raising `run_generation` on the script path called `restore_for_job` once and `release_for_job` **zero** times. `Closes #1793` — the issue's "releases both" — is true against current `main` only with this merge in.

## Adversarial pass — every guard shown red

`pytest backend/tests/test_run_generation_job.py` — **21 passed** on the branch — against eight mutations, each reverted by editing back, never by `git checkout`.

| # | Mutation | Result |
|---|---|---|
| 1 | the release made unreachable (the pre-fix shape) | 6 failed, 15 passed |
| 2 | free-slot call dropped from the seam | 4 failed, 17 passed |
| 3 | serving path calls the helpers directly, around the seam | 1 failed, 20 passed |
| 4 | `except Exception` instead of `finally` | 6 failed, 15 passed |
| A | the naive merge resolution — free slot left in the caller's `finally` | 4 failed, 17 passed |
| B | `list_events` removed from the store double, production code untouched | 3 failed, 18 passed |
| C | `run_job` awaits `_release_credit_if_undelivered` directly instead of the seam | 4 failed, 17 passed |
| D | an unwired, off-convention `_void_free_slot_when_undelivered` added to the module | 1 failed, 20 passed |

Three of these are worth naming.

**A is not hypothetical** — it is what `git merge origin/main` produces if you keep both sides, and the shape `main` is in today:
```
FAILED …test_a_failed_run_hands_back_both_the_credit_and_the_free_slot - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_silent_corpus_failure_shape_also_hands_the_credit_back - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_refund_reads_the_bound_store_not_the_import_time_singleton - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_every_release_helper_is_reached_through_the_shared_seam - a release helper is not called by release_entitlements_if_undelivered
```

**C was fully green (18 passed) before this round.** Nothing pinned `run_job` to the seam — only the serving path was pinned — so the offload entrypoint could drift back to a single helper and silently stop handing the free slot back, with #1793 re-opened on its own path and a green suite. `test_the_script_path_releases_through_the_same_seam` closes it, and asserts the awaited argument is the store the worker **bound**, not `get_job_store()`.

**D passed 21/21 undetected** under the old `_release_\w+_if_undelivered` pattern. Widened to `_(release|void|refund|restore)_\w+_(if|when)_undelivered`, the same mutant fails with `Right contains one more item: '_void_free_slot_when_undelivered'`.

**B is the vacuity check.** `_BoundStore` had no `list_events`, so `_job_persisted_a_strategy` answered "cannot tell" and `_release_free_slot_if_undelivered` returned without releasing — every free-slot assertion would have passed while proving nothing.

The stand-in for the pipeline is deliberately not only a `raise`. `_corpus_failure` pushes the `GENERATION_UNAVAILABLE` event and writes the `error` status **and then returns** — the shape #1785 identified as the reason this class of bug is invisible. `_persist_then_die` pushes a `persisted` event before dying, so the free tier's "delivered a library row" branch is exercised on this path too.

## Verification

```
pytest backend/tests/test_run_generation_job.py                  → 21 passed
  + test_free_generation_gate + test_generation_credits
  + test_generation_timeout + test_generation_concurrency
  + test_api_docs_drift                                          → 107 passed
ruff format --check .                                            → 764 files already formatted
ruff check --select E9,F63,F7,F40,F82 .                          → All checks passed
ruff check <3 touched python files>                              → All checks passed
pytest -m "not integration" --tb=short -q -p no:cacheprovider    → 6218 passed, 4 skipped, 2 deselected in 1124s (exit 0)
```

No `ui/` file touched, so no npm run. No infra/terraform change, so no `fmt`/`validate`/`plan`.

**Pre-existing failures: none, this time.** The full run above is a single process (xdist is not installed) and reports **zero** `FAILED` lines. The earlier revision of this body listed two — `test_generate_job_scoping.py`'s identity-ledger pair and `services/test_generation_pipeline.py`'s — from a run split into six file-chunks. They are a test-isolation artifact of that chunking, not a defect in this PR or on `main`: run in one process, in the CI command's own order, they pass.

## Collision check

- **#1785** — merged to `main`; merged into this branch and resolved as above, with the tripwire's red output as the record of why the naive resolution is wrong.
- **#1813** / **#1835** — both edit `generate_routes.py` only inside `start_generation` (~:422–:445). This PR's edits are the new seam and one line in `_run_with_cleanup`'s `finally`. No overlap.
- No other open PR touches `run_generation_job.py` or `docs/adr/lambda-generation-offload.md`.

## Left out, deliberately

- **A run whose process dies never runs its `finally`.** A container roll mid-run leaves the job `running` and no cleanup executes, on either path. A sweeper over stalled jobs would close it; out of scope, unchanged by this PR, and #1785 has the same gap.
- **`generate_routes` is now imported by the worker at call time.** Deferred inside `run_job` with the other heavy imports, and harmless in the Lambda image (`Dockerfile.lambda` derives from the production backend image, so the module is already present). Flagged rather than buried: it is a new edge from `scripts/` to `api/`. A `services/generation_entitlements.py` would invert it, at the cost of moving both helpers out from under the tests that patch them by that path — a bigger, separable refactor.
- **The tripwire enforces a naming convention, not reachability.** A refund helper named outside `_(release|void|refund|restore)_<thing>_(if|when)_undelivered` is invisible to it. Said plainly in the seam's docstring rather than papered over; closing it properly needs an AST or call-graph check, which is a different tool than this test file should carry.


dbrowneup:
## Fix round — every defect addressed, with the red proof for each

| # | Sev | Defect | Commit | Red proof |
|---|---|---|---|---|
| 1 | blocker | `Closes #1793` false + PR `CONFLICTING`: #1785 merged as `3b900870` 89 min after this PR opened, putting `_release_free_slot_if_undelivered` into `_run_with_cleanup`'s `finally`; the script path released only the credit | `e76a9f7a` | mutation **A** |
| 2 | major | `_BoundStore` had no `list_events`, so `_job_persisted_a_strategy` answered "cannot tell" and every free-slot assertion would have passed vacuously | `9acd923d` | mutation **B** |
| 3 | major | Nothing pinned `run_job` to the seam — only the serving path was pinned | `9acd923d` | mutation **C** |
| 4 | minor | Discovery regex bound to `_release_*_if_undelivered`; an off-convention refund helper escaped in silence | `18b61480` | mutation **D** |
| 5 | minor | ADR bullet, module docstring, `run_job` docstring and the deferred-import comment all overstated against post-#1785 main | `7ca4eeb6`, `f9473fb9` | n/a (prose) |

### 1 — the blocker

`origin/main` is merged in. The two `generate_routes.py` conflict hunks are **not** resolved by keeping both sides: `await _release_free_slot_if_undelivered(job_id, store)` moves *into* `release_entitlements_if_undelivered`, so each caller's `finally` is a single call to the seam. The PR's own tripwire is what says the naive resolution is wrong, which is the strongest evidence I can offer that the guard was worth building.

**Mutation A** — the naive resolution, i.e. the shape `main` is in today:

```
FAILED …test_a_failed_run_hands_back_both_the_credit_and_the_free_slot - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_silent_corpus_failure_shape_also_hands_the_credit_back - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_refund_reads_the_bound_store_not_the_import_time_singleton - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_every_release_helper_is_reached_through_the_shared_seam - a release helper is not called by release_entitlements_if_undelivered
4 failed, 17 passed
```

The body's section "#1785 is NOT on main, and this PR does not pretend otherwise" is **withdrawn**; the body has been rewritten around what is now true.

### 2 — the vacuity check

`list_events` added to `_BoundStore`. **Mutation B** removes it again with production code untouched — if the double cannot see the release, the free-slot assertions prove nothing:

```
FAILED …test_a_failed_run_hands_back_both_the_credit_and_the_free_slot - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_silent_corpus_failure_shape_also_hands_the_credit_back - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_refund_reads_the_bound_store_not_the_import_time_singleton - Expected 'release_for_job' to be called once. Called 0 times.
3 failed, 18 passed
```

Two tests carry the issue's acceptance directly: `test_a_failed_run_hands_back_both_the_credit_and_the_free_slot`, and `test_a_run_that_persisted_a_strategy_keeps_the_free_slot_spent` for the free tier's stricter negative direction — a run that bought a library row keeps its slot spent. Every script-path test now patches both ledgers: `free_generations.release_for_job` opens a DB session, and this module claims no Redis, no DB, no network.

### 3 — the missing mirror

`test_the_script_path_releases_through_the_same_seam` added, asserting the seam is awaited once **with the store the worker bound**, not `get_job_store()`. **Mutation C** is the drift the review found — `run_job`'s deferred import pointed at `_release_credit_if_undelivered` and that awaited instead. It was **fully green (18 passed)** on the old head:

```
FAILED …test_a_failed_run_hands_back_both_the_credit_and_the_free_slot - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_silent_corpus_failure_shape_also_hands_the_credit_back - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_refund_reads_the_bound_store_not_the_import_time_singleton - Expected 'release_for_job' to be called once. Called 0 times.
FAILED …test_the_script_path_releases_through_the_same_seam - Expected mock to have been awaited once. Awaited 0 times.
4 failed, 17 passed
```

### 4 — the convention, widened, and the limit stated where it is read

Regex widened to `_(release|void|refund|restore)_\w+_(if|when)_undelivered`. **Mutation D** — an unwired `async def _void_free_slot_when_undelivered` added to `generate_routes` — **passed 21/21, undetected** before `18b61480`. Now:

```
FAILED …test_every_release_helper_is_reached_through_the_shared_seam
E   AssertionError: a release helper is not called by release_entitlements_if_undelivered …
      Right contains one more item: '_void_free_slot_when_undelivered'
1 failed, 20 passed
```

What the tripwire still cannot do is now written into the seam's docstring, not only the test's: the discovery matches a naming convention, not reachability. The test class already claimed "That limit is written into `release_entitlements_if_undelivered`'s docstring too" — before `18b61480` it was not, so the cross-reference pointed at nothing.

### 5 — the prose, against post-#1785 reality

The ADR correction bullet still read "has already spent a credit at enqueue time" and described the tripwire as discovering `_release_*_if_undelivered` — both stale. The module docstring said a failed run "hands back what the enqueue spent", singular. `run_job`'s docstring said "The refund is evaluated from the job's terminal status" without the free helper's second input, the event log. And the deferred-import comment called a `from … import` a module-attribute read: the patchability conclusion was right, the mechanism was not — what makes patching work is that the import statement runs on every call. All four corrected, and the serving path's `finally` now carries the instruction at the place the mistake actually gets made.

### The four original mutations, re-proved on the merged tree

Every green check on the old head predates #1785, so none of them carried over for free.

| Mutation | Result |
|---|---|
| 1 — the release made unreachable (the pre-fix shape) | 6 failed, 15 passed |
| 2 — free-slot call dropped from the seam | 4 failed, 17 passed |
| 3 — serving path calls the helpers directly, around the seam | 1 failed, 20 passed |
| 4 — `except Exception` instead of `finally` | 6 failed, 15 passed |

Mutation 4 goes red in both directions at once — `DID NOT RAISE` on the raising cases (a worker that absorbs an infrastructure failure reports a successful invocation) and `restore_for_job … Called 0 times` on the silent-corpus case.

### Verification

```
pytest backend/tests/test_run_generation_job.py                  → 21 passed
  + test_free_generation_gate + test_generation_credits
  + test_generation_timeout + test_generation_concurrency
  + test_api_docs_drift                                          → 107 passed
ruff format --check .                                            → 764 files already formatted
ruff check --select E9,F63,F7,F40,F82 .                          → All checks passed
ruff check <3 touched python files>                              → All checks passed
pytest -m "not integration" --tb=short -q -p no:cacheprovider    → 6218 passed, 4 skipped, 2 deselected in 1124s (exit 0)
```

No `ui/` file and no terraform touched, so npm and `fmt`/`validate`/`plan` do not apply.

Pushed head `26a59829` (that last commit is only the pre-push merge of `origin/main` @ `d784733f`, whose delta is `auth/` and does not enter the Python suite).

PR #1783, opened 2026-09-02
[product-truth] Passport card fields come from the validated DSL; honest paper header; raw-vs-deflated sentence

dbrowneup:
Closes #1769.

## The owner's evidence

Dogfood 2026-09-01 (iPhone), strategy `8411f2d044aeacc6` ("Volatility-Relief Swing"), read back through `GET /api/strategies/{id}` with the owner's session:

| field | methodology card (before) | validated `strategy_spec` | served now |
|---|---|---|---|
| `rebalance_frequency` | **weekly** | **monthly** | **monthly** |
| `position_sizing` | **equal_weight** | **full_invested_when_in_market** | **full_invested_when_in_market** |
| `asset_universe` | QQQ, SHV | QQQ, SHV | QQQ, SHV (from the spec) |
| papers header | "FUSED FROM 5 PAPERS" | 5 refs, **0** `contribution` | "5 papers cited · 0 name a mechanism this strategy trades" |
| Sharpe 0.70 vs DSR 0.08 | no sentence connecting them | — | one reconciliation sentence under Backtest |

The backtest ran the DSL. The card did not come from it and was never checked against it.

## Root cause — the issue's diagnosis was wrong

#1769 reads the symptom as **proposal prose leaking into the card**. That is not what happened, and a fix built on that reading would not have worked: there is no code path anywhere that copies proposal prose into `strategy_passports`. Credit to the builder for refusing the stated diagnosis and going to the writers instead — the real cause is narrower, worse, and in a different file:

* `generation_pipeline`'s **three** `StrategyPassport(...)` writers passed `asset_universe=c.asset_universe or []` and passed **neither `rebalance_frequency` nor `position_sizing` at all**. So every generated `strategy_passports` row took its *column defaults*, `weekly` and `equal_weight`, whatever its spec said. Nothing anywhere had ever written a spec-derived cadence. The card was not showing the prose — it was showing a constant.
* `passport_loader._update_record` had **no writer for those two columns either** (the create branch wrote them, the update branch did not). So a corrected value could not survive the post-backtest `force_update` rebuild, and a pre-fix row could never be repaired.
* `PositionSizing` was missing `full_invested_when_in_market` and `volatility_target` — two of the DSL's four sizing types. `StrategyPassportRecord.to_strategy_passport` coerces the stored column through that enum, so writing the true value would have raised `ValueError` on read. That is the mechanical reason nobody had ever written it.

[...]

PR #1826, opened 2026-09-03
Close the 2026-09-03 detection gaps; force a paging destination (#1818 P5)

dbrowneup:
Part of #1818 (P5).

Do not merge — owner vets first. Owner applies (targeted; recipe below).

> ### ⚠️ Corrections after merge — see #1838
>
> A post-merge review round re-derived this PR's numbers from the account and found **the ELB-5xx rationale below is backwards**. On 2026-09-03 the ALB generated **no** 5xx at all (every `HTTPCode_ELB_*` metric returns zero datapoints for the day); the "2/1/1 504s" were four **target**-side 5xx at 13:31–13:33Z, and they are what fired `alb_5xx_rate_high` at 13:39:16Z. `alb_5xx_high` stayed OK because its threshold is 10, not for want of signal. The alarm itself is correct and unchanged; only its justification was wrong, and it is re-stated as general ELB-side cover.
>
> Also corrected: the unhealthy-hosts retune is worth **~3 minutes**, not ~6; `9m47s` here should read **9m46s** (as the code always did); and the platform-admin wallet is truncated below to match every tracked file.
>
> Struck text below is left in place rather than silently rewritten. **#1838 carries the fixes.**

## Read this first: P5's premise is wrong, and it changes the fix

P5 says *"there was no alarm, the owner found out by using the site."* The first half is not what happened. From the account (CloudWatch alarm history + `AWS/SNS` metrics on `archimedes-alerts`, all times UTC):

| time | event |
|---|---|
| 13:29 | outage begins — both targets unhealthy |
| **13:38:46** | `archimedes-alb-unhealthy-hosts` **OK → ALARM** |
| **13:39:16** | `archimedes-alb-5xx-rate-high` **OK → ALARM** (flaps 4× to 13:45) |
| 13:35–13:50 | SNS `NumberOfNotificationsDelivered` = **5**, `NumberOfNotificationsFailed` = **0** |
| 15:03 | outage ends (OOM kill) |
| 15:06:46 | unhealthy-hosts **ALARM → OK** (1 more delivered) |

**Two alarms fired. Six emails were delivered, zero failures, to a confirmed subscriber** — the first 9m46s into a 94-minute outage and 84 minutes before it ended. The owner still found it by loading the site.

Re-derive it:

```bash
aws cloudwatch describe-alarm-history --alarm-name archimedes-alb-unhealthy-hosts \
  --history-item-type StateUpdate --start-date 2026-09-03T00:00:00Z \
  --end-date 2026-09-04T00:00:00Z --query 'AlarmHistoryItems[].[Timestamp,HistorySummary]' --output text

aws cloudwatch get-metric-statistics --namespace AWS/SNS \
  --metric-name NumberOfNotificationsDelivered \
  --dimensions Name=TopicName,Value=archimedes-alerts \
  --start-time 2026-09-03T13:00:00Z --end-time 2026-09-03T16:00:00Z --period 300 --statistics Sum
```

I had already written and pushed the obvious PR — "the topic has no subscriber, add one" — before checking. It was wrong, and it would have shipped a fix for a defect that does not exist while leaving the real one untouched. This is the rewrite.

So P5 is three gaps, and Terraform owns two.

## Gap 1 — detection shape (four new alarms)

Nothing watched what was already abnormal ten hours earlier. The wedge was a DDL lock chain, not saturation, so it presented as numbers the existing alarms are shaped wrong for.

| Alarm | Fires when | Why this number |
|---|---|---|
| `aurora_connections_wedge` | `DatabaseConnections` > 30 for 15 min | Connections went 16 → 33 and sat **flat** at 33 for 11.5 h at 4% CPU — sessions parked on a lock, not load. 30 sits just under the observed wedge and well above the healthy 16. The existing alarm fires at 80 and never fired. |
| `ecs_service_memory_high` | `MemoryUtilization` **max** > 85% for 5 min | Wedged tasks ramped 39% → 100% over 90 min. `Maximum`, not `Average`: two fresh replacements sat near idle and would have averaged the ramp to ~50% at the moment the OOM killer fired. There was no ECS alarm at all before this. |
| `alb_elb_5xx_count` | `HTTPCode_ELB_5XX_Count >= 5` in 5 min | The **balancer's own** 5xx, which nothing watched before. ~~`alb_5xx_high` never fired because the ALB synthesised 504s~~ — **corrected, see the note below**: the ELB-side metric published *no datapoints at all* on 2026-09-03. General cover for a live signal (1,254 ELB-side 5xx on UTC day 2026-08-20), not an #1818 detector. No `TargetGroup` dimension: that metric has none. |
| `ecs_service_cpu_high` | `CPUUtilization` avg > 90% for 10 min | `Average`, because the service's target-tracking policy tracks average against a 60% target — this means "the autoscaler is out of room". `Maximum` would page on one task draining a generation queue. |

**Replayed against the day — three of five would have fired, two would not:**

- `aurora_connections_wedge` at **~03:48** — nearly ten hours before the 13:38:46 alarm that actually fired, while the fleet was still serving. **This is the only reason the set is worth applying.**
- `alb_unhealthy_hosts` at **~13:32** (see gap 2).
- `ecs_service_memory_high` at **~14:39** — 85% crossed ~68 min into the 13:31–15:01 ramp. A real detector, 70 minutes into a 94-minute outage. Not early warning, and not sold as such.
- `alb_elb_5xx_count` — **would not have fired.** ~~The ALB logged 2/1/1 504s per minute, four in the window, under a threshold of five.~~ **Corrected:** the ELB-side metric had *zero* datapoints all day. The 2/1/1 at 13:31–13:33Z was `HTTPCode_Target_5XX_Count`, and it is what put `alb_5xx_rate_high` into ALARM at 13:39:16Z. See the note below.
- `ecs_service_cpu_high` — **would not have fired.** CPU was 3%.

The last two are general saturation cover. That split is written into the file, the runbook and a test, so "five alarms" is never read as five nets under this failure.

## Gap 2 — detection latency

`alb_unhealthy_hosts` retuned from 5-of-5 minutes to 2-of-2. ~~worth **about six minutes**~~ — **corrected: ~3 minutes.** `UnHealthyHostCount` first read 2 at 13:31Z and the alarm transitioned at 13:38:46Z, so 2m46s of that 9m46s is evaluation lag the retune does not remove. Dropping three of five one-minute datapoints buys ~3 min — useful, and small.

**Known cost, stated rather than discovered.** The 2026-08-21 flapping that forced the 5-minute window was real, and `deployment_minimum_healthy_percent = 100` means every rollout registers new targets while old ones drain. Expect transient fires on deploys, into a mailbox that already has an attention problem. Both numbers are now in the file so the trade can be re-argued with evidence; the revert is `datapoints_to_alarm`.

## Gap 3 — the page did not reach the owner

**No Terraform change fixes this.** What the code can do is refuse to leave the destination unchosen.

`var.owner_alert_email` has **no default** and its subscription has no `count`, so `terraform plan` errors until a destination is named. A `validation` rejects `""`:

```
$ terraform plan -var 'owner_alert_email='
│ Error: Invalid value for variable
│   var.owner_alert_email is ""
│ owner_alert_email must be a real email address, and one you will actually see
│ — issue #1818 P5's measured finding is that six alarm emails were delivered
│ on 2026-09-03 and still did not reach the owner. Set it in
│ infra/terraform.tfvars.

$ terraform plan            # no value at all
│ Error: No value for required variable
│ The root module input variable "owner_alert_email" is not set, and has no default value.
```

A `lifecycle { precondition }` also refuses the plan if it equals `alarm_email` — the address that already received six ignored emails:

```
$ terraform plan -var 'owner_alert_email=ops@example.com' -var 'alarm_email=ops@example.com'
│ Error: Resource precondition failed
│   condition = var.owner_alert_email != var.alarm_email
│   ├── var.alarm_email is "ops@example.com"
│   └── var.owner_alert_email is "ops@example.com"
│ owner_alert_email must differ from alarm_email. Both would resolve to ONE SNS
│ subscription ARN (SNS keys by topic+protocol+endpoint) and a later apply
│ dropping either resource would unsubscribe the other, leaving the topic with
│ no subscriber while state says otherwise. …
```

That is a precondition rather than a `count` stand-down on purpose: standing `alerts_email` down would have **destroyed the live, confirmed, delivering subscription**, and because both resources would hold one ARN, the `Unsubscribe` could leave zero subscribers with state claiming one — #1818's own failure mode manufactured by its fix. Failing the plan has no such path.

**Choosing the channel is the owner's call and is the open half of P5.** A mailbox that is actually watched, an SMS gateway, a PagerDuty/Opsgenie intake address. Nothing here decides it.

The live `alerts_email` subscription is deliberately untouched. Its `count` gate *is* a real landmine — no tfvars holds the applied address, so a bare apply unsubscribes the only destination this topic has ever delivered to — but that predates P5 and closing it means capturing the address, which is the owner's to supply. The README row for it is corrected from "did not confirm whether a real value is applied" to **confirmed applied and working**.

## A live foot-gun this turned up

The ECS alarms reach the cluster and service through `data` sources, not resource references. Measured, not stylistic: with `aws_ecs_service.backend.name` in `dimensions`, the alarm depends on `aws_ecs_task_definition.backend`, and `terraform plan -target=` on either alarm proposed:

```
  # aws_ecs_task_definition.backend must be replaced
      ~ PAPER_ADVANCE_ENABLED   "false" -> "true"
      ~ PLATFORM_ADMIN_WALLETS  "0x2a29…5105" -> ""
  Plan: 6 to add, 1 to change, 1 to destroy.
```

Applying an **observability** alarm would have re-armed the paper-advance loop that caused this outage (prod is pinned `false` — the #1778 / #1818 pull-back — while `ecs.tf` still says `true`) and blanked the live admin-wallet list. Data sources are reads and carry no such edge, while still failing the plan loudly on a rename (the precedent is `ecs.tf`'s own `data "aws_lb_target_group" "backend"`). A test fails if the alarms drift back onto that chain. **The `ecs.tf` drift itself is not fixed here** — see "Left out".

## Targeted plan (read-only; owner applies)

`AWS_PROFILE=ArchimedesDanAdmin`, plan only, never apply. `owner_alert_email` passed as `owner@example.com` for the plan; the real address goes in `terraform.tfvars`.

```
  # aws_cloudwatch_metric_alarm.alb_elb_5xx_count will be created
  # aws_cloudwatch_metric_alarm.alb_unhealthy_hosts will be updated in-place
  # aws_cloudwatch_metric_alarm.aurora_connections_wedge will be created
  # aws_cloudwatch_metric_alarm.ecs_service_cpu_high will be created
  # aws_cloudwatch_metric_alarm.ecs_service_memory_high will be created
  # aws_sns_topic_subscription.owner_alerts_email will be created

Plan: 5 to add, 1 to change, 0 to destroy.
```

`aws_sns_topic_subscription.alerts_email[0]` does not appear — the working subscription is not touched. The one in-place change:

```
  ~ resource "aws_cloudwatch_metric_alarm" "alb_unhealthy_hosts" {
      ~ alarm_description   = "One or more backend targets are unhealthy." -> "One or more backend targets are unhealthy for 2 min (#1818 P5)."
      ~ comparison_operator = "GreaterThanThreshold" -> "GreaterThanOrEqualToThreshold"
      ~ datapoints_to_alarm = 5 -> 2
      ~ evaluation_periods  = 5 -> 2
      ~ threshold           = 0 -> 1
```

Resolved dimensions: `LoadBalancer = app/archimedes-alb/955aeff03e643d11`, `ClusterName = archimedes-cluster`, `ServiceName = archimedes-backend`, `DBClusterIdentifier = archimedes-aurora`. The apply recipe and expected plan line are in `infra/runbooks/disaster-recovery.md`.

**After applying, click the confirmation link AWS emails.** Until then the subscription is `PendingConfirmation` and pages nobody, and that is indistinguishable from working on Terraform's side.

## Guards, each shown red on the input it rejects

`backend/tests/test_outage_alarms_1818.py` — 41 tests, hermetic (tracked files only; no AWS, no terraform binary). Fifteen mutations were applied, run, captured and reverted by editing:

| Mutation | Test that went red |
|---|---|
| `default = ""` on `owner_alert_email` | `test_owner_alert_email_has_no_default` |
| `count`-gate the owner subscription | `test_the_owner_subscription_exists_and_is_unconditional` |
| Weaken the precondition so the addresses may match | `test_the_owner_subscription_refuses_to_duplicate_the_working_one` |
| Widen `alerts_email`'s `count` (a new destroy case on the live subscription) | `test_the_working_subscription_is_left_alone` |
| Loosen the validation regex so `""` passes | `test_the_validation_rejects_an_unusable_destination` |
| Real address into the tracked `tfvars.example` | `test_no_personal_address_is_committed` |
| `alb_unhealthy_hosts` back to 5-of-5 | `test_the_window_is_the_duration_the_issue_asks_for[alb_unhealthy_hosts]` |
| Aurora threshold 30 → 50 | `test_metric_coordinates_match_the_issue` + `test_the_connections_alarm_sits_between_healthy_and_wedged` |
| `TargetGroup` dimension on the ELB-5xx alarm | `test_the_elb_5xx_alarm_watches_the_balancer_not_the_target` |
| `alarm_actions = []` on the memory alarm | `test_it_pages_and_is_tagged[ecs_service_memory_high]` |
| M-of-N damping (`10 of 12`) on the CPU alarm | `test_the_breaching_datapoints_are_consecutive[ecs_service_cpu_high]` |
| Direct ECS resource refs in `dimensions` | `test_the_ecs_alarms_stay_off_the_task_definitions_dependency_chain` |
| Hardcode the service name in the data source | `test_the_ecs_data_sources_fail_loudly_on_a_rename` |
| Drop `owner_alert_email` from `tfvars.example` | `test_apply_sh_will_block_on_an_unset_value` |
| Strip `owner_alert_email` from the DR runbook | `test_the_operator_docs_name_it[disaster-recovery.md]` |

Plus the four `terraform plan` results quoted above (empty value, missing value, colliding addresses, valid value), which exercise the validation and precondition themselves rather than their spelling.

## Tests

- `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` (repo root): **5843 passed, 4 skipped, 2 deselected, 0 failed** (35m06s, commit `8976626f`). No pre-existing failures to attribute.
- `backend/tests/test_outage_alarms_1818.py` + `test_cloudwatch_alarms.py`: 93 passed
- `cd ui && npm ci && npm test`: **945 passed, 0 failed** (no UI files changed by this PR; the count rose because #1783 merged into the branch)
- CI on `8976626f`: **Backend — unit tests pass**, Frontend + Better Auth pass, Analytics engine pass, all three `Infra — terraform fmt + validate` roots pass, Ruff gate pass
- `terraform fmt -check -recursive .` clean; `terraform init -backend=false && terraform validate` → *Success! The configuration is valid.*
- `ruff format --check .` clean; `ruff check --select E9,F63,F7,F40,F82 .` clean

## Left out, and why

- **Gap 3's actual answer.** Which channel the owner will genuinely notice is a decision, not a config value. This PR forces the decision and blocks the obvious wrong answer (the same mailbox again); it does not make it. That is the open half of P5.
- **P1–P4 of #1818.** Separate PRs, as the issue specifies. P5 is ops only: it makes the next wedge visible ten hours earlier, it does not prevent one.
- **The alarm drill.** Its assertion is now "I noticed it", which needs the owner and a deliberate prod disturbance. The checklist names the step; it is not run here.
- **`alarm_email`'s missing tfvars value.** Confirmed applied and working, but the address is Dan's and belongs in the gitignored file, not in this PR. Until it is captured, a non-targeted apply still unsubscribes it — which is why the apply recipe is `-target`ed.
- **`ecs.tf`'s `PAPER_ADVANCE_ENABLED = "true"` drift.** Surfaced and written into the runbook as an apply hazard, but flipping it is a decision about the #1778 tourniquet, not an observability change.
- **The linear 39% → 100% memory ramp** (#1818 open question). The alarm detects it; nothing here explains it.
- **Slack/PagerDuty as SNS protocols.** The issue asks for an email subscription. Another protocol on the same topic is a later change — and, given gap 3, probably the right one.


dbrowneup:
**Correction to the first push.** The original version of this PR (commit `d87cbfaf`) was built on P5's stated premise — "the SNS topic has no subscriber, every alarm has been changing colour in a console nobody watches" — and its whole framing followed from that. I checked the account afterwards and it is false.

The topic has one confirmed email subscription. It delivered six notifications on 2026-09-03 with zero failures, starting 9m47s into the outage.

Had that shipped, it would have added a subscription to a topic that already had one, declared the root cause closed, and left the actual gaps — nothing watching the 03:33 connections wedge, and a delivered page that nobody acted on — untouched. Force-pushed as `e587eb57` with the measured record instead. Flagging it here rather than quietly rewriting, because "the fix matched the issue's description of the bug" is exactly the kind of thing that reads fine in review.

PR #1830, opened 2026-09-03
[backend][infra] Readiness: /health/ready goes 503 when the DB probes have been stale too long (#1818 P3)

dbrowneup:
Part of #1818 (P3).

## The gap this closes

From **03:32Z to 13:28Z on 2026-09-03** `/health` answered **200 in ~1.05 s** on both tasks while neither could read its database. It was not lying: `/health` reports what it knows (#1592/#1594), and what it knew was a set of `stale_cached` readings, each honestly labelled, whose `cached_age` reached **35,797 s**. Liveness was true the whole time — the loop was turning, the endpoint answered. Readiness was false and was **represented nowhere**, so ECS and the ALB stayed green over a wedged fleet for ten hours, and the wedge broke only when a task was OOM-killed at 15:02Z.

This PR adds the missing representation and wires it to the one consumer that can act on it proportionately.

## The rule

`GET /health/ready` (also `/api/health/ready`) runs exactly the five DB-backed `/health` probes — `corpus`, `corpus_db`, `corpus_meta`, `paper_rag`, `risk_data` — through the **same** `HealthProbeCache`, the same 0.8 s budget and the same `absent` values, and answers:

| condition | status | body |
|---|---|---|
| any probe `stale_cached` with `age_s > HEALTH_STALE_UNREADY_S` (default **900**) | **503** | `ready: false`, `stale_probes: [{probe, age_s}]`, per-probe `state`/`age_s`/`reason` |
| stale but within the threshold | 200 | `ready: true`, every probe still labelled `stale_cached` with its age |
| fresh | 200 | `ready: true`, every probe `live` |

`/health` itself is **unchanged in status** and now publishes the same verdict as data: `ready`, `stale_unready_probes`, `stale_unready_threshold_s`.

## Why a second endpoint, and why the ALB check does not move

Failing the **ALB** check pulls a target out of rotation immediately, and a shared cause pulls **all** of them at once — that is literally the 13:29Z line of the incident (`HealthyHostCount=0`, 504s to the only user of the day). `infra/alb.tf` is therefore untouched: path `/health`, `matcher = "200"`, and a guard test pins that it stayed that way.

Failing the **container** check hands the task to the ECS scheduler, which is bound by `deployment_minimum_healthy_percent = 100` — it brings a replacement up and healthy *before* draining the wedged one. Same verdict, proportionate response. `retries = 3 × interval = 30 s` ⇒ ~90 s of continuous 503 before ECS acts, so detection is ~16.5 min against the 11.5 h this actually took.

Liveness stays `/health`'s job. A 503 here means "alive, and unable to read its database since `age_s` seconds ago" — the endpoint answering at all is still the liveness proof.

## Four judgement calls, stated

**ANY probe, not all.** The five share one database and one probe executor, so in the incident they went stale together — but requiring unanimity would let a wedged database hide behind `paper_rag`, the one member that can keep answering with Postgres gone. Cost of ANY, plainly: one permanently stuck probe replaces the task about once per threshold interval, forever. That is loud in the ECS event log, every replacement logs `HEALTH_READINESS_STALE` naming the probe, and `HEALTH_STALE_UNREADY_S=0` stops it with a restart and no deploy.

**`risk_data` is the member of that set whose cost grows with the data, and it is named so nobody has to rediscover it mid-incident.** Four of the five are scalar `COUNT`s or a single-row read. `risk_data_health()` (`backend/archimedes/api/risk_routes.py`) is `list_strategies()` plus one `get_backtest_result(s.id)` **per strategy** — an N+1 — under the same shared 0.8 s `_LOCAL_PROBE_BUDGET_SECONDS`, and it catches every exception internally, so it can never be `probe_error`: past the budget it goes `stale_cached` and *stays* there with its age growing without bound, and ANY then replaces every backend task about once per threshold interval on a fleet that is serving fine. Prod is nowhere near that today — a read-only `GET https://archimedes-arc.com/api/health` (re-read 2026-09-03) returns all five probes `live`, `strategy_count: 34`, `risk_data: live, 30 persisted backtest(s) with equity curves`; the same read also shows `ready` **absent** from the payload, which is the live confirmation that the running revision predates this PR — so this is a **tripwire, not a defect**:

- **The signal:** `risk_data_probe_state` on `/health` sitting persistently at `stale_cached` while the other four stay `live`. That shape is the N+1 outgrowing its budget, *not* a wedged database.
- **The fix when it fires:** bound or drop the per-strategy read — **not** raise the threshold.
- **The escape hatch:** `risk_data` is the one member of `_READINESS_PROBES` removable without weakening the incident's own coverage; the other four read the exact tables (`papers`, `strategy_store`) the DDL wedged.

All of this is now in `_stale_unready_probes`' docstring, where the next person to touch the set will read it.

**Only `stale_cached` counts.** `probe_timeout` (missed, nothing ever cached) is the cold-start shape — the cache is process-local and empty at boot — and replacing a task for not having completed a first read turns a slow boot into a replacement loop; that case belongs to `startPeriod`. `probe_error` (the probe raised) is a defect in the probe, not a verdict about the database, and a broken probe must not be able to take the fleet down — **pinned by mutation E below**, on the dangerous shape specifically: a probe whose memo is *already* ten hours stale and which now raises. A useful consequence: a **total** DB outage does not cause a replacement loop, because a replacement task never completes a first read and so is never "stale" — only a task that *could* read and stopped gets replaced, which is the incident's exact shape.

**The readiness fields on `/health` are reporting, and cannot outrank the endpoint reporting them.** `/health` is the ALB target-group check with `matcher = "200"`; anything raising in the block that computes `ready` would answer **500 on every target at once** and drain the service — the exact fleet-wide failure this PR avoids by putting the 503 on `/health/ready`. The block is therefore wrapped, like the reveal-reconcile read directly above it, and the fallback is `ready: null`, never `ready: true`: "we could not compute the verdict" is not "the task is ready" (same rule as the `absent` values). `/health/ready` computes it independently and stays the authority. **Pinned by mutation F below.**

**It runs its own probes rather than reading the cache someone else fills.** A read-only readiness check has the mirror-image failure: with nobody polling `/health` the recorded ages grow while nothing is ever attempted, and a healthy idle task gets replaced for being idle. Probe load is unchanged-to-lower — the container check used to run all nine `/health` probes every 30 s and now runs five.

## Wiring

- **`.github/scripts/ecs_rewrite_task_def.py` — the effective writer.** It rewrites the backend container's `healthCheck.command` to probe `/health/ready` and pins `HEALTH_STALE_UNREADY_S=900`, exactly the way it pins `PAPER_ADVANCE_ENABLED`. See the section below for why this, and not terraform, is what ships it.
- `backend/Dockerfile` `HEALTHCHECK` → `/health/ready` (documented twin; an image-only `HEALTHCHECK` is invisible to the ECS agent, so this alone gives nginx's `dependsOn: HEALTHY` nothing to wait on)
- `infra/ecs.tf` backend container `healthCheck` → `/health/ready` plus `HEALTH_STALE_UNREADY_S = "900"` (documented twin, labelled as one — the block now opens with `THIS FILE IS NOT WHAT SHIPS IT` and names the script)
- `infra/alb.tf` — **untouched**
- nginx and CloudFront need no change: `location /health` is a prefix proxy and `/health/*` is already an uncached CloudFront behaviour
- `docs/api/vaults-and-chain.md` documents the endpoint (`test_api_docs_drift` gates this); `docs/operations/feature-flag-fliplist.md`'s CI-task-def-rewrite row said the script injects `PAPER_ADVANCE_ENABLED` and nothing else, which is no longer true

## Tests — shown RED before pushing

`backend/tests/test_health_readiness_stale_1818.py`, 14 tests, hermetic (no network/DB/Redis; staleness injected by seeding the memo, reads blocked with `threading.Event.wait` in the calling thread so the loop is genuinely at risk).

**Mutation A — the rule removed** (`_stale_unready_probes` → `return []`):

```
FAILED ...::TestStaleBeyondThresholdIsUnready::test_all_db_probes_stale_for_ten_hours_reports_503
FAILED ...::TestStaleBeyondThresholdIsUnready::test_one_stale_db_probe_is_enough
FAILED ...::TestTheThresholdIsOperable::test_env_threshold_is_honoured
FAILED ...::TestTheThresholdIsOperable::test_an_unparseable_threshold_falls_back_to_the_default
FAILED ...::TestHealthItselfIsUnchanged::test_health_stays_200_and_publishes_the_same_verdict
=================== 5 failed, 7 passed, 1 warning in 25.03s ====================
```
```
E   AssertionError: a task 9.9h into a DB wedge answered 200
E   assert 200 == 503
```

**Mutation B — every probe treated as unready** (state and age checks defeated):

```
FAILED ...::TestStaleBeyondThresholdIsUnready::test_one_stale_db_probe_is_enough
FAILED ...::TestStaleWithinThresholdStaysReady::test_sixty_seconds_stale_answers_200_with_the_labels_intact
FAILED ...::TestStaleWithinThresholdStaysReady::test_a_never_completed_probe_is_not_unready
FAILED ...::TestFreshProbesAreReady::test_a_healthy_task_answers_200_and_live
=================== 4 failed, 8 passed, 1 warning in 15.91s ====================
```

**Mutations D+E — unwire the container check back to `/health`, and move readiness onto the ALB target group**:

```
FAILED ...::TestTheProbeIsActuallyWiredToTheContainerCheck::test_dockerfile_healthcheck_targets_the_readiness_endpoint
FAILED ...::TestTheProbeIsActuallyWiredToTheContainerCheck::test_ecs_task_definition_mirrors_it
FAILED ...::TestTheProbeIsActuallyWiredToTheContainerCheck::test_the_alb_target_group_still_checks_plain_health_for_200
========================= 3 failed, 1 warning in 3.50s =========================
```

`backend/tests/test_ecs_readiness_deploy_pin.py`, **21 tests**, hermetic, in the deploy-pin style of `test_ecs_paper_advance_deploy_pin.py`. It runs the same function `deploy.yml` invokes, against a last-good-clone fixture that states the pre-P3 `/health` command explicitly — plus an anti-vacuity class (`TestTheFixtureIsThePreP3Shape`) asserting the fixture really is that shape, so no rewrite assertion can pass by accident of the input already agreeing with it.

**Mutation P1 — the health-check rewrite removed** from `rewrite_registered_task_definition`:

```
FAILED ...::TestTheRewritePointsTheContainerCheckAtReadiness::test_a_clone_still_probing_plain_health_ships_the_readiness_command
FAILED ...::TestTheRewritePointsTheContainerCheckAtReadiness::test_the_timing_of_the_live_revision_is_preserved_not_reset
FAILED ...::TestTheRewritePointsTheContainerCheckAtReadiness::test_a_clone_with_no_health_check_at_all_gets_the_whole_block
FAILED ...::TestTheCliShipsIt::test_cli_writes_both_the_command_and_the_pin
=================== 4 failed, 17 passed, 1 warning in 3.35s ====================
```
```
E   assert '/health/ready' in 'python -c "import urllib.request; urllib.request.urlopen(\'http://localhost:8000/health\')" || exit 1'
```

**Mutation P2 — the `HEALTH_STALE_UNREADY_S` pin removed**: `4 failed, 17 passed` (`KeyError: 'HEALTH_STALE_UNREADY_S'`; and `assert '0' == '900'` on the emergency-stop-is-taken-back case).

**Mutation P3 — the live revision's timing clobbered by the defaults** (`dict(_DEFAULT_HEALTH_CHECK_SHAPE)` instead of `dict(health_check or {})`): `2 failed, 19 passed` — `assert (30, 5, 3, 30) == (45, 9, 5, 120)` and `KeyError: 'someNewKnob'`.

**Mutations P4+P5 — nginx given the readiness check, and a top-level field written** (`out["memory"] = "4096"`, outside #1799's ignore list): `3 failed, 28 passed` across both deploy-pin files, including `AssertionError: the readiness pin gave the top-level field 'memory' a second registrar`.

**Mutation P6 — the `ecs.tf` effective-writer note removed**: `1 failed, 20 passed` (`test_ecs_tf_says_the_pipeline_script_is_the_effective_writer`) — without it the `healthCheck` block reads as the thing that ships and the next person edits it and waits for a deploy that will never carry it.

### Review round — two more guards, both shown RED

The review found two claims the PR *asserted* but nothing pinned. Both now have a guard, and each guard was shown red on the mutation it exists to catch.

**Mutation E — a raising probe treated as unready** (in `_stale_unready_probes`, the `isinstance(outcome, BaseException)` branch appends `(name, threshold_s + 1.0)` instead of skipping). Before the new test this mutation was **survivable**: 62 tests passed across the three health files. Now:

```
FAILED ...::TestStaleWithinThresholdStaysReady::test_a_raising_probe_is_not_unready_even_when_its_cache_is_hours_old
E   AssertionError: a raising probe must not replace a task that can still read
E   assert 503 == 200
------------------------------ Captured log call -------------------------------
ERROR    archimedes.main:main.py:1353 HEALTH_READINESS_STALE: unready — corpus_db=901s stale beyond 900s; container health check fails
=================== 1 failed, 13 passed, 1 warning in 10.12s ===================
```

The test seeds `corpus_db` **10 h stale** and *then* makes its read raise, so it cannot pass by the age being small — only by the state being `probe_error`.

**Mutation F — the `try/except` dropped from `/health`'s readiness block.** The exception escapes the handler entirely (through starlette's task group), i.e. a **500** on the ALB's check rather than a 200:

```
FAILED ...::TestHealthItselfIsUnchanged::test_a_raising_readiness_verdict_cannot_turn_health_into_a_500
E   RuntimeError: readiness verdict blew up
=================== 1 failed, 13 passed, 1 warning in 10.02s ===================
```

Every mutation was reverted by editing the source back; the tree is clean of them (`grep -c MUTATION` = 0) and both files are green — 14/14 and 21/21.

## Suite

Re-run in full on the final tree (`057640cf`, `origin/main` merged in), from the repo root:

- `backend/tests/test_ecs_readiness_deploy_pin.py` — **21 passed** (the pipeline-pin file)
- `backend/tests/test_health_readiness_stale_1818.py` — **14 passed** (12 + the two review-round guards)
- Union with the pin and drift suites on the merged tree — **112 passed**: `test_ecs_readiness_deploy_pin` + `test_ecs_paper_advance_deploy_pin` + `test_ecs_free_generations_pin` + `test_health_readiness_stale_1818` + `test_health_always_answers` + `test_api_docs_drift` + `test_feature_flag_fliplist_drift`. The three env pins compose — `pin_backend_health_stale_unready(pin_backend_free_generations(pin_backend_paper_advance(env)))` — and both pin suites pass against the merged script, which is the union test for the `ecs_rewrite_task_def.py` conflict.
- `test_health_is_uncacheable.py` + `test_health_llm_probe_bounded.py` + `test_cloudfront_health_uncached.py` + `test_api_surface_status.py` — all green
- **CI backend, whole repo, on the final merged tree**: `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` — **`6250 passed, 4 skipped, 2 deselected, 337 warnings in 1125.93s (0:18:45)`**, **exit 0**, zero `FAILED`/`ERROR` lines. (6250, not the 5888 quoted earlier — `origin/main` moved 54 commits.)
- **GitHub CI on `057640cf`: 21 checks pass, none failing** — including `Backend — unit tests`, `Frontend + Better Auth`, `Ruff — format + critical lint rules` and `Infra — terraform fmt + validate (infra)`.
- `ui`: not re-run — the diff touches **no `ui/` file** (`git diff origin/main --name-only | grep '^ui/'` is empty), and `Frontend + Better Auth` passes on CI.
- `ruff format --check .` — **766 files already formatted**; `ruff check --select E9,F63,F7,F40,F82 .` — All checks passed (ruff **0.15.13**, the version `quality-gate.yml` pins)
- `terraform -chdir=infra fmt -check -recursive` clean, `terraform validate` — Success

## Ships through the pipeline rewrite (#1799): no terraform apply needed

`deploy.yml` clones the **live** task definition and `.github/scripts/ecs_rewrite_task_def.py` rewrites the clone. It does not apply terraform — so a `healthCheck` that exists only in `ecs.tf` is not live until somebody applies, and a clone registered before P3 goes on shipping the old `/health` command forever. That is task-def `:211`'s shape applied to a probe instead of a flag. On top of that, #1799 (PR #1833) puts `lifecycle { ignore_changes = [container_definitions] }` on `aws_ecs_task_definition.backend`, so terraform stops writing container settings **at all**: inside `containerDefinitions` the pipeline is not merely first, it is the only writer there is.

So the readiness check ships the only way it can. The rewrite script now:

```
healthCheck.command[1]  "…urlopen('http://localhost:8000/health')…"
                     -> "…urlopen('http://localhost:8000/health/ready')…"
environment            + { name = "HEALTH_STALE_UNREADY_S", value = "900" }
```

**The next deploy after merge carries it. There is nothing to apply.** Only the command moves — `interval`/`timeout`/`retries`/`startPeriod` and any unknown fields come through from the live revision untouched, because what the check *asks* is the script's decision and how often it asks is the registered revision's. A clone carrying no `healthCheck` at all gets the whole block from the script's documented default shape rather than being left with none (an absent key means ECS sees no container health, which is the same hole as an unset env name).

Everything written stays inside `containerDefinitions`, so #1799's ignore list stays correct — that branch's own `TestThePipelineWritesOnlyContainerDefinitions` passes unmodified against this script.

**The targeted apply is explicitly not the path.** Re-run on the merged tree (2026-09-03, `AWS_PROFILE=ArchimedesDanAdmin`, `-lock=false`, read-only): `terraform plan -target=aws_ecs_task_definition.backend` → **`Plan: 1 to add, 0 to change, 1 to destroy`**. What it would carry:

```
      ~ command = "…urlopen('http://localhost:8000/health')…"
              -> "…urlopen('http://localhost:8000/health/ready')…"     <- what we want
      + { name = "HEALTH_STALE_UNREADY_S",       value = "900" }             <- what we want
      + { name = "FREE_GENERATIONS_PER_ACCOUNT", value = "3"   }             <- intended (#1643, on main)
      ~ { name = "PAPER_ADVANCE_ENABLED",  value = "false" -> "true" }       <- intended (#1632 lift, on main)
      ~ { name = "PLATFORM_ADMIN_WALLETS", value = "0x2a29…5105" -> "" }      <- DRIFT, would break admin
      - secrets: GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET,
                 GOOGLE_CLIENT_ID, GOOGLE_CLIENT_SECRET                      <- DRIFT, would break sign-in
```

**Correcting an earlier revision of this section:** it listed `PAPER_ADVANCE_ENABLED false → true` as a hazard, on #1818's "keep it pinned `false` until P1 ships". That is now stale — P1 (#1819) merged, #1632's lift (`7ca54a7f`) merged, and both `infra/ecs.tf` and `ecs_rewrite_task_def.PAPER_ADVANCE_VALUE` pin `"true"` on purpose. The live revision is simply behind on it. The two lines that remain genuine hazards are the last two: emptying `PLATFORM_ADMIN_WALLETS` and dropping the four OAuth secret references would break platform admin and GitHub/Google sign-in respectively.

So the apply is unnecessary either way — the pipeline ships the readiness check on the next deploy — and it is still not free. **Nothing was applied. No AWS writes were made** (`-lock=false`, plan only; the SSM read of the Aurora password was read-only and the value was never printed).

**The ALB check is unchanged.** `infra/alb.tf` still checks path `/health` with `matcher = "200"`; readiness stays on the container check only, and a guard test pins that it stayed that way.

**Pulling the rule back.** `HEALTH_STALE_UNREADY_S=0` on the live revision disables it immediately and holds until the next deploy — the right shape for an incident. The *durable* pull-back is `HEALTH_STALE_UNREADY_VALUE = "0"` in the script, since the pipeline re-pins the name on every deploy. `main.py`'s docstring said an env change was enough on its own; corrected to say this.

## Left out (deliberately)

- **The CloudWatch alarm — and note that P5 landing did not supply it.** `HEALTH_READINESS_STALE` is emitted as a loud greppable literal in the shape of `HEALTH_CHAIN_DISCONNECTED` / `HEALTH_ORACLE_STALE`, but **no `aws_cloudwatch_log_metric_filter` or alarm keys off it**. P5 (#1826) has since MERGED; its four alarms are `alb_elb_5xx_count`, `ecs_service_memory_high`, `ecs_service_cpu_high` and `aurora_connections_wedge` — none of them reads this literal (`grep -n HEALTH_READINESS_STALE infra/cloudwatch.tf` → nothing). **So the residual is real and worth naming:** once this ships, a wedged task is *replaced* but nobody is *paged* for the replacement. `aurora_connections_wedge` would likely fire on the incident's own cause and the ECS event log records every replacement, so this is a gap in attribution rather than in detection — but it is a gap, and it belongs in a follow-up rather than smuggled into this PR.
- **P2 and P4** — the second session inside the cycle transaction, and `list_generated_strategies` on the event loop. Separate PRs per the issue's own split. (**P1 (#1819) and P6 have since merged** and are ancestors of this branch; P5 (#1826) too.)
- **`regime_detector`, `chain`, `oracle`** are not in the readiness set. A dark Arc RPC is not a reason to destroy a task that is serving pages, and the GMM probe reads an artifact off disk — neither says anything about reaching Postgres.
- **`docker-compose.production.yml`'s backend healthcheck stays on `/health`** (commented in place so it doesn't read as an oversight). Readiness exists so a *scheduler* replaces a wedged task; that ASG tier has none — compose evaluates `depends_on: service_healthy` at startup only and never re-routes — so a 503 there would take the replica out with nothing behind it, which is the fleet-wide shape this PR is avoiding. `dev.sh` and `deploy.yml` also stay on `/health`: they wait for the app to answer and parse its payload, which is liveness.
- **A "never completed a read since boot" rule.** There is no recorded "stale since" for `probe_timeout`, and inventing one from process start would fire on a genuinely cold task. Called out above; it is the one shape this rule does not cover.

## Merge note

#1733 also touches `backend/archimedes/main.py` and `infra/ecs.tf`. Union-test the pair before landing them together.

**#1833 (#1799) is the other one to union-test**, in both directions: it adds the `lifecycle` block and the `ignore_changes` guard to the same `aws_ecs_task_definition.backend` resource this PR comments, and its `TestThePipelineWritesOnlyContainerDefinitions` asserts against the same script this PR extends. Both sides were checked here — #1833's guard, taken verbatim from its branch, passes unmodified against this script — but the two ecs.tf hunks are close enough together to conflict textually. Order does not matter for correctness: this PR ships through the pipeline whether or not #1833 has landed.


dbrowneup:
## Review round — every defect addressed

Five defects were raised (1 major, 4 minor). All five are addressed; one was already closed by a commit that landed after the reviewed SHA, and I verified that rather than take it on faith. One extra defect of the same class as #3 was found while doing it and is fixed too.

| # | Sev | Defect | Commit | Guard |
|---|-----|--------|--------|-------|
| 1 | **major** | `ecs.tf` / "Wiring" reads as done, but the rule has no production effect until a (currently unsafe) `terraform apply` | **`0a20fc50`** (pre-existing on the branch) — verified this round | `test_ecs_readiness_deploy_pin.py`, 21 tests |
| 2 | minor | `probe_error` carve-out asserted but unpinned | **`518cebc8`** | new test + **mutation E red** |
| 3 | minor | `_READINESS_ABSENT` comment gives a false mechanism | **`518cebc8`** | n/a (comment) |
| 4 | minor | `/health`'s readiness block runs bare on a must-answer-200 endpoint | **`518cebc8`** | new test + **mutation F red** |
| 5 | minor | ANY-vs-ALL never names `risk_data` as the growth-sensitive member | **`518cebc8`** + body | n/a (docstring) |
| — | extra | `#1799`'s `ignore_changes` stated in the present tense while **#1833 is still open** | **`057640cf`** | n/a (comment) |

---

### 1 (major) — resolved by `0a20fc50`, and I checked the mechanism rather than the claim

The defect is exactly right about the reviewed SHA. It is no longer the shape of the branch: `0a20fc50` moved the readiness check off terraform and onto the deploy pipeline, which is the only writer inside `containerDefinitions`.

I traced the whole path myself rather than trusting the commit message:

- `.github/workflows/deploy.yml:738` → `aws ecs describe-task-definition` → **`:746`** `python3 .github/scripts/ecs_rewrite_task_def.py` → `:752` `aws ecs register-task-definition`.
- `ecs_rewrite_task_def.rewrite_backend_health_check()` sets `healthCheck.command` to `READINESS_HEALTH_CHECK_COMMAND` (`/health/ready`) and `pin_backend_health_stale_unready()` pins `HEALTH_STALE_UNREADY_S=900`, both inside the backend container.

So: **the next deploy after merge carries it; there is nothing to apply**, and the detection gap closes then rather than on an owner's apply. Both of the defect's mechanics are now stated in the body and at both twin sites — the `ecs.tf` block opens `THIS FILE IS NOT WHAT SHIPS IT`, and the `Dockerfile` bullet says an image-only `HEALTHCHECK` is invisible to the ECS agent.

**I also reproduced the plan the reviewer could not** (their SSO token had expired). Read-only, `-lock=false`, nothing applied:

```
terraform plan -target=aws_ecs_task_definition.backend
Plan: 1 to add, 0 to change, 1 to destroy.

  ~ command  "…urlopen('http://localhost:8000/health')…"
          -> "…urlopen('http://localhost:8000/health/ready')…"          <- wanted
  + { name = "HEALTH_STALE_UNREADY_S",       value = "900" }            <- wanted
  + { name = "FREE_GENERATIONS_PER_ACCOUNT", value = "3"   }            <- intended (#1643)
  ~ { name = "PAPER_ADVANCE_ENABLED",  value = "false" -> "true" }      <- intended (#1632 lift)
  ~ { name = "PLATFORM_ADMIN_WALLETS", value = "0x2a29…5105" -> "" }    <- DRIFT
  - secrets: GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET,
             GOOGLE_CLIENT_ID, GOOGLE_CLIENT_SECRET                     <- DRIFT
```

This **corrected the body**: it had listed `PAPER_ADVANCE_ENABLED false → true` as an apply hazard on #1818's "keep it `false` until P1 ships". Stale — P1 (#1819) and #1632's lift (`7ca54a7f`) both merged, and `ecs.tf` and `PAPER_ADVANCE_VALUE` pin `"true"` on purpose. The two genuine hazards are the last two lines. Body updated to say exactly that.

### 2 (minor) — `probe_error` now pinned, and the mutation that survived now fails

The reviewer's mutation (raising probe → unready) left all 62 health tests green. New test, seeded on the **dangerous** shape — `corpus_db` **10 h stale** *and then* raising, so it cannot pass by the age being small:

```
FAILED ...::TestStaleWithinThresholdStaysReady::test_a_raising_probe_is_not_unready_even_when_its_cache_is_hours_old
E   AssertionError: a raising probe must not replace a task that can still read
E   assert 503 == 200
------------------------------ Captured log call -------------------------------
ERROR    archimedes.main:main.py:1353 HEALTH_READINESS_STALE: unready — corpus_db=901s stale beyond 900s; container health check fails
=================== 1 failed, 13 passed, 1 warning in 10.12s ===================
```

### 3 (minor) — the false mechanism replaced with the true one

Confirmed the reviewer's reading in `services/health_cache.py`: `absent` goes out at **:152** in a `return ProbeOutcome(...)` and only a completed read reaches `self._entries[name]` at **:176**. The described divergence cannot occur. The comment now says so explicitly *and* gives the real reason — two endpoints answering the same question must publish the same "we do not know" value, and neither may publish an optimistic one.

### 4 (minor) — wrapped, and the fallback is `null`, not `true`

The block is now wrapped like the reveal-reconcile read directly above it. One thing beyond the suggested fix: defaulting `_stale_unready = []` alone would publish **`ready: true`** on a failure, which is the optimistic-substitute failure class this codebase treats as primary. So `ready` goes **`null`** (with a `null` threshold) when the verdict could not be computed. Mutation F — the `try/except` dropped — shows what it is protecting against:

```
FAILED ...::TestHealthItselfIsUnchanged::test_a_raising_readiness_verdict_cannot_turn_health_into_a_500
E   RuntimeError: readiness verdict blew up
=================== 1 failed, 13 passed, 1 warning in 10.02s ===================
```

The exception escapes the handler through starlette's task group — a **500** on the ALB's `matcher = "200"` check, i.e. the whole fleet drained. `docs/api/vaults-and-chain.md` follows the type: `ready: bool|null`, `stale_unready_threshold_s: float|null`.

### 5 (minor) — `risk_data` named, with the tripwire

Verified the N+1 at `api/risk_routes.py:98` and that it catches every exception internally, so it can only ever be `live` or `stale_cached`, never `probe_error`. I re-read prod myself rather than inherit the reviewer's reading — read-only `GET https://archimedes-arc.com/api/health`, 2026-09-03: all five probes `live`, `strategy_count: 34`, `risk_data: 30 persisted backtest(s) with equity curves`, and `ready` **absent** from the payload (the live confirmation that the running revision predates this PR). Now in `_stale_unready_probes`' docstring and in the body: the **signal** (`risk_data_probe_state` persistently `stale_cached` while the other four stay `live`), the **fix** (bound or drop the per-strategy read — *not* raise the threshold), and the **escape hatch** (`risk_data` is the one member droppable without weakening the incident's own coverage; the other four read `papers` / `strategy_store` directly).

### Extra — `#1799` stated as landed when #1833 is still open

Found while verifying #1: `infra/ecs.tf` and the script's module docstring both said "since #1799, `aws_ecs_task_definition.backend` carries `lifecycle { ignore_changes = [container_definitions] }`" in the present tense. **PR #1833 is OPEN** — `grep -n ignore_changes infra/ecs.tf` finds it only on `aws_ecs_service.backend`. The argument never needed it (deploy.yml clones the live revision and never applies, which is true today on its own); #1799 *strengthens* it. Both comments now say that, with #1833's state and the date. Same defect class as #3, which is why I did not leave it.

### Two body corrections the review round forced out

Not defects that were raised, but things the body asserted that stopped being true while main moved 54 commits:

- **`PAPER_ADVANCE_ENABLED false → true` is no longer an apply hazard** (see #1 above) — it is the intended value now.
- **P5 landing did not supply this PR's alarm.** The "Left out" bullet said alarms belong to P5. #1826 has since **merged**, and its four alarms (`alb_elb_5xx_count`, `ecs_service_memory_high`, `ecs_service_cpu_high`, `aurora_connections_wedge`) do **not** read `HEALTH_READINESS_STALE` — `grep -n HEALTH_READINESS_STALE infra/cloudwatch.tf` returns nothing. So the residual is real and now stated plainly: once this ships, a wedged task is **replaced but nobody is paged for the replacement**. Attribution gap rather than detection gap (`aurora_connections_wedge` would likely fire on the incident's own cause, and every replacement is in the ECS event log), but a gap, and a follow-up rather than something to smuggle in here.

---

### Verification

- **Mutations E and F**: each applied by editing the source, run, captured, reverted **by editing back**. `grep -c MUTATION backend/archimedes/main.py` → `0`, `git status --porcelain` clean after each.
- `backend/tests/test_health_readiness_stale_1818.py` — **14 passed** (12 → 14).
- Union with the pin + drift suites on the merged tree — **112 passed** (`test_ecs_readiness_deploy_pin` + `test_ecs_paper_advance_deploy_pin` + `test_ecs_free_generations_pin` + `test_health_readiness_stale_1818` + `test_health_always_answers` + `test_api_docs_drift` + `test_feature_flag_fliplist_drift`).
- `ruff format --check .` — **766 files already formatted**; `ruff check --select E9,F63,F7,F40,F82 .` — **All checks passed** (ruff 0.15.13, the version quality-gate.yml pins).
- `terraform -chdir=infra fmt -check -recursive` clean; `validate` — **Success**.
- No `ui/` file is touched by the diff.

**Merged `origin/main` in** (`e1c6f06a`). It had moved 54 commits and **conflicted in two files**, both resolved by keeping *both* sides:

- `.github/scripts/ecs_rewrite_task_def.py` — `b32d0417` added `pin_backend_free_generations` (#1643). The three pins now compose: `pin_backend_health_stale_unready(pin_backend_free_generations(pin_backend_paper_advance(env)))`, plus the `healthCheck` rewrite. **Both pin suites pass on the merged script** — that is the union test.
- `docs/api/vaults-and-chain.md` — main dropped `fusion_enabled` from `/health`; the readiness fields were re-applied onto main's line, so the drop survives.

**Do not merge — owner vets first.** The merge notes about #1733 and #1833 still stand.

PR #1831, opened 2026-09-03
#1818 P2: the paper-advance cycle is one connection (resolve_paper_hashes takes the caller's session)

dbrowneup:
Part of #1818 (P2).

## What this is

Mechanism item 3's other half. P1 (merged) took the DDL off the paper-advance cycle path; this takes the **second connection** off it.

`resolve_paper_hashes` (`services/paper_trace.py`) opened its own session — `with get_session() as session:` — while the cycle transaction was still open. On 2026-09-03 that was the wedge:

- the cycle transaction held `AccessShareLock` on `papers` for the whole replay,
- the sibling task's `ALTER TABLE papers` queued an `AccessExclusiveLock` behind it,
- this lookup's session then queued behind the ALTER.

The cycle waited on itself through two connections, which PostgreSQL cannot see. No deadlock detection, no timeout, no log line — ten hours wedged behind a `/health` that answered 200 from `stale_cached` probes.

## The change

1. **`resolve_paper_hashes(session, arxiv_ids)`** — `session` is REQUIRED, no `= None` default, because an optional one leaves a live route back to the second connection. Its single production caller, `_publish_decision_traces` in `services/paper_trading.py`, already holds the cycle session and passes it down, so `advance_all` / `advance_deployment` did not have to change and the `POST /api/paper/deployments` path inherits the fix.
2. **The corpus query runs inside `session.begin_nested()`.** Sharing the session is what makes the SAVEPOINT necessary: on PostgreSQL a failed statement aborts the whole transaction, so a missing `papers` table would stop costing this lookup and start costing the *cycle* — `PendingRollbackError` for every remaining deployment and a raising `session.commit()`. Without the savepoint the function's documented fail-soft return would be a lie about a wedged cycle. A connection-level failure is still fatal, as it must be: the cycle's writes were never going to commit either way.

Also updates `docs/incidents/2026-09-03-paper-advance-ddl-wedge.md` — adds a "What the P2 PR changes" section and removes P2 from the not-yet-done list.

## Guards, each shown RED on the input it rejects

**Mutation 1** — reinstate the old second session inside `resolve_paper_hashes` (`with get_session() as session:`):

```
E   AssertionError: the advance cycle opened 2 sessions, not 1 — a second connection inside the
    cycle transaction is the 2026-09-03 wedge shape (#1818 P2).
    Openers: ['.../backend/archimedes/services/paper_trace.py:472 in resolve_paper_hashes()']
E   assert 2 == 1

FAILED tests/test_paper_trace_pipeline.py::test_the_advance_cycle_opens_no_second_session
FAILED tests/services/test_paper_trace.py::TestResolvePaperHashesAgainstTheCorpus::test_it_reads_on_the_session_it_was_given_and_opens_none
        - AssertionError: resolve_paper_hashes opened a second session (#1818 P2)
FAILED tests/services/test_paper_trace.py::TestResolvePaperHashesAgainstTheCorpus::test_a_row_attribute_this_code_is_wrong_about_RAISES
FAILED tests/services/test_paper_trace.py::TestResolvePaperHashesAgainstTheCorpus::test_a_real_database_outage_IS_still_soft
```

`test_the_advance_cycle_opens_no_second_session` instruments `archimedes.db.get_session` for the whole of `advance_all` and allows exactly one call — the cycle's own — and names the file and line of every extra opener. It is a guard about the CYCLE being one connection, not about `papers` or about DDL, so it also catches whatever is added to the cycle path next.

**Mutation 2** — delete `begin_nested` (keep the shared session):

```
FAILED tests/services/test_paper_trace.py::...::test_the_outage_is_contained_by_a_savepoint_not_by_luck
        - AssertionError: the corpus lookup must run inside a savepoint that is rolle...
FAILED tests/services/test_paper_trace.py::...::test_the_savepoint_is_released_not_leaked_on_the_healthy_path
        - AssertionError: assert [] == ['savepoint',...se_savepoint']
```

The savepoint test drops the `papers` table for real and asserts the savepoint/rollback through SQLAlchemy's own `savepoint` / `rollback_savepoint` connection events. The events are the guard rather than "the caller could still write afterwards", because **SQLite does not abort a transaction on a statement error** — that weaker assertion passes with or without `begin_nested`, so it is asserted but is not the guard. Both mutations were reverted by editing the file back, and the suites re-run green.

## Not in this PR

P3 (readiness — `/health` still answers 200 from `stale_cached`), P4 (`list_generated_strategies` on the event loop), P5 (CloudWatch alarms), P6 (`init_db()` on eleven serving handlers), and the unexplained memory ramp. All still listed in the incident doc.

One second connection remains in the cycle deliberately: `paper_advance_loop` holds the fleet-lock session open for the whole cycle, which is what `pg_try_advisory_xact_lock` requires. It issues `SELECT pg_try_advisory_xact_lock(:key)` and no table reads, so it cannot join a table lock queue; the guard counts sessions opened DURING the advance, which is the thing that wedged.

## Tests

- CI backend command from the repo root, on this branch with `origin/main` (`1a40aa48`) merged in: `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` → **5854 passed, 4 skipped, 2 deselected** in 24:05.
- Targeted: `tests/services/test_paper_trace.py tests/test_paper_trace_pipeline.py tests/services/test_paper_trading.py tests/services/test_paper_advance_kill_switch.py tests/test_db_boot_patches.py` → 166 passed, 1 skipped.
- One earlier full-suite run (before the merge of main) had 2 failures, both in `backend/tests/test_post_rollout_probe.py` (`test_the_ceiling_bounds_the_whole_probe_not_just_the_gap_between_samples` — "probe ran 8.0s against a 4s ceiling" — and `test_a_single_slow_request_cannot_run_past_the_ceiling`). Those two assert wall-clock elapsed time around a `bash`/`curl` deploy probe and import nothing from `archimedes`; they passed 33/33 in isolation on this branch and on a clean `origin/main` worktree, and did not recur in the green run above. Load-sensitive flakes, not this diff.
- `ruff format --check .` → 742 files already formatted; `ruff check --select E9,F63,F7,F40,F82 .` → All checks passed.
- No UI files in the diff, so `ui` tests were not run.


dbrowneup:
**Post-merge review round — 3 defects, all fixed in #1841.**

This PR merged as `42cc1294` before the review landed, so the corrections are a follow-up branch off `origin/main` rather than a push here (same shape as #1838 for #1826). All three map to one commit, `66feeace` on `dbrowneup/1818-p2-followup`.

| # | severity | defect | fix |
|---|---|---|---|
| 1 | major | `session.begin_nested()` was **inside** the `try`, so the savepoint was never emitted in the case the docstring promised and the corpus handler swallowed the **caller's** errors | `66feeace` — `nested = session.begin_nested()` before `try:`, then `with nested:` inside |
| 2 | major | the cycle guard patched `archimedes.db.get_session`, which is blind to module-level-bound imports — the repo's dominant idiom | `66feeace` — instrument `archimedes.db.SessionLocal` |
| 3 | minor | body reports `ruff format --check .` → "742 files already formatted"; the pushed head's figure was 750 | corrected below; body left intact rather than rewriting a merged record |

---

**1.** `SessionTransaction.__init__` calls `_take_snapshot`, which runs `session.flush()` for a `BEGIN_NESTED` origin **before** the transaction is installed and before any `SAVEPOINT` SQL is emitted (sqlalchemy 2.0.51, `orm/session.py:962` → `:1090`), and `SessionLocal` is `autoflush=False`. `IntegrityError` **is** a `DBAPIError`. Probed on this PR's merged head with one pending `PaperDailyReturn` violating the unique constraint, against a **healthy** corpus:

```
LOG WARNING archimedes.services.paper_trace: paper trace: corpus content-hash lookup failed — recording no consulted hashes
RESULT: []
STATEMENTS: ['INSERT INTO paper_daily_returns (deployment_id, date, daily_']
SAVEPOINT EVENTS: []
```

Zero savepoint events; the corpus query never ran. The trace is stored and **hashed** with `consulted_paper_hashes=[]` while the corpus is fine, and on PostgreSQL the caller's transaction is already aborted with no savepoint to roll back to — the exact outcome §2 of this body says `begin_nested()` prevents. Reachable via `_publish_decision_traces` → `PaperTraceCoverageError` → `advance_all` catching per deployment without a rollback.

So **this correction applies to the body above**: "the caller's work before it survives" is not what happens — `begin_nested()` flushes that work first, outside the savepoint. The savepoint scopes the corpus query only. Corrected in the `paper_trace.py` docstring, the sibling test, and `docs/incidents/2026-09-03-paper-advance-ddl-wedge.md` item 2.

**2.** Mutation on this PR's head — a module-level `from archimedes.db import get_session as _mgs` in `paper_trading.py` plus a genuinely second session opened with it inside `_publish_decision_traces` (executing `SELECT 1`) — left the guard **GREEN** (`1 passed`). 14 modules bind the name at import time (`api/user_routes.py:29`, `api/wallet_routes.py:18`, `marketplace/service.py:26`, `services/corpus_service.py:23`, `scripts/run_backtests.py:19`, …), and that binding is invisible to the patch.

So **"it also catches whatever is added to the cycle path next"** (body, test docstring, incident doc) was not true for the way most of this repo opens sessions. `get_session()` is only `return SessionLocal()`, resolving the module global at call time, so #1841 instruments the factory. Now RED on all three idioms — module-level-bound `get_session`, direct `db.SessionLocal()`, and the original in-function `get_session()` — GREEN on clean code, message unchanged.

**3.** Correct figure for this PR's head (`8e8ca678`): **750 files already formatted**, not 742. On #1841's head it is 764.

Nothing here changes P2's actual delivery: `resolve_paper_hashes` still reads on the cycle session, the signature is still `session`-required, and the second-connection route is still closed. The defects are in the savepoint's placement and in the guard's reach.

PR #1833, opened 2026-09-03
[infra] #1799: container_definitions is pipeline-owned, plus a terraform drift gate

### LINKED ISSUE #1799: [infra][terraform] aws_ecs_task_definition.backend has drifted from the pipeline-owned revisions — an untargeted apply would replace it from infra/ecs.tf

dbrowneup:
Closes #1799.

Owner decision of 2026-09-03 encoded: **container settings are pipeline-owned.**

---

## 1. The ownership split (`infra/ecs.tf`)

`lifecycle { ignore_changes = [container_definitions] }` on
`aws_ecs_task_definition.backend`, with the reasoning in a comment above it.

**Why exactly that attribute and nothing more.** I read
`rewrite_registered_task_definition` in `.github/scripts/ecs_rewrite_task_def.py` rather
than guessing. The only keys the pipeline writes are inside `containerDefinitions` — the
`backend`/`nginx`/`auth` `image` tags and the backend container's `PAPER_ADVANCE_ENABLED`
environment pin. (That enumeration was true of the parent; `main` has since added more
container-level writes — see *Merged `main`* below. The list has grown, the boundary has
not.) It then drops the seven describe-only fields. Everything else in a
registered revision (`cpu`, `memory`, `executionRoleArn`, `taskRoleArn`, `runtimePlatform`,
the corpus-artifact `volume`, `networkMode`, `requiresCompatibilities`) is copied through the
clone untouched and authored by nobody but Terraform, so it stays Terraform-owned and still
plans. Adding any of those to the ignore list would hide drift on attributes that have
exactly one writer — strictly worse than the bug.

`tags` is also **not** ignored, deliberately: `describe-task-definition --query
taskDefinition` omits tags entirely, so the pipeline's revisions carry none, while Terraform
refreshes the specific revision ARN it registered (which has them). There is no diff to
suppress.

**Confirmed the service already ignores `task_definition`** — `aws_ecs_service.backend` has
`ignore_changes = [task_definition, desired_count]` and it is kept unchanged. A guard test
now pins it.

## 2. Evidence — before and after plan

Both runs: account `037613907429`, `infra/apply.sh` (which is `terraform plan`),
**untargeted, no `-target`**, from a clean worktree of this branch, with
`infra/terraform.tfvars` carrying the real production values (derived read-only from the live
task definition and SNS subscription: `platform_admin_wallets` =
`0x2a2968…15105`, `archimedes_treasury_wallet` = `""`, both OAuth flags `true` — the auth
container really does carry the Google/GitHub secret pairs — and `alarm_email` =
the address subscribed to `archimedes-alerts`).

### BEFORE (this branch's parent, `d3a157cd`)

```
Terraform will perform the following actions:

  # aws_ecs_task_definition.backend must be replaced
-/+ resource "aws_ecs_task_definition" "backend" {
      ~ arn                      = "arn:aws:ecs:us-east-1:037613907429:task-definition/archimedes-backend:213" -> (known after apply)
      ~ container_definitions    = jsonencode(
          ~ [
              ...
                  ~ environment      = [
                      ~ {
                            name  = "PAPER_ADVANCE_ENABLED"
                          ~ value = "false" -> "true"
                        },
                    ]
              ...
            ] # forces replacement
        )
      ~ revision                 = 213 -> (known after apply)
    }

  # aws_lambda_function.deploy_drift will be updated in-place
  # local_sensitive_file.private_key will be created

Plan: 2 to add, 1 to change, 1 to destroy.
```

State revision **213**; the live family is on **233** and the service runs 233 — exactly the
report in #1799.

### AFTER (this branch, `f9e29d80`)

```
Terraform will perform the following actions:

  # aws_lambda_function.deploy_drift will be updated in-place
  ~ resource "aws_lambda_function" "deploy_drift" {
        id                             = "archimedes-deploy-drift"
      ~ last_modified                  = "2026-09-01T21:19:17.000+0000" -> (known after apply)
      ~ source_code_hash               = "WaMLxBJV+Vw7S9ogvBc9/OSu0RNDGGys+scK/MJnzq4=" -> "yw2Sn4T/M/su/aDFikXK9B5eLBa92Y3hHFxWrVkHAQE="
        # (27 unchanged attributes hidden)
    }

  # local_sensitive_file.private_key will be created
  + resource "local_sensitive_file" "private_key" {
      + content              = (sensitive value)
      + directory_permission = "0700"
      + file_permission      = "0600"
      + filename             = "./archimedes-deploy-key.pem"
      + id                   = (known after apply)
    }

Plan: 1 to add, 1 to change, 0 to destroy.
```

**No `aws_ecs_task_definition.backend` line at all.** The two remaining entries are reported
verbatim above and are not mine — both pre-date this branch:

- **`local_sensitive_file.private_key` — create.** `infra/main.tf` writes the deploy SSH key
  to `./archimedes-deploy-key.pem` in Terraform's working directory. The `local` provider
  drops the resource from state when the file is absent, so **every** checkout that is not
  the working copy which first applied plans it as a create. It will not appear in your own
  checkout if that file is still there. Nothing in AWS is drifting.
- **`aws_lambda_function.deploy_drift` — update in place**, `source_code_hash` only. Root
  cause found, not guessed: I downloaded the deployed zip (`aws lambda get-function`) and it
  contains `__pycache__/index.cpython-312.pyc` alongside a **byte-identical** `index.py`,
  swept in from the applier's working copy by `archive_file`'s `source_dir`. One untargeted
  apply clears it.

### Adversarial checks on the claim

- Planned again with all four container-only operational variables at their `variables.tf`
  defaults (`google_oauth_enabled=false`, `github_oauth_enabled=false`, both wallets `""`).
  Result: **identical, resource for resource** — proof the ignore is really absorbing them,
  and the basis for the CI variable list below.
- Planned again with `alarm_email=""`. Result: an extra
  `aws_sns_topic_subscription.alerts_email[0] will be destroyed` — which is why the CI job
  needs that one value and why it is a documented secret rather than an omission.

## 3. The CI drift check

`.github/workflows/terraform-drift.yml` — `infra/**` pull requests, pushes to `main`, weekly
`cron: "17 7 * * 1"`, plus `workflow_dispatch`. Runs
`terraform plan -detailed-exitcode -input=false -lock=false -no-color -out=…`, uploads the
plan as an artifact, and fails on drift.

**It crosses a documented anti-goal, so it is a separate file.** `infra-gate.yml` is boxed
with "WHY fmt + validate AND NOT plan": a plan needs credentials and reads a state file
holding `tls_private_key.deploy`'s private key. That reasoning is about what *syntax
checking* is worth, not what *drift detection* is worth — and drift cannot be seen without a
plan. So `infra-gate.yml` is untouched and stays credential-free (a guard test asserts this),
and the new workflow carries the whole cost in one place, bounded:

- a dedicated read-only role, never the deploy role;
- `-lock=false`, so the role needs no write on the state bucket at all;
- **only the human-readable `plan.txt` is uploaded.** The `-out` planfile embeds the full
  prior state unredacted, so uploading it would hand the private key to anyone with repo
  read access. The `show -json` rendering has the same problem; both are deleted on the
  runner and a guard test asserts neither reaches `upload-artifact`.

**Exit code 2 alone cannot be the verdict**, because `local_sensitive_file.private_key`
guarantees it forever on an ephemeral runner. So `-detailed-exitcode` stays the trigger and
`.github/scripts/tf_drift_gate.py` is the verdict: it fails on **every** planned change
except one named, action-pinned exemption (that resource, `create` only — the same address
planning a `delete` fails). Stdlib-only, hermetic, 11 unit tests.

### The OIDC role: it does not exist, and here is exactly what is missing

The task said to add the missing permissions to "the terraform that defines the role." **No
Terraform defines it** — `infra/scripts/setup-github-oidc.sh` says so in its own header:
`archimedes-github-deploy` is deliberately not Terraform-managed. So the grants go in a
script, and in a **new, separate** one:

1. **Trust.** The deploy role's trust policy pins `:ref:refs/heads/main`. A `pull_request`
   run mints `…:pull_request`, which cannot match. Widening *that* role to accept pull
   requests would let any PR branch assume something that can push to ECR and
   `ssm:SendCommand` into production — a bigger hole than #1799.
2. **Permissions.** Its inline policy is ECR push, SSM SendCommand, `ec2:DescribeInstances`
   and the docs-site S3/CloudFront pair. A plan of this root refreshes ~70 resources across
   ~25 services and reads `s3://archimedes-tfstate-037613907429`. It has none of that.

`infra/scripts/setup-github-plan-role.sh` (dry-run by default, same subject-resolution logic
and the same both-forms trust as the existing script — its dry run correctly resolved this
repo's id-qualified subject `repo:aprin-labs@284008417/archimedes@1236816811`) creates
`archimedes-github-plan` with:

- **`arn:aws:iam::aws:policy/ReadOnlyAccess`.** I checked the live policy document rather
  than assuming: it covers every service this root touches (ec2, elasticloadbalancing, ecs,
  ecr, rds, elasticache, elasticfilesystem, iam, logs, cloudwatch, events, scheduler, lambda,
  sns, wafv2, cloudfront, route53, acm, application-autoscaling, autoscaling, budgets, ses,
  s3, ssm). The script's header is honest about its size — `s3:Get*` on every bucket,
  `ssm:Get*` on every parameter — and names the tighter alternative.
- **`kms:Decrypt`**, scoped to the resolved `alias/aws/ssm` key **and**
  `kms:ViaService = ssm.us-east-1.amazonaws.com`. **This is the one real gap:**
  ReadOnlyAccess grants `ssm:Get*` but deliberately not `kms:Decrypt`, so
  `get-parameter --with-decryption` fails without it.
- explicit read on the state bucket (redundant with ReadOnlyAccess today, written out so
  tightening the attachment later does not silently break `terraform init`).

Nothing writeable is granted anywhere.

### Secrets and variables — documented, never hardcoded

| Setting | Kind | Why it is needed |
|---|---|---|
| `TF_DRIFT_ENABLED` | repo variable | Arms the job. Unset ⇒ skipped, so this PR is safe to merge before the role exists. Same self-activating convention as `DEPLOY_ENABLED`. Row added to the flip-list. |
| `TF_PLAN_ROLE_ARN` | repo variable | A role ARN is not a secret — `setup-github-oidc.sh` says so itself. A guard test fails if a literal ARN appears in a workflow step. |
| `TF_VAR_ALARM_EMAIL` | repo **secret** | Measured, not assumed: unset ⇒ `aws_sns_topic_subscription.alerts_email[0]` plans as a destroy. A personal address, hence a secret. |
| `aurora_master_password` | **neither** | Read from SSM by the job itself, exactly as `infra/README.md` documents for a human operator, so no DB password is stored in GitHub. Needed because `outputs.tf`'s `database_url` interpolates it (`aws_rds_cluster.main` itself ignores `master_password`). |

The other four operational variables from `terraform.tfvars.example` are read only inside
`container_definitions` and are now provably unnecessary in CI — see the adversarial check
above.

**Expect the gate's first armed run to be RED**, on `aws_lambda_function.deploy_drift`. That
is the residual above and it is the gate working. One untargeted `infra/apply.sh --apply`
clears it.

## 4. Runbook

`docs/runbooks/terraform-apply-and-task-definition-ownership.md` (indexed; `make docs-check`
is clean — 0 broken links, 0 orphans). It states the ownership table, that editing
`container_definitions` in `ecs.tf` no longer reaches production and that the env pins live
in `ecs_rewrite_task_def.py`, three ways to make a deliberate task-definition change (with
the warning that `ecs.tf` must be updated too — #1798 is exactly that shape), what a clean
untargeted plan now looks like with both residuals named, how to arm the drift gate, and how
to roll the whole thing back. `cloudfront-cache-behaviour-apply.md`'s "CI never runs
`terraform plan`" line is corrected in place.

## 5. One change beyond the three deliverables, flagged

`infra/cloudwatch.tf` gains `excludes = ["__pycache__"]` on
`data.archive_file.deploy_drift`. It is load-bearing for deliverable 2: without it, anyone
who imports that module locally before applying re-bakes a `__pycache__` into the zip and the
drift gate flaps red on a clean tree. **It does not change the current plan** — verified by
re-running the after-plan with and without it: the same
`WaMLxBJV… -> yw2Sn4T…` diff either way.

## Verification run

- `terraform fmt -check -recursive -diff infra` — clean.
- `terraform -chdir=infra validate` — `Success! The configuration is valid.`
- Two untargeted plans (before/after) plus two adversarial variable plans. **No apply, no
  AWS write of any kind** — every AWS call in this work was a describe/get.
- `backend/tests/test_task_definition_ownership.py` — 33 tests, all passing, and each guard
  was mutation-tested: removing the lifecycle block, widening the ignore list with `cpu`,
  gutting the service's `ignore_changes`, making the rewrite script write a top-level field,
  dropping the schedule, uploading the binary planfile, removing the enable gate, dropping
  `-lock=false`, dropping `-detailed-exitcode`, un-watching `infra/**`, and hardcoding the
  role ARN each turn it red. Two of those mutations initially passed — the comment prose was
  satisfying the assertion — and both tests were fixed to read the executable lines only.
- `test_ecs_backend_secrets`, `test_ecs_generation_timeout`, `test_ecs_paper_advance_deploy_pin`,
  `test_infra_gate_workflow`, `test_docs_site`, `test_feature_flag_fliplist_drift`,
  `test_cloudwatch_alarms`, `test_canonical_repo_links` — 140 passing.
- `ruff format --check .` and `ruff check --select E9,F63,F7,F40,F82 .` — clean.
- The flip-list guard caught `TF_DRIFT_ENABLED` as an undocumented flag; a row was added
  rather than the guard weakened.

## Review round — `f1bc1f89`

Four defects were raised against `f9e29d80`; all four are fixed in `f1bc1f89`.

**Major — the plan role could decrypt every prod secret.** `ReadOnlyAccess` grants
`ssm:Get*` on Resource `*` (read from the live document: three `Allow` statements, no `Deny`
of any kind), and this PR added the one thing it withholds — `kms:Decrypt` on
`alias/aws/ssm`. All **19** SecureStrings in the account are encrypted under that key, so
the pairing was permission to read `CIRCLE_ENTITY_SECRET`, `BETTER_AUTH_SECRET`,
`DATABASE_URL` and the rest in cleartext, from a role assumable by any in-repo pull-request
branch. The `AuroraPasswordParameter` statement scoped to one parameter was cosmetic. Two
`NotResource` **Deny** statements now re-scope the managed attachment; a `Deny` beats an
`Allow` from any policy. Confirmed with `aws iam simulate-custom-policy`, run over the
document the script actually renders (checked equal to the simulated one) plus the live
`ReadOnlyAccess` grant:

| call | decision |
| --- | --- |
| `ssm:GetParameter` on `…/AURORA_MASTER_PASSWORD` | `allowed` |
| `ssm:GetParameter` on `…/CIRCLE_ENTITY_SECRET`, `…/BETTER_AUTH_SECRET` | `explicitDeny` |
| `ssm:GetParameters` on `…/DATABASE_URL` | `explicitDeny` |
| `ssm:GetParametersByPath` on `…/archimedes/prod` | `explicitDeny` |
| `ssm:GetParameterHistory` on `…/TIINGO_API_TOKEN` | `explicitDeny` |
| `s3:GetObject` on the state object and the state lockfile | `allowed` |
| `s3:GetObject` on another bucket | `explicitDeny` |
| `s3:GetBucketPolicy` / `GetBucketAcl` (bucket config refresh) | `allowed` |
| `kms:Decrypt` `ViaService=ssm` / `ViaService=s3` | `allowed` / `implicitDeny` |

Neither Deny costs the plan anything, and that is checked rather than assumed: `infra/`
declares no `aws_ssm_parameter` and no `aws_s3_object` (the SSM ARNs in `ecs.tf` are
`valueFrom` **strings** handed to ECS, never read by terraform), and both lambdas load code
from a local `filename`. The header sentence calling the marginal exposure "small" is gone,
replaced by a plain statement of the real reach and of what deliberately survives — parameter
*names* via `ssm:Describe*`, bucket *configuration* via `s3:Get*`.

**Minor — the gate was red by construction on any intentional infra PR.** A PR that changes
`infra/**` plans its own resources as `create`, and the classifier fails on every planned
change outside the one exemption. The classify step is now `continue-on-error` on the
`pull_request` arm only. Nothing is hidden: the verdict is still logged, the step is still
annotated failed, `plan.txt` still uploads. `push` / `schedule` / `workflow_dispatch` still
fail the job, and a *broken* plan still fails on every arm — that is the `plan` step, which
carries no `continue-on-error`. Documented in the workflow's warning box and in the runbook,
including the consequence: **on a PR, a green tick does not mean "no infra change here".**

**Minor — a push to `main` could cancel the Monday run.** `github.event.pull_request.number
|| github.ref` resolves to `refs/heads/main` for both a push and the schedule, and
`cancel-in-progress: true` then let a push kill the one trigger that catches drift nobody
committed. `github.event_name` is now part of the group key *and* cancellation is scoped to
pull requests.

**Minor — the test count.** `92 passing` was wrong; the eight files collect 140. Corrected
above.

Nine guards were added for the above (33 tests total, from 24). Each was mutation-tested:
12 mutations, every one caught by the intended test and by no other — both Denys deleted,
an action dropped from the ssm Deny, the `NotResource` → `Resource` inversion (which reads
as protection and is its exact opposite), each Deny's scope pulled out of step with its
`Allow`, the `ViaService` pin removed, the concurrency key reverted, cancellation switched
off, `continue-on-error` removed and widened to `true`, and the plan step made advisory.

**Plan re-run after these changes** (untargeted, `-lock=false`, exit 2): `Plan: 1 to add,
1 to change, 0 to destroy`, no `aws_ecs_task_definition` entry, `source_code_hash
WaMLxBJV… -> yw2Sn4T…` — identical to the after-plan above, hash for hash. This round
touched no `.tf` file. Local terraform was v1.15.3 (the workflow pins 1.15.8;
`required_version` is `>= 1.0`). No apply, no AWS write — every AWS call was a
describe/get/simulate.

## Merged `main` — `f94818f4`, `cba64741`

`origin/main` (`e81b60f0`) merged in. One conflict, in `docs/runbooks/README.md`: #1806's
`market-data-provider-proof.md` index row landed at the same position as this branch's
runbook row. **Both rows kept** — `docs_links.py` and `docs_index.py` are clean (1864 links,
0 broken; 183 docs, 0 orphaned).

**The pipeline script grew four more container-level writes since the parent — every one of
them inside `containerDefinitions`, so `ignore_changes = [container_definitions]` is
unchanged and still exactly right.** They are `FREE_GENERATIONS_PER_ACCOUNT` (#1643 A5), the
`TIINGO_API_TOKEN` secret entry (#1798/#1806), and #1818 P3/#1830's pair — the backend
container's `healthCheck.command` repointed at `/health/ready` plus the
`HEALTH_STALE_UNREADY_S` env pin. `healthCheck` is a **container** field
(`containerDefinitions[*].healthCheck`), not a top-level one, so it falls inside the ignore
just as the image tags do.

Verified rather than reasoned about: `rewrite_registered_task_definition` from the merged
script, run over a revision fixture carrying `tags`, `ephemeralStorage` and
`placementConstraints`, returns **zero** changed top-level keys, **zero** invented top-level
keys and zero dropped ones beyond `_DROP_FIELDS`.
`backend/tests/test_task_definition_ownership.py` — 33 tests, green on the merge result, and
the two load-bearing guards were
re-mutated against main's rewritten script: making the script write a top-level `cpu` reds
`test_no_top_level_field_is_rewritten`, and widening the ignore list to
`[container_definitions, cpu]` reds `test_it_ignores_nothing_else`. Both reverted by editing.

**The before/after plans in section 2 are left exactly as they were.** They were run on the
parent (`d3a157cd` / `f9e29d80`) against the live account and this merge adds no `.tf` change
of its own — re-running them would need credentials this environment does not have, and
restating them as if re-measured would be the dishonest move. `terraform fmt -check
-recursive -diff infra` and `terraform -chdir=infra validate` are both clean on the merge
result.

Also re-run on the merge: `ruff format --check .` (776 files), `ruff check --select
E9,F63,F7,F40,F82 .`, and the full backend suite — `pytest -m "not integration"`, **6640
passed, 4 skipped, 0 failed** in 13m48s. No `ui/` file is in this PR's diff against `main`.

**One correction to the runbook, because `main` made it false** (`cba64741`, docs only).
Section 4's runbook used #1798's Tiingo key as the worked example of a *one-time live
registration* — step 2's shape. #1806 did not do that; it pinned the secret in
`ecs_rewrite_task_def.py`, which is step 1's shape. Both that example and the closing "#1798
is exactly that shape" line are corrected, route 1 is now named as the preferred one with the
reason (a one-time addition survives only until somebody registers a revision from a clone
that predates it), and step 1's single-guard reference (`test_ecs_paper_advance_deploy_pin.py`)
is widened to the four pin guards that now exist. Nothing else in the runbook moved.

## Owner decisions left open

1. Arm the gate (run the plan-role script, set the three settings) — or don't, and the
   workflow stays inert.
2. ~~Whether `ReadOnlyAccess` is the right blast radius for a CI role.~~ **Addressed in
   `f1bc1f89`** — the two `NotResource` Denys cut it back to the one Aurora parameter and the
   state bucket's objects, simulator-verified. What remains open is narrower and is a
   judgement call, not a hole: `ssm:Describe*` still lists parameter *names* account-wide and
   `s3:Get*` still reads bucket *configuration* account-wide. Say the word and the managed
   attachment becomes an explicit per-service policy; nothing else in the workflow changes.
3. The `deploy_drift` hash residual clears on your next untargeted apply. Nothing here
   requires you to do it now.


dbrowneup:
## Fix round — all four defects addressed in `f1bc1f89`

| # | Severity | Defect | Fix | Where |
| --- | --- | --- | --- | --- |
| 1 | **major** | Plan role's real blast radius is every prod secret — `ReadOnlyAccess` grants `ssm:Get*` on `*`, and the added `kms:Decrypt` on `alias/aws/ssm` decrypts all 19 SecureStrings; the one-parameter `Allow` was cosmetic. Header called the marginal exposure "small". | Two `NotResource` **Deny** statements in the inline policy + header rewritten to state the real reach | `f1bc1f89` — `infra/scripts/setup-github-plan-role.sh` |
| 2 | minor | PR arm red by construction: a PR that deliberately changes `infra/**` plans `create` and fails the classifier | `continue-on-error` scoped to the `pull_request` arm, documented in the warning box **and** the runbook | `f1bc1f89` — `.github/workflows/terraform-drift.yml`, `docs/runbooks/terraform-apply-and-task-definition-ownership.md` |
| 3 | minor | A push to `main` could cancel the weekly scheduled run (same group key, `cancel-in-progress: true`) | `github.event_name` in the group key **and** cancellation scoped to pull requests | `f1bc1f89` — `.github/workflows/terraform-drift.yml` |
| 4 | minor | PR body said "92 passing"; the eight files collect 140 | Corrected to `140 passing` (and `24 tests` → `33 tests`) | PR body, plus a `Review round` section |

### On defect 1 — the fix is verified, not asserted

You were right that the `AuroraPasswordParameter` statement was decoration. I confirmed the
mechanism before fixing it: the live `ReadOnlyAccess` document (v188) is **three `Allow`
statements on `Resource: "*"` and contains no `Deny` of any kind**, one of which carries
`ssm:Get*` and `s3:Get*`. And `describe-parameters` returns **19** SecureStrings, every one
`KeyId = alias/aws/ssm`.

Since a `Deny` beats an `Allow` from any policy, the two `NotResource` Denys re-scope the
managed attachment without enumerating it. I checked the result with
`aws iam simulate-custom-policy` — a read-only API — over the live `ReadOnlyAccess` grant
statement plus **the document the script actually renders** (I diffed the dry-run output
against the simulated document; they are equal):

```
ssm:GetParameter        …/AURORA_MASTER_PASSWORD          -> allowed
ssm:GetParameter        …/CIRCLE_ENTITY_SECRET            -> explicitDeny
ssm:GetParameter        …/BETTER_AUTH_SECRET              -> explicitDeny
ssm:GetParameters       …/DATABASE_URL                    -> explicitDeny
ssm:GetParametersByPath …/archimedes/prod                 -> explicitDeny
ssm:GetParameterHistory …/TIINGO_API_TOKEN                -> explicitDeny
s3:GetObject            <state bucket>/…/terraform.tfstate  -> allowed
s3:GetObject            <state bucket>/…tfstate.tflock      -> allowed
s3:ListBucket           <state bucket>                      -> allowed
s3:GetObject            archimedes-docs-site/index.html     -> explicitDeny
s3:GetBucketPolicy      archimedes-docs-site                -> allowed
s3:GetBucketAcl         archimedes-alb-logs                 -> allowed
kms:Decrypt             ViaService=ssm / ViaService=s3      -> allowed / implicitDeny
```

Bucket-level reads stay `allowed`, so refreshing bucket configuration is untouched. **Neither
Deny costs the plan anything**, and I checked that rather than assuming it: `infra/` declares
no `aws_ssm_parameter` and no `aws_s3_object` — the SSM ARNs in `ecs.tf` are `valueFrom`
*strings* handed to ECS, never read by terraform — and both lambdas load code from a local
`filename`, not from S3.

I took your `s3:GetObject` suggestion as well as the ssm one. What I deliberately left is
named in the header rather than papered over: `ssm:Describe*` still lists parameter **names**
account-wide (no values) and `s3:Get*` still reads bucket **configuration** account-wide.

### On defect 2 — I took the `continue-on-error` option, and documented it anyway

You offered either/or; the split that seemed right is that the *classifier* is advisory on
PRs while the *plan* stays blocking everywhere. So a broken plan — bad credentials, an
unparseable root — still fails on every arm including PRs; only "this plan contains changes"
is advisory, and only on `pull_request`. I still wrote both doc lines you asked for, because
the consequence needs saying out loud: **on a PR, a green tick does not mean "no infra change
here"** — read `plan.txt`.

### On defect 3 — both halves, since each is independently justified

Group key now separates the triggers, *and* `cancel-in-progress` is limited to pull requests
(the only trigger that genuinely races itself — a force-push makes the older plan worthless).
A second push to `main`, a dispatch, or a schedule now queues rather than replacing, so no
drift verdict is discarded.

### Guards

Nine new tests (33 total, from 24), all stdlib-only to match this suite's contract — PyYAML
is not in `backend/requirements.txt`. **12 mutations, every one caught by the intended test
and by no other:**

| mutation | caught by |
| --- | --- |
| concurrency key reverted to the shared form | `test_a_push_to_main_cannot_cancel_the_weekly_scheduled_run` |
| cancellation switched off entirely | `test_a_pull_request_still_cancels_its_own_superseded_run` |
| classify step made blocking again | `test_an_intentional_infra_pr_is_not_red_by_construction` |
| classify advisory on **every** arm | `test_an_intentional_infra_pr_is_not_red_by_construction` |
| the *plan* step made advisory | `test_a_broken_plan_still_fails_on_every_arm` |
| ssm Deny deleted | `test_every_value_returning_ssm_action_is_denied_elsewhere` |
| one value-returning ssm action dropped | `test_every_value_returning_ssm_action_is_denied_elsewhere` |
| `NotResource` → `Resource` (the inversion) | `test_the_denys_are_scoped_with_notresource_never_resource` |
| Deny exempts a different parameter | `test_each_deny_carves_out_exactly_what_its_allow_grants` |
| s3 Deny deleted / widened to `*/*` | `test_object_reads_outside_the_state_bucket_are_denied`, `test_each_deny_carves_out_exactly_what_its_allow_grants` |
| `kms:ViaService` pin removed | `test_the_decrypt_grant_is_still_pinned_to_ssm` |

The inversion mutation is the one I most wanted covered: `{"Effect":"Deny","Resource": <the
Aurora parameter>}` denies the one thing the workflow needs and permits the other eighteen
secrets, while still reading as "there is a Deny here" to anyone skimming.

The concurrency guards are behavioural rather than greps — they render the group key GitHub
would actually compute for `push`, `schedule` and `pull_request` and compare them.

### Verification

- **Read-only plan re-run, untargeted, `-lock=false`, exit 2:** `Plan: 1 to add, 1 to change,
  0 to destroy`; no `aws_ecs_task_definition` entry; `source_code_hash WaMLxBJV… -> yw2Sn4T…`
  — identical to the after-plan in the body, hash for hash. This round touched no `.tf` file.
  Local terraform was v1.15.3 (the workflow pins 1.15.8; `required_version` is `>= 1.0`).
- `terraform fmt -check -recursive -diff infra` clean; `terraform -chdir=infra validate`
  `Success!`
- New suite **33 passed**; the eight named suites **140 passed** (your count, confirmed);
  `test_local_mode_contract` / `test_post_rollout_probe` / `test_single_openssl` **73 passed,
  1 skipped**; `ruff format --check .` 747 files formatted; `ruff check --select
  E9,F63,F7,F40,F82 .` clean; `make docs-check` 0 broken links of 1816, 181/181 indexed, no
  new stale docs.
- **No apply, no AWS write.** Every AWS call this round was `describe` / `get` /
  `simulate-custom-policy`. `infra/terraform.tfvars` stays gitignored and untracked; no
  `archimedes-deploy-key.pem` or planfile was produced.

One process note worth flagging: mid-round I restored a mutation with `git checkout --` while
the fixes were still uncommitted, which silently reverted two of them. I caught it, re-applied
both, and redid the whole mutation pass with file snapshots plus a SHA-256 check that every
file comes back byte-identical. The 12 results above are from that clean run.

Issue #1804, opened 2026-09-02
[email][signal] SES bounces and complaints are invisible: no configuration set, no event destination, MessageId discarded — a bounced address looks like an unverified user and is locked out of free generations forever


dbrowneup:
Owner direction (Dan, 2026-09-02): a bounce means a fake email; the system should record it; keep the ability to remove an address from suppression for the listed cases.

**Facts (main @ 9cb868eb):** no `aws_ses_configuration_set` / event destination / identity notification topic in `infra/` (only inbound receipt rules in `infra/ses_inbound.tf`); `auth/mailer.js:31-36` sends without `ConfigurationSetName` and drops the returned `MessageId`. The users table has only `emailVerified` (`models/account.py:40`), which is also the free-generation gate (`free_generations.py:91-103`), so a bounced address is indistinguishable from "hasn't clicked yet" and stays locked out with no explanation. #1790 (unmerged, tranche 3) adds `auth_email_deliveries` with per-send rows and a `suppressed` state read from the SES suppression list at status time — that is a pull; this issue is the push.

**Minimal loop:** terraform: `aws_sesv2_configuration_set` + `aws_sesv2_configuration_set_event_destination` (BOUNCE, COMPLAINT, REJECT, DELIVERY) → `aws_sns_topic.ses_events` → SQS queue (durable) → a small consumer (scheduled ECS task or the auth service polling) that writes `bounced`/`complained` rows into `auth_email_deliveries` (address, message_id, type, sub-type, timestamp) and sets a `users.email_bounced_at`; mailer passes `ConfigurationSetName` and stores the `MessageId` (#1790 already records it). UI: resend control shows `bounced` ("this address rejected our mail — use a different one"); signup refuses known hard-bounced addresses; the #1780 script stays the operator path for the three legitimate removal cases; docs runbook updated. Guard: a test that the configuration set name in terraform equals the one the mailer sends with.

PR #1837, opened 2026-09-04
#1804 SES bounce push loop: configuration set → SNS → SQS → user stamp → typed refusal

dbrowneup:
Part of #1804.

> **Fix round (review pass 1).** Five defects addressed, plus one I found myself. The migration had forked against main's new chain head (**blocked every deploy**); the name-drift guard scanned the whole of `ecs.tf` and so could not tell **which container** an env var landed in — the loop could be made deaf with every test green; **nothing invoked the consumer**, so the loop as shipped was inert, and that is now wired rather than disclaimed (a scheduled Fargate one-off + two alarms); the terraform note here understated `aws_ecs_task_definition.backend`'s pre-existing drift, and now states it with the plan hunks and two explicit options; and the audit-trail residual is concrete instead of implied. Re-serializing the migration also broke two of #1283's tests, fixed here. Details in the sections below and in the fix-round comment.

Nothing in this account subscribed to SES delivery feedback: no `aws_sesv2_configuration_set`, no event destination, no topic (`ses_inbound.tf` next door is inbound mail for privacy@, a different thing). A hard bounce therefore reached us only as an entry AWS silently added to the account suppression list — visible to an operator running `ses_suppression list` and to nobody else, ever. The product's own view was `emailVerified = false`, which is the same value it holds for someone who simply has not clicked the link yet, and that boolean is also the free-generation gate. **A dead or fake address was indistinguishable from an impatient human and was locked out with no explanation and no path forward.**

## The loop this builds

```
auth container ──SendEmail(ConfigurationSetName = archimedes-mail)──▶ SES
                                                                      │
                                                    BOUNCE / COMPLAINT / REJECT / DELIVERY
                                                                      ▼
                                              SNS  archimedes-ses-events
                                                                      ▼
                                              SQS  archimedes-ses-events   (14-day retention + DLQ)
                                                                      ▼
                          EventBridge Scheduler ──every 15 min──▶ Fargate one-off
                                   python -m archimedes.scripts.ses_events drain
                                                                      ▼
                                   auth_users.emailBouncedAt / emailBounceKind
                                                                      ▼
                       signup + the signed-in resend refuse, with a typed code
```

**Why SNS→SQS and not SES→SQS.** SESv2 event destinations cannot target SQS. SNS alone is not enough either — a topic with no healthy subscriber *drops* the notification, so a consumer that runs periodically would lose every bounce that arrived while it was not listening. The queue is the buffer that makes those losses impossible, and its 14-day retention is what licenses a periodic consumer: **a bounce that arrives between drains is late, never lost.**

**Something calls it.** A first draft shipped the consumer as a management command with nothing invoking it — which is the defect this issue opened against, rebuilt with more terraform in it: bounces would accumulate in SQS and be deleted unread by the 14-day retention. The issue names the shape, so it is here: `aws_scheduler_schedule.ses_events_drain` runs a dedicated single-container `archimedes-ses-events-drain` Fargate task every 15 minutes (`var.ses_events_drain_schedule_expression`).

**Not a command override on the service family.** `aws_ecs_task_definition.backend` defines three containers, and its nginx sidecar `dependsOn` the backend being HEALTHY. Overriding the backend's command to the drain means it never serves HTTP, never turns HEALTHY, and ECS is liable to kill the task before the drain finishes — the same reasoning, and the same conclusion, as `infra/ecs_migrate.tf`, whose one-off shape this copies. Fargate has no native SQS trigger either; the alternatives were a Lambda (a second runtime and dependency set for one file that already runs in the backend image) or an always-on poller (a new singleton to reason about) — both buying latency that nothing here needs, since the deadline is a human eventually retrying a signup.

**And two alarms, so the schedule is not the next unwatched link.** `archimedes-ses-events-not-drained` watches the queue's `ApproximateAgeOfOldestMessage` (a drain that stopped running is otherwise indistinguishable from a quiet week); `archimedes-ses-events-dlq-not-empty` watches DLQ depth (a parser behind an SES payload change). Both route to `cloudwatch.tf`'s existing alerts topic, and neither treats missing data as breaching — an empty queue publishes no age datapoints, and empty is the healthy steady state.

Cost, stated rather than waved at: 96 invocations/day at Fargate's 1-minute minimum billing ≈ 48 h/month at 0.25 vCPU + 1 GB — `48 × (0.25 × $0.04048 + 1 × $0.004445)` ≈ **$0.70/month**, plus SQS requests inside the free tier. The interval is a variable precisely so this is the owner's dial.

## What the consumer records, and what it refuses to record

| SES event | recorded as | why |
|---|---|---|
| Bounce / **Permanent** | `bounce` | the mailbox does not exist |
| Bounce / Transient | **nothing** | a full mailbox is a real person — stamping it is this issue's own defect, pointed the other way |
| Bounce / Undetermined | **nothing** | SES itself does not know |
| Complaint | `complaint` | a human told their provider to stop us |
| Reject / Delivery | **nothing** | carried on the same stream as the control group: they are how you tell "no bounces" from "no events at all" |

Three properties the code makes and the tests execute rather than assert by inspection:

- **First stamp wins.** SQS is at-least-once, so a redelivery must be a no-op — and an operator who has just run `ses_events clear` must not be silently overwritten by an old event still on the queue. A genuinely new bounce after a clear finds a NULL and stamps again.
- **An unparseable body stays on the queue.** SQS's own `maxReceiveCount` then routes it to the DLQ where somebody can see it. Deleting it would make an AWS schema change look exactly like a quiet week.
- **The write commits BEFORE the delete.** A crash in between costs a duplicate (which is a no-op); the other order costs the event permanently, and SES will never resend it.

## The refusal, and the trade it makes deliberately

Signup for a stamped address answers **422** `EMAIL_ADDRESS_BOUNCED` / `EMAIL_ADDRESS_COMPLAINED` instead of better-auth's generic duplicate 200 (`autoSignIn: false` turns that on — verified against the installed better-auth@1.6.25, not assumed).

That is an account-existence disclosure and it is a trade, not an oversight:

- the disclosed set is **exactly the addresses that cannot receive mail**. Knowing an account exists there buys an attacker nothing they can act on — no reset link, no verification link, nothing can be delivered to it;
- the alternative is the defect: someone who typos their address gets the cheerful 200, never receives the mail, retries the same typo, and gets the same 200 forever with the free tier stuck shut. This refusal is the only moment anything in the product can tell them;
- timing is unaffected — the lookup runs for every signup, not only refused ones.

**The anonymous resend keeps today's uniform 200.** `/send-verification-email` is reachable with no session and answers identically for unknown, verified and genuine sends; that uniformity is what stops it being an account-existence oracle (better-auth backs it with a 500ms constant-time floor). A 4xx there would hand any stranger "this address is registered AND its mail bounces" — strictly more than they knew, and a worse trade than the one this issue is about. Only a caller signed in **as** the address is told.

The two kinds get different words because they are different facts. A permanent bounce is a mailbox that does not exist — "use a different address" is actionable. A complaint is a real person who told their provider our mail is spam, and calling their address broken would be a lie.

## Also in scope

- **`GET /api/auth/verification-status` gains a `bounced` state, above `suppressed`.** Same conclusion from a stronger source (SES pushed it rather than us asking), and it still answers when the suppression lookup is not permitted — which is exactly when the pull half correctly refuses to guess. Without it the panel would say "accepted" beside a button that returns 422.
- **The resend control renders it and disables the button** (both mount points, through the component #1790 already shares).
- **Mailer** passes `ConfigurationSetName`, omitted entirely when blank — SES rejects an empty string, so the pre-apply state degrades to "deaf", not "every verification email fails".
- **`ses_suppression remove` is still the operator path** for the legitimate removal cases, and the runbook is explicit that it is only **half** the unblock: the suppression list and the user stamp are separate facts in separate places, and clearing one alone changes nothing the user can see.

## Guards, each shown RED on the input it rejects

Every one of these was mutated, run, and reverted by editing before pushing.

| Mutation | Test that went red |
|---|---|
| drop `ConfigurationSetName` from `SendEmailCommand` | `a configured set is actually named on the SendEmailCommand SES receives` |
| send `ConfigurationSetName: ""` instead of omitting | `with no configured set the property is ABSENT, not an empty string` |
| terraform `SES_CONFIGURATION_SET` drifts from the resource name | `test_the_auth_container_is_given_the_configuration_sets_own_name` |
| event destination drops BOUNCE/COMPLAINT | `test_the_event_destination_carries_bounce_and_complaint` |
| stamp transient bounces too | `test_non_permanent_bounce_records_nothing[Transient]` + `[Undetermined]` |
| delete unparseable messages instead of leaving them for the DLQ | `test_an_unparseable_body_is_left_on_the_queue` + one more |
| **delete before commit** | `test_the_write_is_committed_before_the_delete_so_a_crash_loses_nothing` — **this one initially survived; the test is new in this PR** |
| remove the signup refusal | 3 tests in `bounced-address.test.js` |
| drop `input: false` on the bounce fields | `a signup body cannot set its own bounce state` |
| refuse the anonymous resend too | `the ANONYMOUS resend still answers 200 for a bounced address — no existence oracle` |
| migration omits `emailBounceKind` | `test_alembic_auth_users_email_bounce_columns_added_and_removed` (new) |
| migration makes the column NOT NULL with a default | same test — "NULL means 'SES has never reported anything'" |
| resolver ignores the pushed stamp | 2 tests in `delivery-feedback.test.js` |
| `bounced` demoted below `suppressed` | `a bounce SES pushed to us is reported as BOUNCED, above suppressed` |
| UI describes a complaint as a dead mailbox | `a COMPLAINT is never described as a broken mailbox — that would be a lie` |
| UI leaves the button enabled for a bounced address | 3 tests in `verification-delivery.test.js` |

**Fix round — eleven more, all of them mine on the reviewer's finding that the wiring guard scanned the whole file instead of the right container:**

| Mutation | Test that went red |
|---|---|
| `SES_CONFIGURATION_SET` moved from the auth container to the backend container | `test_the_auth_container_is_given_the_configuration_sets_own_name` — **this survived before the fix; twelve tests, `terraform fmt`, and 151 tests across every ecs.tf-reading file all stayed green while no send would name the set** |
| `SES_CONFIGURATION_SET` copied onto both auth and backend | same test, second assert |
| `SES_EVENTS_QUEUE_URL` moved from the backend container to nginx | `test_the_backend_container_knows_which_queue_to_drain` — **also survived before the fix** |
| `SES_EVENTS_QUEUE_URL` copied onto both backend and nginx | same test, second assert |
| delete `aws_scheduler_schedule.ses_events_drain` | `test_something_invokes_the_consumer_on_a_clock` |
| drop the `drain` subcommand from the task's `command` | `test_the_scheduled_task_actually_runs_the_drain_command` |
| schedule `state = "DISABLED"` | `test_something_invokes_the_consumer_on_a_clock` |
| `ecs:RunTask` grant widened to a wildcard family | `test_the_scheduler_may_only_run_this_one_family` |
| drop the `DATABASE_URL` secret from the drain task | `test_the_scheduled_task_is_told_which_queue_and_which_database` |
| delete either alarm | `test_a_drain_that_stops_running_sets_off_an_alarm` / `test_a_message_the_parser_cannot_handle_sets_off_an_alarm` |
| age alarm watches message COUNT instead of age, or `treat_missing_data = "breaching"` | `test_a_drain_that_stops_running_sets_off_an_alarm` |
| migration `down_revision` left on the stale head | `.github/scripts/migration_chain_guard.py` exits 1 (see Migration) |

Every one reverted **by editing**, then `diff` against a pre-mutation copy and `git status --porcelain` confirmed clean.

## Migration

`e6b2a19c4d70` chains on **`d3a71f5c9e28`** — the orphaned-legacy-row adoption that PR #1817 landed for issue #1283, which merged into main while this PR was in review. It has now been re-pointed twice for the same reason (first onto #1790's `d4b1f7c8e206`, then onto this), which is what the header warns about and what the CI guard catches. Before the fix, verbatim:

```
$ python3 .github/scripts/migration_chain_guard.py
FAIL: backend/migrations/versions/e6b2a19c4d70_auth_users_email_bounce.py has
down_revision='d4b1f7c8e206', but origin/main's migration chain head is 'd3a71f5c9e28'.
exit=1

>>> ScriptDirectory.get_heads()
['d3a71f5c9e28', 'e6b2a19c4d70']
```

Two heads is not a cosmetic problem: `alembic upgrade head` refuses to run, which fails `deploy.yml`'s pre-rollout `migrate` ECS task, which is a `needs:` of `deploy-ecs` — **every deploy blocked**. After the one-line re-point the guard exits 0 and `get_heads()` returns `['e6b2a19c4d70']`. Safe on the general ground the header already gave, and on a specific one now recorded there: `d3a71f5c9e28` INSERTs an `auth_users` row naming its columns explicitly, so two more nullable columns appearing after it cannot change what that insert does. Two nullable columns, no server default, **no backfill**: the bounces that predate the configuration set were published nowhere and cannot be reconstructed, so inventing timestamps for them would be fabricating evidence. NULL means "SES has never told us anything bad about this address", which is true of every row that exists today.

**One unrelated break fixed along the way.** `test_alembic_migrations.py`'s `_seed_phase1_fixture` seeds a **mid-chain** schema using the **head** ORM model, so any column added to `AuthUser` breaks three phase-1 tests with `no column named emailBouncedAt`. It now reflects `auth_users` at that revision — the same technique the `paper_deployments` block directly beside it already uses, and which its own comment explains. This was a real regression from this branch, caught by running the full suite, not pre-existing.

## Terraform

`terraform fmt -check -recursive` clean. `terraform validate` — `Success! The configuration is valid.` Read-only, `-lock=false`, `AWS_PROFILE=ArchimedesDanAdmin`, `TF_VAR_aurora_master_password` read from SSM in-shell and never printed, **never applied**.

**The plan recipe gained a required variable while this PR sat in review, and it is not this PR's.** `cdad9193` (#1818 P5) landed `owner_alert_email` in `variables.tf` with **no default** — deliberately, so an apply cannot succeed without someone deciding where a page goes. It is the only defaultless variable in the file besides `aurora_master_password`, it is set in the gitignored `infra/terraform.tfvars`, and a fresh checkout does not carry it. So the two exports named above are no longer enough to produce *any* plan here, this one included:

```
Error: No value for required variable

  on variables.tf line 296:
 296: variable "owner_alert_email" {

The root module input variable "owner_alert_email" is not set, and has no
default value.
```

The plan below was re-run with `-var 'owner_alert_email=plan-only-placeholder@archimedes-arc.com'` supplied purely to let the read-only plan resolve. **That placeholder is safe here only because it is a plan and only because of the targets:** `aws_sns_topic_subscription.owner_alerts_email` is the sole consumer of the variable and is not among them, so nothing that reads it appears in the output. Anyone *applying* must use the real address from `infra/terraform.tfvars`, which must differ from `alarm_email` — `cloudwatch.tf`'s own `precondition` enforces that, and the reasoning is in the variable's doc comment.

```
Plan: 15 to add, 1 to change, 0 to destroy.

  + aws_sesv2_configuration_set.mail                        ("archimedes-mail")
  + aws_sesv2_configuration_set_event_destination.feedback
  + aws_sns_topic.ses_events                                ("archimedes-ses-events")
  + aws_sns_topic_policy.ses_events
  + aws_sns_topic_subscription.ses_events_sqs
  + aws_sqs_queue.ses_events                                (14d retention, DLQ, SSE)
  + aws_sqs_queue.ses_events_dlq
  + aws_sqs_queue_policy.ses_events
  + aws_iam_role_policy.ecs_task_ses_events_queue
  + aws_ecs_task_definition.ses_events_drain                ("archimedes-ses-events-drain")
  + aws_iam_role.ses_events_scheduler
  + aws_iam_role_policy.ses_events_scheduler_run_task
  + aws_scheduler_schedule.ses_events_drain                 (rate(15 minutes), ENABLED)
  + aws_cloudwatch_metric_alarm.ses_events_not_being_drained
  + aws_cloudwatch_metric_alarm.ses_events_dlq_not_empty
  ~ aws_iam_role_policy.ecs_task_ses_send                   (in-place, additive)

Changes to Outputs:
  + ses_configuration_set_name              = "archimedes-mail"
  + ses_events_dlq_url                      = (known after apply)
  + ses_events_drain_task_definition_family = "archimedes-ses-events-drain"
  + ses_events_queue_url                    = (known after apply)
```

**Nothing is replaced and nothing is destroyed in that plan.** The one in-place change adds the configuration-set ARN to the send policy's `Resource` list — **required**, because naming a configuration set on `SendEmail` needs permission on the configuration set as well as the identity, and without it every send from the auth container starts failing `AccessDeniedException` the moment `SES_CONFIGURATION_SET` is set. The plan also shows #1790's `ReadAccountSuppressionListForDeliveryFeedback` statement, which merged but has not been applied yet.

### The one target that is NOT safe to apply today, in full

Delivering `SES_CONFIGURATION_SET` and `SES_EVENTS_QUEUE_URL` needs `aws_ecs_task_definition.backend`, and **that resource already carries drift this PR did not create.** Proven pre-existing: the same targeted plan on a clean `origin/main` worktree (`78041c37`, no changes from this branch) already says `must be replaced` — `Plan: 1 to add, 0 to change, 1 to destroy` — with these live-vs-config deltas:

```hcl
# LIVE HAS, CONFIG DOES NOT — these would be REMOVED by an apply
~ container_definitions = jsonencode(
    ~ { name = "auth"
        ~ secrets = [
          - { name = "GITHUB_CLIENT_ID"     valueFrom = ".../archimedes/prod/GITHUB_CLIENT_ID" },
          - { name = "GITHUB_CLIENT_SECRET" valueFrom = ".../archimedes/prod/GITHUB_CLIENT_SECRET" },
          - { name = "GOOGLE_CLIENT_ID"     valueFrom = ".../archimedes/prod/GOOGLE_CLIENT_ID" },
          - { name = "GOOGLE_CLIENT_SECRET" valueFrom = ".../archimedes/prod/GOOGLE_CLIENT_SECRET" },
          ]
      }
    ~ { name = "backend"
        ~ { name = "PLATFORM_ADMIN_WALLETS"
            ~ value = "0x2a2968d5F0E1Dc2d527f5017c36eE97C33A15105" -> "" }
      }
```

`var.github_oauth_enabled` and `var.google_oauth_enabled` both default `false` and `var.platform_admin_wallets` defaults `""` — the live values came from an apply that supplied a `terraform.tfvars` this repo does not carry (it is gitignored, and absent from a fresh checkout). **An apply of this target without those variables set breaks OAuth sign-in in production and drops the admin wallet.** That is #1799's territory, not this PR's, and it is the reason the note in the previous revision of this body ("also changes: it gains `SES_CONFIGURATION_SET` and `SES_EVENTS_QUEUE_URL`") was misleading enough to be called out in review.

The same target also carries six changes that ARE intended, from PRs already merged and awaiting an apply — worth knowing you are switching them all on in the same revision:

```hcl
- { name = "ARCHIMEDES_FUSION_ENABLED", value = "true" }        # flag retired
+ { name = "FREE_GENERATIONS_PER_ACCOUNT", value = "3" }
+ { name = "HEALTH_STALE_UNREADY_S", value = "900" }            # #1818 P3
+ { name = "TIINGO_API_TOKEN", valueFrom = ".../TIINGO_API_TOKEN" }   # #1798
~ { name = "PAPER_ADVANCE_ENABLED", value = "false" -> "true" }
~ healthCheck: .../health -> .../health/ready                   # #1818
```

(The `mountPoints = []`, `systemControls = []`, `volumesFrom = []` and `hostPort` removals in that plan are API-filled defaults terraform normalises away, not changes.)

**Two ways forward, owner's call:**

**(a) Hold the two env vars.** Apply the fifteen resources above today; they are purely additive and touch nothing live. The loop is then built and scheduled but **deaf** — no send names the configuration set, so SES publishes nothing, which is precisely today's behaviour (`auth/mailer.js` omits `ConfigurationSetName` entirely when the variable is blank, so this degrades to the status quo rather than to a broken send path). The drain runs every 15 minutes against an empty queue and reports `received 0`. Flip the vars on with the task definition once #1799 settles ownership.

**(b) Apply the task definition too, with the drift supplied**, i.e. `-var github_oauth_enabled=true -var google_oauth_enabled=true -var platform_admin_wallets=0x2a2968d5F0E1Dc2d527f5017c36eE97C33A15105` (or the equivalent `terraform.tfvars`), so the apply does not push the drift. That switches the whole loop on in one step, and switches on `PAPER_ADVANCE_ENABLED`, the readiness health check and the rest of the intended list with it — including the `TIINGO_API_TOKEN` secret, whose SSM parameter I confirmed is seeded (`aws ssm get-parameter --name /archimedes/prod/TIINGO_API_TOKEN` resolves), so the new revision will launch.

`(a)` is the default this PR assumes; `(b)` is one command away if you want the signal live now.

## Tests

All of these on the pushed HEAD (`6ab11e65`), with `origin/main` (`e81b60f0`) merged in.

**Re-verified from scratch on 2026-09-03**, in a clean worktree reset to `origin/dbrowneup/1804-ses-bounce-push-loop`. `origin/main` (`e81b60f0`) was already an ancestor of HEAD — `git log HEAD..origin/main` empty — so no merge was needed, the migration was already re-pointed correctly (`alembic heads` → `['e6b2a19c4d70']`, one head, `down_revision='d3a71f5c9e28'`), and every number below reproduced identically. The targeted plan reproduced `Plan: 15 to add, 1 to change, 0 to destroy` and all four outputs. Two things came out of the round:

1. **the plan recipe**, corrected in the Terraform section above — main's `cdad9193` made `owner_alert_email` required, so the exports this body named no longer produce a plan on their own;
2. **`6ab11e65`, the three ruff findings this branch added over main.** The Quality Gate comment reports `48` and is informational, so nothing blocked — but that number is a *delta*, and `ruff check .` on `origin/main` finds **45**. Three were this branch's: `SIM108` and `SIM910` in `ses_events.py`, and a `RUF100` unused `# noqa: N803` in `test_ses_bounce_consumer.py` for a rule this repo does not enable. All three are behaviour-preserving (the ternary evaluates the same two branches on the same predicate; `.get` with no default already returns `None`; an unused noqa suppresses nothing either way). `ruff check .` now finds **45 — exactly main's baseline, so this branch's lint delta is zero.**

**Guards re-verified adversarially, not taken on trust.** Four of the table's mutations were re-applied to a separate worktree at this HEAD, run, and reverted **by editing**, with `diff` against a pre-mutation copy and `git status --porcelain` confirming a clean tree after each:

| Mutation re-applied | Result |
|---|---|
| `SES_CONFIGURATION_SET` moved from the auth container to the backend container | **RED** — `test_the_auth_container_is_given_the_configuration_sets_own_name` (1 failed, 18 passed) |
| `session.commit()` moved to AFTER `delete_message` | **RED** — `test_the_write_is_committed_before_the_delete_so_a_crash_loses_nothing` (1 failed, 18 passed) |
| `bounceRefusal()` short-circuited to `return null` | **RED** — 6 of 12 in `bounced-address.test.js`. The anonymous-resend test stayed **green**, which is correct: it asserts the *absence* of a refusal |
| migration drops the `emailBounceKind` column | **RED** — `test_alembic_auth_users_email_bounce_columns_added_and_removed`, "without the kind, a complaint and a dead mailbox get the same…" |

- `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` → **6,646 passed, 4 skipped, 2 deselected, 0 failed** (735s)
- `cd auth && npm test` → **115 passed, 0 failed**
- `cd ui && npm ci && npm test` → **1,067 passed, 0 failed**; `npm run lint` clean; `npm run build` exit 0
- `ruff format --check .` → 778 files already formatted; `ruff check --select E9,F63,F7,F40,F82 .` → All checks passed (`F82` is in `quality-gate.yml`'s set and was missing from the previous revision of this list)
- `python3 .github/scripts/migration_chain_guard.py` → OK, one head
- `docs_links.py` → 1,872 links in 188 files, 0 broken; `docs_index.py` → 183/183 indexed, 0 orphaned
- `infra/scripts/check-user-data-size.sh` → under the gate

## Deliberately left out — the one residual `Part of #1804` carries

**The per-event audit trail.** The issue's minimal loop names `(address, message_id, type, sub-type, timestamp)`. The consumer **parses** `MessageId` and the SES sub-type (`bounceSubType` for a bounce, `complaintFeedbackType` for a complaint) — they reach a `logger.info` line and are then dropped. So the loop can answer "is this address dead" forever, but cannot answer "**which** verification email died" once the log rotates. Closing it needs a table keyed on `MessageId`: a new `bounce_events`, or `auth_email_deliveries` once its `seq` is reachable from Python.

**Which is why the rows do not go into `auth_email_deliveries` today.** #1790 was unmerged when the issue was written; the table it landed with has `seq` as `BIGINT GENERATED BY DEFAULT AS IDENTITY`, and this repo's alembic/ORM tests all run on SQLite, which has no IDENTITY and no sequences:

- an ORM insert omitting `seq` fails `NOT NULL constraint failed` **on SQLite** (verified, not assumed) — on Postgres alone that same insert is correct and advances the sequence, so the obstacle is the test substrate, not the database;
- supplying `seq` explicitly satisfies SQLite but **does not advance the Postgres sequence**, so the sidecar's next insert collides on the UNIQUE constraint, `record()` swallows it, and a real send receipt is lost — a regression on #1790's core guarantee;
- the table's own model docstring states "Nothing in the Python backend inserts here."

The product behaviour that half was sketched to enable — the resend control showing `bounced` — is delivered here more directly, off the user stamp, which the status endpoint already has in hand. `docs/runbooks/ses-bounce-signal.md` § 6 states the residual in the same words.

Hence **`Part of #1804`, not `Closes`.** Everything else the issue lists is here: configuration set, event destination, SNS, SQS, the consumer **and the schedule that invokes it**, the user stamp, the mailer's `ConfigurationSetName`, the signup and resend refusals, the UI state, `#1780` kept as the operator path, the runbook, and the terraform-vs-mailer name-drift guard — now pinned to the container the mailer actually runs in.


dbrowneup:
## Fix round — every defect, and where it landed

| # | Severity | Defect | Commit | How it is shown, not asserted |
|---|---|---|---|---|
| 1 | **blocker** | migration chain forked against current `origin/main` | [`e3f4e8f4`](https://github.com/aprin-labs/archimedes/pull/1837/commits/e3f4e8f4) | `migration_chain_guard.py` exit **1** verbatim in the commit message → exit **0**; `get_heads()` `['d3a71f5c9e28', 'e6b2a19c4d70']` → `['e6b2a19c4d70']` |
| 2 | major | the wiring guard scanned the whole of `ecs.tf`, so the loop could be made deaf with every test green | [`04c19332`](https://github.com/aprin-labs/archimedes/pull/1837/commits/04c19332) | **your two mutations reproduced and now RED**, plus two more you did not run (the var copied onto *both* containers — the presence assert alone would still pass) |
| 3 | major | the PR body's one-line note understated the `aws_ecs_task_definition.backend` drift | body, rewritten | my own `-target` plan on a clean `origin/main` worktree (`78041c37`), with the hunks pasted and two explicit options |
| 4 | major | nothing invoked the consumer — the loop as shipped was inert | [`04c19332`](https://github.com/aprin-labs/archimedes/pull/1837/commits/04c19332) | **wired, not disclaimed**: a scheduled Fargate one-off + two alarms, with nine mutations RED |
| 5 | minor | the per-event audit data is parsed and dropped | [`04c19332`](https://github.com/aprin-labs/archimedes/pull/1837/commits/04c19332) (runbook § 6) + body | the residual is now concrete — `MessageId` + `bounceSubType`/`complaintFeedbackType`, a `bounce_events` table or `auth_email_deliveries` once `seq` is reachable — and the SQLite framing is corrected in both places |

### On #2 — I took your finding one level further

You proved the guard could not tell *which container*. Both of your mutations are red now, and so are the two symmetric ones you did not run (the variable **copied onto both** containers, where the presence assert passes and only the absence assert can fire):

```
SES_CONFIGURATION_SET moved auth -> backend   → test_the_auth_container_is_given_the_configuration_sets_own_name
SES_CONFIGURATION_SET copied onto both        → same test, absence assert
SES_EVENTS_QUEUE_URL  moved backend -> nginx  → test_the_backend_container_knows_which_queue_to_drain
SES_EVENTS_QUEUE_URL  copied onto both        → same test, absence assert
```

Sliced on the **object braces**, not on the container-name markers, so an entry smuggled above a container's own `name` line still lands in the right slice — and `_container()` asserts loudly if the shape ever moves, rather than degrading to a slice that matches nothing and turns every per-container assert into a pass.

### On #4 — I wired it rather than disclosing it

You offered both. The issue's own words are "a small consumer (**scheduled ECS task** or the auth service polling)", so shipping the runnable unit and a paragraph explaining its absence was delivering the sentence, not the requirement.

`infra/ses_events.tf` now carries a dedicated single-container `archimedes-ses-events-drain` task definition — `ecs_migrate.tf`'s one-off shape, **not** a command override on the three-container service family, whose nginx sidecar `dependsOn` a HEALTHY backend that a drain never becomes — invoked by `aws_scheduler_schedule.ses_events_drain` every 15 minutes (`var.ses_events_drain_schedule_expression`), reusing both existing IAM roles (the task role already carried the queue grant; its comment already anticipated "a future scheduled task").

And because an unwatched schedule is just the next silent link, two alarms on free SQS metrics: `ApproximateAgeOfOldestMessage` on the queue (the drain stopped running — the only externally visible symptom) and depth on the DLQ (the parser is behind an SES payload change). Both `notBreaching` on missing data, because an empty queue publishes no age datapoints and empty is the healthy steady state.

Nine mutations RED before pushing: schedule deleted; `drain` subcommand dropped; `state = "DISABLED"`; `RunTask` widened to a wildcard family; `DATABASE_URL` dropped; each alarm deleted; the age alarm switched to a message *count*; the age alarm switched to `treat_missing_data = "breaching"`.

Cost stated rather than waved at: ≈ **$0.70/month** (96 invocations/day × Fargate's 1-minute minimum at 0.25 vCPU + 1 GB).

### One thing I found that you did not, and it was mine

Re-serializing onto `d3a71f5c9e28` broke two of #1283's tests — [`7b0662ba`](https://github.com/aprin-labs/archimedes/pull/1837/commits/7b0662ba). Caught by the full suite, and **not** pre-existing: `test_orphan_row_adoption.py` is 10/10 green on a clean `origin/main` worktree and was 2-failed/8-passed here.

```
FAILED test_dry_run_on_a_zero_population_database_also_writes_absolutely_nothing
  AssertionError: assert [('e6b2a19c4d70',)] == [('d3a71f5c9e28',)]

FAILED test_running_the_upgrade_a_second_time_is_a_no_op
  sqlite3.OperationalError: duplicate column name: emailBouncedAt
```

The second is the interesting one. That test `stamp`s the version pointer back **without** downgrading, so the next upgrade re-executes the revision under test against already-adopted data — the only way to reach its second-run branch. With `"head"` it re-executes every *later* revision too, against a database that already has their DDL, and `add_column` is not idempotent. Fixed by naming `_REVISION` in all eleven `upgrade` calls, with a comment at the constant so the next branch to chain a migration does not rediscover it. Same class as this PR's earlier `_seed_phase1_fixture` fix: a test reaching for "head" is really asking for one specific revision.

### Everything re-run on the merged tree (`origin/main` = `e81b60f0`, merged in; HEAD `8f656822`)

- `python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider` → **6,646 passed, 4 skipped, 2 deselected, 0 failed** (735s)
- `cd auth && npm test` → **115 passed, 0 failed**
- `cd ui && npm ci && npm test` → **1,067 passed, 0 failed**; `npm run lint` clean; `npm run build` exit 0
- `ruff format --check .` → 778 files already formatted; `ruff check --select E9,F63,F7,F40,F82 .` → All checks passed
- `terraform fmt -check -recursive` exit 0; `terraform validate` → `Success!`; targeted read-only plan → **`Plan: 15 to add, 1 to change, 0 to destroy`** (nothing replaced, nothing destroyed), **never applied**
- `migration_chain_guard.py` → OK, one head; docs gates → 1,872 links / 0 broken, 183/183 indexed
- `infra/scripts/check-user-data-size.sh` → under the gate

Still `Part of #1804`, not `Closes` — the per-event audit trail is the one thing left, and it is now written down concretely in the body and in the runbook rather than implied.

Issue #1504, opened 2026-08-29
[infra] Move DMARC from p=none to enforcement once aggregate reports justify it


onder-akkaya:
Follow-up to #1462 (PR #1503), which publishes DMARC at `p=none`. Item 3 of that issue asked for this to be tracked separately rather than folded in, because moving the policy is evidence-led and jumping straight to enforcement risks silently dropping our own sign-in and password-reset mail.

## Why `p=none` is not the finished state

`p=none` tells receivers "here is my policy: do nothing". It closes the *observability* gap (aggregate reports) and satisfies Gmail's and Yahoo's bulk-sender requirement that a DMARC record exist, but it does not actually stop a forged message from being delivered. Only `p=quarantine` or `p=reject` does that.

So #1462's fix makes spoofing *visible*. This one makes it *fail*.

## Preconditions

1. **#1460 must land first.** The `rua` address (`dmarc-reports@archimedes-arc.com`) has an MX pointing at SES inbound, but no receipt rule delivers it — so no reports are being collected today. Without reports there is no evidence, and without evidence this change is a guess.
2. **At least two weeks of aggregate reports**, covering a full send cycle: sign-up verification, password reset, and whatever else is sending by then.
3. **Every legitimate source aligned.** The reports must show SPF *and/or* DKIM passing with alignment for 100% of legitimate volume. Anything failing is a real sending path that will start bouncing at `p=reject`.

## Steps

1. Read the aggregate reports; enumerate every source IP sending as `@archimedes-arc.com`.
2. For each, confirm it is ours and that it aligns. Fix or remove anything that doesn't.
3. Move to `p=quarantine; pct=25`, and hold. Watch for a drop in delivered auth mail.
4. Ramp `pct` to 100 over roughly a week if nothing regresses.
5. Move to `p=reject`.

Each step is a one-line change to `aws_route53_record.dmarc` in `infra/dns_email.tf`.

## Acceptance

- `dig +short TXT _dmarc.archimedes-arc.com @8.8.8.8` contains `p=reject`.
- A test send still arrives (not quarantined) at Gmail, with `dmarc=pass` in the received headers.
- The aggregate reports for the fortnight before the switch show zero legitimate sources failing alignment — attach that evidence to the PR, since it is the whole justification for the change.

## Anti-goals

- **Do not jump to `p=reject`.** The staged ramp exists because the failure mode is our own auth mail disappearing, and it disappears silently at the receiver.
[...]

PR #1849, opened 2026-09-04
infra: mail a weekly DMARC aggregate-report summary (#1504)

dbrowneup:
Part of #1504.

#1836 gave `dmarc-reports@` a mailbox and a parser. **Nothing looked in it.** The owner's
call on this issue (2026-09-03) was explicit — *"the aggregate-report parser runs on a
schedule and posts a weekly summary (per-source pass/fail table) … the issue is not done
without it"* — and on 2026-09-04 the destination was settled: **scheduled, weekly, email to
the alert address, sent from `no-reply@archimedes-arc.com` through the SES identity the
verification mail already uses.** This is that job.

**Does not touch the DMARC policy.** `p=none` stays exactly as it is; the ramp to
`quarantine`/`reject` is the rest of #1504 and is an owner decision after a fortnight of
these reports, which is why this is `Part of` and not `Closes`.

## The design decision worth arguing with

**A quiet week still sends.** If no reports landed, the job mails a one-line
`NO REPORTS RECEIVED` summary and exits **0**. That is not a formality:

- an empty bucket and an un-forged domain look identical, and only one of them is good news;
- a job that sends nothing on a quiet week is indistinguishable from a job that stopped
  running, a task role that lost its S3 grant, or a schedule somebody disabled.

**So the arrival of the Monday mail is the heartbeat, and its absence is the alarm.** That is
also why this adds **no CloudWatch alarm**, unlike `infra/ses_events.tf` next door: the thing
being watched there is a queue nobody looks at, and the thing being watched here is an email
a person reads. A second, weaker monitor on task exit code would be one more thing to mute.

The residual is stated rather than waved at, in the runbook and in the terraform comment: **a
run that fails to send leaves nothing but a non-zero task exit in the log group.** Exit codes
are `0` sent · `2` misconfigured · `3` could not read the bucket · `4` could not send — `3`
and `4` deliberately distinct from `0` **and from each other**, because *"I could not look"*
and *"I looked and found nothing"* are different facts and only one is about DMARC.

## Why the parser moved

`backend/Dockerfile` copies `backend/` **and nothing else**. The scheduled job runs in that
image, so it cannot import `scripts/dmarc_report_summary.py` — that directory is not there.
Duplicating the parser would give the two readers of the same evidence two different answers
to "is the domain being spoofed".

So the parsing, aggregation and rendering are now
`backend/archimedes/scripts/dmarc_reports.py`, and `scripts/dmarc_report_summary.py` is the
CLI over it, keeping its documented `python scripts/dmarc_report_summary.py --path …`
invocation (a `sys.path` insert, the same one `scripts/purge_orphan_generated.py` uses).

Not `archimedes/services/`: that package's `__init__.py` re-exports the generation pipeline,
the portfolio agent and redis state, so importing anything under it drags the application in
and wants `DATABASE_URL` at import time. `archimedes/scripts/__init__.py` is empty — which is
why **the scheduled task carries no secrets block at all**, unlike `ecs_migrate.tf` and the
ses-events drain. A test pins that import surface, so a future
`from archimedes.services…` fails here rather than crash-looping a Monday task.

## What the summary looks like

The **plain-text body is `render_table`'s output verbatim** — literally the same rendering an
operator gets from §2 of the runbook, not a second one that could disagree with it. The
subject carries the verdict so a clean week and a forged one do not read the same in an inbox
list:

```
DMARC weekly [archimedes-arc.com]: all 173 messages aligned across 3 source(s)
DMARC weekly [archimedes-arc.com]: 3 of 45 messages FAILED alignment, 1 source(s), 1 new source(s)
DMARC weekly [archimedes-arc.com]: NO REPORTS RECEIVED in 7 days
DMARC weekly [archimedes-arc.com]: NO READABLE REPORTS (7 unreadable) in 7 days
```

The last two are not the same line — see **Review round 2** below.

Real output, from the fixtures in the test suite:

```
DMARC weekly summary — the last 7 days of aggregate reports.

DMARC aggregate reports: 1 parsed
Window:   2025-08-24 01:46 UTC → 2025-08-25 01:46 UTC
Domain:   archimedes-arc.com
Policy:   none (as the reporters saw it)

SOURCE IP     MSGS  PASS  FAIL  DKIM-A  SPF-A  DISPOSITION  REPORTERS
------------  ----  ----  ----  ------  -----  -----------  ----------
203.0.113.77  3     0     3     0       0      none:3       google.com
54.240.8.1    42    42    0     42      42     none:42      google.com

VERDICT: 3 of 45 messages FAILED DMARC alignment, from 1 source(s): 203.0.113.77.
...

CHANGES SINCE THE PREVIOUS WINDOW
  NEW sources (1): 203.0.113.77 (FAILING)
  A source sending as this domain that was not here last week is either a
  sending path someone added, or a forgery. Name it before the policy moves.
```

The **week-on-week diff** comes from one listing pass covering both windows, so an object
that crosses the boundary between two listings can never be reported as new. A first week of
collection says `nothing to compare against` rather than announcing our own SES egress as an
unexplained sender — a bug the tests caught in the subject line during this build, where the
body said one thing and the subject said another.

The **HTML body escapes every cell.** `dmarc-reports@` is an address published in DNS; anyone
can mail it a hand-written "report", and a crafted `org_name` must not rewrite the summary
the owner reads to decide whether the domain is being forged.

## Infra

Appended to `infra/dmarc_reports.tf`, the same way `infra/ses_events.tf` keeps its own
invocation next to the wiring it drives:

| Piece | Note |
|---|---|
| `aws_ecs_task_definition.dmarc_weekly_summary` | Single container on the **existing backend image** — no new build. No secrets block (see above). |
| `aws_iam_role_policy.ecs_task_dmarc_reports_read` | `s3:ListBucket` (prefix-conditioned) + `s3:GetObject`, on this bucket and prefix only. **Read-only** — the 180-day lifecycle rule stays the only thing that removes a report. |
| `aws_iam_role.dmarc_summary_scheduler` + its RunTask policy | Scoped to this one family, this one cluster, `PassRole` scoped to `ecs-tasks.amazonaws.com`. Same shape as `ses_events_scheduler`. |
| `aws_scheduler_schedule.dmarc_weekly_summary` | `cron(0 13 ? * MON *)`, `ENABLED`, `flexible_time_window` OFF — a predictable arrival is what makes a *missing* summary noticeable. |

**No SES IAM is added.** `aws_iam_role_policy.ecs_task_ses_send` (`infra/ecs.tf`) already
grants this task role `ses:SendEmail` on `identity/${var.domain_name}` — which is precisely
the identity this sends as, and is what makes reusing `no-reply@` free rather than a second
identity to provision. A test asserts that grant still exists and still names the identity,
so deleting it in `ecs.tf` fails a test rather than silently turning the Monday mail off.

One small refactor: the receipt rule's `object_key_prefix` became
`local.dmarc_object_key_prefix`, because four things now have to agree on it and
`aws_ses_receipt_rule.s3_action` is a **set** in the provider schema whose elements have no
addressable index. The refreshed plan below shows the live rule unchanged.

Cost: 52 invocations a year, Fargate's 1-minute minimum at 0.25 vCPU + 1 GB ≈ **$0.01/year**,
plus a handful of S3 GETs and 52 SES messages.

## Read-only targeted plan

`AWS_PROFILE=ArchimedesDanAdmin`, `terraform plan -lock=false` (no state lock taken),
**refreshed against live state**, targeted at the five new resources plus
`aws_ses_receipt_rule.dmarc_reports` to prove the prefix refactor is a no-op. **Nothing was
applied.** `DMARC_SUMMARY_TO`'s value is the live `owner_alert_email` and is redacted below;
it was read from the confirmed `archimedes-alerts` SNS email subscription and never printed.

<details>
<summary><code>terraform plan</code> — 5 to add, 0 to change, 0 to destroy</summary>

```
Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # aws_ecs_task_definition.dmarc_weekly_summary will be created
  + resource "aws_ecs_task_definition" "dmarc_weekly_summary" {
      + arn                      = (known after apply)
      + arn_without_revision     = (known after apply)
      + container_definitions    = jsonencode(
            [
              + {
                  + command          = [
                      + "python",
                      + "-m",
                      + "archimedes.scripts.dmarc_weekly_summary",
                    ]
                  + environment      = [
                      + {
                          + name  = "AWS_REGION"
                          + value = "us-east-1"
                        },
                      + {
                          + name  = "DMARC_REPORTS_BUCKET"
                          + value = "archimedes-dmarc-reports-037613907429"
                        },
                      + {
                          + name  = "DMARC_REPORTS_PREFIX"
                          + value = "reports/"
                        },
                      + {
                          + name  = "DMARC_SUMMARY_TO"
                          + value = "<owner_alert_email>"
                        },
                      + {
                          + name  = "EMAIL_SENDER"
                          + value = "no-reply@archimedes-arc.com"
                        },
                    ]
                  + essential        = true
                  + image            = "037613907429.dkr.ecr.us-east-1.amazonaws.com/archimedes-backend:latest"
                  + logConfiguration = {
                      + logDriver = "awslogs"
                      + options   = {
                          + awslogs-group         = "/archimedes/app"
                          + awslogs-region        = "us-east-1"
                          + awslogs-stream-prefix = "dmarc-weekly-summary"
                        }
                    }
                  + name             = "dmarc-weekly-summary"
                },
            ]
        )
      + cpu                      = "256"
      + enable_fault_injection   = (known after apply)
      + execution_role_arn       = "arn:aws:iam::037613907429:role/archimedes-ecs-task-execution-role"
      + family                   = "archimedes-dmarc-weekly-summary"
      + id                       = (known after apply)
      + memory                   = "1024"
      + network_mode             = "awsvpc"
      + requires_compatibilities = [
          + "FARGATE",
        ]
      + revision                 = (known after apply)
      + skip_destroy             = false
      + tags                     = {
          + "Project" = "archimedes"
        }
      + tags_all                 = {
          + "Project" = "archimedes"
        }
      + task_role_arn            = "arn:aws:iam::037613907429:role/archimedes-ecs-task-role"
      + track_latest             = false

      + runtime_platform {
          + cpu_architecture        = "X86_64"
          + operating_system_family = "LINUX"
        }
    }

  # aws_iam_role.dmarc_summary_scheduler will be created
  + resource "aws_iam_role" "dmarc_summary_scheduler" {
      + arn                   = (known after apply)
      + assume_role_policy    = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "sts:AssumeRole"
                      + Condition = {
                          + StringEquals = {
                              + "aws:SourceAccount" = "037613907429"
                            }
                        }
                      + Effect    = "Allow"
                      + Principal = {
                          + Service = "scheduler.amazonaws.com"
                        }
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
      + create_date           = (known after apply)
      + force_detach_policies = false
      + id                    = (known after apply)
      + managed_policy_arns   = (known after apply)
      + max_session_duration  = 3600
      + name                  = "archimedes-dmarc-summary-scheduler-role"
      + name_prefix           = (known after apply)
      + path                  = "/"
      + tags                  = {
          + "Project" = "archimedes"
        }
      + tags_all              = {
          + "Project" = "archimedes"
        }
      + unique_id             = (known after apply)

      + inline_policy (known after apply)
    }

  # aws_iam_role_policy.dmarc_summary_scheduler_run_task will be created
  + resource "aws_iam_role_policy" "dmarc_summary_scheduler_run_task" {
      + id          = (known after apply)
      + name        = "archimedes-dmarc-summary-scheduler-run-task"
      + name_prefix = (known after apply)
      + policy      = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = [
                          + "ecs:RunTask",
                        ]
                      + Condition = {
                          + ArnLike = {
                              + "ecs:cluster" = "arn:aws:ecs:us-east-1:037613907429:cluster/archimedes-cluster"
                            }
                        }
                      + Effect    = "Allow"
                      + Resource  = "arn:aws:ecs:us-east-1:037613907429:task-definition/archimedes-dmarc-weekly-summary:*"
                      + Sid       = "RunDmarcWeeklySummaryTask"
                    },
                  + {
                      + Action    = "iam:PassRole"
                      + Condition = {
                          + StringEquals = {
                              + "iam:PassedToService" = "ecs-tasks.amazonaws.com"
                            }
                        }
                      + Effect    = "Allow"
                      + Resource  = [
                          + "arn:aws:iam::037613907429:role/archimedes-ecs-task-execution-role",
                          + "arn:aws:iam::037613907429:role/archimedes-ecs-task-role",
                        ]
                      + Sid       = "PassDmarcWeeklySummaryRoles"
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
      + role        = (known after apply)
    }

  # aws_iam_role_policy.ecs_task_dmarc_reports_read will be created
  + resource "aws_iam_role_policy" "ecs_task_dmarc_reports_read" {
      + id          = (known after apply)
      + name        = "archimedes-ecs-dmarc-reports-read"
      + name_prefix = (known after apply)
      + policy      = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = [
                          + "s3:ListBucket",
                        ]
                      + Condition = {
                          + StringLike = {
                              + "s3:prefix" = [
                                  + "reports/*",
                                ]
                            }
                        }
                      + Effect    = "Allow"
                      + Resource  = "arn:aws:s3:::archimedes-dmarc-reports-037613907429"
                      + Sid       = "ListDmarcReportsBucket"
                    },
                  + {
                      + Action   = [
                          + "s3:GetObject",
                        ]
                      + Effect   = "Allow"
                      + Resource = "arn:aws:s3:::archimedes-dmarc-reports-037613907429/reports/*"
                      + Sid      = "ReadDmarcReports"
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
      + role        = "archimedes-ecs-task-role"
    }

  # aws_scheduler_schedule.dmarc_weekly_summary will be created
  + resource "aws_scheduler_schedule" "dmarc_weekly_summary" {
      + arn                          = (known after apply)
      + group_name                   = (known after apply)
      + id                           = (known after apply)
      + name                         = "archimedes-dmarc-weekly-summary"
      + name_prefix                  = (known after apply)
      + schedule_expression          = "cron(0 13 ? * MON *)"
      + schedule_expression_timezone = "UTC"
      + state                        = "ENABLED"

      + flexible_time_window {
          + mode = "OFF"
        }

      + target {
          + arn      = "arn:aws:ecs:us-east-1:037613907429:cluster/archimedes-cluster"
          + role_arn = (known after apply)

          + ecs_parameters {
              + launch_type         = "FARGATE"
              + task_count          = 1
              + task_definition_arn = "arn:aws:ecs:us-east-1:037613907429:task-definition/archimedes-dmarc-weekly-summary"

              + network_configuration {
                  + assign_public_ip = false
                  + security_groups  = [
                      + "sg-0f21c2773901a14e7",
                    ]
                  + subnets          = [
                      + "subnet-010efb75b1093ba92",
                      + "subnet-0f412b89a025ca15b",
                    ]
                }
            }

          + retry_policy {
              + maximum_event_age_in_seconds = 86400
              + maximum_retry_attempts       = 1
            }
        }
    }

Plan: 5 to add, 0 to change, 0 to destroy.

Changes to Outputs:
```

</details>

Note the receipt rule refreshed cleanly and is **not** in the plan: `local.dmarc_object_key_prefix`
resolves to the same `reports/` the live rule already has.

### Apply this only after the image is pushed

The task definition pins `archimedes-backend:latest`, and `:latest` in ECR today (pushed
2026-09-04 09:01, `bcfd224e` = `origin/main`) **does not contain**
`archimedes/scripts/dmarc_weekly_summary.py` — that module ships with this branch. Applying
before the branch is merged and `deploy.yml` has pushed a new `:latest` gives the first
scheduled task a `ModuleNotFoundError`, mails nothing, and — by the deliberate no-alarm
design above — says so nowhere but a task exit code in `/archimedes/app`. The habit on this
issue has been a targeted apply from a checkout (that is how #1836 went in), so the ordering
is not hypothetical.

**Order: merge → let CI push `:latest` → apply → prove it** with the `--dry-run` invocation or
a manual `aws ecs run-task`, rather than waiting for a Monday to find out. Written into
`infra/dmarc_reports.tf` next to the `image` line and into runbook §7.

## Tests

`backend/tests/scripts/test_dmarc_weekly_summary.py` (23) and
`backend/tests/test_dmarc_summary_wiring.py` (13), plus the parser suite (18) moved
onto shared fixtures in `backend/tests/dmarc_fixtures.py` so both consumers of the parser
test the same report shapes.

**Every guard verified RED under mutation before pushing** (20 mutations, each applied and
reverted, each caught):

| Mutation | Caught by |
|---|---|
| `dmarc_pass` reads `auth_results` instead of `policy_evaluated` | parser + summary suites |
| HTML escaping removed | `test_report_content_is_escaped_before_it_reaches_the_html_body` |
| Zero reports sends nothing | `test_zero_reports_still_sends_a_message_that_says_so` |
| Send failure swallowed / missing `MessageId` treated as a send | `test_a_refused_send_exits_four…`, `test_a_response_with_no_message_id_is_not_a_send` |
| Read failure rendered as "no reports" | `test_a_read_failure_exits_three_and_sends_nothing` |
| Window widened to everything; previous-window guard removed | `test_only_the_last_seven_days_are_summarised`, `test_no_reports_last_week_does_not_make_every_source_new` |
| Schedule → daily / `rate(1 hour)` / Mon+Thu | `test_the_schedule_is_weekly` |
| `DMARC_SUMMARY_TO` removed, or hardcoded to a literal | `test_the_task_names_the_owner_alert_variable_as_the_summary_destination` |
| Bucket or prefix hardcoded | `test_the_task_is_told_which_bucket_and_prefix_off_the_resources_themselves` |
| Schedule set `DISABLED`; command pointed at the drain | `test_a_schedule_invokes_the_summary_task`, `test_the_scheduled_task_runs_the_summary_module` |
| Read grant gains `s3:DeleteObject`; RunTask widened to `task-definition/*` | `test_the_task_role_can_read_the_reports_bucket_and_nothing_more`, `test_the_scheduler_may_only_run_this_one_family` |
| `ses:SendEmail` grant deleted in `ecs.tf` | `test_the_send_grant_still_exists_where_it_is_reused_from` |

`terraform fmt -check` and `terraform validate` clean. Full backend CI
(`python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider`) from the repo root
on the current head: **7212 passed, 4 skipped, 2 deselected, exit 0** in 594s — the 7206 of the
previous round plus exactly the six tests round 2 adds.

## Review round 2

Five defects from the review, all addressed in `f1a878f5`.

**Major — "I could not read any of it" was being mailed as "there was nothing to read."**
`_no_reports_lines()` and `subject_line()` both branched on `reports_parsed == 0` alone and
never read `summary.unreadable`; `render_html()` never rendered it on *either* path. The
reviewer reproduced it against this branch: seven real aggregate reports landed in the window
and every one tripped `MAX_ARCHIVE_MEMBERS`, so the subject read
`NO REPORTS RECEIVED in 7 days`, the body said *"the absence of evidence … either no receiver
had mail from us to report on, or the collection path is broken"* and pointed at §5's arrival
ladder — MX, receipt rule, active rule set, prefix — **none of which was the fault**. The only
trace was `(7 object(s) in the window)`, while the parser was holding seven lines reading
`zip holds 69 members (limit 64); refusing`.

That is the exact conflation this file argues against, arriving one door along: the operator
CLI has always printed those errors (`for item in summary.unreadable`, before `return 2`), and
the mail did not. `"I could not look"`, `"I looked and found nothing"` and `"I looked, I found
seven things, and I could not read any of them"` are three facts, and the third now has:

- its own subject line — `NO READABLE REPORTS (7 unreadable) in 7 days`, so the inbox list is
  not a false all-clear;
- its own body — `N object(s) landed in the window and NOT ONE of them held a report this
  parser could read`, then every parse error named;
- its own runbook section — **§6**, about contents rather than arrival, because §5's six steps
  are all about mail that never came.

It still exits **0**: the run succeeded, so the fact belongs in the subject line, not an exit
code. And the **HTML part now carries the `UNREADABLE` list on both paths** — that part is what
nearly every mail client actually displays, so a fact reaching only the text body was a fact
the owner did not see. (Second symptom of the same root: a mixed good/bad window rendered
`UNREADABLE (n) — these were NOT counted above` into the text part and nothing into the HTML.)

**Minor — the IAM guard could not see a resource over-grant.** It asserted
`"aws_s3_bucket.dmarc_reports.arn" in grant` as a substring over the whole two-statement
policy, so *either* statement's `Resource` could be widened and the other still satisfied it.
The reviewer mutated `ListBucket`'s `Resource` to `"*"` (list every bucket in the account):
green. Then `GetObject`'s to `"*"` (read every object in every bucket): green. Both silent,
past a guard whose message reads *"the grant must name this bucket, not a wildcard"*. Now
split on `Sid` and asserted **per statement** — each `Resource` pinned to its exact
expression, the statement set itself pinned, and no statement allowed a bare `"*"`. The
shipped policy was correctly scoped throughout; this was a guard hole, not a live over-grant.

**Minor — one oversized cell could make the message unsendable.** `render_table` pads every
row to its column's widest cell, so a single attacker-supplied cell is multiplied by the row
count: measured, a 219 KB report carrying one 200 KB `<source_ip>` plus 30 ordinary records
rendered a **6.6 MB** table, and the factor grows with the number of sources.
`dmarc-reports@` is published in DNS and the receipt rule stores whatever anyone mails it.
Past SES's per-message limit that is a refused send — exit 4, **no Monday mail** — which by
this PR's own no-alarm design is indistinguishable from the silence it exists to break. Now
bounded in three places: cells clipped at 64 characters (longer than any real IPv6 address,
org_name or disposition), `unreadable` notes at 200 (they embed archive member names chosen by
the sender), and each assembled body capped at 512 KB with a visible `(TRUNCATED …)` note
naming the operator command. **Clipped, never dropped** — a source missing from the table is
the quiet undercount everything here is built against.

**Minor — the runbook's live-state claim**, and **minor — the apply ordering**: both above.

### Red proofs

Eight mutations, each applied to a clean tree, run, and reverted. **All eight caught:**

| Mutation | Caught by |
|---|---|
| `subject_line` ignores `unreadable` (the behaviour the reviewer reproduced) | `test_reports_that_all_fail_to_parse_are_not_reported_as_no_reports_received` |
| `_no_reports_lines` ignores `unreadable` | same |
| `render_html` drops the unreadable block from the populated path | `test_the_html_body_carries_the_unreadable_list_whenever_the_text_body_does` |
| `ListBucket` `Resource` → `"*"` | `test_the_task_role_can_read_the_reports_bucket_and_nothing_more` |
| `GetObject` `Resource` → `"*"` | same |
| `clip_cell` stops clipping | `test_an_oversized_cell_cannot_inflate_the_rendered_table` — failed on `len(...) == 6602506`, reproducing the reviewer's measurement exactly |
| `bounded_body` stops bounding | `test_a_body_past_the_size_limit_is_truncated_rather_than_left_unsendable` |
| runbook renames the §6 heading the mail points at | `test_the_runbook_sections_the_summary_mails_people_to_actually_exist` (new — the mail's only next step is a section name, so a renamed heading turns it into a dead reference) |

**Terraform is unchanged** apart from one HCL comment, which does not render into
`container_definitions`. The plan above was re-run read-only on the new head
(`AWS_PROFILE=ArchimedesDanAdmin`, `-lock=false`, refreshed) and is **byte-identical** to the
one pasted above — still `5 to add, 0 to change, 0 to destroy`, receipt rule still absent.
Nothing applied. `terraform fmt -check -recursive` and `terraform validate` clean; `ruff check`
and `ruff format --check` clean on every changed file.

## Runbook

`docs/runbooks/dmarc-reports.md` gains **§7 The weekly summary**: what is wired, what a
healthy summary looks like (including the two things that are *not* findings — `SPF-A 0` on
our own SES egress, and `nothing to compare against` in the first fortnight), how to force one
with `aws ecs run-task`, how to see one without sending anything (`--dry-run`), and how to
read the last run's log stream. It also gains **§6 Reports arrive but cannot be parsed**, the
diagnosis for the third state (round 2, below), with the three refusal messages you will
actually see and what each one is worth doing about.

**The opening paragraph is corrected.** It still said "the bucket is empty until it is
applied"; #1836 was applied on 2026-09-04 at 13:19Z (`6 added, 0 changed, 0 destroyed`, test
object under `reports/`), so that sentence was false on a published page. It is now split:
collection is live and §§2–5 describe today; §7's weekly summary is not live until this PR's
terraform is applied, and §7 says so in its own first line.

The page stays **published** — nav entry unchanged, no secrets on it.
`docs/runbooks/README.md` and `docs/doc-index.md` updated; the "not built yet"
bullet in the old §6 is gone.


dbrowneup:
## Work-leader vet (2026-09-04) — SHIP, held for the owner's word

Vetted the content commit `5718998e` (the second commit is only the merge of main after #1847):

- `backend/tests`: **48 passed** across `test_dmarc_weekly_summary.py`, `test_dmarc_report_summary.py`, `test_dmarc_summary_wiring.py`; ruff check + format clean on all six files.
- **Adversarial:** default schedule mutated to `rate(1 day)` → `test_the_schedule_is_weekly` red, restored.
- Exit-code claims in the body checked against `main()`: `0` sent (quiet week included), `2` misconfigured, `3` `ReportsUnreadable`, `4` `SummaryNotSent` — all true.
- **Read-only targeted `terraform plan`** from this branch: **5 to add, 0 to change, 0 to destroy** (task definition on `archimedes-backend:latest`, scheduler `cron(0 13 ? * MON *)` UTC, S3 read policy scoped to `reports/*`, scheduler role trusting this account only). Destination resolves to the owner alert address; sender `no-reply@archimedes-arc.com`; no secrets block on the task.
- The inbox side is already live and proven (probe mail in the bucket at 13:20:28Z, on #1504).

**Non-blocking nit:** `dmarc_reports.py` parses with `xml.etree.ElementTree.fromstring`. The 64 MB decompression cap plus the image's Python 3.12 expat (amplification limits on by default) mitigate entity blow-up, but a one-line `<!DOCTYPE` reject before `fromstring` would make that explicit rather than inherited from libexpat. Fine as a follow-up.

After merge: targeted apply of the five resources (plan file already prepared), then `aws ecs run-task` once by hand to see the first mail land before Monday.

dbrowneup:
## Test results

**Full backend CI, from the repo root, on this branch's head (`5545715`), after merging `origin/main`:**

```
$ python -m pytest -m "not integration" --tb=short -q -p no:cacheprovider
==== 7206 passed, 4 skipped, 2 deselected, 339 warnings in 589.05s (0:09:49) ====
```

Clean. Worth recording how it got there, because the first run was not:

An earlier run on this branch reported **33 failed, 5 errors** — every one of them in
`test_alembic_migrations.py` and `test_account_deletion_cascade.py`, all `alembic upgrade
head failed`. I checked them against a detached worktree at `origin/main` before assuming
anything, and they failed identically there: the same 33, the same 5. Then #1847
(`hotfix-alembic-single-head`) landed on `main`, I merged it in, and all 51 of those tests
pass. So they were a pre-existing multiple-heads condition on `main`, not this change, and
they are now fixed by someone else's PR rather than tolerated.

**Targeted:**

```
$ python -m pytest backend/tests/scripts/test_dmarc_weekly_summary.py \
    backend/tests/scripts/test_dmarc_report_summary.py \
    backend/tests/test_dmarc_summary_wiring.py \
    backend/tests/test_ses_event_wiring.py -q
67 passed
```

`test_ses_event_wiring.py` is in there deliberately: this PR touches the shared ECS task role
and reads `infra/ecs.tf`, so the sibling loop's guards are the check that nothing was
disturbed next door.

**Terraform:**

```
$ terraform fmt -check infra/     # clean
$ terraform validate              # Success! The configuration is valid.
```

**Mutation pass** — the 20 mutations in the PR body were applied one at a time to a clean
tree and reverted after each, and all 20 went red. Two of them are worth naming because they
found real defects during the build rather than confirming existing ones:

- the subject line counted **every** source as new when the previous window held no reports,
  while the body correctly said `nothing to compare against`. A first week of collection
  would have announced our own SES egress as an unexplained sender.
- the HTML-escaping test initially passed for the wrong reason: the hostile `org_name` was
  raw `<script>` in the XML, which is not well-formed and never reached the renderer at all.
  Fixed the fixture to XML-escape it, so the string the renderer actually receives is the
  dangerous one.

**Not tested here, and it cannot be until the owner applies:** that SES accepts a real send
from the Fargate task role, and that the private subnets' egress reaches the SES endpoint.
The task role's `ses:SendEmail` grant is the same one the auth container uses in production
today, and S3 reads go over the existing gateway endpoint — but neither is a live
observation on *this* task, and the first real Monday is what confirms it. The runbook's
"Run one by hand" section exists so that does not have to wait for a Monday.

dbrowneup:
Head sha correction: the merge commit was amended to carry the required commit trailers, so the branch head is now `c651f243`, not `5545715` as the test-results comment above says. The tree is byte-identical — `git diff 5545715 c651f243` is empty; only the merge commit's message changed. Every test result reported above stands unchanged.

dbrowneup:
## Fix round — all five defects addressed in `f1a878f5`

One commit; the mapping below is defect → what changed. Full backend CI on the new head:
**7212 passed, 4 skipped, 2 deselected, exit 0** (594s) — the 7206 you verified plus exactly
the six tests this round adds. `ruff check` / `ruff format --check` clean on every changed
file; `terraform fmt -check -recursive` and `terraform validate` clean.

### 1 · MAJOR — unreadable reports mailed as `NO REPORTS RECEIVED` → `f1a878f5`

You were right that this is a third fact, and right that the operator CLI already had it.

- `subject_line()` — new branch: `NO READABLE REPORTS (n unreadable) in 7 days`. The exit code
  stays **0**: the run succeeded, so the fact belongs in the subject line, not an exit code.
- `_no_reports_lines()` — takes `unreadable` and has two shapes now. The new one leads with
  `N object(s) landed in the window and NOT ONE of them held a report this parser could read`,
  says in words that the arrival ladder is not the fault, lists every entry, and points at a
  **new runbook §6 'Reports arrive but cannot be parsed'** instead of §5's arrival ladder.
- `render_html()` — renders `summary.unreadable` on **both** paths, escaped: the zero path
  through the shared lines, the populated path as its own tinted `<ul>` block after the
  changes div. Your second symptom (mixed good/bad pair: text had it, HTML did not) is that
  one.
- New runbook **§6** carries the three refusal messages an operator will actually see, what
  each means, what to do, and the read-only commands to fetch the object. §§7–8 renumbered.

Your exact scenario, replayed against the new head — 7 objects, each a `<feedback>` document
zipped into 69 members:

```
SUBJECT: DMARC weekly [archimedes-arc.com]: NO READABLE REPORTS (7 unreadable) in 7 days

NO READABLE REPORTS in the last 7 days: 7 object(s) landed in the window
and NOT ONE of them held a report this parser could read.

This is neither a quiet week nor a clean one. Something IS arriving, so the
collection path is not the fault and the arrival ladder is not where to look.
An unreadable object is counted nowhere: not in the table, not in the failures.

UNREADABLE (7) — every object in the window:
  s3://…/reports/real-report-0.eml: …::real-report.zip: zip holds 69 members (limit 64); refusing
  … ×7

Looked at: s3://archimedes-dmarc-reports-1/reports/
Diagnose (read-only): docs/runbooks/dmarc-reports.md § 'Reports arrive but cannot be parsed'.
```

HTML carries the same list (`'refusing' in html` → `True`). Exit 0, one message sent.

### 2 · MINOR — the IAM guard could not see a resource over-grant → `f1a878f5`

`test_the_task_role_can_read_the_reports_bucket_and_nothing_more` now splits the policy on
`Sid` and asserts **per statement**: `ListDmarcReportsBucket`'s `Resource` is exactly
`aws_s3_bucket.dmarc_reports.arn` (plus its `s3:prefix` condition, which is the only thing
that actually scopes a listing), `ReadDmarcReports`'s is exactly
`"${aws_s3_bucket.dmarc_reports.arn}/${local.dmarc_object_key_prefix}*"`, the statement set
itself is pinned so a third one cannot be added unseen, and no statement may carry a bare
`"*"`. Both of your mutations now go red — see the table below.

### 3 · MINOR — render amplification → `f1a878f5`

Bounded in three places, in `archimedes/scripts/dmarc_reports.py` and the summary:

- `clip_cell()` + `MAX_CELL_CHARS = 64` on every table cell in **both** renderers — including
  the counts, since `<count>` is an integer out of untrusted XML and a 4,000-digit one renders
  a 4,000-column table.
- `MAX_NOTE_CHARS = 200` on `unreadable` entries — they embed archive member names chosen by
  the sender, and this round puts them in the email.
- `MAX_BODY_BYTES = 512 KB` per body in `bounded_body()`, called in `main()` before
  `send_summary`, with a visible `(TRUNCATED …)` note naming the operator command. Cell
  clipping bounds the table's *width*; the row count is still unbounded on its own (one
  hand-written report may declare thousands of distinct source IPs), so the body cap is what
  actually bounds it.

**Clipped, never dropped** — a source that vanished from the table is the quiet undercount
everything in this file is built against. Your 6.6 MB measurement is now the assertion message
of a test (see M6).

### 4 · MINOR — the runbook's opening paragraph → `f1a878f5`

Split, as you suggested. Collection has been live since 2026-09-04 13:19Z (`6 added, 0
changed, 0 destroyed`, test object under `reports/`) so §§2–5 describe today; §7's weekly
summary is not live until this PR's terraform is applied, and §7 opens by saying so.

### 5 · MINOR — apply-before-deploy ordering → `f1a878f5`

Written in three places: a comment on the `image` line in `infra/dmarc_reports.tf`, a callout
at the top of runbook §7, and a section in the PR body. **Merge → let CI push `:latest` →
apply → prove it** with `--dry-run` or `aws ecs run-task`, rather than waiting for a Monday to
find a `ModuleNotFoundError` that surfaces nowhere.

### Red proofs — 8 mutations, each applied to a clean tree, run, reverted. All 8 caught.

| # | Mutation | Result |
|---|---|---|
| M1 | `subject_line` ignores `unreadable` (the behaviour you reproduced) | RED — 1 failed |
| M2 | `_no_reports_lines` ignores `unreadable` | RED — 1 failed |
| M3 | `render_html` drops the unreadable block from the populated path | RED — 1 failed |
| M4 | `ListBucket` `Resource` → `"*"` (**your mutation**) | RED — 1 failed |
| M5 | `GetObject` `Resource` → `"*"` (**your mutation**) | RED — 1 failed |
| M6 | `clip_cell` stops clipping | RED — failed on `6602506 = len(rendered)`, your figure exactly |
| M7 | `bounded_body` stops bounding | RED — 1 failed |
| M8 | runbook renames the §6 heading the mail points at | RED — 1 failed |

M8 is a new guard, `test_the_runbook_sections_the_summary_mails_people_to_actually_exist`: the
mail's only actionable line is a section name, so a renamed heading turns it into a dead
reference for someone who is already looking at something they do not understand.

### Terraform: unchanged, plan re-verified

The only `.tf` edit is one HCL comment, which does not render into `container_definitions`. I
re-ran your read-only targeted plan on the new head (`AWS_PROFILE=ArchimedesDanAdmin`,
`terraform plan -lock=false`, refreshed, `TF_VAR_owner_alert_email` read in-shell from the
`archimedes-alerts` SNS subscription and `TF_VAR_aurora_master_password` from SSM, neither
printed, output redacted) and diffed it against the block in the PR body: **byte-identical**.
Still `Plan: 5 to add, 0 to change, 0 to destroy`; `aws_ses_receipt_rule.dmarc_reports`
refreshes and stays out of the plan. **Nothing applied, no AWS writes, no merges.**

Also unchanged and still true: `p=none` untouched, so this stays **Part of #1504** — the
ramp is your call after a fortnight of these reports.

dbrowneup:
## Work-leader vet, corrected for the fix head `f1a878f5` — SHIP, held for the owner's word

My earlier comment vetted the pre-review commit `5718998e` (48 tests). The builder's verifier then found five defects and the fixer landed them in `f1a878f5`; this supersedes that comment.

**The five defects and their fixes, re-checked by me on `f1a878f5`:**
1. *(major)* Objects that arrive but cannot be parsed were mailed as `NO REPORTS RECEIVED`. Now a third state: subject `NO READABLE REPORTS (n unreadable)`, both bodies list every failure, runbook § 6 is the diagnosis target.
2. *(minor)* IAM guard only checked actions. Now per-statement Resource assertions.
3. *(minor)* One oversized cell amplified the table ~30×. Cells clipped at 64 chars, notes at 200, bodies capped at 512 KB with a visible truncation note.
4. *(minor)* Runbook claimed the bucket was empty until applied. Now: collection live since 2026-09-04 13:20Z; only § 7's weekly job waits on this apply.
5. *(minor)* Apply-before-image hazard undocumented. Runbook § 7 now says: merge → let deploy push `:latest` → apply → prove with one `run-task`.

**Evidence on `f1a878f5`:** 54 targeted tests pass; ruff clean; read-only targeted plan **5 to add, 0 to change, 0 to destroy**. Adversarial, my own two mutations: the unreadable branch of `subject_line` made to fall through → **2 failed / 21 passed**; the ListBucket statement's Resource → `"*"` → wiring guard **red**. Both restored. Builder's own 8 mutations also reported red.

**After your word:** merge → wait for the deploy to push `:latest` → targeted apply of the five resources → one hand `run-task` and read the mail before Monday 13:00 UTC. The DOCTYPE-reject nit from my first comment stands as a follow-up.

